PR #96, opened 2018-09-02
Enforce profile name: set AWS.config.credentials.

onlybakam:
*Issue #, if available:*

*Description of changes:*

amplify-cli does not seem to apply the profile name specified on init.
Looking at: packages/amplify-provider-awscloudformation/lib/{configuration-manager.js,initializer.js}

process.env.AWS_PROFILE is being set to the profile name.
However, the AWS library has already been loaded and AWS_PROFILE env variable will not be read again.

Need to explicitly reconfigure credentials.
see ref: https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/loading-node-credentials-shared.html


onlybakam:
will do. closing this PR though and will open a new one. some other changes were included on the branch.

Issue #99, opened 2018-09-03
GraphQl API creation failed


psycura:
**Do you want to request a feature or report a bug?**
Bug
**What is the current behavior?**
After initialisation of GraphQl api, trying to do amplify push, but receive error:

CREATE_FAILED               ---------                AWS::CloudFormation::Stack Mon Sep 03 2018 16:13:56 GMT+0300 (Israel Daylight Time) Template format error: Parameter count 170 is greater than max allowed 60

my schema.graphql:

```
`type User  @model{
    id: ID!
    categories:[Category] @connection(name:"UserCategories")
    company: String!
    email: String!
    firstName: String!
    lastName: String!
    logo:String
    site:String!
    role:Int!
    type:String!
    package:Package! @connection(name:"UserPackage")
    customers:[Customer] @connection(name:"UserCustomers")
    resources:[Resource] @connection(name:"UserResources")
    proposals:[Proposal] @connection(name:"UserProposals")
    deliverables:[Deliverable] @connection(name:"UserDeliverables")
}

type Package @model{
    id:ID!
    deleted:Boolean!
    name:String!
    price:Float!
    type:Int!
    users: [User] @connection(name:"UserPackage")
}

type Category  @model{
    id:ID!
    name:String!
    user: User! @connection(name:"UserCategories")
}

type ContactPerson @model{
    company:String!
    email:String!
    firstName:String!
    id:ID!
    lastName:String!
    type:String!
    customer:Customer! @connection(name:"CustomerContactPersons")
    proposals:[Proposal] @connection(name:"ContactPersonProposals")
}

type Customer @model{
    id:ID!
    logo:String
    name:String!
    user:User! @connection(name:"UserCustomers")
    contactPersons:[ContactPerson] @connection(name:"CustomerContactPersons")
    proposals:[Proposal] @connection(name:"CustomerProposals")
}

type Resource @model{
    category:String!
    cost:Float!
    description:String!
    id:ID!
    name:String!
    price:Float!
    user:User! @connection(name:"UserResources")
    workItems:[WorkItem] @connection(name:"WorkItemResources")
}

type Proposal @model{
    createdAt: AWSDateTime!
    company:Customer! @connection(name:"CustomerProposals")
    contactPerson:ContactPerson! @connection(name:"ContactPersonProposals")
    id:ID!
    owner:User! @connection(name:"UserProposals")
    parameters:ProposalParameters! @connection(name:"ProposalParameters")
    sections:AWSJSON!
[...]

Issue #11, opened 2018-08-24
Cloudformation Parameter Limit AppSync


pigmanbear:
Amplify CLI version: 0.1.13
OS: macOS 10.13.6
 
Any non-trivial AppSync application will have may have more than 60 parameters added to the nested api stack, which is beyond the limits of CloudFormation and causing amplify push to fail.

mikeparisstuff:
This is a great point. My initial thought is that we can get around this by leveraging nested stacks more. If instead of outputting a single stack, the transformer output a stack per @model, for example, we could avoid violating the parameter count limit.

More generally, I think there is a need for a "Serializer" concept in the transform-core. Right now, the **graphql-appsync-transformer** handles taking the transformer context and writing it out in the correct format which seems wrong. I could see it working like this:

```javascript
interface TransformerSerializer {
  // Returns the CloudFormation template as a string or buffer.
  serialize(ctx: TransformerContext): string | Buffer
}

// You could then optionally override the default serializer.
const transform = new GraphQLTransform({ transformers: [...], serializer: new CustomSerializer() })
```

A little off topic but with something like this it would be easier to swap out the logic that writes to one or multiple CF docs (or even terraform in the future) as these we run into these constraints.

pigmanbear:
A stack output for each @model is the direction I was exploring after running into the limit. 

So, each model stack would be nested under the primary api stack, which is nested under the primary root? How would you handle the file creation?

As for the broader case, the "Serializer" makes sense, do you see this with just the default creation of per @model stacks unless a "CustomSerializer" is provided? 

I was able to go through the source and forked/clone and will play around a bit. 

Thanks for the tip on where in the source to start looking.

mikeparisstuff:
Yeah I was thinking along the same lines. I'll do a little research as well and see if something seems to make more sense than a stack per @model. One thing to think about is that the schema currently has to be a single file (or single CF resource). Thus the primary api stack will have to contain the full schema and will have to deploy before the nested @model stacks which contain the resolvers, datasources, roles, tables etc.

pigmanbear:
Yes, back in March, I created a deploy process for tearing down and building up resources using CloudFormation/AppSync, et al, etc. (essentially, a lot of what amplify is doing for me now). So, it was great to see amplify as 'infrastructure as code' with AppSync was a pain. 

[...]

PR #115, opened 2018-09-06
Fixing 60 parameter limit error on graphql api create.

mikeparisstuff:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-cli/issues/99
https://github.com/aws-amplify/amplify-cli/issues/11

*Description of changes:*

These changes stop the bleeding on issues https://github.com/aws-amplify/amplify-cli/issues/99 and https://github.com/aws-amplify/amplify-cli/issues/11. This reorganizes the transform so only 3 parameters are needed to configure the resolver s3 locations instead of (2 x the number of resolvers). This is still not a long-term fix and I suspect the next limit we hit will be the max of 200 resources per stack. I am also working on another solution that uses nested stacks but this will take a bit longer to finish.

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/upload-appsync-files.js
+++ b/packages/amplify-provider-awscloudformation/lib/upload-appsync-files.js
@@ -5,6 +5,15 @@ const pythonStreamingFunctionFileName = 'python_streaming_function.zip';
 const schemaFileName = 'schema.graphql';
 const S3 = require('../src/aws-utils/aws-s3');
 
+const ROOT_APPSYNC_S3_KEY = 'amplify-appsync-files'
+const providerName = require('./constants').ProviderName;
+
+function getProjectBucket(context) {
+  const projectDetails = context.amplify.getProjectDetails();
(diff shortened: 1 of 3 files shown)


kaustavghosh06:
@mikeparisstuff  We should document these limitations.

packages/amplify-provider-awscloudformation/lib/upload-appsync-files.js
```diff
@@ -79,7 +94,7 @@ function uploadAppSyncFiles(context, resources) {
 function uploadAppSyncFile(context, fileName, filePath, s3LocationMap, buildTimeStamp) {
```
kaustavghosh06: Don't need this anymore I guess?

mikeparisstuff: This is still used to upload the schema file.

mikeparisstuff:
@kaustavghosh06 We can document the CF limitations but ideally will get the nested stack stuff implemented so its not an issue.

kawallis:
if we have an older project that is hitting this limit will updating the cli solve this issue or do we need to create a new project ?

Issue #78, opened 2018-08-29
amplify add notifications


minbi:
**Do you want to request a feature or report a bug?**
bug

**What is the current behavior?**
does not add identity pool / IAM access when notifications is added to the project

If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal amplify add notifications

**What is the expected behavior?**
add identity pool / IAM access when notifications is added to the project

**Additional environment details**

* **OS version (ie Windows 10 build X, macOS Sierra 10.12.6, etc.)** macOS Sierra
* **Output of ``amplify --version``**: 0.1.15
* **Did this work in previous versions?**: N

rohandubal:
@haverchuck tagging as requested. Cannot assign as I do not have permissions.

muellerfr:
What's the status of this issue?

UnleashedMind:
The update has already been implemented and sent out in the PR #123 
Once it's approved, it'll be merged and published in our next release.

muellerfr:
@UnleashedMind Thanks for the update.

kaustavghosh06:
We just published to npm with a fix for this. Please use the version -> 0.1.19.
Closing the issue, feel free to re-open if the issue persists.

PR #123, opened 2018-09-07
notifications and publish updated.

UnleashedMind:
*Issue #, if available:*
#78

*Description of changes:*
- resolved issue #78, when a Pinpoint app is created by notifications category, auth is checked, which include creating the identitypool if not already created, and creating the necessary policy and attach it to the auth and unauth roles. 
- addressed issue raised by mlab on publish
- removed some dependencies


packages/amplify-category-notifications/lib/auth-helper.js
```diff
@@ -0,0 +1,125 @@
...
+        Resource: [
+          `arn:aws:mobiletargeting:*:${accountNumber}:apps/${pinpointAppId}*`,
+        ],
+      },
+    ],
+  };
+  return JSON.stringify(policy);
+}
+
+function getPlicyName(context) {
```
kaustavghosh06: Typo: getPolicy?

packages/amplify-cli/src/cli.js
```diff
@@ -1,6 +1,6 @@
 const { build } = require('gluegun');
 const path = require('path');
-const globalPrefix = require('global-prefix');
+const globalPrefix = require('./lib/get-global-prefix');
```
kaustavghosh06: @mlabieniec  I can't seem to remember the reason, but why had you recommended against using the global-prefix package?

packages/amplify-category-notifications/lib/auth-helper.js
UnleashedMind: check

PR #137, opened 2018-09-08
feat: support yarn global packages

ifyoumakeit:
*Issue #, if available:*
Fixes #6

*Description of changes:*
Uses `yarn-global-modules` to detect global package path whether the cli
was installed there. If so, it uses the node_modules there instead.

Previously yarn was unable to find sibling packages for looking up
plugins and categories. This was due to `global-prefix` support for
only `npm`.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #148, opened 2018-09-10
amplify-cli: support separation of bin  and 

ifyoumakeit:
I tried my patch on another computer with a symlinked dotfile setup and it was unable to find the `yarn` `node_modules` due to the `__dirname` being different from the global folder directory. This is a simpler test (maybe too simple) for whether it was installed by `yarn` over `npm`.

In the future it might help to ask if you are using `yarn` as a flag or question in the CLI. Optionally you can also search both global folders for `yarn` and `npm`, ideally only one will return the correct plugins, although it could generate a false positive.

Tested with an `npm` and yarn `global` install.

This is the filepath that caused the issue.
```
globalYarnModulesDirPath => '/Users/dave.garwacke/.config/yarn/global/node_modules',
__dirname => '/Users/dave.garwacke/src/dotfiles/config.dotfile/yarn/global/node_modules/@aws-amplify/cli/src' 
```
updates #137


kaustavghosh06:
Looks good. Minor lint nitpick- because of which the testing phase fails in circleci:

lerna ERR! /home/circleci/repo/packages/amplify-cli/src/lib/get-global-prefix.js
lerna ERR!   5:32  error  Trailing spaces not allowed  no-trailing-spaces

kaustavghosh06:
@ifyoumakeit 
Looks good. Minor lint nitpick- because of which the testing phase fails in circleci:

lerna ERR! /home/circleci/repo/packages/amplify-cli/src/lib/get-global-prefix.js
lerna ERR! 5:32 error Trailing spaces not allowed no-trailing-spaces

packages/amplify-cli/src/lib/get-global-prefix.js
```diff
@@ -2,7 +2,7 @@ const fs = require('fs-extra');
 const path = require('path');
 const os = require('os');
 const ini = require('ini');
-const which = require('which'); 
+const which = require('which');
```
ifyoumakeit: Fixed this up @kaustavghosh06

PR #149, opened 2018-09-10
Semicolon removal

EliasSuzarte:
I think its not necessary to use semicolon. Am i right?


kaustavghosh06:
We follow the commonly used Airbnb ES6 style - and it recommends using semicolons as a part of the style.

PR #150, opened 2018-09-11
Add support for GraphQL document generation

yuth:
*Issue #, if available:*

*Description of changes:*
Adding support for GraphQL document generation using introspection schema. This generates all possible queries, mutations and subscriptions for a given introspection schema.

Updated the code gen cli to trigger the document generation


packages/amplify-graphql-docs-generator/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+  const parserMap = {
+    javascript: 'babylon',
+    graphql: 'graphql',
+    typescript: 'typescript',
+    flow: 'flow',
+  }
+  try {
+    return prettier.format(str, { parser: parserMap[language] })
+  } catch (e) {
+    console.log('Error')
```
kaustavghosh06: Do we want this? should probably use the logger

yuth: Yea, should change this.

kaustavghosh06:
Pulled in and tried to cover some common user-flows and found some small issues with a few of the flows:
1) When running "amplify codegen add —apiId xxxxxxxxxxxxx53ui" command, the documents (queries, mutations, subscriptions) doesn't get generated. Although the interesting bit is that the graphqlconfig.yml and the corresponding schema.json is fetched correctly

2) When adding or updating the API and pushing the API, always seeing the following warning.( Need to suppress this since its incorrect)

_**WARNING:** You have modified your schema locally and not pushed to the cloud, which may result in incomplete type generation.
We recommend you first run $amplify push_

packages/amplify-codegen/commands/codegen/add.js
```diff
@@ -1,14 +1,15 @@
...
 
-const featureName = 'add'
+const featureName = 'add';
 
 module.exports = {
   name: featureName,
   run: async (context) => {
     try {
-      await codeGen.add(context)
+      const apiId = context.parameters.options.apiId || null;
```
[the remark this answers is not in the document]
yuth: `context.parameters.options` will always be defined. It might be an empty object but its always be present if the context is passed by gluegun

`apiId` can be undefined in which case it becomes null

packages/amplify-codegen/src/constants.js
```diff
@@ -24,10 +31,19 @@ module.exports = {
     'No API is pushed to cloud. Did you forget to do \n amplify api push',
   WARNING_CODEGEN_PENDING_API_PUSH:
     'The APIs listed below are not pushed to the cloud. Run amplify api push',
+  ERROR_APPSYNC_API_NOT_FOUND:
+    'Could not find the AppSync API. If you have removed the AppSync API in the console run amplify codegen remove',
+  MSG_CODEGEN_PENDING_API_PUSH: `${chalk.bold(
+    'WARNING:',
+  )} You have modified your schema locally and not pushed to the cloud, which may result in incomplete type generation.\nWe recommend you first run ${chalk.underline(
```
kaustavghosh06: nitpick: in an incomplete

packages/amplify-codegen/src/constants.js
```diff
@@ -27,9 +27,9 @@ module.exports = {
   CMD_DESCRIPTION_NOT_SUPPORTED: 'invalid subcommand',
   CMD_DESCRIPTION_CONFIGURE: 'Change/Update codegen configuration',
   ERROR_CODEGEN_NO_API_CONFIGURED:
-    'code generation is not configured. Configure it by running \namplify codegen add',
+    'code generation is not configured. Configure it by running \n$amplify codegen add',
   ERROR_CODEGEN_PENDING_API_PUSH:
-    'No API is pushed to cloud. Did you forget to do \n amplify api push',
+    'AppSync API is not pushed to the cloud. Did you forget to do \n$amplify api push',
```
kaustavghosh06: GraphQL API* Lets be consistent.

PR #178, opened 2018-09-19
Initial Native development guide for iOS and Android using Amplify CL…

undefobj:
…I and new codegen features

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/native_guide.md
+++ b/native_guide.md
@@ -1,0 +1,572 @@
+# Native development with Amplify CLI and AWS AppSync
+
+The AWS Amplify CLI supports a rich workflow for Native application development on iOS or Android. This guide will take you through the code generation capabilities of Amplify with an AWS AppSync GraphQL API. Where appropriate, it will also explain general Amplify concepts for native app development.
+
+### Table of Contents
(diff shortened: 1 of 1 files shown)


native_guide.md
```diff
@@ -0,0 +1,572 @@
...
+
+```bash
+$amplify codegen types
+```
+A new updated Swift file will be created (or run Gradle Build on Android for the same). You can then use the updates in your application code.
+ 
+**Flow 4: Shared schema, modified elsewhere (e.g. console or team workflows)**
+
+Suppose you are working in a team and the schema is updated either from the AWS AppSync console or on another system. Your types are not out of date because your GraphQL statement was generated off an outdated schema. The easiest way to resolve this is to regenerate your GraphQL statements, update them if necessary, and then generate your types again. Modify the schema in the console or on a separate system, then run:
```
m14t: "not out of date" should be "now out of date"

undefobj:
Fixed "not out of date" in flow 4.

native_guide.md
undefobj: Fixed

PR #192, opened 2018-09-21
Add support to add cross region AppSync API to codegen

yuth:
*Issue #, if available:*

*Description of changes:*
amplify codegen add --apiId <api-id>` fails if you run this command in a project which is set to use `<region1 >` and the AppSync API is in <region2>`. 

Updated add command to detect this and prompt user to choose different region.

### DIFF
--- a/packages/amplify-codegen/src/commands/add.js
+++ b/packages/amplify-codegen/src/commands/add.js
@@ -6,15 +6,19 @@ const generateTypes = require('./types');
 const {
   AmplifyCodeGenNoAppSyncAPIAvailableError: NoAppSyncAPIAvailableError,
   AmplifyCodeGenAPIPendingPush,
+  AmplifyCodeGenAPINotFoundError,
 } = require('../errors');
 const {
   downloadIntrospectionSchemaWithProgress,
   getAppSyncAPIDetails,
   getAppSyncAPIInfo,
+  getProjectAwsRegion,
 } = require('../utils');
 const addWalkThrough = require('../walkthrough/add');
+const changeAppSyncRegion = require('../walkthrough/changeAppSyncRegions');
 
 async function add(context, apiId = null) {
+  let region = getProjectAwsRegion(context);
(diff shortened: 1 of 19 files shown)


packages/amplify-codegen/src/commands/add.js
```diff
@@ -31,17 +35,36 @@ async function add(context, apiId = null) {
...
+          context.print.info(`AppSync API was not found in region ${region}`);
+          ({ shouldRetry, region } = await changeAppSyncRegion(context, region));
+        } else {
+          throw e;
+        }
+      }
+    }
+  }
+
+  if (!apiDetails) {
```
kaustavghosh06: Whats this condition for?

packages/amplify-codegen/src/constants.js
```diff
@@ -7,10 +7,12 @@ module.exports = {
   PROMPT_MSG_GQL_FILE_PATTERN:
     'Enter the file name pattern of graphql queries, mutations and subscriptions',
   PROMPT_MSG_GENERATE_CODE: 'Do you want to generate code for your newly created GraphQL API',
+  PROMPT_MSG_CHANGE_REGION: 'Do you want to choose a different region',
```
kaustavghosh06: According to me, this message should be like. 
Could not find API in <region>. Do you want to choose a different region?

yuth: We do show that API was not found and then we show this question asking them if they want choose a different region. So the question looks like this
```
amplify codegen add --apiId hahq7hxq5fbsto3ldozg6ewaje
✖ Getting API details
AppSync API was not found in region us-east-1
? Do you want to choose a different region Yes
? Choose AWS Region
  US East (N. Virginia)
❯ US East (Ohio)
  US West (Oregon)
  EU (Ireland)
  EU (Frankfurt)
  Asia Pacific (Tokyo)
  Asia Pacific (Singapore)
```

PR #227, opened 2018-09-27
Allow users to provide their own Query, Mutation, and Subscription root object types.

mikeparisstuff:
*Description of changes:*

These changes allow users to provide their own Query, Mutation and Subscription root object types.

For example,

```graphql
type Post @model {
  id: ID!
  title: String
}
type Query {
  customQuery: String
}
```

With this input, after compilation the root query type will be:

```graphql
type Query {
  customQuery: String
  getPost(id: ID!): Post
  listPosts(...): PostConnection
}
```

You can also now turn off mutations/subscriptions entirely simply by leaving them out in your schema definition without having to update every @model or @searchable:

```graphql
schema {
  query: Query
  # explicitly leaving out mutation so no mutations will be created
}
type Query {
  customQuery: String
}
type Post @model { # NOTE: I do not have to set mutations: null to turn off mutations.
  id: ID!
  title: String!
}
```

You can also change the name of the root object types.

```graphql
schema {
  query: MyQueryType
}
type MyQueryType {
  customQuery: String
}
type Post @model {
  id: ID!
  title: String!
}
```

and the generated query fields will be placed on the `MyQueryType` type instead of the default `Query`.

### DIFF
--- a/packages/graphql-transformer-core/src/TransformerContext.ts
+++ b/packages/graphql-transformer-core/src/TransformerContext.ts
@@ -296,10 +296,7 @@ export default class TransformerContext {
      * @param fields The fields to add the query type.
      */
     public addQueryFields(fields: FieldDefinitionNode[]) {
-        const schemaNode = this.getSchema();
-        const queryTypeOperation = schemaNode.operationTypes.find((op: OperationTypeDefinitionNode) => op.operation === 'query');
-        const queryTypeName = queryTypeOperation && queryTypeOperation.type && queryTypeOperation.type.name && queryTypeOperation.type.name.value ?
-            queryTypeOperation.type.name.value : undefined;
+        const queryTypeName = this.getQueryTypeName();
         if (queryTypeName) {
             if (!this.getType(queryTypeName)) {
                 this.addType(blankObject(queryTypeName))
@@ -316,12 +313,7 @@ export default class TransformerContext {
      * @param fields The fields to add the mutation type.
      */
     public addMutationFields(fields: FieldDefinitionNode[]) {
-        const schemaNode = this.getSchema();
-        const mutationTypeOperation = schemaNode.operationTypes.find((op: OperationTypeDefinitionNode) => op.operation === 'mutation');
-        const mutationTypeName = (
-            mutationTypeOperation && mutationTypeOperation.type &&
-            mutationTypeOperation.type.name && mutationTypeOperation.type.name.value
-        ) ? mutationTypeOperation.type.name.value : undefined;
+        const mutationTypeName = this.getMutationTypeName();
         if (mutationTypeName) {
             if (!this.getType(mutationTypeName)) {
                 this.addType(blankObject(mutationTypeName))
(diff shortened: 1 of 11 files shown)


mikeparisstuff:
@kaustavghosh06 Can you add @mwudtke as a reviewer

kaustavghosh06:
@mwudtke, @mlabieniec Sent you an invite for adding you to the list of collaborators.

packages/graphql-transformer-core/src/TransformerContext.ts
```diff
@@ -167,6 +289,70 @@ export default class TransformerContext {
...
+     * Extends the context query object with additional fields.
+     * If the customer uses a name other than 'Query' this will proxy to the
+     * correct type.
+     * @param fields The fields to add the query type.
+     */
+    public addQueryFields(fields: FieldDefinitionNode[]) {
+        const schemaNode = this.getSchema();
+        const queryTypeOperation = schemaNode.operationTypes.find((op: OperationTypeDefinitionNode) => op.operation === 'query');
+        const queryTypeName = queryTypeOperation && queryTypeOperation.type && queryTypeOperation.type.name && queryTypeOperation.type.name.value ?
+            queryTypeOperation.type.name.value : undefined;
```
mwudtke: It looks like you're duplicating the work from getQueryTypeName() here. Shouldn't you just call that function?

packages/graphql-transformer-core/src/TransformerContext.ts
```diff
@@ -167,6 +289,70 @@ export default class TransformerContext {
...
+     * correct type.
+     * @param fields The fields to add the mutation type.
+     */
+    public addMutationFields(fields: FieldDefinitionNode[]) {
+        const schemaNode = this.getSchema();
+        const mutationTypeOperation = schemaNode.operationTypes.find((op: OperationTypeDefinitionNode) => op.operation === 'mutation');
+        const mutationTypeName = (
+            mutationTypeOperation && mutationTypeOperation.type &&
+            mutationTypeOperation.type.name && mutationTypeOperation.type.name.value
+        ) ? mutationTypeOperation.type.name.value : undefined;
```
mwudtke: Same here, with getMutationTypeName().

packages/graphql-transformer-core/src/TransformerContext.ts
```diff
@@ -167,6 +289,70 @@ export default class TransformerContext {
...
+     * correct type.
+     * @param fields The fields to add the subscription type.
+     */
+    public addSubscriptionFields(fields: FieldDefinitionNode[]) {
+        const schemaNode = this.getSchema();
+        const subscriptionTypeOperation = schemaNode.operationTypes.find((op: OperationTypeDefinitionNode) => op.operation === 'subscription');
+        const subscriptionTypeName = (
+            subscriptionTypeOperation && subscriptionTypeOperation.type &&
+            subscriptionTypeOperation.type.name && subscriptionTypeOperation.type.name.value
+        ) ? subscriptionTypeOperation.type.name.value : undefined;
```
mwudtke: getSubscriptionTypeName()?

blbradley:
Yo, Michael. This looks awesome. Thanks!

vedransisic:
This allows us to add custom queries, but how do we allow amplify to keep generating resolvers for the default queries, but stop it from removing our .request and .response files for our custom ones? It seems like doing `amplify push --no-gql-override` stops all overrides, including the default queries/mutations.

alexandrecanuto:
@vedransisic I'm wondering that as well. I need to add a custom query, but would like to keep the generated queries as well.

The custom query is added, but my custom resolvers are not.

edfinegan:
Is there any documentation on this? I'm hoping to clear up how this should be used. I need to create a second query on an object type. Do I use this syntax in the schema and then define my own resolver and index in Dynamo or is there a way to do that within amplify. I'm not sure the best approach to this problem.

lorensr:
Here are the docs on creating a custom resolver: 

https://aws-amplify.github.io/docs/cli/graphql#add-a-custom-resolver-that-targets-a-dynamodb-table-from-model

PR #214, opened 2018-09-25
Add support for non-model types for mutations. Issues 72 & 190

mwudtke:
https://github.com/aws-amplify/amplify-cli/issues/190
https://github.com/aws-amplify/amplify-cli/issues/72

Adding support for non-model types and enums on the @model transformer for create and update mutations.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #230, opened 2018-09-28
Ensures we don't add multiple non-model input types

snvishna:
*Issue #, if available:*
This is related to [Issue #214](https://github.com/aws-amplify/amplify-cli/pull/214)

*Description of changes:*
When a non-model type is referenced within multiple @model types, the transformer ran into issues.

For example, this test used to fail before:

```graphql
    type Post @model {
        id: ID!
        title: String!
        createdAt: AWSDateTime
        updatedAt: AWSDateTime
        metadata: PostMetadata
        entityMetadata: EntityMetadata
        appearsIn: [Episode!]
    }
    type Author @model {
        id: ID!
        name: String!
        postMetadata: PostMetadata
        entityMetadata: EntityMetadata
    }
    type EntityMetadata {
        isActive: Boolean
    }
    type PostMetadata {
        tags: Tag
    }
    type Tag {
        published: Boolean
        metadata: PostMetadata
    }
    enum Episode {
        NEWHOPE
        EMPIRE
        JEDI
    }
```

### DIFF
--- a/packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
+++ b/packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
@@ -105,10 +105,9 @@ export class DynamoDBModelTransformer extends Transformer {
         nonModelArray.forEach(
             (value: ObjectTypeDefinitionNode) => {
                 let nonModelObject = makeNonModelInputObject(value, nonModelArray, ctx)
-                if (this.typeExist(nonModelObject.name.value, ctx)) {
-                    return
+                if (!this.typeExist(nonModelObject.name.value, ctx)) {
+                    ctx.addInput(nonModelObject)
                 }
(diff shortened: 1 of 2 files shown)


packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -105,6 +105,9 @@ export class DynamoDBModelTransformer extends Transformer {
         nonModelArray.forEach(
             (value: ObjectTypeDefinitionNode) => {
                 let nonModelObject = makeNonModelInputObject(value, nonModelArray, ctx)
+                if (this.typeExist(nonModelObject.name.value, ctx)) {
```
mikeparisstuff: Instead of the return why not just check for the negative. `if(!this.typeExists(...)) { ctx.addInput() }`

PR #240, opened 2018-10-01
Ability to add external AppSync API to an Amplify project as a part of the codegen add command

kaustavghosh06:

### DIFF
--- a/packages/amplify-category-api/commands/api/remove.js
+++ b/packages/amplify-category-api/commands/api/remove.js
@@ -1,4 +1,5 @@
 const path = require('path');
+
 const subcommand = 'remove';
 const category = 'api';
 const gqlConfigFilename = '.graphqlconfig.yml';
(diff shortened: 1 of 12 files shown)


packages/amplify-category-api/commands/api/remove.js
```diff
@@ -8,6 +10,17 @@ module.exports = {
     const resourceName = parameters.first;
 
     return amplify.removeResource(context, category, resourceName)
+      .then((resourceValues) => {
+        if(resourceValues.service === "AppSync") {
+          const {projectPath} = amplify.getProjectDetails().projectConfig;
```
yuth: linter

packages/amplify-codegen/src/utils/updateAmplifyMeta.js
```diff
@@ -0,0 +1,40 @@
+
```
yuth: Can we use a helper function available in  cli core to update the amplify meta file

Issue #127, opened 2018-09-07
Enable sorting on @connection


onlybakam:
**Is your feature request related to a problem? Please describe.**

take this schema:

```graphql
type Conversation @model {
  id: ID!
  messages: [Message] @connection(name: "ConvoMsgs")
}

type Message @model {
  id: ID!
  conversation: Conversation @connection(name: "ConvoMsgs")
}
```

this produces

```graphql
type Conversation {
  id: ID!
  messages(filter: ModelMessageFilterInput, sortDirection: ModelSortDirection, limit: Int, nextToken: String): ModelMessageConnection
}
```

messages resolver works by doing query on GSI with `messageConversationId` as the **HASH** key.
There is no **RANGE** key, so the sort has no impact (since resolver is using `ScanIndexForward`) and returned order is not guaranteed. This has a huge impact if we fetch messages in bunches. Think of a long conversation or a long todo list. the list order is important.

**Describe the solution you'd like**

Update the `@connection` directive to include a sortField

example

```graphql
type Message
  @model(
    mutations: { create: "createMessage" }
    queries: null
    subscriptions: null
  ) {
  id: ID!
  conversation: Conversation
    @connection(name: "ConvoMsgs", sortField: "createdAt")
}
```

which creates:

```javascript
{
...
  "GlobalSecondaryIndexes": [
    {
      "IndexName": "gsi-ConvoMsgs",
      "KeySchema": [
        {
          "AttributeName": "messageConversationId",
          "KeyType": "HASH"
        },
        {
          "AttributeName": "createdAt",
          "KeyType": "RANGE"
        }
      ],
...
  ]
...
}
```

If `sortField` is excluded, a **RANGE** key should not be created. 
Adding the `sortField` does not impact the generated schema or resolver.
setting the `createdAt` or `updatedAt` attribute as **RANGE** could be interesting defaults.

**Describe alternatives you've considered**

Not sure there's an alternative here. any directions?

CodySwannGT:
I get that this is a feature, but this is the single-handed reason we're not using Amplify in production. 

If there's any sort of good work around, would love to hear it.

mikeparisstuff:
@CodySwannGT This feature has been added with the merged PR above. We can close this issue.

stanleycyang:
I'm seeing issues with this after the new cloudformation fix:

```
[...]

Issue #56, opened 2018-08-28
Allow specifying strategy to implement @model id


onlybakam:
**Do you want to request a feature or report a bug?**

Feature Request

**What is the current behavior?**

currently when using `@model`, resolver for create mutation sets id to `$util.autoId()`.
There are situations where this is not wanted. e.g.: we may want to use the current user identity information as the ID (use case: build user setting store).

could the directive be updated to something like this:

```graphql
directive @model(
    id: idStrategy
    queries: ModelQueryMap,
    mutations: ModelMutationMap,
    subscriptions: ModelSubscriptionMap
) on OBJECT
enum idStrategy {auto sub username cognitoIdentityId}
input ModelMutationMap { create: String, update: String, delete: String }
input ModelQueryMap { get: String, list: String }
input ModelSubscriptionMap {
    onCreate: [String]
    onUpdate: [String]
    onDelete: [String]
}
```

where:

```
auto: util.autoId()
sub: ctx.identity.sub
username: ctx.identity.username
cognitoIdentityId: ctx.identity.cognitoIdentityId
```

with `auto` as the default

onlybakam:
The problem with what you suggest is now you get into the necessity for mobile and web developers to understand distributed systems concepts in databases, which is exactly what we want to avoid. 
> i agree we should try to avoid this if possible

 This also has implications to offline storage and persistence in the clients
> not necessarily

The identity of a user can be used as a key to retrieve a single object. This is made available by CUP and does not require persistence beyond what is already done. Suppose i want to store information about a User in dynamodb table and want to put auth around it. right now i could do

```graphql
type User @model @auth(rules: [{ allow: owner }]) {
id: ID!
owner: String!
}
```

there is no direct way to get this a User. i would have to use `listUsers` and filter on owner field. If there are 100s of users, that means scanning the all the rows before filtering the returned data. seems costly.

this might be a limited use case, but i could see this coming up for collaborative applications where user information is maintained with appsync.

thoughts?

undefobj:
Not sure I understand that use case directly because you could just have a query to get that user's specific data, either with an expression or an index on the userId per my previous post. And this *does* impact offline storage tracking as the semantics by which we perform CRUD operations in local DBs is dependent on the key schema and references. 

I'm still not convinced of your use case, but if we were to do something like this it would have to be an override and not the default, with caveats that it can break some things (hotkeys, offline).

onlybakam:
[...]

PR #277, opened 2018-10-06
Implement sorting on connection by implementing RANGE key

onlybakam:
Allows to specify `sortField` on a named connection crated by the
`@connection` directive. Has no effect on an unnamed connection.

```graphql
type Conversation @model {
  id: ID!
  messages: [Message]
    @connection(name: "ConvoMsgs", sortField: "createdAt")
}

type Message @model {
  id: ID!
  createdAt: AWSDateTime
  conversation: Conversation @connection(name: "ConvoMsgs")
}
```

Creates a GSI on the message table with hash key: `messageConversationId`
and range key `createdAt`

*Issue #, if available:* #127 

*Description of changes:*

- Update directive to accept sortField
- creates a range key for the GSI of the table of the associated type
- No effect if the connection is not named (need name to find the associated type)
- the sortField can be of any type that maps to a DynamoDB Scalar (not a list or boolean)
- Adds the sortField attribute to the attributeDefinitions list if it does not exist


undefobj:
@mikeparisstuff when reviewing this let's think about the interface consistency when compared to the @model type discussion from the other day. For instance in this PR `sortField` is in the argument, but an alternative implementation might be a directive on the field:

type Conversation @model {
  id: ID!
  messages: [Message]
    @connection(name: "ConvoMsgs")
}

type Message @model {
  id: ID!
  createdAt: AWSDateTime @sort 
  conversation: Conversation @connection(name: "ConvoMsgs")
}

This would be more consistent with the proposal in https://github.com/aws-amplify/amplify-cli/issues/56#issuecomment-416736520

Otherwise I think this does a good job at not leaking implementation of DDB into the transform as `@sort` or `sortField` are generic to the use case.

mikeparisstuff:
I think this is a reasonable way to go about this to solve an evident customer pain point. In the future, I still think we can look into more advanced ways to configure key schemas of model types and can see a future where @connection takes an index as an argument instead of a sortField.

I.E. If I have a model type that looks something like this:

```
type Post @model {
  id: ID! @primary
  createdAt: String! @sort @sort(index: "blogIdIndex")
  blogId: @primary(index: "blogIdIndex")
}
```

Then I might use my pre-defined index:

```
type Blog @model {
  id: ID!
  posts: [Post] @connection(index: "blogIdIndex")
}
```

With this being said, I think this PR solves a pain point now and doesn't stop us from pursuing these advanced scenarios in the future so think we can work on getting this merged.

mikeparisstuff:
Can we add an e2e test in the graphql-transformers-e2e-tests package for this?

amlcodes:
this has killed me for weeks. would be helpful in docs

PR #291, opened 2018-10-10
Fixing integration tests & moving amplify init to seperate expect script

nidsharm:
*Issue #, if available:*

*Description of changes:*


.circleci/amplify_init.sh
```diff
@@ -1,3 +1,5 @@
 #!/bin/sh -xv
+echo $1
```
kaustavghosh06: What are we echoing here?

nidsharm: The repo name, to make sure it picks up the correct repo. This script is used by both auth and api repos

PR #326, opened 2018-10-19
fix: update pinpoint IAM policy to use mobiletargeting instead of mobileanalytics

kaustavghosh06:
*Description of changes:*
Due to a recent change with Pinpoint API's, IAM policies need to be changed for the newly created Pinpoint API's. This PR modifies the IAM policies generated for a pinpoint endpoint using the CLI.

### DIFF
--- a/packages/amplify-category-notifications/lib/auth-helper.js
+++ b/packages/amplify-category-notifications/lib/auth-helper.js
@@ -96,7 +96,7 @@ function getPolicyDoc(context) {
       {
         Effect: 'Allow',
         Action: [
-          'mobileanalytics:PutEvents',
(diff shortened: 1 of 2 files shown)


packages/amplify-category-notifications/lib/auth-helper.js
```diff
@@ -96,7 +96,7 @@ function getPolicyDoc(context) {
       {
         Effect: 'Allow',
         Action: [
-          'mobileanalytics:PutEvents',
+          'mobiletargeting:PutEvents',
```
powerful23: Shouldn't move this line to be next to line 108?

Issue #340, opened 2018-10-23
Cannot set identityField to custom attribute using @auth


mkrn:
**Describe the bug**
When using @auth directive and "owner" type rule with identityField set to `custom:fieldName` the generator creates invalid template code

**To Reproduce**
Create a model 
```
type PigRun @model @auth(rules: [{
  allow: owner,
  ownerField: "pigRunClientId",
  identityField: "custom:company"
}]) {
  id: ID!
  client: Company! @connection(name: "CompanyPigRuns")
}

type Company @model {
  id: ID!
  name: String!
}
```
The generated resolver templates will use invalid `$ctx.identity.custom:company` syntax in Ownership check sections

**Expected behavior**
I would expect the generators to use `$ctx.identity.claims.get("custom:company_id")` syntax that have worked for me with ID token before. 

**Desktop (please complete the following information):**
 - amplify -v 
0.1.28

mkrn:
The workaround is to set claims in Pre Token Generation Trigger lambda,

like this:
```
exports.handler = (event, context, callback) => {
    const { request } = event;
    const { userAttributes } = request;

    event.response = {
        "claimsOverrideDetails": {
            "claimsToAddOrOverride": {
                "company": userAttributes['custom:company'],
            }
        }
    };

    // Return to Amazon Cognito
    callback(null, event);
};```

But leaving this here as a potential to improve for newcomers and first time users. The `identityField` is largely undocumented as well, so could be worth extending documentation for cases like this.

**Comment by mkrn** [comment]
Another gotcha for newcomers is to make sure to use `idToken`:

In your front-end code:
```
Amplify.configure({
  ...aws_exports,
  graphql_headers: async () => {
    try {
      const token = (await Auth.currentSession()).idToken.jwtToken;
      return { Authorization: token }
    }
    catch (e) {
        console.error(e);
        return {};
        // return { Authorization: await asyncLocalStorage.getItem("jwt") }
    }
  }
});```

mikeparisstuff:
This is interesting. I think I agree that we should use `.get(identityField)` instead of dot notation to allow for this use case. This will also require changing the default identityField from `username` to `cognito:username` which is how it shows up in the claims object.

mikeparisstuff:
Addressed in https://github.com/aws-amplify/amplify-cli/pull/354. @mkrn it would be great to get your eyes on it if you have a few minutes.

mkrn:
[...]

PR #354, opened 2018-10-25
Bug/340

mikeparisstuff:
*Issue #, if available:*

#340 

*Description of changes:*

This tweaks the behaviour of the "identityField" to allow users to use any attributes of the `$ctx.identity.claims` map. Before if I set the "identityField" to be "customField" it would work out to be `$ctx.identity.customField` and now it will be `$ctx.identity.claims.get("customField")`.

One side note. Before this change, the default value for identityField was **"username"** and is now **"cognito:username"** as that is how it appears in the claims object. I have inserted a check that if you provide the value **"username"**, it will be automatically aliased to **cognito:username** to prevent breaking existing projects. This is open for discussion.


mikeparisstuff:
I want to add a few more e2e tests before this is merged.

mikeparisstuff:
Ok, I have added the e2e tests. This should be gtg.

PR #367, opened 2018-10-29
Fixing GraphQL e2e tests

mikeparisstuff:
*Issue #, if available:*

N/A

*Description of changes:*

This fixes the graphql-e2e-tests. The tests failed when run in parallel because of a race condition on the local file system when writing resolvers to disk. Each test now has its own local directory to save resolvers in. This also removed a defective test.


mikeparisstuff:
This is incorrect. Closing then re-opening with the correct branch.

PR #380, opened 2018-10-31
Fix codegen to handle all permutations of LIST args

OtterFlip:
This is a tested fix for issues with the code-gen module so that it properly handles all four permutations of GraphQL LIST arguments, which are:

Nullable List Type: [] e.g. [String]
Nullable List of Non-null Types: [!] e.g . [String!]
Non-null List Type: []! e.g. [String]!
Non-null List of Non-null Types: [!]! e.g [String!]!

### DIFF
--- a/packages/amplify-graphql-docs-generator/src/generator/types.ts
+++ b/packages/amplify-graphql-docs-generator/src/generator/types.ts
@@ -37,6 +37,8 @@ export type GQLTemplateArgDeclaration = {
   name: string
   type: string
   isRequired: boolean
+  isList: boolean
+  isListRequired: boolean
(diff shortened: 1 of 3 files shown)


OtterFlip:
Out of curiosity, can anyone explain to me why the ci/circledci: test is failing?  It says "isRequired.ts[8, 3]: misplaced \'else\'\n\n',"     

The code is syntactically good, as far as I can tell, and I tested this code as well.  So is the linter just complaining about a style of bracing?

OtterFlip:
@yuth 

Okay, this PR is now passing the CI tests (tslint was picky about spacing/bracing/etc).

PR #384, opened 2018-11-01
Codegen angular

yuth:
*Issue #, if available:*

*Description of changes:*


packages/amplify-codegen/src/utils/getFrontEndFramework.js
```diff
@@ -0,0 +1,13 @@
+const getFrontendHandler = require('./getFrontEndHandler');
+
+function getFrontEndFramework(context) {
```
kaustavghosh06: Shouldn't it just be getFrontendJavascript framework? Seems like very specific to javascript.

packages/amplify-codegen/src/utils/getOutputFileName.js
```diff
@@ -1,18 +1,27 @@
 const path = require('path');
 
-function getOutputFileName(inputFileName = 'API', target) {
+function getOutputFileName(inputFileName, target) {
   const extensionMap = {
     swift: 'swift',
     typescript: 'ts',
     flow: 'js',
+    angular: 'service.ts',
```
kaustavghosh06: Is this the extension or the filename? the values for swift/typescript/flow etc looks like just extensions, unlike that for angular

packages/amplify-codegen/src/utils/getOutputFileName.js
```diff
@@ -1,18 +1,27 @@
...
   };
+  const folderMap = {
+    javascript: 'src',
+    typescript: 'src',
+    flow: 'src',
+    angular: 'src/app',
+  };
+
+  inputFileName = inputFileName || target === 'angular' ? 'api' : 'API';
```
kaustavghosh06: Can you explain what this is doing? and in what cases target = api vs API? This check seems a bit fragile.

packages/amplify-graphql-types-generator/CONTRIBUTING.md
```diff
@@ -0,0 +1,82 @@
+# Apollo Contributor Guide
```
kaustavghosh06: Should remove this file? Looks like apollo contributor guide. We have a contributing block in our main readme, so we can get rid of this.

packages/amplify-graphql-types-generator/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
kaustavghosh06: Remove this file. We have a License file on our mono repo whihc is Apcha Software and not MIT.

packages/amplify-graphql-types-generator/NOTICE
```diff
@@ -0,0 +1,13 @@
+AWS AppSync Codegen for iOS
```
kaustavghosh06: Can get rid fo this as well.

packages/amplify-graphql-types-generator/README.md
```diff
@@ -0,0 +1,226 @@
+
+
+
+
+# Amplify GraphQL types generator
```
kaustavghosh06: Is this file up to date with the latest changes to the cli?

packages/amplify-codegen/src/utils/getOutputFileName.js
yuth: Yes. Angular adds `service.ts` to Services

kaustavghosh06:
The licensing and author info should also be consistent with the other packages in the repo:

```
  "author": "Amazon Web Services",
  "license": "Apache-2.0",
```

packages/amplify-codegen/src/utils/getFrontEndFramework.js
yuth: I don't think this code really belongs here. Everything should move to frontend handler eventually. But in the interest of time I am keeping it here

packages/amplify-codegen/src/utils/getOutputFileName.js
yuth: For every other language other then angular we are using `API` as the base file name. For angular its lowercase. Also the first check is if the user has already specified the `inputFileName` we provide no default file name

packages/amplify-graphql-types-generator/LICENSE
yuth: Are these 2 compatible with one another?

packages/amplify-graphql-types-generator/NOTICE
yuth: OK

packages/amplify-graphql-types-generator/README.md
yuth: I think I can get rid of this all together

kaustavghosh06:
Ignore the licensing comments. Should retain the MIT license.

packages/amplify-graphql-types-generator/LICENSE
yuth: Removed

Issue #365, opened 2018-10-28
Tag Support


officeguy:
please add tag support to AWS Amplify cli, so that all resources can be tagged automatically at creation time with the tags you specify via a --tags argument

mtliendo:
If anyone has insight into how one would approach raising a PR against this, perhaps it would help drive development? For context, my company allows us devs to have a sandbox AWS account so long as the resources are tagged, thus having to figure out what resources were made or update a cloudformation file after the fact isn't the best DX.

tzdesign:
@mtliendo We ran into the same issues. We use aws-nuke to keep the costs low on testing. 

Please make this possible!

mtliendo:
@tzdesign Oh, I've never used that package before.  It's like rimraf for aws hehe.  My solution has been to keep a notepad of the repos I create that use amplify and when I feel they go idle run [`amplify delete`](https://github.com/aws-amplify/amplify-cli#commands-summary).  

In practice are there advantages that aws-nuke provides over that command?

tzdesign:
@mtliendo We are using a lot of Cloudformation and serveral team members are deploying with amplify, serverless, etc. It's impossible to keep track on each one of these things without tags.

It has no advantages except that it implements a workflow for testing and destruct stuff which is not in production. 

At the moment it's okay to nuke all inside the sandbox sub account, but in the end when it comes to cost consumption you will struggle without tags as well.

It's a feature of aws cloud formation, why not supporting it here.

BeneyKim:
Please, I do need this.

jbabe:
@kaustavghosh06 it would be awesome if this was added, right now we have environments that haven't been committed to `amplify/team-provider.json` (e.g. abandoned in a branch) that we'd like to periodically clean up.

mtliendo:
I can work on drafting an RFC.  From what I can tell from our use cases, the process would be the following:

1. When running `amplify init` prompt the user to enter their set of tags that would be used to _all_(?) created resources.

2. If the resource allows tagging (not all resources do), then update the necessary cloudformation templates to include the desired tags. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-resource-tags.html

Possibly 3: Allow the user to edit/add tags when running `amplify configure`.

It's been a while since this issue was first opened and the fact that it hasn't gone stale means there is continued interest.  Looking forward to all of your feedback to make sure we can get a more directed PR up soon.

[...]

PR #390, opened 2018-11-03
Fix #355 - Consolidating inquirer package versions to ^6.0.0

### LINKED ISSUE #355: amplify init stuck after choosing editor
**Describe the bug**

**To Reproduce**
amplify init  and then choose any editor ( doesnt matter which, result is same)

**Expected behavior**
obv init project, continue with other questions etc

**Screenshots**
https://i.imgur.com/wXyfkeO.png

 - OS:  Windows 10 x64
Node 11
```
amplify -v
0.1.29
```

attilah:
inquirer <5.0.0 is not functioning properly under windows, after first prompt the console hangs and 3.2.1 was used by various modules and the main cli too.

*Issue #355 

*Description of changes:*

* Update package versions to 6.0.0 so it is the same now across the whole repo
* Update the setup-dev-win npm script to be in sync with the non-windows version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "clean": "rm -rf node_modules && lerna clean --yes",
     "production-build": "lerna exec -- rm -f package-lock.json && lerna bootstrap --hoist && lerna run build",
     "setup-dev": "lerna exec -- rm -f package-lock.json && lerna bootstrap && cd packages/amplify-cli && rm -f -- package-lock.json && npm link && cd ../.. && lerna run build",
-    "setup-dev-win": "lerna bootstrap && cd packages/amplify-cli && del /f package-lock.json && npm link",
+    "setup-dev-win": "lerna exec -- del /f package-lock.json && lerna bootstrap && cd packages/amplify-cli && del /f package-lock.json && npm link && cd ../.. && lerna run build",
     "publish:master": "lerna publish --canary --yes --independent  --preid=alpha --message 'chore(release): Publish [ci skip]' --no-git-tag-version --no-push",
     "publish:beta": "lerna publish --conventional-commits --cd-version=prerelease --yes --independent --npm-tag=beta --preid=beta --message 'chore(release): Publish [ci skip]'",
     "publish:release": "lerna publish --conventional-commits --cd-version=patch --yes --independent --message 'chore(release): Publish [ci skip]'",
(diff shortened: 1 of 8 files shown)


kaustavghosh06:
@attilah Were you able to successfully get the  CLI to work with just the inquirer package change?
I was under the impression the issue was the Gluegun package.

attilah:
@kaustavghosh06 I think stuff is adding app with these diverse set of "undependent" dependencies ;-)

With this package update fix I was able to work my way through the full creation with customization options (naming everything 1 by 1) of the init.

I saw that there is a separate issue for 'push', then gluegun invokes a different command, I did not push my solution yet, but PR is not addressing that obviously.

I added the tech details of WHAT is causing the hang to the #365 issue itself.

kaustavghosh06:
@attilah  Thanks for the fix. 
I'm in touch with @jamonholmgren regarding the issue with gluegun and we're working towards fixing the issue with 'push'. I've outlined the main causes for it out here -https://github.com/infinitered/gluegun/issues/384.

attilah:
@kaustavghosh06 I tried to repro the mentioned issue infinitered/gluegun#384 with my latest sync of amplify-cli sources and when I patched the 2 push commands to just output 'push' and 'storage push'. I also added a 'hello storage' log to storage.js when they're executed and what I see was this:

```
> amplify push
push
> amplify storage push
storage push
> amplify storage
hello storage
and the amplify header + help was printed

So based on this, gluegun does its job and invoking the storage push command directly without calling storage.js first...

Can you confirm this or point to the issue in more detail?

**Comment by attilah** [comment]
I think this can be removed now as gluegun seems to work correctly:

```
    if (/^win/.test(process.platform)) {
      try {
        const { run } = require(`./${featureName}/${context.parameters.first}`);
        return run(context);
      } catch (e) {
        context.print.error('Command not found');
      }
    }
```

attilah:
@kaustavghosh06 I found an interesting issue. I've no storage resources configured for the project, just auth, and when I say ```amplify storage push``` it lists the table with no entries and ask for confirmation...but there is nothing to push. But if I say ```Y``` it pushes "successfully" the nothing.

kaustavghosh06:
@attilah That's the default behavior for the CLI (not related to the OS), but it doesn't provide any resources in the cloud when you execute this command.
I agree there should be a validation check out here and a message saying there are no resources to push.
From your other responses, looks like it works fine right now on Windows natively (same as Unix based OS's). 
I'll give it a try on my windows workspace right away and validate it.

kaustavghosh06:
@attilah  Just validated the usage on a windows VM using PowerShell and it WORKS!! :D  (npm install -g @aws-amplify/cli@0.2.0-alpha.323ce626).

Thanks for finding the root cause and providing a fix for the same :)
Will publish to stable version to NPM in the next couple of days.

PR #398, opened 2018-11-06
headless setup for Push and Publish

UnleashedMind:
*Issue #, if available:*

*Description of changes:*


packages/amplify-cli/src/lib/input-params-manager.js
```diff
@@ -2,7 +2,12 @@ function normalizeInputParams(context) {
   const inputParams = {};
   Object.keys(context.parameters.options).forEach((key) => {
     const normalizedKey = normalizeKey(key);
+<<<<<<< HEAD
```
kaustavghosh06: Need to remove this. (merge conflict)

packages/amplify-cli/src/commands/configure.js
```diff
@@ -25,8 +25,7 @@ module.exports = {
 };
 
 function constructExeInfo(context) {
```
kaustavghosh06: Should get the localEnvInfo out here as well?

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
```diff
@@ -438,6 +438,7 @@ function getCurrentConfig(context) {
         configLevel: 'general',
         config: {},
       };
+      fs.removeSync(configInfoFilePath);
```
kaustavghosh06: We don't need to remove this. Its not a single config now..there would be configs for several environments now. And it would cause problems (in other environments) if we remove the whole file.

PR #412, opened 2018-11-07
Add multienv support for hosting

UnleashedMind:
*Issue #, if available:*

*Description of changes:*


packages/amplify-category-hosting/lib/S3AndCloudFront/parameters.json
```diff
@@ -0,0 +1,4 @@
+{
+    "env": "NONE",
```
kaustavghosh06: You don't need to add this.

UnleashedMind: My thought was, when user sees 'env' parameter in the template file, he might search for the parameter in the parameters.json file, and without it, it can confuse the user.

PR #422, opened 2018-11-09
Multienv support in the transform.

mikeparisstuff:
*Description of changes:*

This adds support for multiple environments in the transform.

### DIFF
--- a/packages/graphql-dynamodb-transformer/src/resources.ts
+++ b/packages/graphql-dynamodb-transformer/src/resources.ts
@@ -160,11 +160,16 @@ export class ResourceFactory {
             RoleName: Fn.If(
                 ResourceConstants.CONDITIONS.HasEnvironmentParameter,
                 Fn.Join('-', [
-                    tableId, 'role',
-                    Fn.GetAtt(ResourceConstants.RESOURCES.GraphQLAPILogicalID, 'ApiId'),
-                    Fn.Ref(ResourceConstants.PARAMETERS.Env)
+                    tableId.slice(0, 21), // max of 64. 64-10-26-4-3 = 21
+                    'role', // 4
+                    Fn.GetAtt(ResourceConstants.RESOURCES.GraphQLAPILogicalID, 'ApiId'), // 26
+                    Fn.Ref(ResourceConstants.PARAMETERS.Env) // 10
                 ]),
-                Fn.Join('-', [tableId, 'role', Fn.GetAtt(ResourceConstants.RESOURCES.GraphQLAPILogicalID, 'ApiId')])
+                Fn.Join('-', [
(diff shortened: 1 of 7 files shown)


packages/graphql-dynamodb-transformer/src/resources.ts
```diff
@@ -145,7 +157,15 @@ export class ResourceFactory {
      */
     public makeIAMRole(tableId: string) {
         return new IAM.Role({
-            RoleName: Fn.Join('-', [tableId, 'role', Fn.GetAtt(ResourceConstants.RESOURCES.GraphQLAPILogicalID, 'ApiId')]),
+            RoleName: Fn.If(
+                ResourceConstants.CONDITIONS.HasEnvironmentParameter,
+                Fn.Join('-', [
```
kaustavghosh06: Doesn't roles have a limitation of 64 chars as well? This name seems to be fragile.

packages/graphql-dynamodb-transformer/src/resources.ts
```diff
@@ -160,7 +180,20 @@ export class ResourceFactory {
             },
             Policies: [
                 new IAM.Role.Policy({
-                    PolicyName: 'DynamoDBAccess',
+                    PolicyName: Fn.If(
```
kaustavghosh06: You don't need to modify policy names. Only role names can't clash. There can be policies with the same name under different IAM roles.

packages/graphql-elasticsearch-transformer/src/resources.ts
```diff
@@ -348,54 +334,37 @@ export class ResourceFactory {
...
-                "-",
+    private domainName() {
+        return Fn.If(
+            ResourceConstants.CONDITIONS.HasEnvironmentParameter,
+            Refs.NoValue,
+            Fn.Join(
+                '-',
                 [
-                    "d",
+                    'd',
```
kaustavghosh06: What is this for?

packages/graphql-dynamodb-transformer/src/resources.ts
mikeparisstuff: Makes sense. I'll revert this.

packages/graphql-elasticsearch-transformer/src/resources.ts
mikeparisstuff: The api id is 26 characters so this was just to give the cluster a slightly different name than api id.

packages/graphql-dynamodb-transformer/src/resources.ts
mikeparisstuff: I have limited the length of the table id.

PR #430, opened 2018-11-10
feat: added amplify env sync and amplify env checkout command & added help commands for env 

kaustavghosh06:

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/path-manager.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/path-manager.js
@@ -130,7 +130,14 @@ function getProviderInfoFilePath(projectPath) {
 
 function getBackendConfigFilePath(projectPath) {
   return path.normalize(path.join(
-    getDotConfigDirPath(projectPath),
+    getBackendDirPath(projectPath),
+    amplifyCLIConstants.BackendConfigFileName,
+  ));
+}
+
+function getCurrentBackendConfigFilePath(projectPath) {
+  return path.normalize(path.join(
+    getCurrentCloudBackendDirPath(projectPath),
     amplifyCLIConstants.BackendConfigFileName,
   ));
 }
@@ -173,4 +180,5 @@ module.exports = {
   getLocalEnvFilePath,
   getProviderInfoFilePath,
   getBackendConfigFilePath,
+  getCurrentBackendConfigFilePath,
 };
--- a/packages/amplify-provider-awscloudformation/lib/initialize-env.js
+++ b/packages/amplify-provider-awscloudformation/lib/initialize-env.js
@@ -9,13 +9,10 @@ function run(context, providerMetadata) {
   }
 
   const zipFilename = '#current-cloud-backend.zip';
-  const backendDir = context.amplify.pathManager.getBackendDirPath();
-  const tempDir = `${backendDir}/.temp`;
+  const amplifyDir = context.amplify.pathManager.getAmplifyDirPath();
+  const tempDir = `${amplifyDir}/.temp`;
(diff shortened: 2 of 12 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/path-manager.js
```diff
@@ -130,7 +130,14 @@ function getProviderInfoFilePath(projectPath) {
...
-    getDotConfigDirPath(projectPath),
+    getBackendDirPath(projectPath),
+    amplifyCLIConstants.BackendConfigFileName,
+  ));
+}
+
+function getCurrentBackendConfigFilePath(projectPath) {
+  return path.normalize(path.join(
+    getCurrentCloudBackendDirPath(projectPath),
     amplifyCLIConstants.BackendConfigFileName,
```
yuth: should it be starting with capital `B`. Anything that starts with capital letter gives me an impression that is an class that can be initialized. If its a constant, it better to have the name in capital letters

packages/amplify-provider-awscloudformation/lib/initialize-env.js
```diff
@@ -9,13 +9,10 @@ function run(context, providerMetadata) {
   }
 
   const zipFilename = '#current-cloud-backend.zip';
-  const backendDir = context.amplify.pathManager.getBackendDirPath();
-  const tempDir = `${backendDir}/.temp`;
+  const amplifyDir = context.amplify.pathManager.getAmplifyDirPath();
+  const tempDir = `${amplifyDir}/.temp`;
```
yuth: May be we should use [temp](https://www.npmjs.com/package/temp) module to generate temporary files and content instead of creating one in the amplify directory

kaustavghosh06: Yeah, we had a discussion regarding this a while back of using amplify dir vs using the users systems local tmp dir and decided to use the amplify dir due to variation in every OS's tmp/ dir location.

packages/amplify-cli/src/extensions/amplify-helpers/path-manager.js
kaustavghosh06: This is a higher level change. Right now we have this constants file with the fillowing info.

```
module.exports = {
  AmplifyCLIDirName: 'amplify',
  DotAmplifyDirName: '.amplify',
  DotConfigamplifyCLISubDirName: '.config',
  BackendamplifyCLISubDirName: 'backend',
  CurrentCloudBackendamplifyCLISubDirName: '#current-cloud-backend',
  PluginConfigFileName: 'plugin-config.json',
  ProjectConfigFileName: 'project-config.json',
  amplifyMetaFileName: 'amplify-meta.json',
  BrandName: 'AWS Amplify',
  CliName: 'amplify',
  DateTimeFormatString: 'YYYY-MM-DD-HH-mm-ss',
  DateTimeFormatStringCompact: 'YYYYMMDDHHmmss',
  DefaultAWSAccessKeyId: '<YOUR_ACCESS_KEY_ID>',
  DefaultAWSSecretAccessKey: '<YOUR_SECRET_ACCESS_KEY>',
  DefaultAWSRegion: 'us-east-1',
  LocalEnvFileName: 'local-env-info.json',
  ProviderInfoFileName: 'team-provider-info.json',
  BackendConfigFileName: 'backend-config.json',

};
```

We could decide as a team as to if we want to change all of these to caps since these are basically constants.

packages/amplify-provider-awscloudformation/lib/initialize-env.js
yuth: the module I linked handles all of that. Need not be now, but we should use this later just keep our directories clean

PR #445, opened 2018-11-13
feat(amplify-category-auth) support multienv for dependent services

yuth:
Auth can be added automatically by other categories like storage or interactions
as they are depenedent on auth. Updated auth add method to support multienv

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-category-auth/index.js
+++ b/packages/amplify-category-auth/index.js
@@ -3,7 +3,12 @@ const fs = require('fs');
 const _ = require('lodash');
 const uuid = require('uuid');
 const defaults = require('./provider-utils/awscloudformation/assets/cognito-defaults');
-const { updateConfigOnEnvInit } = require('./provider-utils/awscloudformation');
+const {
+  updateConfigOnEnvInit,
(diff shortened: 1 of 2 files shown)


packages/amplify-category-auth/index.js
```diff
@@ -109,38 +112,30 @@ async function externalAuthEnable(context, externalCategory, resourceName, requi
...
+      ENV_SPECIFIC_PARAMS,
+    );
+    if (!authExists) {
       const options = {
         service: 'Cognito',
         providerPlugin: 'awscloudformation',
       };
       await amplify.updateamplifyMetaAfterResourceAdd(category, authProps.resourceName, options);
     }
-    const action = context.updatingAuth ? 'updated' : 'added';
```
kaustavghosh06: this updating auth is defined somewhere I believe? Did you check dependency plugins setting it?

yuth: Its not set anywhere other then the update command for auth. Also auth should be updated if its already present or should be added if its not.

PR #448, opened 2018-11-14
fix: codegen headless fix & ask AWS profile info when info not present when checking out a new env

kaustavghosh06:
Codegen headless fix & ask AWS profile info when info is not present when checking out a new env

### DIFF
--- a/packages/amplify-cli/src/commands/env/checkout.js
+++ b/packages/amplify-cli/src/commands/env/checkout.js
@@ -1,5 +1,7 @@
 const fs = require('fs');
+const sequential = require('promise-sequential');
 const { initializeEnv } = require('../../lib/initialize-env');
+const { getProviderPlugins } = require('../../extensions/amplify-helpers/get-provider-plugins');
 
 module.exports = {
   name: 'checkout',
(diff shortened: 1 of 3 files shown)


packages/amplify-cli/src/commands/env/checkout.js
```diff
@@ -22,12 +24,40 @@ module.exports = {
...
+        allEnvs[envName][provider],
+      ));
+    });
+
+    await sequential(initializationTasks);
+
+    const onInitSuccessfulTasks = [];
+    context.exeInfo.projectConfig.providers.forEach((provider) => {
+      const providerModule = require(providerPlugins[provider]);
+      onInitSuccessfulTasks.push(() => providerModule.onInitSuccessful(
```
yuth: may be name it `afterInit`

yuth: How is this related to codegen headless?

kaustavghosh06: Take a look at the PR heading - fix: codegen headless fix & ask AWS profile info when info not present when checking out a new env

kaustavghosh06: Also onInitSuccessful is a function defined in providers and it’s used in multiple places - also in the provider from the beginning - not something introduced in this PR

yuth: May be these should be 2 sepearate PRs then. Just so that we can create the right change log

PR #455, opened 2018-11-14
 fix amplify env checkout to ask AWS profiles if its not present locally

kaustavghosh06:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-cli/src/commands/env/checkout.js
+++ b/packages/amplify-cli/src/commands/env/checkout.js
@@ -1,5 +1,7 @@
 const fs = require('fs');
+const sequential = require('promise-sequential');
 const { initializeEnv } = require('../../lib/initialize-env');
+const { getProviderPlugins } = require('../../extensions/amplify-helpers/get-provider-plugins');
 
 module.exports = {
   name: 'checkout',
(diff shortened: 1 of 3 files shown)


packages/amplify-cli/src/commands/env/checkout.js
```diff
@@ -22,12 +24,40 @@ module.exports = {
...
+        allEnvs[envName][provider],
+      ));
+    });
+
+    await sequential(initializationTasks);
+
+    const onInitSuccessfulTasks = [];
+    context.exeInfo.projectConfig.providers.forEach((provider) => {
+      const providerModule = require(providerPlugins[provider]);
+      onInitSuccessfulTasks.push(() => providerModule.onInitSuccessful(
```
yuth: `afterInit` or `postInitSuccess` would be a better name. Also we may want to run `afterInitFailure` just in case we want to give category plugins to clean things up

kaustavghosh06: I agree the name is a bit odd, but its the interface method name which the provider exposes and changing that would mean changing it in a bunch of places which is out of scope for this PR.

yuth: Can we put a todo and create an issue for that. Long term name should be changed

PR #458, opened 2018-11-15
feat: amplify env remove and ampify delete command for multi envs

kaustavghosh06:


packages/amplify-cli/src/commands/env/remove.js
```diff
@@ -1,16 +1,21 @@
...
 
 module.exports = {
   name: 'remove',
   run: async (context) => {
-    const envName = context.parameters.options.name;
+    const envName = context.parameters.first;
+    const currentEnv = context.amplify.getEnvInfo().envName;
+
     if (!envName) {
       context.print.error('You must pass in the name of the environment using the --name flag');
```
UnleashedMind: need to change this line because it's inconsistent with the help message.
if help message says 
`remove <env-name>`
It means there's no --name flag anymore, and the first parameter <env-name> is mandatory.

packages/amplify-cli/src/extensions/amplify-helpers/remove-env-from-cloud.js
```diff
@@ -0,0 +1,35 @@
...
+    throw e;
+  }
+
+  const categoryPlugins = getCategoryPlugins(context);
+  if (categoryPlugins.notifications) {
+    const notificationsModule = require(categoryPlugins.notifications);
+    await notificationsModule.deletePinpointApp(context);
+  }
+}
+
```
UnleashedMind: The logic for notifications deletePinpointApp is incorrect. 
Currently, that function will search the amplifyMeta and find the pinpoint App and delete it. 
For env specific delete of a pinpoint App, the pinpoint app should be searched in the team-provider-info.json. 
A new function for the notifications should be implemented and invoked here for env specific app delete.

kaustavghosh06: As discussed offline, I'll comment this out and create a stub out there and @UnleashedMind would take over the implementation of it.

packages/amplify-cli/src/commands/env/remove.js
kaustavghosh06: good catch. Will change.

PR #472, opened 2018-11-17
fix(amplify-category-auth):remove unsupported chars from auth defaults

yuth:
amplify lets users use the default settings while adding auth. These defaults make use of project
names. Auth uses cognito user pools which accepts have rules not to accept certain chars in the
name. Updating the code to replace unacceptable chars with underscore

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-category-auth/provider-utils/awscloudformation/assets/cognito-defaults.js
+++ b/packages/amplify-category-auth/provider-utils/awscloudformation/assets/cognito-defaults.js
@@ -99,7 +99,9 @@ const entityKeys = {
 };
 
 const getAllDefaults = (name) => {
-  const projectName = name.projectConfig ? `${name.projectConfig.projectName.toLowerCase()}${sharedId}` : name;
+  const disallowedChars = /[^A-Za-z0-9_]+/g;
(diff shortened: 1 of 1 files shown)


packages/amplify-category-auth/provider-utils/awscloudformation/assets/cognito-defaults.js
```diff
@@ -99,7 +99,9 @@ const entityKeys = {
 };
 
 const getAllDefaults = (name) => {
-  const projectName = name.projectConfig ? `${name.projectConfig.projectName.toLowerCase()}${sharedId}` : name;
+  const disallowedChars = /[^A-Za-z0-9_]+/g;
+  let projectName = name.projectConfig ? `${name.projectConfig.projectName.toLowerCase().substring(0, 100)}${sharedId}` : name.substring(0, 100);
```
kaustavghosh06: Why is it 0,100? The max number of chars you can have in a project name is 20.

yuth: Cognito has limitations of 100 characters. Since we are validating against that this should not hurt

PR #473, opened 2018-11-17
fix(amplify-category-auth): get env specific data in externalAuthEnable

yuth:
With muti enviornment category data is stored in both parameter.json and team-provider-info.json. The externalAuthEnable did not retrive data from team-provider-info.json as it was directly reading the data from the disk. Update the method to use util method to load the resource data 


packages/amplify-category-auth/provider-utils/awscloudformation/assets/cognito-defaults.js
```diff
@@ -99,7 +99,9 @@ const entityKeys = {
 };
 
 const getAllDefaults = (name) => {
-  const projectName = name.projectConfig ? `${name.projectConfig.projectName.toLowerCase()}${sharedId}` : name;
+  const disallowedChars = /[^A-Za-z0-9_]+/g;
```
kaustavghosh06: This doesn't belong to this PR.

yuth: This change is gone now after merging with upstream/multienv

PR #474, opened 2018-11-17
fix(amplify-category-notifications): fix redundant pinpoint creation

UnleashedMind:
*Issue #, if available:*

*Description of changes:*


packages/amplify-category-notifications/lib/channel-APNS.js
```diff
@@ -156,11 +156,11 @@ function pull(context, pinpointApp) {
     ApplicationId: pinpointApp.Id,
   };
   spinner.start(`Pulling ${channelName} Channel.`);
-  return new Promise((resolve, reject) => {
+  return new Promise((resolve, reject) => {//eslint-disable-line
```
yuth: why eslint disable line? If we keep on disabling linter where ever it fails to fix automatically there is no value in using it at all. We should avoid doing this

packages/amplify-category-notifications/lib/channel-APNS.js
```diff
@@ -156,11 +156,11 @@ function pull(context, pinpointApp) {
...
     context.exeInfo.pinpointClient.getApnsChannel(params, (err, data) => {
       if (err) {
-        spinner.fail(`get channel ${channelName} error`);
-        reject(err);
+        spinner.succeed(`no channel data retrieved for: ${channelName}`);
```
yuth: Why is this a success instead of err? Why are we resolving this?

packages/amplify-category-notifications/lib/channel-Email.js
```diff
@@ -128,11 +128,11 @@ function pull(context, pinpointApp) {
...
+  return new Promise((resolve, reject) => {//eslint-disable-line
     context.exeInfo.pinpointClient.getEmailChannel(params, (err, data) => {
       if (err) {
-        spinner.fail(`get channel ${channelName} error`);
-        reject(err);
+        spinner.succeed(`no channel data retrieved for: ${channelName}`);
+        resolve(err);
       } else {
         spinner.succeed(`get ${channelName} channel successful`);
         pinpointApp[channelName] = data.EmailChannelResponse;
```
yuth: The same code is repeated for all the channels. May be we can consolidate the logic to a single function and it take the response key as one of the params

UnleashedMind: The API method names are different for different channels, and also the name of the property we want to grab are different in the returned data for different channels.
I'm sure there are many 'smart' ways to write this, but explicit and simple coding style makes it easier to understand and maintain by all.

packages/amplify-category-notifications/lib/channel-APNS.js
UnleashedMind: In the Pinpoint API, if a channel has not been properly setup (has never been enabled), the get channel method errs. For our purpose, we just don't set the channel in the pinpointApp object and resolve it.

packages/amplify-category-notifications/lib/channel-APNS.js
UnleashedMind: A new Promise looks much better with both (resolve, reject) in it, but we don't need to reject in this, and eslint complains unused parameter.

packages/amplify-category-notifications/lib/channel-APNS.js
UnleashedMind: This part has been changed so be more precise on that err to ignore.

packages/amplify-category-notifications/lib/channel-APNS.js
yuth: 👍

PR #476, opened 2018-11-19
 feat: core metadata migration of projects using the old version of the CLI

kaustavghosh06:
* Files generated as a part of the migration process:
amplify/backend/backend-config.jsin
amplify/.config/local-aws-info.json
amplify/.config/local-env-info.json
amplify/team-provider-info.json

* Files deleted as a part of the migration process:
amplify/aws-info.json (which is replaced by amplify/.config/local-aws-info.json)

* Files modified as a part of the migration process:
amplify/.config/project-config.json

### DIFF
--- a/packages/amplify-cli/src/lib/migrate-project.js
+++ b/packages/amplify-cli/src/lib/migrate-project.js
@@ -1,6 +1,7 @@
 const fs = require('fs-extra');
 const path = require('path');
 const ora = require('ora');
+const os = require('os');
 
 const spinner = ora('');
 const { prompt } = require('gluegun/prompt');
@@ -12,6 +13,7 @@ const {
   getLocalEnvFilePath,
   getProviderInfoFilePath,
   getBackendConfigFilePath,
+  getGitIgnoreFilePath,
 } = require('../extensions/amplify-helpers/path-manager');
 
 async function migrateProject(plugins) {
@@ -32,6 +34,7 @@ async function migrateProject(plugins) {
       generateAwsLocalInfo();
       generateTeamProviderInfo();
       generateBackendConfig();
+      generateGitIgnoreFile();
     }
 
     // Give each category a chance to migrate their respective files
@@ -68,7 +71,7 @@ async function migrateProject(plugins) {
     await Promise.all(categoryMigrationTasks);
     spinner.succeed('Migrated your project successfully.');
   } catch (e) {
-    spinner.fail('There was an error migrating your environment.');
+    spinner.fail('There was an error migrating your project.');
     throw e;
   }
 }
@@ -156,6 +159,27 @@ function generateBackendConfig() {
   const jsonString = JSON.stringify(backendConfig, null, 4);
   const backendConfigFilePath = getBackendConfigFilePath();
   fs.writeFileSync(backendConfigFilePath, jsonString, 'utf8');
+}
+
+
+function generateGitIgnoreFile() {
+  const getGitIgnoreAppendString = () => {
+    const toAppend = `${os.EOL + os.EOL
+    }amplify/\\#current-cloud-backend${os.EOL
+    }amplify/.config/local-*${os.EOL
+    }amplify/backend/amplify-meta.json${os.EOL
+    }aws-exports.js${os.EOL
+    }awsconfiguration.json`;
+
+    return toAppend;
+  };
+
+  const gitIgnoreFilePath = getGitIgnoreFilePath();
+  if (fs.existsSync(gitIgnoreFilePath)) {
+    fs.appendFileSync(gitIgnoreFilePath, getGitIgnoreAppendString());
+  } else {
+    fs.writeFileSync(gitIgnoreFilePath, getGitIgnoreAppendString().trim());
+  }
 }
 
 
--- a/packages/amplify-cli/src/lib/migrate-project.js
+++ b/packages/amplify-cli/src/lib/migrate-project.js
@@ -16,6 +16,7 @@ const {
 } = require('../extensions/amplify-helpers/path-manager');
 
 const { getGitIgnoreBlob } = require('../extensions/amplify-helpers/get-git-ignore-blob');
+const { PROJECT_CONFIG_VERSION } = require('./constants');
 
 async function migrateProject(plugins) {
   try {
@@ -28,7 +29,7 @@ async function migrateProject(plugins) {
     }
 
     const projectConfig = JSON.parse(fs.readFileSync(projectConfigFilePath));
-    if (projectConfig.projectPath && await prompt.confirm('We detected the project was initialized using an older version of the CLI. Do you want to migrate the project, so that it is compatible with the latest version of the CLI?')) {
+    if (!projectConfig.version && await prompt.confirm('We detected the project was initialized using an older version of the CLI. Do you want to migrate the project, so that it is compatible with the latest version of the CLI?')) {
       // This is an older project & migration is needed
       generateNewProjectConfig(projectConfig, projectConfigFilePath);
       generateLocalEnvInfo(projectConfig);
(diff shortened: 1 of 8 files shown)


packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -0,0 +1,188 @@
...
+
+    // Give each category a chance to migrate their respective files
+
+    const categoryMigrationTasks = [];
+    const amplifyMetafilePath = getAmplifyMetaFilePath();
+    const amplifyMeta = JSON.parse(fs.readFileSync(amplifyMetafilePath));
+    const initializedCategories = Object.keys(amplifyMeta);
+
+    const categoryPlugins = {};
+    plugins.forEach((plugin) => {
```
yuth: Don't we have a method for this?

packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -0,0 +1,188 @@
...
+        categoryPlugins[categoryName] = plugin.directory;
+      }
+    });
+
+    const availableCategory = Object.keys(plugins).filter(key =>
+      initializedCategories.includes(key));
+
+    availableCategory.forEach((category) => {
+      try {
+        const { migrateCategoryFiles } = require(categoryPlugins[category]);
```
yuth: may be `migrateCategory`

packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -0,0 +1,188 @@
...
+
+  Object.assign(newProjectConfig, projectConfig);
+  // These attributes are now stores in amplify/.config/local-env-info.json
+  delete newProjectConfig.projectPath;
+  delete newProjectConfig.defaultEditor;
+
+  // Modify frontend handler
+  const frontendPluginPath = Object.keys(projectConfig.frontendHandler)[0];
+  const frontendPlugin = frontendPluginPath.split('/')[frontendPluginPath.split('/').length - 1];
+  const frontend = frontendPlugin.split('-')[frontendPlugin.split('-').length - 1];
```
yuth: We have this code in bunch of places where we split and get the plugin name. We should move this to a helper

packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -0,0 +1,188 @@
...
+
+  newProjectConfig.frontend = frontend;
+  delete newProjectConfig.frontendHandler;
+
+  // Modify provider handler
+  const provider = Object.keys(projectConfig.providers)[0];
```
yuth: Is this wise?

packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -0,0 +1,188 @@
...
+
+  const jsonString = JSON.stringify(newProjectConfig, null, 4);
+  fs.writeFileSync(projectConfigFilePath, jsonString, 'utf8');
+}
+
+function generateLocalEnvInfo(projectConfig) {
+  const envInfo = {
+    projectPath: projectConfig.projectPath,
+    defaultEditor: projectConfig.defaultEditor,
+    envName: 'NONE',
```
yuth: Should we ask the user what would they like to name this env instead of naming it none?

packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -0,0 +1,188 @@
...
+        backendConfig[category][resourceName] = {};
+        backendConfig[category][resourceName].service = amplifyMeta[category][resourceName].service;
+        backendConfig[category][resourceName].providerPlugin =
+        amplifyMeta[category][resourceName].providerPlugin;
+      });
+    }
+  });
+
+
+  const jsonString = JSON.stringify(backendConfig, null, 4);
```
yuth: I see these repeated, can we move this a function. I know its 2 lines but still its worth moving it in case if we have to add more checks

packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -0,0 +1,188 @@
...
+  const backendConfigFilePath = getBackendConfigFilePath();
+  fs.writeFileSync(backendConfigFilePath, jsonString, 'utf8');
+}
+
+
+function generateGitIgnoreFile() {
+  const getGitIgnoreAppendString = () => {
```
yuth: Isn't there a function already which generates `.gitIgnore`

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: There's no helper function. The helper functions use context.

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Planning to change this to migrateResourceFiles.

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Yes, there's no other provider other cloudformation. Checked npm.

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Don't think so. We've changed our CFN files accordingly.

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Yeah, but it uses context.

yuth: can we make it not to use the context

packages/amplify-cli/src/lib/migrate-project.js
yuth: Internally it would be none. But for the user it could be `prod` env or `dev` env

packages/amplify-cli/src/lib/migrate-project.js
yuth: For now 😉

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Yeah, we can. As you mentioned that this is just two lines, but don't have a strong opinion on this one.

packages/amplify-cli/src/lib/migrate-project.js
yuth: 👍

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Adding a TODO out here:

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Moving to a helper function.

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Adding a TODO here.

packages/amplify-cli/src/lib/migrate-project.js
kaustavghosh06: Will change the logic for this.

packages/amplify-cli/src/lib/migrate-project.js
UnleashedMind: This breaks the cli pluggability.

packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -0,0 +1,188 @@
...
+    let projectConfigFilePath;
+    try {
+      projectConfigFilePath = getProjectConfigFilePath();
+    } catch (e) {
+      // New project, hence not able to find the amplify dir
+      return;
+    }
+
+    const projectConfig = JSON.parse(fs.readFileSync(projectConfigFilePath));
+    if (projectConfig.projectPath && await prompt.confirm('We detected the project was initialized using an older version of the CLI. Do you want to migrate the project, so that it is compatible with the latest version of the CLI?')) {
```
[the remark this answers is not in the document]
yuth: I agree. Lets plug-in the version number in the `projectConfigFilePath` going forward. It need not be the version of the CLI itself, but the version of the config which we could update independently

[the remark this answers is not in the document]
kaustavghosh06: Yes, we'll add this. I've discussed with @yuth and we'll be adding a version number to the new CLI and would be using that. Right now, the projectPath is not present in the newer project-config.json files and is enough to detect old vs new version of the CLI's.

Issue #375, opened 2018-10-30
Add support for complex objects in the codegen generated swift code


yuth:
**Is your feature request related to a problem? Please describe.**
AWS Appsync SDK supports complex objects by passing `--add-s3-wrapper` flag to `generate` command. But the swift code generated by amplify codegen doesn't support complex types

**Describe the solution you'd like**
Ask users if they want to support complex types in the generated code

yuth:
The feature has been released in `0.1.36`

PR #478, opened 2018-11-20
feat(amplify-graphql-types-generator): detect complex types

yuth:
Types generator expected users to pass add-s3-wrapper param to CLI if they wanted to include the S3
Wrapper code when generating swift code. Changed the param to complex-object-support and accept
either yes, no or detect. When user passes detect, the types generator loops through the types used
and adds the S3 wrapper code if the types use S3 Object

#375

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-graphql-types-generator/src/cli.js
+++ b/packages/amplify-graphql-types-generator/src/cli.js
@@ -89,10 +89,11 @@ yargs
         default: true,
         type: 'boolean'
       },
-      "add-s3-wrapper": {
+      "complex-object-support": {
         demand: false,
-        describe: "Adds S3 wrapper code to the output",
-        default: false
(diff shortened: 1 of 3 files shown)


packages/amplify-graphql-types-generator/src/cli.js
```diff
@@ -118,7 +119,7 @@ yargs
         addS3Wrapper: argv["add-s3-wrapper"]
```
UnleashedMind: either change this line to or add a new line as: `complexObjectSupport: argv["complex-object-support"]`
You don't have a `complexObjectSupport` in the options object, which is referenced in the generate method if the target is swift.

yuth: Good catch. Fixed

PR #491, opened 2018-11-22
refactor(cli): refactor the migration workflow

UnleashedMind:
*Issue #, if available:*

*Description of changes:*


packages/amplify-cli/src/cli.js
```diff
@@ -16,10 +17,14 @@ async function run(argv) {
     .version() // provides default for version, v, --version, -v
     .create();
 
+  if (argv[2] !== MIGRATE) {
+    const yesFlag = argv.includes('-y') || argv.includes('--yes');
```
kaustavghosh06: I think providing a --yes flag support for this could be dangerous since this would bypass the critical questions we ask the users about the migration (since this involves a risk of them losing their current changes to files). For eg: If someone runs `amplify push --yes`, the migration shouldn't trigger automatically and the users need to answer the migration questions. We should be careful about this since its a special case.

packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -17,63 +20,147 @@ const {
...
+    projectPath = searchProjectRootPath();
   } catch (e) {
     // New project, hence not able to find the amplify dir
     return;
   }
 
+  const projectConfigFilePath = getProjectConfigFilePath(projectPath);
+  const projectConfig = JSON.parse(fs.readFileSync(projectConfigFilePath));
+  if (projectConfig.version !== constants.PROJECT_CONFIG_VERSION) {
+    if (context.parameters.options.y || context.parameters.options.yes ||
```
kaustavghosh06: Should remove this condition, based on comments made above.

packages/amplify-cli/src/cli.js
UnleashedMind: I see your point. 
The other side of this is that, without the yes flag, migration would normally need to be run with human interaction, and it becomes inconvenient to run it within a CICD script.

PR #502, opened 2018-11-24
fix(amplify-provider-awscloudformation): no region

MadsBogeskov:
*Issue #, if available:*
`amplify configure` would during the creation of an IAM role not get the right region. It would try to fetch it from `answers` which only contains the username at that point.

*Description of changes:*
Instead of fetching from `answers` it now fetches from the global `awsConfig`


MadsBogeskov:
Seems that the tests failed inside of CircleCI because of ` image cache not found on this host, downloading circleci/node:10.8` any admin that can look into this? :)

PR #511, opened 2018-11-27
Improves performance by lazy-loading modules

jamonholmgren:
When talking to the Amplify team, Amplify CLI's startup performance has been less than desirable. While there are probably some things to be done on Gluegun's side (see also: https://github.com/infinitered/gluegun/issues/391), I took a stab at moving as many `require` statements into function bodies as possible, to see if this improved startup performance.

From what I can tell, it has. Running `time amplify --help`:

*before*: ~2.079s
*after*: ~0.886s

That takes about 40% as long to run the same command -- over double the speed. Tests all pass.

_Note: While working on this, I didn't follow the commit strategy perfectly, so if you want to merge this it should probably be squashed and the commit message edited._

✅ By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.


packages/amplify-cli/src/extensions/amplify-helpers/path-manager.js
```diff
@@ -1,23 +1,21 @@
-const path = require('path');
-const fs = require('fs');
-const homedir = require('os').homedir();
-const amplifyCLIConstants = require('./constants.js');
```
jamonholmgren: This file I may have gotten a little carried away with the code duplication.

jamonholmgren:
Note that https://github.com/infinitered/gluegun/pull/436 and https://github.com/infinitered/gluegun/pull/437 will also improve startup performance by about 300ms.

kaustavghosh06:
@jamonholmgren Thanks for your help. We're rearchitecting our plugin-based architecture and we'll keep the lazy loading concept in mind for the same.

PR #514, opened 2018-11-27
feat: Add second level of messaging when migrating projects from old version to newer version

kaustavghosh06:
*Issue #, if available:*

*Description of changes:*


packages/amplify-cli/src/lib/migrate-project.js
```diff
@@ -39,7 +41,9 @@ async function migrateProject(context) {
     if (await prompt.confirm(confirmMigrateMessage)) {
     // Currently there are only two project configuration versions, so call this method directly
     // If more versions are involved, switch to apropriate migration method
-      await migrateFrom0To1(context, projectPath, projectConfig);
+      if(await prompt.confirm(secondConfirmMessage)) {
```
UnleashedMind: looks fine, just move this condition immediately after the first if condition to put the above two lines of comments immediately before the actual migration method is called.

Issue #613, opened 2018-12-20
'amplify init' ignores AWS profiles with role_arn


ovcharenko:
**Describe the bug**
`amplify init` ignores AWS profiles with role_arn.

**To Reproduce**
1. Run `amplify init`
2. During *awscloudformation* configuration there will be only profiles with credentials

**Expected behavior**
All supported by AWS CLI profiles should be listed.

**Desktop (please complete the following information):**
```
$ amplify -v
0.2.1-multienv.14
```

ovcharenko:
@UnleashedMind Does it support running on instance or container with IAM role? I tried with different credentials providers, but always getting during `amplify env checkout`:

```
 uncaughtException: Missing credentials in config
   ...
   execPath=/usr/local/bin/node
   version=v10.14.1
   argv=[
      /usr/local/bin/node
      /usr/local/bin/amplify
      env
      checkout
      master
   ]
   ...
   stack=[
     CredentialsError: Missing credentials in config
       at Object.openSync (fs.js:436:3)
       at Object.readFileSync (fs.js:341:35)
       at Object.readFileSync (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/node_modules/aws-sdk/lib/util.js:95:26)
       at IniLoader.parseFile (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/node_modules/aws-sdk/lib/shared-ini/ini-loader.js:7:40)
       at IniLoader.loadFrom (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/node_modules/aws-sdk/lib/shared-ini/ini-loader.js:56:46)
       at SharedIniFileCredentials.load (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/node_modules/aws-sdk/lib/credentials/shared_ini_file_credentials.js:110:41)
       at SharedIniFileCredentials.coalesceRefresh (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/node_modules/aws-sdk/lib/credentials.js:205:12)
       at SharedIniFileCredentials.refresh (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/node_modules/aws-sdk/lib/credentials/shared_ini_file_credentials.js:190:10)
[...]

PR #645, opened 2019-01-01
feat(amplify-provider-awscloudformation): assume role

UnleashedMind:
*Issue #, if available:*
#613 
*Description of changes:*
allows cli to assume IAM role to access aws resources, as described in [this doc](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-role.html#cli-configure-role-mfa)

Note that this implementation does not include using MFA

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/configuration-manager.js
+++ b/packages/amplify-provider-awscloudformation/lib/configuration-manager.js
@@ -168,9 +168,9 @@ async function promptForProjectConfig(context) {
   const { projectConfigInfo } = context;
 
   let availableProfiles = [];
-  const systemConfig = systemConfigManager.getNamedProfiles();
-  if (systemConfig) {
-    availableProfiles = Object.keys(systemConfig);
+  const namedProfiles = systemConfigManager.getNamedProfiles();
+  if (namedProfiles) {
+    availableProfiles = Object.keys(namedProfiles);
   }
 
   const useProfileConfirmation = {
--- a/packages/amplify-provider-awscloudformation/lib/configuration-manager.js
+++ b/packages/amplify-provider-awscloudformation/lib/configuration-manager.js
@@ -168,7 +168,7 @@ async function promptForProjectConfig(context) {
   const { projectConfigInfo } = context;
 
   let availableProfiles = [];
-  const systemConfig = systemConfigManager.getFullConfig();
+  const systemConfig = systemConfigManager.getNamedProfiles();
   if (systemConfig) {
     availableProfiles = Object.keys(systemConfig);
   }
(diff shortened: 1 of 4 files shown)


packages/amplify-provider-awscloudformation/lib/configuration-manager.js
```diff
@@ -357,44 +357,39 @@ function removeProjectConfig(context) {
...
-  const { config } = projectConfigInfo;
-  if (config.useProfile) {
-    return systemConfigManager.getProfile(config.profileName).region;
+  const profileName = process.env.AWS_PROFILE || 'default';
+  const profileRegion = systemConfigManager.getProfileRegion(profileName);
+  if (profileRegion) {
+    return profileRegion;
   }
-  return config.region;
+  return 'us-east-1';
```
kaustavghosh06: Why should it always return us-east-1?

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
```diff
@@ -357,44 +357,39 @@ function removeProjectConfig(context) {
...
-  }
+
   return awsClient;
 }
 
-function getRegion(projectConfigInfo) {
+function getRegion() {
   // For details of how aws region is set, check the following link
   // https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html
-  if (projectConfigInfo.configLevel === 'general') {
```
kaustavghosh06: Why is this if statement remove?

kaustavghosh06:
Looking at the PR right now, could you also make the corresponding documentation changes for this change?

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -54,76 +55,107 @@ function setProfile(awsConfig, profileName) {
...
-
-    awsConfigs = {};
 
     Object.keys(credentials).forEach((key) => {
-      const profileName = key.trim();
-      awsConfigs[profileName] = {
-        accessKeyId: credentials[key].aws_access_key_id,
-        secretAccessKey: credentials[key].aws_secret_access_key,
-      };
+      const keyName = key.trim();
```
kaustavghosh06: This variable could be more descriptive.

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -54,76 +55,107 @@ function setProfile(awsConfig, profileName) {
...
+    if (!isRoleSourceProfile && profileConfig.role_arn && profileConfig.source_profile) {
+      const roleCredentials =
+        await getRoleCredentials(profileConfig);
+      delete profileConfig.role_arn;
+      delete profileConfig.source_profile;
+      awsConfig = {
+        ...profileConfig,
+        ...roleCredentials,
+      };
+    } else {
```
kaustavghosh06: Not sure as to in which case this branch would get hit?

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
kaustavghosh06: If it's 'general', shouldn't the region be present in either the env variables or the default?

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
UnleashedMind: it is hit for regular profile that does not use roles.

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
UnleashedMind: removed default and throw error if region can not be resolved.

UnleashedMind: I refactored this part. 
I removed attach region flag for the method, and let the caller (aws-pinpoint.js) handle the specifics, this is now much cleaner.

PR #687, opened 2019-01-10
fix(amplify-provider-awscloudformation): fix profile loading

UnleashedMind:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/system-config-manager.js
+++ b/packages/amplify-provider-awscloudformation/lib/system-config-manager.js
@@ -69,13 +69,9 @@ async function getProfiledAwsConfig(profileName, isRoleSourceProfile) {
       };
     } else {
       const profileCredentials = getProfileCredentials(profileName);
-      const normalizeCredentials = {
-        accessKeyId: profileCredentials.aws_access_key_id,
-        secretAccessKey: profileCredentials.aws_secret_access_key,
-      };
       awsConfig = {
         ...profileConfig,
-        ...normalizeCredentials,
+        ...profileCredentials,
       };
     }
   } else {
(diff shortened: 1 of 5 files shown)


packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -92,10 +94,11 @@ async function getRoleCredentials(profileConfig) {
...
     ExternalId: profileConfig.external_id,
   }).promise();
-  // accessKeyId: data.Credentials.AccessKeyId,
-  // secretAccessKey: data.Credentials.SecretAccessKey,
-  // sessionToken: data.Credentials.SessionToken
-  return roleData.Credentials;
+  return {
+    accessKeyId: roleData.Credentials.AccessKeyId,
+    secretAccessKey: roleData.Credentials.SecretAccessKey,
+    sessionToken: roleData.Credentials.SessionToken,
```
kaustavghosh06: Is it okay if the sessionToken is undefined? Did you test that case?

UnleashedMind: no, it's not OK, I tested and it fails without the session token. 
but if the assumeRole is successful, there's always going to be data.Credentials.SessionToken.

kaustavghosh06: Okay, so in that case, if my profile doesn't assume an STS role, will the SDK calls fail?

UnleashedMind: If the profile does not assume role, the credentials will be plain old credential with key id and key, there's no session token. 
AWS checks if the credential is a temp credential or a normal credential, for temp credentials, it requires session token.

PR #688, opened 2019-01-10
fix(amplify-provider-awscloudformation): fix profile loading

UnleashedMind:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/system-config-manager.js
+++ b/packages/amplify-provider-awscloudformation/lib/system-config-manager.js
@@ -92,10 +92,12 @@ async function getRoleCredentials(profileConfig) {
     RoleSessionName: 'amplify',
     ExternalId: profileConfig.external_id,
   }).promise();
-  // accessKeyId: data.Credentials.AccessKeyId,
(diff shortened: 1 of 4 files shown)


kaustavghosh06:
@UnleashedMind Why doesn't this PR triggering the unit tests and build?

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -124,7 +126,10 @@ function getProfileCredentials(profileName) {
       }
     });
   }
-  return profileCredentials;
+  return {
+    accessKeyId: profileCredentials.aws_access_key_id,
+    secretAccessKey: profileCredentials.aws_secret_access_key,
+  };
```
kaustavghosh06: Why don't we want to return the region out here?

UnleashedMind: this part has been updated.
region is in the config file, and credentials can be in the config file or the credentials file, although it's recommended to store it in the credentials file.
the refactor allows both, and jus normalize the key.

Issue #708, opened 2019-01-14
`amplify init` sometimes excludes lambda dependencies for REST api


mrcoles:
**Describe the bug**

I have two lambda functions underneath a REST API in my project. One requires auth and the other doesn’t. I have discovered that when `amplify init` runs it generates an `amplify/backend/amplify-meta.json` file that is missing the second lambda function as a `dependsOn` item for my api.

Since the lambda function is not showing up as a dependsOn in this file, its name and arn are not available in the clouformation script and I ultimately get this error when I `amplify push` or build in the AWS Amplify console:

> UPDATE_FAILED               myapi                                          AWS::ApiGateway::RestApi   Mon Jan 14
> 2019 22:35:05 GMT+0000 (UTC) Errors found during import:  Unable to put integration on 'ANY' for resource at path '/webhooks': Invalid lambda function  Unable to put integration on 'ANY' for resource at path '/webhooks/{proxy+}': Invalid lambda function (Service: AmazonApiGateway; Status Code: 400; Error Code: BadRequestException; Request ID: a381e261-184c-11e9-898d-17093dff19de)

**To Reproduce**

I haven’t figured out exactly how to reproduce this yet, but I have a private repo that is reliably creating this issue. I’m trying to get an example that I can share as a public repo.

**Versions:**

Amplify 0.2.1-multienv.23

**Additional context**

I went ahead and hand-edited the `amplify-meta.json` file to add the additional `dependsOn` and it allowed me to do an `amplify push`, but it obviously still broke when it built on the AWS Amplify Console.

kaustavghosh06:
This is a bug with the `amplify update api` command when adding an an additional path and tyeing it to a lambda function. The operatioon doesn't update the `backend-config.json` file. Looking at it now and will provide a PR for this soon.

erodrig:
2021 and having this issue with a new non authenticated function added into an existing API

PR #722, opened 2019-01-16
fix: update dependsOn block when updating api

kaustavghosh06:
Fixes issue - https://github.com/aws-amplify/amplify-cli/issues/708

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/update-backend-config.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/update-backend-config.js
@@ -25,9 +25,6 @@ function updateBackendConfigDependsOn(category, resourceName, value) {
   } else if (!backendConfig[category][resourceName]) {
     backendConfig[category][resourceName] = {};
   }
-  if (!backendConfig[category][resourceName].dependsOn) {
-    backendConfig[category][resourceName].dependsOn = [];
-  }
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/update-backend-config.js
```diff
@@ -15,6 +15,25 @@ function updateBackendConfigAfterResourceAdd(category, resourceName, options) {
...
+  if (!backendConfig[category]) {
+    backendConfig[category] = {};
+    backendConfig[category][resourceName] = {};
+  } else if (!backendConfig[category][resourceName]) {
+    backendConfig[category][resourceName] = {};
+  }
+  if (!backendConfig[category][resourceName].dependsOn) {
+    backendConfig[category][resourceName].dependsOn = [];
+  }
+  backendConfig[category][resourceName].dependsOn = value;
```
UnleashedMind: When meta data file is updated, it's Object.assign, and here it's complete replacement? 
if you are going to assign `value` to the `dependOn` property, you don't need to check its existence, the above `if` statement is not needed.

kaustavghosh06: The value has all the dependsOn properties. Can remove the check.

Issue #705, opened 2019-01-14
Disastrous Throttling: Rate exceeded error


leantide:
**Describe the bug**
After using amplify-cli to add  several more AWS resources to my configuration, I observe amplify push will create and execute over 20 nested stacks and then have an error as it hits the CloudFormation throttling limit, possibly because of the concurrency and the rate of the CloudFormation service calls that amplify is executing. Console log says "uncaughtException: Rate exceeded" and "stack=[Throttling: Rate exceeded". Verified by AWS support that CloudFormation logs indicate the Throttling limit occurs.

Upon seeing this 'rate exceeded' response, can amplify slow its push efforts down instead of terminating on an error? 

**To Reproduce**
1. Make a sufficiently large backend using aws amplify (see my example)
2. amplify push

| Category | Resource name   | Operation | Provider plugin   |
| -------- | --------------- | --------- | ----------------- |
| Storage  | settings        | Create    | awscloudformation |
| Function | settings        | Create    | awscloudformation |
| Api      | settings        | Create    | awscloudformation |
| Auth     | cognito3efe9085 | No Change | awscloudformation |
| Hosting  | S3AndCloudFront | No Change | awscloudformation |
| Storage  | membership      | No Change | awscloudformation |
| Storage  | products        | No Change | awscloudformation |
| Storage  | releases        | No Change | awscloudformation |
| Storage  | features        | No Change | awscloudformation |
| Storage  | items           | No Change | awscloudformation |
| Storage  | authorization   | No Change | awscloudformation |
| Storage  | counters        | No Change | awscloudformation |
| Storage  | users           | No Change | awscloudformation |
| Storage  | files           | No Change | awscloudformation |
| Function | membership      | No Change | awscloudformation |
| Function | products        | No Change | awscloudformation |
| Function | releases        | No Change | awscloudformation |
| Function | features        | No Change | awscloudformation |
| Function | items           | No Change | awscloudformation |
| Api      | membership      | No Change | awscloudformation |
| Api      | products        | No Change | awscloudformation |
| Api      | releases        | No Change | awscloudformation |
| Api      | features        | No Change | awscloudformation |
| Api      | items           | No Change | awscloudformation |

**Expected behavior**
The backend is pushed without error

**Console log:**
[...]

PR #733, opened 2019-01-18
fix(amplify-provider-awscloudformation): batch cfn status polling

yuth:
Cloud formation provider polls cloud formation to get the list of events to show when pushing the
stack. When a project has more than 20 resources, this causes throtlling. Fixing this issue by
batching the polling and retrying

*Issue #, if available:*
#705
*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
@@ -117,7 +117,11 @@ class CloudFormation {
             nestedStackQueue = [...nestedStackQueue, ...newStacks];
           }
 
-          for (let i = 0; i < Math.min(CFN_MAX_CONCURRENT_REQUEST, nestedStackQueue.length); i++) {
+          for (
+            let i = 0;
+            i < Math.min(CFN_MAX_CONCURRENT_REQUEST, nestedStackQueue.length);
+            i += 1
+          ) {
             nestedStackEventPromises.push(self.getNestedStackEvents(nestedStackQueue[0]));
             if (nestedStackQueue.length > 1) {
               nestedStackQueue.shift();
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
@@ -7,22 +7,32 @@ const providerName = require('../../lib/constants').ProviderName;
 const columnify = require('columnify');
 const { formUserAgentParam } = require('./user-agent');
 
+const CFN_MAX_CONCURRENT_REQUEST = 15;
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -75,106 +88,102 @@ class CloudFormation {
...
-                // Just get the new events to display to the user
-
-                newEvents = _.differenceBy(newEvents, allShownEvents, 'EventId');
-                newEvents = _.uniqBy(newEvents, 'EventId');
-                showEvents(newEvents);
-                allShownEvents = allShownEvents.concat(newEvents);
-              });
           }
 
           let newEvents = stackEvents;
```
kaustavghosh06: Shouldn't there be an else out here?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -75,106 +88,102 @@ class CloudFormation {
...
-        nestedstackEvents = nestedstackEvents.filter(event =>
-          self.eventStartTime < new Date(event.Timestamp));
+        nestedstackEvents = nestedstackEvents
+          .filter(event => self.eventStartTime < new Date(event.Timestamp));
 
         return nestedstackEvents;
-      });
+      })
+      .catch(() =>
+        // ignore. We dont fail if we can't get the nested stack status
```
kaustavghosh06: I think we should have a check for the CFN errors. There might be other errors as well which will go uncaught and hard to debug in a production environment.

yuth: OK

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -7,22 +7,32 @@ const providerName = require('../../lib/constants').ProviderName;
...
 
+const CFN_MAX_CONCURRENT_REQUEST = 15;
+const CFN_POLL_TIME = 7000; // 7sec
+
+const CFN_SUCCESS_STATUS = [
+  'UPDATE_COMPLETE',
+  'CREATE_COMPLETE',
+  'DELETE_COMPLETE',
+  'DELETE_SKIPPED',
+];
```
UnleashedMind: Maybe rename it TERMINAL_STATUS and add all the statuses that do not need further checking?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
yuth: We don't need else block as we dedupe the events when showing

PR #784, opened 2019-01-29

slaat:

*Description of changes:*
I have added another global path, without the lib subfolder, for windows


slaat:
Done, see #797

Issue #768, opened 2019-01-25
How to upload S3Objects with GraphQL AppSync (iOS)


peterfennema:
** Which Category is your question related to? **
Uploading S3Objects with GraphQL AppSync and iOS
I am not able to store complex objects (in my case an image) to S3.

** What AWS Services are you utilizing? **
I use Amplify iOS with GraphQL Transform and Codegen to generate a Swift api for AppSync.

** Provide additional details e.g. code snippets **

The docs say:

> The GraphQL Transform, Amplify CLI, and Amplify Library make it simple to add complex object support with Amazon S3 to an application.

> The AppSync SDKs and Amplify library handle uploading the files to S3 transparently.

But I can not make it work. I do not even know if this is caused by poor documentation, or by lacking support for iOS, or by bugs. 
I will walk you through the steps I made and the issues I encountered.

My setup is best described by the output of the `amplify status` command:

```
| Category | Resource name   | Operation | Provider plugin   |
| -------- | --------------- | --------- | ----------------- |
| Api      | myservice  | No Change | awscloudformation |
| Auth     | mycognito | No Change | awscloudformation |
| Storage  | mystorage   | No Change | awscloudformation |
```

So far, so good. The api is working, I use the generated Swift api, my model objects are stored in DynamoDB as expected.
Authorisaton via user pools works too.

Here's my schema:

```
type User 
  @model   
  @auth(
    rules: [
      { 
        allow: owner, 
        ownerField: "id", 
        identityField: "sub", 
        mutations: [create, update, delete],
        queries: null
      }
    ]
  ) 
{
  id: ID!
  name: String
  about: String
  createdAt: String!
  updatedAt: String! 
}
```

Now I want to add an image property to the User. I follow the iOS tutorial and add an S3Object. The new schema:

```
type User 
  @model   
  @auth(
    rules: [
      { 
        allow: owner, 
        ownerField: "id", 
        identityField: "sub", 
        mutations: [create, update, delete],
        queries: null
      }
    ]
  ) 
{
  id: ID!
  name: String
  about: String
  createdAt: String!
  updatedAt: String! 
  image: S3Object
}

type S3Object {
    bucket: String!
    region: String!
    key: String!
}
```

After `amplify push` I have a new Swift api, and a new generated schema.
The generated schema now has an auto generated S3ObjectInput type

```
input S3ObjectInput {
	bucket: String!
	region: String!
	key: String!
}
```

[...]

PR #796, opened 2019-01-30
 feat: generate s3 fragment to support autoupload in Swift

yuth:
Generate s3 fragments to support complex object upload. Swift codegen requires the S3Object to be a fragment to automatically upload files to s3. Updating statement generator to render fragments when there is an S3 object

*Issue #, if available:*
re #768

### DIFF
--- a/packages/amplify-graphql-docs-generator/src/generator/getBody.ts
+++ b/packages/amplify-graphql-docs-generator/src/generator/getBody.ts
@@ -1,18 +1,19 @@
 import { GraphQLField, GraphQLSchema } from 'graphql'
 
 import getFields from './getFields'
-import { GQLTemplateOpBody, GQLTemplateArgInvocation, GQLTemplateField } from './types'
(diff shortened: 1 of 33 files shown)


packages/amplify-graphql-docs-generator/src/generator/getBody.ts
```diff
@@ -1,18 +1,19 @@
...
 
 import getFields from './getFields'
-import { GQLTemplateOpBody, GQLTemplateArgInvocation, GQLTemplateField } from './types'
+import { GQLTemplateOpBody, GQLTemplateArgInvocation, GQLTemplateField, GQLDocsGenOptions } from './types'
 
 export default function getBody(
   op: GraphQLField<any, any>,
   schema: GraphQLSchema,
-  maxDepth: number = 3
+  maxDepth: number = 3,
```
kaustavghosh06: isn't this variable now?

yuth: yes with default value 3

PR #797, opened 2019-01-30

slaat:

Description of changes:
I have added another global path, without the lib subfolder, for windows

Same changes, just for multienv as instructed by #784 


wizage:
I looked over this and why add another plugin search section
```
.plugins(globalNodeModulesDirPath, { matching: 'amplify-*', hidden: false })
.plugins(globalWindowsNodeModulesDirPath, { matching: 'amplify-*', hidden: false [})
```
when you can just add a simple check in the global-prefix file like so:
```
if (process.platform === 'win32') {
    return path.join(getNpmPrefix(), 'node_modules');
} else {
    return path.join(getNpmPrefix(), 'lib', 'node_modules');
}
```
This seems like an addition that can be avoided by adding the correct location straight to the global location.

wizage:
@slaat Any thoughts on this change?

slaat:
> @slaat Any thoughts on this change?

yes that makes way more sense. I will get to it ASAP

slaat:
@kaustavghosh06 what do you think? Any idea when this will be merged?

kaustavghosh06:
Hey @slaat,any reason why you got this closed. I was testing this PR and getting it to merge sometime this week.

slaat:
I was under the impression that it should be done in the master branch,
since the whole env branch has been merged to it. I was about to do it
again there and submit another pull request.

kaustavghosh06:
Yes, you're correct. Thanks!

slaat:
Here you go: #941

Issue #373, opened 2018-10-30
Amplify Push for Storage (Resource is not in the state stackUpdateComplete)


dougcpr:
** Which Category is your question related to? **
I am trying to add storage to my app with the Amplify CLI. I am following the instructions at https://aws-amplify.github.io/docs/js/storage.

** What AWS Services are you utilizing? **
I am using Amplify CLI v0.1.29.

** Provide additional details e.g. code snippets **
Douglass-MacBook-Pro:hobbit-tracker docooper$ **amplify push**
| Category | Resource name   | Operation | Provider plugin   |
| -------- | --------------- | --------- | ----------------- |
| Storage  | hobbitDB        | Create    | awscloudformation |
| Auth     | cognito509c6c77 | No Change | awscloudformation |
? Are you sure you want to continue? true
⠏ Updating resources in the cloud. This may take a few minutes...Error updating cloudformation stack
✖ An error occurred when pushing the resources to the cloud

**_Resource is not in the state stackUpdateComplete_**

What am I doing wrong?

kaustavghosh06:
@dougcpr  Could you provide the full Cloudformation stack trace?

dougcpr:
@kaustavghosh06 thank you for the quick reply.

I looked up the failed stacks and deleted stacks for storageHobbitDB wasn't in the complete or active stack.

<img width="930" alt="screen shot 2018-10-30 at 12 29 16 pm" src="https://user-images.githubusercontent.com/15065698/47737724-878cc480-dc3f-11e8-827a-fff842fd7909.png">

I had this log where the create failed for storage DB in my console.

UPDATE_ROLLBACK_IN_PROGRESS hobbit-20181028210916 AWS::CloudFormation::Stack Mon Oct 29 2018 21:34:12 GMT-0500 (Central Daylight Time) The following resource(s) failed to create: [storagehobbitDB]. The following resource(s) failed to update: [authcognito509c6c77].                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         
CREATE_FAILED               storagehobbitDB       AWS::CloudFormation::Stack Mon Oct 29 2018 21:34:11 GMT-0500 (Central Daylight Time) Parameters: [partitionKeyType, partitionKeyName, tableName] must have values                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                              
[...]

PR #803, opened 2019-02-01
Improvements to multi-envrionments based on feedback

kaustavghosh06:
* Added `amplify env add` command as a proxy instead of amplify init command (users complained of it being confusing)

* Changed the previous `amplify env add` command to `amplify env import` command to bring in an external cloudformation stack 

* Added further checks on the params needed for the `amplify env import` command.

### DIFF
--- a/packages/amplify-category-interactions/provider-utils/awscloudformation/cloudformation-templates/lex-cloudformation-template.json.ejs
+++ b/packages/amplify-category-interactions/provider-utils/awscloudformation/cloudformation-templates/lex-cloudformation-template.json.ejs
@@ -269,6 +269,7 @@
                                       {
                                         "Ref": "AWS::Region"
                                       },
+                                      ":",
                                       { "Fn::Select" : [ "4", { "Fn::Split": [":", {"Ref": "authRoleArn"}]}]},
                                       ":bot:",
                                       {
--- a/packages/amplify-category-interactions/provider-utils/awscloudformation/cloudformation-templates/lex-cloudformation-template.json.ejs
+++ b/packages/amplify-category-interactions/provider-utils/awscloudformation/cloudformation-templates/lex-cloudformation-template.json.ejs
@@ -269,7 +269,6 @@
                                       {
                                         "Ref": "AWS::Region"
                                       },
-                                      ":",
                                       { "Fn::Select" : [ "4", { "Fn::Split": [":", {"Ref": "authRoleArn"}]}]},
                                       ":bot:",
                                       {
--- a/packages/amplify-cli/src/commands/env/add.js
+++ b/packages/amplify-cli/src/commands/env/add.js
@@ -1,60 +1,10 @@
-const fs = require('fs-extra');
-const path = require('path');
+const { run } = require('../init');
+
+const initRun = run;
 
 module.exports = {
   name: 'add',
   run: async (context) => {
-    const envName = context.parameters.options.name;
-    const config = JSON.parse(context.parameters.options.config);
-    const awsInfo = JSON.parse(context.parameters.options.awsInfo);
-
-    if (!envName) {
-      context.print.error('You must pass in the name of the environment using the --name flag');
-      process.exit(1);
-    }
-
-    if (!config) {
-      context.print.error('You must pass in the configs of the environment using the --config flag');
-      process.exit(1);
-    }
-
-    let envFound = false;
-    const allEnvs = context.amplify.getEnvDetails();
-    Object.keys(allEnvs).forEach((env) => {
-      if (env === envName) {
-        envFound = true;
-      }
-    });
-
-    const addNewEnvConfig = () => {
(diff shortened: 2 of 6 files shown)


packages/amplify-category-interactions/provider-utils/awscloudformation/cloudformation-templates/lex-cloudformation-template.json.ejs
```diff
@@ -269,6 +269,7 @@
                                       {
                                         "Ref": "AWS::Region"
                                       },
+                                      ":",
```
yuth: Is this relevant to push command?

packages/amplify-cli/src/commands/env/add.js
```diff
@@ -1,60 +1,10 @@
-const fs = require('fs-extra');
-const path = require('path');
+const { run } = require('../init');
```
yuth: Do we ask the question if its an existing env when this command is run

packages/amplify-cli/src/commands/env/import.js
```diff
@@ -0,0 +1,70 @@
...
+        awsInfo = JSON.parse(context.parameters.options.awsInfo);
+      } catch (e) {
+        context.print.error('You must pass in the AWS credential info in an object format for intializating your environment using the --awsInfo flag');
+        process.exit(1);
+      }
+    }
+
+    let envFound = false;
+    const allEnvs = context.amplify.getEnvDetails();
+    Object.keys(allEnvs).forEach((env) => {
```
yuth: Nit: instead of using `forEach` you could use [`some`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) to do the search

packages/amplify-cli/src/commands/env/import.js
```diff
@@ -0,0 +1,70 @@
...
+      if (env === envName) {
+        envFound = true;
+      }
+    });
+
+    const addNewEnvConfig = () => {
```
yuth: This can be its own file so that it can be tested with unit test. Longer term we should move all the configuration related thing to a single module

packages/amplify-cli/src/commands/env/import.js
```diff
@@ -0,0 +1,70 @@
...
+      }
+
+      envAwsInfo[envName] = awsInfo;
+      jsonString = JSON.stringify(envAwsInfo, null, 4);
+      fs.writeFileSync(configInfoFilePath, jsonString, 'utf8');
+
+      context.print.success('Successfully added environment from your project');
+    };
+
+    if (envFound) {
```
yuth: Nit: we probably can simply the whole thing with 
```javascript
let shouldAdd = context.paramters.options.yes;
if (envFound && !context.paramters.options.yes) {
shouldAdd = await context.amplify.confirmPrompt.run('We found an environment with the same name. Do you want to overwrite existing enviornment config?')
}
if(shouldAdd) {
   addNewEnvConfig(); 
}

```

packages/amplify-cli/src/lib/init-steps/s0-analyzeProject.js
```diff
@@ -123,7 +123,7 @@ async function getEnvName(context) {
   const isEnvNameValid = (inputEnvName) => {
     let valid = true;
 
-    if (inputEnvName.length > 10 || inputEnvName.length < 2 || /[^a-zA-Z]/g.test(inputEnvName)) {
+    if (inputEnvName.length > 10 || inputEnvName.length < 2 || /[^a-z]/g.test(inputEnvName)) {
```
yuth: why only lower case support?

packages/amplify-cli/src/commands/env/add.js
kaustavghosh06: Yes. We do.

packages/amplify-cli/src/lib/init-steps/s0-analyzeProject.js
kaustavghosh06: because resource names like S3 don't support upper case. This is an artifact of this bug/issue -> https://github.com/aws-amplify/amplify-cli/issues/373#issuecomment-458689994

yuth: OK, Can we have a separate commit for this. So its easier to track

kaustavghosh06: Yes, there is a separate commit for it -> https://github.com/aws-amplify/amplify-cli/pull/803/commits/6c581481c083c3af6f4c7a2eab489755b92ec4b4

packages/amplify-category-interactions/provider-utils/awscloudformation/cloudformation-templates/lex-cloudformation-template.json.ejs
kaustavghosh06: Rebased.

PR #851, opened 2019-02-11
Limit to allow "development" environment name

joelvh:
**Issue #, if available:**

Limit on environment name doesn't allow for "development", which is a common environment name.

**Description of changes:**

Extended the limit on environment name from 10 to 11 characters.

### DIFF
--- a/packages/amplify-cli/src/lib/init-steps/s0-analyzeProject.js
+++ b/packages/amplify-cli/src/lib/init-steps/s0-analyzeProject.js
@@ -123,7 +123,7 @@ async function getEnvName(context) {
   const isEnvNameValid = (inputEnvName) => {
     let valid = true;
 
-    if (inputEnvName.length > 10 || inputEnvName.length < 2 || /[^a-z]/g.test(inputEnvName)) {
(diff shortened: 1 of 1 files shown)


packages/amplify-cli/src/lib/init-steps/s0-analyzeProject.js
```diff
@@ -123,7 +123,7 @@ async function getEnvName(context) {
   const isEnvNameValid = (inputEnvName) => {
     let valid = true;
 
-    if (inputEnvName.length > 10 || inputEnvName.length < 2 || /[^a-z]/g.test(inputEnvName)) {
+    if (inputEnvName.length > 11 || inputEnvName.length < 2 || /[^a-z]/g.test(inputEnvName)) {
```
kaustavghosh06: Thanks for the PR, but we have validation text in the code which says that "Environment name should be between 2 and 10 charecters", like in line 149. You might want to make corresponding changes out there as well.

joelvh:
Hi @kaustavghosh06 sorry for the late reply - I updated the validation messages. Can you re-open?

PR #889, opened 2019-02-18
fix(graphql-dynamodb-transformer): generate filters for connection

### LINKED ISSUE #865: Running amplify push after updating to updated CLI causes error;
**Describe the bug**
Running `amplify push` after updating to updated CLI causes error; 

**To Reproduce**
1. Run amplify push
2. Accept migration instructions
3. See error:

```
Migrating your API. This may take a few minutes.
× An error occured when migrating the API project.
× An error occurred when pushing the resources to the cloud

Cannot read property 's3TemplateURL' of undefined
```

**Expected behavior**
It works.

Would appreciate any ideas on how to solve this or get logs for your use.  Thank you!

yuth:
DyanmoDBTransformer generates filter input type if and only if queries are enbled for a type. This used to work fine
 when transformer did not support enum filters. ID, String, Int, Float are not unique to a schema and are generated all the time. With EnumFilter if the Type does not have queries enabled, these filter input types never gets generated  but they may gets referenced in connections queries. Updated code to generate Filters even when queries are disabled.

fixes #865

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
+++ b/packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
@@ -293,10 +293,9 @@ export class DynamoDBModelTransformer extends Transformer {
             const listResolver = this.resources.makeListResolver(def.name.value, listFieldNameOverride, ctx.getQueryTypeName())
             ctx.setResource(ResolverResourceIDs.DynamoDBListResolverResourceID(typeName), listResolver)
 
-            this.generateFilterInputs(ctx, def)
-
             queryFields.push(makeModelScanField(listResolver.Properties.FieldName, def.name.value))
         }
+        this.generateFilterInputs(ctx, def)
 
         ctx.addQueryFields(queryFields)
     }


packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -293,10 +293,9 @@ export class DynamoDBModelTransformer extends Transformer {
             const listResolver = this.resources.makeListResolver(def.name.value, listFieldNameOverride, ctx.getQueryTypeName())
             ctx.setResource(ResolverResourceIDs.DynamoDBListResolverResourceID(typeName), listResolver)
 
-            this.generateFilterInputs(ctx, def)
-
             queryFields.push(makeModelScanField(listResolver.Properties.FieldName, def.name.value))
         }
+        this.generateFilterInputs(ctx, def)
```
mikeparisstuff: Looks good. My only feedback, as we discussed, is that this will now always generate filter inputs even when they are not used. I think this is ok, but in the future we could refactor this code such that it is called from the @connection transformer to generate filter inputs when needed.

PR #917, opened 2019-02-21
feat(amplify-provider-awscloudformation): show CFN error when push fail 

yuth:
When push fails due to cloud formation errors show the error message from cloud formation in console and provide a link to failed stack events

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
@@ -8,6 +8,7 @@ const chalk = require('chalk');
 const providerName = require('../../lib/constants').ProviderName;
 const columnify = require('columnify');
 const { formUserAgentParam } = require('./user-agent');
+const { print } = require('gluegun/print');
 
 const CFN_MAX_CONCURRENT_REQUEST = 5;
 const CFN_POLL_TIME = 5 * 1000; // 5 secs wait to check if  new stacks are created by root stack
@@ -34,6 +35,7 @@ class CloudFormation {
     this.pollQueue = new BottleNeck({ minTime: 100, maxConcurrent: CFN_MAX_CONCURRENT_REQUEST });
     this.pollQueueStacks = [];
     this.stackEvents = [];
+    this.stackParentMap = {};
 
     return initializeAwsClient.then((awsItem) => {
       if (userAgentAction) {
@@ -104,7 +106,10 @@ class CloudFormation {
       .then((stackEvents) => {
         const uniqueEvents = getUniqueStacksEvents(stackEvents);
         const nestedStacks = filterNestedStacks(uniqueEvents);
-        nestedStacks.forEach(stackId => this.addToPollQueue(stackId));
+        nestedStacks.forEach((stackId) => {
+          this.stackParentMap[stackId] = stackName;
+          this.addToPollQueue(stackId);
+        });
         this.showNewEvents(stackEvents);
       })
       .catch((err) => {
@@ -394,20 +399,24 @@ function showErrorEvents(events) {
   if (events.length > 0) {
     console.log('\n\n');
     console.log(chalk.reset.inverse.red.bold('Following stacks failed '));
-    console.log(columnify(events, {
-      truncate: true,
-      config: {
-        ResourceStatusReason: {
-          minWidth: 40,
-        }
-      },
-      columns: [
-        'LogicalResourceId',
-        'ResourceType',
-        'ResourceStatusReason',
-      ],
-      showHeaders: false,
-    }));
+    const tableContent = [['Id', 'Type', 'Reason' ]];
+    events.forEach((ev) => {
+      const row = [
+        ev.PhysicalResourceId,
+        ev.ResourceType,
+        ev.ResourceStatusReason
+      ];
+      tableContent.push(row);
+    });
+
+    // console.log('\n\n\n\n Markdown')
+    // print.table(tableContent, { format: 'markdown' });
+
+    // console.log('\n\n\n\n Default')
+    // print.table(tableContent, { format: 'default' });
+
+    // console.log('\n\n\n\n lean')
+    print.table(tableContent, { format: 'lean' });
   }
 }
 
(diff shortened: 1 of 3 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -1,14 +1,17 @@
...
 const columnify = require('columnify');
+
+const aws = require('./aws.js');
+const S3 = require('./aws-s3');
+const providerName = require('../../lib/constants').ProviderName;
 const { formUserAgentParam } = require('./user-agent');
 
-const CFN_MAX_CONCURRENT_REQUEST = 15;
-const CFN_POLL_TIME = 7000; // 7sec
+const CFN_MAX_CONCURRENT_REQUEST = 5;
```
kaustavghosh06: Are we sure we want to change this to 5 from 15?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -1,14 +1,17 @@
...
+const CFN_POLL_TIME = 5 * 1000; // 5 secs wait to check if  new stacks are created by root stack
```
kaustavghosh06: Why change this from 7 to 5?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -73,99 +82,115 @@ class CloudFormation {
...
+    const { envName } = this.context.amplify.getEnvInfo();
+    const stackTrees = eventsWithFailure
+      .filter(stack => stack.ResourceType !== 'AWS::CloudFormation::Stack')
+      .map((event) => {
+        const err = [];
+        const resourceName = event.PhysicalResourceId.replace(`-${envName}`, '') || event.LogicalResourceId;
+        const cfnURL = getCFNConsoleLink(event, this.cfn);
+        err.push(`${chalk.bold('Resource Name:')} ${resourceName} (${event.ResourceType})`);
+        err.push(`${chalk.bold('Event Type:')} ${getStatusToErrorMsg(event.ResourceStatus)}`);
+        err.push(`${chalk.bold('Reson:')} ${event.ResourceStatusReason}`);
```
kaustavghosh06: typo: Reason

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -73,99 +82,115 @@ class CloudFormation {
...
+      });
+    });
+  }
 
-          let newEvents = stackEvents;
-          const nestedStackEventPromises = [];
+  generateFailedStackErrorMsgs(eventsWithFailure) {
+    const { envName } = this.context.amplify.getEnvInfo();
+    const stackTrees = eventsWithFailure
+      .filter(stack => stack.ResourceType !== 'AWS::CloudFormation::Stack')
```
kaustavghosh06: Why do we filter this?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -73,99 +82,115 @@ class CloudFormation {
...
+      .map((event) => {
+        const err = [];
+        const resourceName = event.PhysicalResourceId.replace(`-${envName}`, '') || event.LogicalResourceId;
```
kaustavghosh06: why do we tamper with the actual PhysicalResourceId?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
yuth: We are using bottleneck and it can support 5 concurrent requests as there is at least 100ms gap between these requests. Also I have reduced the poll time to 5 seconds instead of 7 seconds and we don't have to wait 5 seconds for queue to fire next round of requests. POLL TIME is now used only to read root stack and not the ones that are in the queue

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
yuth: We are only interested in showing the actual errors that we have on resources and not the stacks themselves.

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
yuth: We don't . Strings are immutable

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
yuth: Will fix

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
kaustavghosh06: So why are we replaceing envName with ''?

yuth: So when we print the resource name, it won't have `-env` appended to it

kaustavghosh06: What if the envName is NONE? and for Lex it is not `-envName` and it is `_envName`. i think we should avoid doing this replace operation.

yuth: It doesn't harm, if there is no env name

kaustavghosh06: I can think about cases (although edge cases) where this could show inconsistent results, one of them would be in case of lex and then one more that I could think of is when the resource name itself has a `-envName` in it's name.

yuth: Are there any other patterns? Or is it just `-` and `_`

kaustavghosh06: From the top of my head, I can think of these two but there could me more in the future

yuth: Accommodated both cases

PR #798, opened 2019-01-30
feat(amplify-provider-awscloudformation): transition to private bucket

wizage:
*Issue #, if available:* N/A

*Description of changes:*  
Transition from public bucket to private bucket with public objects. This is to follow AWS best practices around bucket policies.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #928, opened 2019-02-22
feat(amplify-provider-awscloudformation): transition to private bucket

wizage:
*Issue #, if available:* N/A

*Description of changes:*
Transition from public bucket to private bucket with public objects. This is to follow AWS best practices around bucket policies.

Rebase as asked for here: #798


UnleashedMind:
Could you please provide more information as why it should be configured like that? 
We were following this document [S3WebHosting](https://docs.aws.amazon.com/AmazonS3/latest/dev/WebsiteHosting.html) for the hosting category implementation, which stated that "This bucket must have public read access."

PR #941, opened 2019-02-25

slaat:

*Description of changes:*
I have added an OS check. If on windows, it returns the correct path (without lib subdirectory).


kaustavghosh06:
@wizage Looks good to you?

wizage:
@kaustavghosh06 Looks good to me!

PR #961, opened 2019-02-28
Increase GSI limit from 5 to 20, per the DynamoDB Docs

nagey:
To my [reading](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html#limits-secondary-indexes), GSI limit is 20, not 5 — also, this works for adding more indicies when changed thusly.


nagey:
to be fair, this wouldn't work the the ISO and Gov regions (is that something that will be changed?) — it could be addressed with a user flag asking the user if they are okay with the restriction.

Or, if the value is an active decision to go for least common denominator, then understandable, btu would be nice to have an override.

mikeparisstuff:
This looks good to me. DynamoDB has updated the limit to 20 so we shoudl as well.

PR #963, opened 2019-02-28
perf: speed up push

### LINKED ISSUE #914: Even with No Changes, have 10 minutes of 'Updating resources in the cloud. This may take a few minutes...'
**Enhancement**

In a project with over 20 amplify resources, an 'amplify push' will have about 10 minutes of 'Updating resources in the cloud. This may take a few minutes...' even when there are No Changes in the backend resources to actually be pushed. 

If there are a few backend changes, I see these being updated very quickly in the cloud and then I have the same 10-minute delay.

Similarly, an 'amplify publish' will have 10 minutes of 'Updating resources in the cloud. This may take a few minutes...' above the time it takes to build and publish the client.

I suspect the 10 minutes is associated with the exponential back off to avoid the Cloud Formation throttling limit with large concurrency. See https://github.com/aws-amplify/amplify-cli/issues/705.

Wondering if the 10 minute delay is expected for the exponential back off solution and if there is any other solution that may be more optimal.

yuth:
Make push performant by
1. Deleting older function builds which were retained in the dist folder of function category.
 With this change only the latest build package is maintained
2. Excluding build/dist/node_modules from getting included in the current-cloud snapshot. This decreases the size of the snapshot substantially (when functions are included in the project) and decreases the upload time
3. Bail quickly if there are no resources to be update

closes #914

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/build-resources.js
+++ b/packages/amplify-provider-awscloudformation/lib/build-resources.js
@@ -1,5 +1,4 @@
-const fs = require('fs');
-const fsExtra = require('fs-extra');
+const fs = require('fs-extra');
 const path = require('path');
 const moment = require('moment');
 const archiver = require('archiver');
@@ -97,10 +96,11 @@ function removeOutdatedPackage(currentBuildFile) {
       .readdirSync(distDir)
       .map(p => path.join(distDir, p))
       .filter(p => currentBuildFile !== p)
-      .map(p => fsExtra.remove(p));
+      .map(p => fs.remove(p));
     return Promise.all(deletePromises);
   } catch (e) {
     // nothing to do here
+    console.log(`Failed to clean up outdated packages ${e.message}`);
   }
 }
 
--- a/packages/amplify-provider-awscloudformation/lib/build-resources.js
+++ b/packages/amplify-provider-awscloudformation/lib/build-resources.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const fsExtra = require('fs-extra');
 const path = require('path');
 const moment = require('moment');
 const archiver = require('archiver');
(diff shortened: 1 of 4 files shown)


packages/amplify-provider-awscloudformation/lib/build-resources.js
```diff
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const fsExtra = require('fs-extra');
```
kaustavghosh06: Shouldn't we just use fs-extra and remove the fs import from the line above? fs-extra has all the fn's that are present in fs I beleive.

packages/amplify-provider-awscloudformation/lib/build-resources.js
```diff
@@ -84,6 +90,20 @@ function getSourceFiles(dir, ignoredDir) {
...
+  try {
+    const distDir = path.dirname(currentBuildFile);
+    const deletePromises = fs
+      .readdirSync(distDir)
+      .map(p => path.join(distDir, p))
+      .filter(p => currentBuildFile !== p)
+      .map(p => fsExtra.remove(p));
+    return Promise.all(deletePromises);
+  } catch (e) {
+    // nothing to do here
```
kaustavghosh06: Should probably show some error log out here? although not neccessarily throw it?

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -179,7 +179,14 @@ function storeCurrentCloudBackend(context) {
 
     const zip = archiver.create('zip', {});
     zip.pipe(output);
-    zip.directory(currentCloudBackendDir, false);
+    zip.glob(
+      '**',
+      {
+        cwd: currentCloudBackendDir,
+        ignore: ['*/*/build/**', '*/*/dist/**', '*/*/node_modules/**'],
+        dot: true,
```
kaustavghosh06: What's this for?

yuth: This includes all the files in the directory that are not build or dist or node_modules. This decreases the size of the zip file when a project has functions with lot of dependencies

packages/amplify-provider-awscloudformation/lib/build-resources.js
yuth: I don’t think the error message is useful as there is nothing user can do about it. Also not deleting doesn’t cause any problem except using disk space

kaustavghosh06: The user can manually delete it then maybe? It would also help us know what the issue is (if there's any). If we don't show anything, we would just silently fail and never know if someothing is going wrong

Issue #748, opened 2019-01-22
General plugin support for callback after resource push


wizage:
**Is your feature request related to a problem? Please describe.**
Currently today only 1st party plugins support callbacks at the end of the push of resources. These callbacks include: 

 - Codegen for GraphQL api endpoint
 - GraphQL endpoint and API Key if available
 - Generating the aws_exports file

Beyond these use cases it would be nice to add other callback functionality for example:
 - Configuring development environment (i.e. configuring OBS to the endpoints returned, generating code for plugin returns)
 - Displaying important endpoints/information
 - Ability to run queries and to kick off builds against the new endpoints 

**Describe the solution you'd like**
The ability to have a callback at the end of the `amplify.context.pushResources` to preform actions locally without needing to change the CLI code. This will also enable changes to be made to the first party plugins to utilize this callback.

**Describe alternatives you've considered**
 - Making a pull request to support my specific plugin but this isn't scalable for all plugin developers
 - Having the user run another command after push. A user can forget this and might use old endpoints or misconfigured applications

**Additional context**

wizage:
Added support on #976

PR #976, opened 2019-03-04
Returnhook

wizage:
*Issue #, if available:* #748 

*Description of changes:* Adds functionality for callbacks to be implemented by any plugin that amplify supports (both 3rd party and 1st party plugins can be supported with this). This just builds on top of the callback for frontend development and adds support for the categories.

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/on-category-outputs-change.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/on-category-outputs-change.js
@@ -15,8 +15,8 @@ function onCategoryOutputsChange(context) {
   const pluginNames = Object.keys(context.amplify.getCategoryPlugins(context));
   pluginNames.forEach((pluginName) => {
     const pluginInstance = context.amplify.getPluginInstance(context, pluginName);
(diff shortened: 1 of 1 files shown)


packages/amplify-category-hosting/lib/S3AndCloudFront/template.json
```diff
@@ -46,7 +46,7 @@
                         }      
                     ]
                 },
-                "AccessControl": "PublicRead",
```
kaustavghosh06: Don't think that this is a part of this PR? Make sure to rebase

wizage:
@kaustavghosh06 Fixed. Sorry about that.

derrick3192:
How does one access this hook when creating a plugin? Would be great to include it in here [Plugins](https://aws-amplify.github.io/docs/cli/plugins)

wizage:
@derrick3192 We are working on updating the docs but to use it you just need to add this code to your index.js file.
```
async function onAmplifyCategoryOutputChange(context) {
  //callback code
}
```

derrick3192:
Awesome! Appreciate the replly cheers.

PR #989, opened 2019-03-07
feat(graphql-dynamodb-transformer): add DyanmoDB point in time recovery

johnf:
*Description of changes:*
Adds an option to turn on point in time recovery for DyanmoDB.

### DIFF
--- a/packages/graphql-dynamodb-transformer/src/resources.ts
+++ b/packages/graphql-dynamodb-transformer/src/resources.ts
@@ -36,7 +36,15 @@ export class ResourceFactory {
                     'PAY_PER_REQUEST',
                     'PROVISIONED'
                 ]
-            })
+            }),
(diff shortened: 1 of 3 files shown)


johnf:
I would also be tempted to make this just on by default but added it as a feature for now

mikeparisstuff:
The code changes look good but it would be great to have an e2e test that covers this scenario. Since we have many e2e tests that include @model directives, can you change one of them such that it uses this extra flag in parameters.json? One such place is to pass the ShouldUsePointInTimeRecovery parameter inside the empty object on this line: https://github.com/yuth/amplify-cli/blob/9d19d819cf1f1f1452419465dc0069a2cc781b20/packages/graphql-transformers-e2e-tests/src/__tests__/DynamoDBModelTransformer.e2e.test.ts#L94.

johnf:
Done. Also ran the test and confirmed that the table is being created with PITR on.

mikeparisstuff:
Looks good!

Issue #1099, opened 2019-03-22
[Feature] Inject the GraphQL API Id Output to all the Lambdas


kstro21:
**Is your feature request related to a problem? Please describe.**
The GraphQL API Id is used to generate the DynamoDB table's name, is also used to generate the **export name** of some outputs of other stacks, for example, the `StreamArn` of a table. I've seen a few comments asking how to get the API Id from a Lambda like this one https://github.com/aws-amplify/amplify-cli/issues/1002#issuecomment-474714854 
Additionally, the **export name** of the GraphQL API Id Output is so random that it can not be referenced easily, see this comments https://github.com/aws-amplify/amplify-cli/issues/1002#issuecomment-471125831 https://github.com/aws-amplify/amplify-cli/issues/1002#issuecomment-472120075

**Describe the solution you'd like**
The generated template in `amplify\backend\awscloudformation\nested-cloudformation-stack.yml` should pass the GraphQL API Id Output to all the Lambdas as a parameter. Let's call the API `myamplifytest`, then, the resource identifier will be like `apimyamplifytest`, so, we can get the GraphQL API Id Output and pass as a parameter
```JSON
"functionmyfunction": {
    "Type": "AWS::CloudFormation::Stack",
    "DependsOn": ["apimyamplifytest"],
    "Properties": {
        "TemplateURL": "https://s3.amazonaws.com/myproject-20190306200305-deployment/amplify-cfn-templates/function/functionmyfunction-cloudformation-template.json",
        "Parameters": {
            "env": "dev",
            "GraphQLAPIId": {
                "Fn::GetAtt": ["apimyamplifytest", "Outputs.GraphQLAPIIdOutput"]
            }
        }
    }
},
```
Note that all the Lambdas then will need a `DependsOn` the resource that outputs the GraphQL API Id.  

**Additional context**
This will make it easy for people writing custom logic in the Lambdas and trying to get either the GraphQL API Id, a table name, `StreamArn` or any other output generated using the GraphQL API Id. Implementing this can be a way of dealing with #986

mikeparisstuff:
I like this idea. The endpoint would also be useful. @kaustavghosh06 can you look into what is needed to push it into the function category?

davidbiller:
> The generated template in `amplify\backend\awscloudformation\nested-cloudformation-stack.yml` should pass the GraphQL API Id Output to all the Lambdas as a parameter.

[...]

Issue #1047, opened 2019-03-15
[Feature] Export the Elasticseach Domain Endpoint and Arn in the SearchableStack


kstro21:
**Is your feature request related to a problem? Please describe.**
Currently, the **SearchableStack** has 0 outputs defined. I'm trying to use the Elasticsearch `DomainEndpoint` in a custom Lambda added using `amplify function add` but I had to wait for the service to be created, go to the AWS Elasticsearch Service to copy the value and hardcode it in the Lambda template. Not ideal when we can import the value.

**Describe the solution you'd like**
Add outputs to the **SearchableStack** and export the `DomainArn` and `DomainEndpoint`.

**Additional context**
I'm using **amplify 1.1.7**

mikeparisstuff:
Hey @kstro21. Yes, I don't see any issues with adding the Elasticsearch details to the exports and should be a relatively quick task. The relevant code is to add Outputs here: https://github.com/aws-amplify/amplify-cli/blob/0665508833fbedc0e1a787a90d8b72435fd71696/packages/graphql-elasticsearch-transformer/src/resources.ts#L83. I would be happy to review a PR if you would like to contribute and otherwise will add this to the backlog.

kstro21:
@mikeparisstuff that would be great. I'll be back with a PR. Thanks

kstro21:
Hey @mikeparisstuff. Well, it turns out not to be a quick task, at least for me. I've added the changes in the file you suggested and 2 files more. It looks good, but, the **outputs** are not written to the template `amplify\backend\api\amplifytest\build\stacks\SearchableStack.json` and instead to `amplify\backend\api\amplifytest\build\cloudformation-template.json`, additionally, the **output** is generated using `"Fn::ImportValue"` thus failing when running `amplify push`. Here is an example of what is generated.
```JSON
"ElasticsearchDomainArn": {
    "Description": "Elasticsearch instance Domain ARN.",
    "Value": {
        "Fn::ImportValue": {
            "Fn::Join": [
                ":",
                [
                    {
                        "Ref": "AppSyncApiId"
                    },
                    "GetAtt",
                    "ElasticSearchDomain",
                    "DomainArn"
                ]
            ]
        }
    },
    "Export": {
        "Name": {
            "Fn::Join": [
                ":",
                [
                    {
                        "Ref": "AppSyncApiId"
                    },
                    "GetAtt",
                    "Elasticsearch",
                    "DomainArn"
                ]
            ]
        }
[...]

PR #1215, opened 2019-04-05
feat(graphql-dynamodb-transformer): output table name

### LINKED ISSUE #1145: Add DynamoDB table name as an output for default model stacks
It would be very useful for creating custom resolvers if the table name for a given model were exported as part of the stack for the model. This would be especially useful when creating batch resolvers, which requires using the table name in the resolver templates. Right now one has to copy and paste a large chunk of CloudFormation JSON to determine the name.

I'd like to see each model CloudFormation stack export the DynamoDB table name like so:
```json
"GetAttModelTableName": {
    "Value": {
        "Fn::GetAtt": [
            "ModelTable",
            "Name"
        ]
    },
    "Export": {
        "Name": {
            "Fn::Join": [
                ":",
                [
                    {
                        "Ref": "AppSyncApiId"
                    },
                    "GetAtt",
                    "ModelTable",
                    "Name"
                ]
            ]
        }
    }
}
```

With this solution any stack could import the table name with the following JSON:
```json
"Fn::ImportValue": {
  "Fn::Join": [
    ":",
    [
      {
        "Ref": "AppSyncApiId"
      },
      "GetAtt",
      "ModelTable",
      "Name"
    ]
  ]
}
```

As far as I can tell the only alternative is to put this large chunk of Cloudformation JSON anywhere one needs the table name:
```json
{
  "Fn::If": [
    "HasEnvironmentParameter",
    {
      "Fn::Join": [
        "-",
        [
          "Model",
          {
            "Ref": "AppSyncApiId"
          },
          {
            "Ref": "env"
          }
        ]
      ]
    },
    {
      "Fn::Join": [
        "-",
        [
          "Model",
          {
            "Ref": "AppSyncApiId"
          }
        ]
      ]
    }
  ]
}
```

This is error prone since the default table naming scheme might change in the future.

**Additional context**
I would love to open a PR with this improvement, but I've searched for a while and cannot locate the code that is generating the current outputs. I found the table is generated here:

https://github.com/aws-amplify/amplify-cli/blob/master/packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts#L131

But I can't find where the current streamArn output is generated. If someone can point me in the right direction I'll prepare a PR.

Thanks!

kstro21:
*Issue #, if available:*
fix #1145

*Description of changes:*
Export the DynamoDB table name


mikeparisstuff:
We need to update this to handle situations where existing APIs from the old versions of the CLI are handled gracefully. I.E. if a stack mapping override was given for a DynamoDB table that causes it to remain in the root stack, then we should update the situation where the output for that stack is also in the root stack.

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -101,7 +101,7 @@ export class DynamoDBModelTransformer extends Transformer {
                 ".*" + def.name.value + "Model",
                 ".*" + def.name.value + "DataSource",
                 ".*" + def.name.value + "IAMRole",
-                "^" + def.name.value + "Table",
+                ".*" + def.name.value + "Table",
```
mikeparisstuff: Does this need to change?

packages/graphql-dynamodb-transformer/src/resources.ts
```diff
@@ -115,6 +115,16 @@ export class ResourceFactory {
         }
     }
 
+    public makeTableNameOutput(resourceId: string): Output {
```
mikeparisstuff: We need to make sure that this does not suffer the same issue as the ElasticSearchDomain output in situations where tables are hoisted to the root stack in migration situations.

kstro21: @mikeparisstuff you are right. Still working on the fix for #1047 then will apply it here if needed.

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
kstro21: In order to avoid breaking current stacks, the name of the output is `GetAtt${tableLogicalID}Name`, see line 144. Exist another way to map the output to the stack? Are you suggesting adding another **Stack Mapping** and avoid editing the one you pointed? Something like `"GetAtt" + def.name.value + "Table",`.

mikeparisstuff: Yeah, I would go with the more specific set so its more clear what is going where.

packages/graphql-dynamodb-transformer/src/resources.ts
mikeparisstuff: Thanks

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
kstro21: Ok, thanks for the suggestion.

kstro21:
I've added more specific mappings in order to support projects that were migrated from a previous version of the CLI.

mikeparisstuff:
Can you please try rebasing off of amplify-cli/master? I think that may solve the git commit issue.

kstro21:
_git commit issue_? Sorry, I don't get it. Besides, can you point me how to rebase off of amplify-cli/master?

mikeparisstuff:
> _git commit issue_? Sorry, I don't get it. Besides, can you point me how to rebase off of amplify-cli/master?

If you have the aws-amplify repo as a remote named "upstream" you would be able to run `git pull upstream master --rebase` from your local branch. This will replay your changes on top of the git tree in master and I think should make this look like a single commit again.

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -104,7 +104,9 @@ export class DynamoDBModelTransformer extends Transformer {
                 "^" + def.name.value + "Table",
                 // All resolvers except the search resolver.
                 "^[^S].*" + def.name.value + "Resolver",
-                "^" + def.name.value + ".+Resolver"
+                "^" + def.name.value + ".+Resolver",
+                def.name.value + "Table",
```
mikeparisstuff: This should be handled by the `"^" + def.name.value + "Table",` above. I believe you should be able to remove one of these and prefer the more specific rule as long as it does not interact with the migration issue.

kstro21: @mikeparisstuff you are right, but, the one that solves the issue with the migration is `def.name.value + "Table",`. I will make some tests to make sure that removing `"^" + def.name.value + "Table",` doesn't break anything that is within my reach.

kstro21:
> > _git commit issue_? Sorry, I don't get it. Besides, can you point me how to rebase off of amplify-cli/master?
> 
> If you have the aws-amplify repo as a remote named "upstream" you would be able to run `git pull upstream master --rebase` from your local branch. This will replay your changes on top of the git tree in master and I think should make this look like a single commit again.

Hi, @mikeparisstuff, one last question. The branch in my fork is named **dynamodb-name-output**, actually, it took me 2 commits for this PR, I don't know why you mention _"look like a single commit again"_, but I'm ok with the `rebase`. So, can I `rebase` directly from **master** into **dynamodb-name-output**? It won't break anything? Please, advice.

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
kstro21: @mikeparisstuff I've removed `"^" + def.name.value + "Table",` and confirmed everything works before and after migration from a previous version of the CLI.

achutkiran:
Hi @kstro21 can i use this to import table name in to lambda function cloudformation template

kstro21:
Hi, @achutkiran, sure, use something like this

```JSON
"Fn::ImportValue": "[GraphQLApiId]:GetAtt:TodoTable:Name"
```
`TodoTable` - replace it with yours table resource id.
`[GraphQLApiId]` - replace it with the value of your GraphQL API ID.

Hope it helps

achutkiran:
I am using multienv so for different  env I will have different graphqlapiid right?

kstro21:
@achutkiran yes, it will be different.

For the GraphQL API ID, there is a feature request #1099 open to pass the value to the Lambdas, so, until it is implemented, we need to hardcoded the GraphQL API ID in the template or use some other methods, for example, I use a script to extract the value from `amplify/backend/amplify-meta.json` and replace the string `[GraphQLApiId]` in the Lambdas' template.

achutkiran:
Ohk understood. Thanks

philipmclifton:
It would be awesome to get an example how to automate this easier. 

I have this in my function cf template:

```
	"organisation_table": {
							"Fn::Sub": "Organisation-XXXXXXXXX-${env}"
						},
```

Feels completely hacky

swatipraskar:
Is there an example of how to use this change so DynamoDb table name can be accessed in side lambda code?

droidgeek1:
@kstro21 @mikeparisstuff  and others,
Can someone outline what steps are needed in amplify project s i can access dynamodb table name in Lambda?
I have DataStore generated dynamoDB table and would like access to that in lambda code such that it is transparent when I clone environments etc. i.e. Hope to not have to use Env Variables as then I will need to add one every time I create new environment. Help appreciated. Good work here but need some specific guidance

Issue #1307, opened 2019-04-18
Got the "Cannot find the distribution folder."  in the Get started the react


haruharuharuby:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[official AWS Amplify Console forum](https://forums.aws.amazon.com/forum.jspa?forumID=314&start=0)

**Describe the bug**
it displays "Cannot find the distribution folder." when it runs 'amplify publish' in the Getting started of the react.

**To Reproduce**
do the Getting started of the react by selecting all default.
the error displays when it runs 'amplify publish'

**Expected behavior**
amplify publish command is to be succeeded.

**Screenshots**
https://drive.google.com/file/d/12XTIvm9seJCEtVb2qha7DVdxfmEkBDJ1/view

**Desktop (please complete the following information):**
 - OS: mac OS
 - Browser chrome
 - Version 1.6.3

**Additional context**
Cause of this issue is to differ the default folder between amplify hosting and amplify publish.
If the distribution folder set to 'dist',  all process drive correct.

UnleashedMind:
@haruharuharuby 
Is your application a react app? For react apps, the distribution folder is 'build'. 
The distribution folder is set during the execution of `amplify init`, and it can be later changed by running `amplify configure project`.

haruharuharuby:
@UnleashedMind Yes, I tried to make react app by Getting started document.  
When I executed 'amplify init', Prompt listens to me 'default' is 'build'. But　Infact, the name of distribution folder was 'dist'.
This was cause of the error.

So I guessed that correct name is 'dist'.

UnleashedMind:
if you use the `create-react-app` command line tool, the and run 'yarn build', the built artifacts are in the `build` directory. there's no `dist` directory for an react app created and managed by the `create-react-app` tool

haruharuharuby:
OK. 
I had not been useing the create-react-app. I will use create-react-app next time.
Thanks.

d3r3nic:
to fix this issue, run 
$ amplify configure project
then rather than setting the path of build to default build, set it to :
/amplify/backend/api/<appFolderName>/build
that will fix the issue

PR #1308, opened 2019-04-18
fix: (amplify init) change default destribution folder

haruharuharuby:
*Issue #1307, if available:*
#1307 

*Description of changes:*
set default distribution folder of the Getting started of react from build to dist.


UnleashedMind:
For react apps, the distribution folder is 'build', and since those are just sample scripts, the customer can change them to suite their specific development needs. Those files shouldn't be changed.

PR #1316, opened 2019-04-19
 fix(graphql-elasticsearch-transformer): fix outputs in wrong template

### LINKED ISSUE #1047: [Feature] Export the Elasticseach Domain Endpoint and Arn in the SearchableStack

kstro21:
*Issue #, if available:*
fix #1047 

*Description of changes:*
By adding more specific _Stack Mapping_ we prevent the outputs of the **SearchableStack** being generated in the **root** stack.


kstro21:
Hi, @mikeparisstuff, I merged **master** into my local branch **searchable-stack-outputs** that I've used before for another PR #1098. I don't know why github is listing all that commits when I just pushed 1 to fix the issue.

And yes, I was able to reproduce the issue commented in #1047 and confirmed that the changes in this PR solve it.

mikeparisstuff:
Great. I was able to pull your branch locally and verify that this does indeed fix the problem. 

Now that I think about it I'm wondering if it may be more robust to update `splitStacks.ts` such that outputs are automatically placed in the same stack as the thing that they reference. In the current implementation there would be a problem because the output names reference the `AppSyncApiId` parameter which is only available in nested stacks (the cause of the original bug in #1047). Food for thought in the future would be to remove that reference from the output names and then update splitStacks to move the outputs automatically. For now this looks good and should unblock people. Thanks @kstro21

kstro21:
Thanks, @mikeparisstuff for the comments. The update to `splitStacks.ts` regarding the **outputs** make sense. Probably would have prevented the bug in #1047 and probably will be my next PR :-)

PR #1338, opened 2019-04-23
fix(amplify-codegen): make codegen multienv aware

### LINKED ISSUE #1243: Is the codegen command multi-env aware?
codegen

amplify cli, appsync, amplify codegen

Is the codegen command multi-env aware? .graphqlconfig.yml does not change when I change envs.

yuth:
Amplify codegen relied graphql config file to figure out the AppSync API ID to download the
introspection schema. With the introduction of multienv, the API ID changes based on the current
environment. Updated codegen to use the API ID from amplify meta to make codegen multienv aware

fix #1243

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-codegen/src/commands/add.js
+++ b/packages/amplify-codegen/src/commands/add.js
@@ -80,13 +80,11 @@ async function add(context, apiId = null) {
     excludes: answer.excludePattern,
     schema,
     amplifyExtension: {
-      graphQLApiId: apiDetails.id,
       codeGenTarget: answer.target || '',
       generatedFileName: answer.generatedFileName || '',
       docsFilePath: answer.docsFilePath,
       region,
     },
-    endpoint: apiDetails.endpoint,
   };
 
   if (answer.maxDepth) {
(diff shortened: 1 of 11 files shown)


packages/amplify-codegen/src/commands/add.js
```diff
@@ -80,13 +80,11 @@ async function add(context, apiId = null) {
     excludes: answer.excludePattern,
     schema,
     amplifyExtension: {
-      graphQLApiId: apiDetails.id,
       codeGenTarget: answer.target || '',
       generatedFileName: answer.generatedFileName || '',
       docsFilePath: answer.docsFilePath,
       region,
     },
-    endpoint: apiDetails.endpoint,
```
kaustavghosh06: Did you use the endpoint anywhere in the statement or doc generation? or was it just the appID?

yuth: We don't use it anywhere. Even with the endpoint present user will have to do some custom headers to support the Auth for AppSync.

Issue #1323, opened 2019-04-21
amplify environment confusion


ultradox:
**Describe the bug**
I'm stuck in a rabbit hole... After setting up amplify multi-environment (dev and prod) and successfully doing `amplify push` from git branch dev as well as git branch prod to two seperate AWS accounts / environments, I tried switching between environments and also tearing down the production environment.

`amplify --version` 1.5.3

```
$ amplify env checkout prod
✔ Initialized provider successfully.
✖ There was an error initializing your environment.
2019-04-21T07:00:29.314Z - error: uncaughtException: Unexpected token export 

$ amplify status

Current Environment: prod

| Category | Resource name   | Operation | Provider plugin   |
| -------- | --------------- | --------- | ----------------- |
| Api      | mymedkiwi       | No Change | awscloudformation |
| Hosting  | S3AndCloudFront | No Change | awscloudformation |

GraphQL endpoint: https://xxx.appsync-api.ap-southeast-2.amazonaws.com/graphql
GraphQL API KEY: xxx
Hosting endpoint: http://xxxx-hostingbucket-prod.s3-website-ap-southeast-2.amazonaws.com

$ amplify delete
? Are you sure you want to continue?(This would delete all the environments of the project from the cloud and wipe out all the local amplify r
esource files) Yes

Deleting env:dev

Deleting env:prod
⠴ Deleting resources from the cloud. This may take a few minutes...
Error in deleting env:dev

Stack with id mymedkiwi-xxx does not exist
```

There's clearly some confusion with the configuration... as you can see above the terminal output states `Deleting env:dev` followed directly by `Deleting env:prod`. As you can see at the start when I try to checkout the prod environment I get "error initializing environment". The same error occurs when I try `amplify env checkout dev`.

**To Reproduce**
I'm not sure I know how to reproduce the error, but this is what I did to set up my two environments:
1. After testing my dev environment and happy to push to production, I git merge dev into master...
```
$ git checkout master
$ git pull origin master
$ git merge dev
Updating 370a42c..8062cae

$ amplify init

$ amplify init

Note: It is recommended to run this command from the root of your app directory
? Do you want to use an existing environment? No
? Enter a name for the environment prod
Using default provider awscloudformation

For more information on AWS Profiles, see:
https://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html

? Do you want to use an AWS profile? Yes
✔ Successfully created initial AWS cloud resources for deployments.
[...]

PR #1353, opened 2019-04-25
fix(amplify-provider-awscloudformation): support multiprofile delete

### LINKED ISSUE #978: Cannot remove environment when checked out in another one not in the same region
Hi there !
   When checked out in an environment by example in region eu-central-1, I cannot remove another one in another region. I get an error like this:

Stack with id xxxyyyyzzz-5132437 does not exist

In same region removing one from another one is ok

yuth:
Amplify CLI used to configure credential at the AWS SDK root level. With multi environment support,
while deleting envs the delete request run in parallel. Since there could only be on profile set at
the SDK level, this caused deletion to fail when env used different profiles. Moved the credetianls
per service instance to support deletion of multi env

fix #978 #1323

*Issue #, if available:*

*Description of changes:*


packages/amplify-provider-awscloudformation/lib/configuration-manager.js
```diff
@@ -424,6 +424,11 @@ function persistLocalEnvConfig(context) {
 }
 
 function getCurrentConfig(context) {
+  const { envName } = context.amplify.getEnvInfo();
+  return getConfigForEnv(context, envName);
+}
+
+function getConfigForEnv(context, envName) {
```
UnleashedMind: why are we splitting the getCurrentConfig method into two? while the envName argument is not used at all in the new method getConfigForEnv(.)

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
```diff
@@ -485,19 +490,24 @@ function removeProjectConfig(context) {
...
 
-async function loadConfiguration(context, awsClient) {
-  const projectConfigInfo = getCurrentConfig(context);
+async function loadConfiguration(context) {
+  const { envName } = context.amplify.getEnvInfo();
+  const config = await loadConfigurationForEnv(context, envName);
+  return config;
+}
+
+async function loadConfigurationForEnv(context, env) {
```
UnleashedMind: why adding the env argument if it is not used at all, env is read from the config info file directly.

packages/amplify-provider-awscloudformation/lib/delete-env.js
```diff
@@ -1,9 +1,10 @@
 const Cloudformation = require('../src/aws-utils/aws-cfn');
-
+const { loadConfigurationForEnv } = require('./configuration-manager');
 
 async function run(context, envName) {
-  return new Cloudformation(context)
-    .then(cfnItem => cfnItem.deleteResourceStack(envName));
+  const credentials = await loadConfigurationForEnv(context, envName);
```
UnleashedMind: this does not actually return the credentials associated with the specific env.

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
yuth: This was caused when rebasing from master

packages/amplify-provider-awscloudformation/lib/delete-env.js
yuth: We do return the config for env now

packages/amplify-provider-awscloudformation/lib/initializer.js
```diff
@@ -58,21 +58,18 @@ async function run(context) {
 
 async function getConfiguredAwsCfnClient(context) {
```
UnleashedMind: If this function does not return the configured aws client anymore, but only the aws config, its name should be appropriately updated to reflect its functionality.

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -242,6 +244,16 @@ function getProfileConfig(profileName) {
   return normalizeKeys(profileConfig);
 }
 
+function loadConfigFromPath(profilePath) {
```
UnleashedMind: This method is not needed, and should not be here, system-config-manager is only meant to  deal with system level configuration settings, such as contents in the ~/.aws/ folder and in the values in the environmental variables. 
This method is only used in the configuration-manager, in which is should be directly calling the new context.amplify.readJsonFile(.) to return aws configurations stored in the file.

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
```diff
@@ -485,19 +490,24 @@ function removeProjectConfig(context) {
...
   if (projectConfigInfo.configLevel === 'project') {
     const { config } = projectConfigInfo;
+    let awsConfig;
     if (config.useProfile) {
-      const awsConfig = await systemConfigManager.getProfiledAwsConfig(context, config.profileName);
-      awsClient.config.update(awsConfig);
+      awsConfig = await systemConfigManager.getProfiledAwsConfig(context, config.profileName);
     } else {
-      awsClient.config.loadFromPath(config.awsConfigFilePath);
+      awsConfig = systemConfigManager.loadConfigFromPath(config.awsConfigFilePath);
```
UnleashedMind: change this line to awsConfig =  context.amplify.readJsonFile(config.awsConfigFilePath);

yuth:
> LGTM. Did you test with profiles accross multiple accounts?

Yes. I tried multi account with multi region and it works

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
yuth: we do some additional check in there. I would prefer to keep this

PR #1368, opened 2019-04-30
Adds http headers for @http directive

GustavoEdinger:
*Description of changes:*

Adds support for http headers for the @http directive.

Usage:

```
type Comment {
        id: ID!
        content: String @http(
            url: "https://www.api.com/ping", 
            headers: [
                {
                    key: "X-Header", 
                    value: "X-Header-Value"
                }
            ]
        )
}
```

### DIFF
--- a/packages/graphql-http-transformer/src/HttpTransformer.ts
+++ b/packages/graphql-http-transformer/src/HttpTransformer.ts
@@ -158,7 +158,7 @@ export class HttpTransformer extends Transformer {
 
         let headers : HttpHeader[] = getDirectiveArgument(directive)("headers")
 
-        if (!Array.isArray(headers)) {
+        if (!headers || !Array.isArray(headers)) {
             headers = [];
         }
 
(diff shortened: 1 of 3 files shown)


kaustavghosh06:
@GustavoEdinger Does this work with multiple environments?

GustavoEdinger:
> @GustavoEdinger Does this work with multiple environments?

Just added =) #1409 


> This looks good. @GustavoEdinger Thanks for taking a look at it. I think we can build in more complex support for multi env in the future and have it work similar to the new @function directive. At this point, I think we should add this to the default list of transformers loaded by the CLI so users can start using it and offering better feedback. I can look at this when I have time but if you want to take a look the relevant code is here:
> 
> https://github.com/aws-amplify/amplify-cli/blob/cab24c0d05659b4f41cea8604cdca221f42706c6/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js#L209-L214

Yes, I just create PR for it #1410

PR #1422, opened 2019-05-09
Allow importing of graphql-docs-generator inside other projects

sebflipper:
*Description of changes:*

Currently `graphql-docs-generator` takes an input `schema.json` and will output a `graphql` file.

This change keeps this behaviour for the cli, but changes the main command so that it accepts an `IntrospectionQuery` object (instead of a schema path) and will return the output as a string.

This change allows the usage of `graphql-docs-generator` inside other projects (without the extra step of having to read/write files).


kaustavghosh06:
What is the shape of this Introspection query parameter? Could you provide an example of this usage? We would probably have to document this if we're support this. Could you add docs for the same out here - https://github.com/aws-amplify/docs

kaustavghosh06:
@sebflipper Would be great if you could provide more context about this PR.

sebflipper:
@kaustavghosh06 👍 I've fixed the type error in 6b7ecbc and I've added some context about how to use these new changes in 99c4ad0.

packages/amplify-graphql-docs-generator/src/index.ts
```diff
@@ -24,22 +24,17 @@ const FILE_EXTENSION_MAP = {
 }
 
 function generate(
-  schemaPath: string,
-  outputPath: string,
-  options: { separateFiles: boolean; language: string; maxDepth: number }
-): void {
+  schema: IntrospectionQuery,
```
yuth: The codegen plugin in expects the default export to accept the `schemaPath` and `outputPath`. Could you export a `cliFileHandler` and use that in the codegen 
https://github.com/aws-amplify/amplify-cli/blob/a61237ff51a26fbf93ee423b43a34d89c06acf57/packages/amplify-codegen/src/commands/statements.js#L3-L5

https://github.com/aws-amplify/amplify-cli/blob/a61237ff51a26fbf93ee423b43a34d89c06acf57/packages/amplify-codegen/src/commands/statements.js#L37-L45

sebflipper: 👍 Good spot, fixed in a87e413

PR #1440, opened 2019-05-11
fix(amplify-category-notifications): remove env name

### LINKED ISSUE #1372: backend-config.json has env specific stack name for notifications
**Describe the bug**
Notifications section of backend-config.json is being named with env instead of this dynamically happening during deploy.

Eg:

Checkout dev:
```
    "notifications": {
        "amplifyreactappdemoudAOX-dev": {
            "service": "Pinpoint",
            "channels": []
        }
    }
```

Checkout test:
```
    "notifications": {
        "amplifyreactappdemobrrHH-test": {
            "service": "Pinpoint",
            "channels": []
        }
    }
```
It's also changing some random string there too.  The rest of the file doesn't change when switching between environments.  This obviously is an issue given this file is managed under source control and shouldn't change just because we switch between environments.

UnleashedMind:
remove env name from the resource name in the notification category
in both the amplify-meta.json file and the backend-config.json file

projects configured by the previous versions will be automatically handled without customers' efforts. 

fix #1372

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-category-notifications/index.js
+++ b/packages/amplify-category-notifications/index.js
@@ -13,10 +13,6 @@ async function initEnv(context) {
   await multiEnvManager.initEnv(context);
 }
 
-async function initEnvPush(context) {
-  await multiEnvManager.initEnvPush(context);
-}
-
 async function migrate(context) {
   await multiEnvManager.migrate(context);
 }
@@ -25,7 +21,6 @@ module.exports = {
   console,
   deletePinpointAppForEnv,
   initEnv,
-  initEnvPush,
   migrate,
 };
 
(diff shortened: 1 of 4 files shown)


packages/amplify-category-notifications/index.js
```diff
@@ -25,7 +21,6 @@ module.exports = {
   console,
   deletePinpointAppForEnv,
   initEnv,
-  initEnvPush,
```
kaustavghosh06: Why is this removed? I beleive this would break a few things including migration?

UnleashedMind: that was an unneeded leftover from last year, we were talking about possibly separating initEnv locally and then push, but that's not the workflow we used.

PR #1462, opened 2019-05-15
feat: flow to add policies to access amplify resources from Lambda

kaustavghosh06:
*Issue #, if available:*

*Description of changes:*


packages/amplify-category-hosting/lib/category-manager.js
```diff
@@ -74,8 +76,54 @@ async function migrate(context) {
   await sequential(migrationTasks);
 }
 
+
+function getIAMPolicies(resourceName, crudOptions) {
+  let policy = {};
+  let actions = new Set();
+
+  crudOptions.forEach((crudOption) => {
+    switch (crudOption) {
```
haverchuck: In the storage category we have 'create/update' option instead of both.  I think it makes sense because it's illustrative of what's actually happening for the user, even though it's at the expense of slight inconsistency.   Not a blocker for merging this but something that I think should be considered.

packages/amplify-category-storage/provider-utils/awscloudformation/service-walkthroughs/s3-walkthrough.js
```diff
@@ -384,5 +384,50 @@ function convertToCRUD(parameters, answers) {
   }
 }
 
+function getIAMPolicies(resourceName, crudOptions) {
+  let policy = {};
+  let actions = new Set();
 
-module.exports = { addWalkthrough, updateWalkthrough, migrate };
+  crudOptions.forEach((crudOption) => {
```
haverchuck: In askReadWrite function, we have 'create/update' instead of 'create' and 'update'.

PR #1483, opened 2019-05-18
fix(amplify-graphql-types-generator): set AWSTimestamp to number type

### LINKED ISSUE #1348: How to use the AWSTimestamp scalar
When I `amplify push` a schema with the following model and hit yes to also generate my mutations I get a conflicting type.

`type Swipe
  @model
  @auth(rules: [{allow: owner}])
{
  id: ID!
  userId: ID!
  recipe: Recipe! @connection
  liked: Boolean!
  timestamp: AWSTimestamp
}`

My mutation expect a timestamp in form of a string, but graphQL want an Integer.

This works but is ugly and typescript linter is complaining here:
`timestamp: Math.round(new Date().getTime() / 1000)`

What is the proper way to send a AWSTimestamp value in my mutations ?

yuth:
AWSTimestamp is serialized to number type. Updating codegen to use number type instead of string

fix #1348

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-graphql-types-generator/src/flow/types.js
+++ b/packages/amplify-graphql-types-generator/src/flow/types.js
@@ -27,6 +27,10 @@ const builtInScalarMap = {
   [GraphQLID.name]: 'string',
 }
 
+const appSyncScalars = {
+  AWSTimestamp: 'number',
+};
+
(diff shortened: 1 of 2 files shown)


yuth:
> LGTM. This shouldn't break exisiting customers right?

They should already be broken. This would fix that

PR #1199, opened 2019-04-03
Add project name to cognito user & identity pool name

dabit3:
*Issue #, if available:*
Issue #389

*Description of changes:*
Updated cognito project names to reflect names of client project.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1492, opened 2019-05-20
fix(amplify-category-auth): add policy name char length limit

yuth:
In pr #1199 the name of coginto userpool was changed to include the project name and a random
number. We use the this to name to the resource and - char is not supported in resource name.
Updated the resource name not to include -. Updated the policyname to use substring of resource name
as there is a limit of 65 chars in policy name

*Issue #, if available:*

*Description of changes:*


yuth:
> Approved, assuming that you have validated that previously-created resources are not being overwritten.

These values are stored in `parameters.json` and we don't generate them unless its adding of new resource

Issue #1485, opened 2019-05-19
Multi-auth support for GraphQL Transformer


potatopotaro:
** Which Category is your question related to? **
GraphQL Transformer

Hey fellas, with the new multi-auth support for Appsync, I was wondering what the ETA will be like for supporting this in the Amplify CLI. I've seen comments suggesting at least 1+ month(s), and the blog post states that support is being worked on.

If it is indeed the case that it'll be 1+ month(s)...

- Is there a hacky solution we can use for multi-auth right now?
- I assume editing our Amplify CLI generated Appsync from the Appsync console (or cloudformation, etc.) would be overwritten with the next `amplify push` correct?
- I'm guessing the best bet is to transform our GraphQL code, copy-paste the generated build code into a new Appsync project created outside of the Amplify CLI, and replace/drop-in the new multi-auth directives (e.g. @aws_api_key)?

Apologies if there's already an ongoing discussion elsewhere, on this specific subject. I couldn't find anything. Multi-auth support would be huge for my project, as currently, I'm using a very slow junction lambda function to make "unauthorized" requests.

kaustavghosh06:
@potatopotaro We're actively working on supporting multi-auth as a part of the GraphQL Transformer and I'll have an update this thread as soon as we release this feature.

kaustavghosh06:
We launched multi-auth support for AppSync API (which included public APIs - with API Keys + IAM) as a part of our CLI version 3.7+.
Please take a look at our documentation around it out here - https://aws-amplify.github.io/docs/cli-toolchain/graphql#public-authorization

PR #1524, opened 2019-05-27
[WIP] Add support for new multi-auth AppSync directives

attilah:
*Issue #, if available:*
Enables part of #1485

*Description of changes:*
Enable the usage of new AppSync new authorization directives in the GraphQL schema.

### DIFF
--- a/packages/graphql-transformers-e2e-tests/src/__tests__/CustomRoots.e2e.test.ts
+++ b/packages/graphql-transformers-e2e-tests/src/__tests__/CustomRoots.e2e.test.ts
@@ -1,6 +1,6 @@
 import {
     ObjectTypeDefinitionNode, parse, FieldDefinitionNode, DocumentNode,
-    DefinitionNode, Kind, InputObjectTypeDefinitionNode
+    DefinitionNode, Kind, InputObjectTypeDefinitionNode, ListValueNode, ValueNode, StringValueNode
 } from 'graphql'
 import GraphQLTransform from 'graphql-transformer-core'
 import DynamoDBModelTransformer from 'graphql-dynamodb-transformer'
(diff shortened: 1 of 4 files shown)


packages/graphql-transformers-e2e-tests/src/__tests__/CustomRoots.e2e.test.ts
```diff
@@ -128,6 +128,11 @@ test('Test custom root query, mutation, and subscriptions.', () => {
 
         authedField: String
             @aws_auth(cognito_groups: ["Bloggers", "Readers"])
```
mikeparisstuff: I don't believe this test is actually deployed through CFN and I believe this would fail since the server has validation that does not allow @aws_auth and @aws_cognito_user_pools to be used at the same time. Can we move multi auth tests to their own e2e test file that actually deploys to CFN so we can test the full e2e flow?

mikeparisstuff:
This looks good as it relates to allowing the pass through but here is a question. Are these directives usable when there is currently no support to configure the "AdditionalAuthProviders" property on the GraphQLApi resource? If the directives are not really usable until we can make the updates that allow you configure the additional auth providers then I say we just wait and roll these out together.

attilah:
Based on the comments/discussion I'm adding the additional functionality for multi auth to this PR.

janhesters:
Will we fine grained control be supported?

E.g. allow Read and Write access (Queries and Mutations) to `I_AM` and Read and Subscriptions (Queries and Subscriptions) to Cognito User Pools?

houmark:
Any updates on this? Would be great to get this merged soon...

attilah:
@houmark this is actively worked on, bear with us please ;-)

houmark:
Thanks for the update @attilah. I tried using the workarounds mentioned elsewhere by other AWS devs, by adding the @aws_iam directly in the AppSync schema, but I hit not only issues with my Cognito Pool not having access and the fact that one has to remember to add the @aws_iam in the cloud is a complete showstopper for a production app. Amplify Console runs automatically and will overwrite if there are schema changes — and even if you try to remember updating this manually, you'll still have seconds or minutes of downtime in parts or the entire app.

In the end, I gave up and went back to my stable Cognito User Pool systems user which I am already using in several Lambdas.

Will this also propagate the access to all related Connections etc. automatically or is the dev responsible for adding the directive in all places?

attilah:
Closing in favor of #1916

ggmon:
any update on this, its is still an issue, @aws_api_key direction on schema.graphql, doesnt seem help

Issue #1403, opened 2019-05-06
amplify api add-graphql-datasource ⠸ Fetching Aurora Serverless cluster...BadRequestException: This API is deprecated and not available. Use ExecuteStatement API instead


oste:
**Describe the bug**
Running `amplify api add-graphql-datasource` gives this fatal error

Full stack trace:
```
at Object.extractError (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/protocol/json.js:51:27)
    at Request.extractError (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/protocol/rest_json.js:55:8)
    at Request.callListeners (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/sequential_executor.js:106:20)
    at Request.emit (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/sequential_executor.js:78:10)
    at Request.emit (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/request.js:683:14)
    at Request.transition (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/request.js:22:10)
    at AcceptorStateMachine.runTo (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/state_machine.js:14:12)
    at /usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/state_machine.js:26:10
    at Request.<anonymous> (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/request.js:38:9)
    at Request.<anonymous> (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/aws-sdk/lib/request.js:685:12)
There was an error adding the datasource
⠧ Fetching Aurora Serverless cluster...
```

It continues to fetch aurora serverless cluster indefinitely.

**To Reproduce**
Follow steps here https://aws-amplify.github.io/docs/cli/graphql#automatically-import-existing-datasources

```
❯ amplify api add-graphql-datasource
Using datasource: Aurora Serverless, provided by: awscloudformation
? Provide the region in which your cluster is located: us-east-1
? Select the Aurora Serverless cluster that will be used as the data source for your API: amplify-serverless-cluster
```

[...]

PR #1556, opened 2019-05-31
Add new available regions for Aurora Serverless

adav:
I bumped into a missing availability region while playing with Aurora Serverless and AppSync.

I was trying the `amplify api add-graphql-datasource` command in the CLI.

According to [this blog](https://aws.amazon.com/blogs/aws/new-data-api-for-amazon-aurora-serverless/), the necessary Data API feature has only recently been added to more regions.

I only had a brief look through the codebase, but I think this is the only change necessary - hope it's a helpful first contribution :) Btw, thanks to the AppSync and Amplify team for the great work!

~

### DIFF
--- a/packages/amplify-category-api/provider-utils/supported-datasources.json
+++ b/packages/amplify-category-api/provider-utils/supported-datasources.json
@@ -34,6 +34,6 @@
 		"serviceWalkthroughFilename": "appSync-rds-walkthrough.js",
 		"cfnFilename": "appSync-rds-cloudformation-template-default.yml.ejs",
 		"provider": "awscloudformation",
-		"availableRegions": ["us-east-1"]
+		"availableRegions": ["us-east-1", "us-east-2", "us-west-2", "ap-northeast-1", "eu-west-1"]
 	}
 }


adav:
After a little more playing around with this CLI, this change is not ready to merge live because actually selecting any region other than `us-east-1` returns exactly the same error as reported in issue #1403. 

The fix for that issue, or reason why it is now closed is very ambiguous. From first glance, it looks like the library dependency `aws-sdk` might be at fault for the newly supported Data API regions...

TrueLecter:
Is ExecuteStatement still not available in aws-sdk? It is available in documentation: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/RDSDataService.html#executeStatement-property

iler:
It should be available because you can find this from the changelog https://github.com/aws/aws-sdk-js/blob/master/CHANGELOG.md#24660.

> The RDS Data API is generally available for the MySQL-compatible edition of Amazon Aurora Serverless in the US East (N. Virginia and Ohio), US West (Oregon), EU (Ireland), and Asia Pacific (Tokyo) regions. This service enables you to easily access Aurora Serverless clusters with web services-based applications including AWS Lambda and AWS AppSync. The new APIs included in this SDK release are **ExecuteStatement**, BatchExecuteStatement, BeginTransaction, CommitTransaction, and RollbackTransaction. **The ExecuteSql API is deprecated; instead use ExecuteStatement** which provides additional functionality including transaction support.

TrueLecter:
So this pull request can be finally merged? I really need RDS on eu-west-1, but I afraid that patching file myself may lead to some unpredictable results.

iler:
@ashwindevendran is this already implemented in amplify-cli?

Issue #1600, opened 2019-03-06
Support @deprecated directive


kevin-lindsay-1:
**Is your feature request related to a problem? Please describe.**
Cannot hide a piece of my schema

**Describe the solution you'd like**
Support of the common @deprecated directive

SwaySway:
Closing this issue as the PR https://github.com/aws-amplify/amplify-cli/pull/1612 for it has been merged.

PR #1612, opened 2019-06-07
fix(graphql-transformer-core): Disable @deprecated directive stripping from schema  

attilah:
*Issue #, if available:*

Fixing #1600

*Description of changes:*

Enable @deprecated directive support on field definitions and enum value nodes in the schema. Previously this directive was stripped from the schema during the transformation, now it will be preserved and will be deployed to AppSync Service.


undefobj:
Will this actually work in AppSync?

attilah:
@undefobj Yes, it works with some caveats:
- AppSync Service is stripping the directives from `ENUM_VALUE` nodes, so it's not preserved yet
- We're correctly "hiding" `FIELD_DEFINITION` nodes in the Query Designer within AppSync Console and put them into a "Deprecated Fields" section.
- AppSync Service correctly sets the `isDeprecated` and `deprecationReason` in the compiled JSON schema.

PR #1661, opened 2019-06-13
test(amplify-e2e-tests): add headless init e2e test

yuth:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-e2e-tests/__tests__/init.test.ts
+++ b/packages/amplify-e2e-tests/__tests__/init.test.ts
@@ -4,7 +4,8 @@ import {
   deleteProject,
   initProjectWithAccessKey,
   initNewEnvWithAccessKey,
-  initNewEnvWithProfile
+  initNewEnvWithProfile,
+  initHeadless
 } from '../src/init';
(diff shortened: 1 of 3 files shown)


packages/amplify-e2e-tests/__tests__/init.test.ts
```diff
@@ -100,4 +101,34 @@ describe('amplify init', () => {
     await expect(newEnvAuthRoleName).toBeIAMRoleWithArn(newEnvAuthRoleArn);
     await expect(newEnvDeploymentBucketName).toBeAS3Bucket();
   });
+  it('should init project in headless mode', async () => {
+    const { ACCESS_KEY_ID, SECRET_ACCESS_KEY } = getEnvVars();
```
attilah: I see that in the init tests we're expecting ```ACCESS_KEY_ID``` and ```SECRET_ACCESS_KEY``` without the ```AWS_``` prefix, but other parts of the test code and the sample env file also has ```AWS_ACCESS_KEY_ID``` and ```AWS_SECRET_ACCESS_KEY```. Do you agree that it makes sense to add the prefix everywhere, or is there a reason why we use different names here?

PR #1668, opened 2019-06-14
fix(amplify-provider-awscloudformation): generate consistent S3 keys

### LINKED ISSUE #1666: Integration in multi-env and multi-branch project

Handling multiple environments and multiple branches

Amplify

Hello, I have a project with a couple of environments linked to a couple of git branches (dev -> dev, prod -> master). I develop features in dev branch (and push to amplify dev branch), and when they're ready I merge them to the master branch (and push to prod amplify env). The problem that I'm having is that I have several functions, and every time I push to either env I get changes like this: 

```
-                                       "S3Bucket": "portal-20190528145021-deployment",
-                                       "S3Key": "amplify-builds/myfn-1559227132-latest-build.zip"
+                                       "S3Bucket": "portal-20190328173653-deployment",
+                                       "S3Key": "amplify-builds/myfn-1560450865-latest-build.zip"
```
but then, if I commit those changes and later merge to master and push, then I'll also have another set of changes in master, so I commit those to master. However, that means that every change that I merge from dev to master will have a merge conflict (even if it doesn't modify amplify resources). Have anyone experience this annoyance? How did you work around it?

Thanks!

yuth:
Amplify CLI used to use date to ensure that the functions S3 key was changing every time it got
pushed to the cloud. This caused multi env with multi branch workflow to have git merge conflict as
these keys kept changing based on the time when the function gets pushed. Updated the code to use the
folder hash of function so the key is consistent if the content has not changed

fix #1666

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/build-resources.js
+++ b/packages/amplify-provider-awscloudformation/lib/build-resources.js
@@ -38,7 +38,12 @@ async function buildResource(context, resource) {
     !resource.distZipFilename ||
     isPackageOutdated(resourceDir, resource.lastPackageTimeStamp)
   ) {
-    const { hash: folderHash } = await hashElement(resourceDir);
+    // generating hash, ignoring node_modules as this can take long time to hash
+    // the content inside node_modules change only when content of package-lock.json changes
+    const { hash: folderHash } = await hashElement(resourceDir, {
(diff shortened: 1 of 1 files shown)


kaustavghosh06:
We can probably get rid of the `removeOutdatedPackage` function since we're overwriting the zip files now with the same name.

PR #1669, opened 2019-06-14
refactor(graphql-transformer-core) - Refactoring amplifyUtil.ts

mikeparisstuff:
*Description of changes:*

Refactors the creepingly large `amplifyUtils.ts` file into `amplifyUtils.ts`, `fileUtils.ts`, and `transformConfig.ts`. This change adds more structural clarity to how the `transform.conf.json` file is read and saved. These changes enable future changes that will be using the `transform.conf.json` file to enable more detailed user configuration options. This change makes no significant functional changes.

### DIFF
--- a/packages/graphql-transformer-core/src/util/amplifyUtils.ts
+++ b/packages/graphql-transformer-core/src/util/amplifyUtils.ts
@@ -1,4 +1,4 @@
-import * as fs from 'fs';
+const fs = require('fs-extra');
 import * as path from 'path';
 import { CloudFormation, Fn, Template } from "cloudform-types";
 import GraphQLTransform from '..';
@@ -6,7 +6,7 @@ import Transformer from '../Transformer';
 import DeploymentResources from '../DeploymentResources';
 import { StackMappingOption } from '../GraphQLTransform';
 import { ResourceConstants } from 'graphql-transformer-common';
-import { exists, readFile, walkDirPosix, clearAtPath, readFromPath, writeToPath, emptyDirectory, throwIfNotJSONExt } from './fileUtils';
+import { walkDirPosix, readFromPath, writeToPath, throwIfNotJSONExt, emptyDirectory } from './fileUtils';
 import { writeConfig, TransformConfig, TransformMigrationConfig, loadProject, readSchema } from './transformConfig';
 
 const CLOUDFORMATION_FILE_NAME = 'cloudformation-template.json';
@@ -316,7 +316,7 @@ export async function migrateAPIProject(opts: MigrationOptions) {
     }
 }
 export async function revertAPIMigration(directory: string, oldProject: AmplifyApiV1Project) {
-    await clearAtPath(directory);
+    await fs.remove(directory);
     await writeToPath(directory, oldProject);
 }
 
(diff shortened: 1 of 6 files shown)


packages/graphql-transformer-core/src/util/fileUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+            err ? rej(err) : res(d)
+        })
+    })
+}
+
+/**
+ * Helpers
+ */
+
+export async function deleteDirectory(directory: string): Promise<void> {
```
yuth: We could use `removeSync` from `fs-extra` which handles this

packages/graphql-transformer-core/src/util/fileUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+            await unlink(fullPath);
+        }
+    }
+}
+
+/**
+ * Make if not exists.
+ */
+export async function mkdirIfNone(dir: string) {
+    const pathExists = await exists(dir);
```
yuth: use `ensureDirSync` from `fs-extra`

packages/graphql-transformer-core/src/util/fileUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+    if (!pathExists) {
+        await mkdir(dir)
+    }
+}
+
+export async function writeToPath(directory: string, obj: any): Promise<void> {
+    if (Array.isArray(obj)) {
```
yuth: use `outputFile` from `fs-extra`

packages/graphql-transformer-core/src/util/fileUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+        accum[fileName] = value;
+    }
+    return accum;
+}
+
+/**
+ * Recursively clear the contents at some path. Might be a file or directory.
+ * @param clearPath The path to clear.
+ */
+export async function clearAtPath(clearPath: string) {
```
yuth: `remove` / `removeSync` from `fs-extra` should do what we are doing here

packages/graphql-transformer-core/src/util/fileUtils.ts
```diff
@@ -0,0 +1,183 @@
...
+
+export async function walkDir(dir: string, handler: (file: { Key: string, Body: Buffer | string}) => Promise<string>) {
+    return await walkDirRec(dir, handler, '', path.join);
+}
+export async function walkDirPosix(dir: string, handler: (file: { Key: string, Body: Buffer | string}) => Promise<string>) {
+    return await walkDirRec(dir, handler, '', path.posix.join);
+}
+
+export function throwIfNotJSONExt(stackFile: string) {
+    const nameParts = stackFile.split('.');
```
yuth: we could use `path.extname` instead

mikeparisstuff: This package does not currently depend on fs-extra. I agree that we could move to use these but since we don't have the dependency right now I think I prefer the self contained changes for now. Thoughts?

packages/graphql-transformer-core/src/util/fileUtils.ts
mikeparisstuff: `outputFile` is not the same as this method which will lay an object out onto the file system.

packages/graphql-transformer-core/src/util/fileUtils.ts
mikeparisstuff: Updated.

packages/graphql-transformer-core/src/util/fileUtils.ts
mikeparisstuff: Replaced

packages/graphql-transformer-core/src/util/fileUtils.ts
mikeparisstuff: Replaced

packages/graphql-transformer-core/src/util/fileUtils.ts
mikeparisstuff: replaced

packages/graphql-transformer-core/src/util/amplifyUtils.ts
```diff
@@ -1,12 +1,12 @@
-import * as fs from 'fs';
+const fs = require('fs-extra');
```
yuth: why common js require instead of es6 import?

packages/graphql-transformer-core/src/util/fileUtils.ts
```diff
@@ -1,79 +1,35 @@
...
-export async function deleteDirectory(directory: string): Promise<void> {
-    const pathExists = await exists(directory);
-    if (!pathExists) {
-        return;
-    }
-    await emptyDirectory(directory);
-    await rmdir(directory);
-}
-
 export async function emptyDirectory(directory: string) {
```
yuth: may be we can directly use [emptyDir](https://github.com/jprichardson/node-fs-extra/blob/master/docs/emptyDir.md)

Issue #1652, opened 2019-06-12
amplify push creating rather than updating a table


aehringer:
** Which Category is your question related to? ** 
graphql-transform, cloudformation, amplify push

** What AWS Services are you utilizing? **
dynamodb, amplify api

** Provide additional details e.g. code snippets **
I'm getting this failure after I try to push an updated api schema with just one extra (empty) line.
I see that it is failing because it is trying to create a table that already exists. Why is it not just updating the table, as it does with all the other tables I have? Nothing in this table is being modified at all.

CREATE_IN_PROGRESS          SavedRestaurantTable                                             AWS::DynamoDB::Table       Wed Jun 12 2019 16:31:26 GMT+0200 (CEST)                                                                                                                                                                                                                                   
CREATE_FAILED               SavedRestaurantTable                                             AWS::DynamoDB::Table       Wed Jun 12 2019 16:31:26 GMT+0200 (CEST) SavedRestaurant-muz7cgqce5hmtkneh5jqg242f4 already exists in stack arn:aws:cloudformation:us-east-1:************:stack/reco-************-apireco-************-SavedRestaurant-*************/************************************
UPDATE_ROLLBACK_IN_PROGRESS reco-************-apireco-************-Restaurant-************ AWS::CloudFormation::Stack Wed Jun 12 2019 16:31:29 GMT+0200 (CEST) The following resource(s) failed to create: [SavedRestaurantTable].

yuth:
@aehringer could you share what change you made to your schema. A minimal repro repo would make it a lot faster for us to diagnose the issue

aehringer:
I added an empty line.
Before: 
type User 
	@model 
	@searchable 
	@versioned
{
	id: ID!
	savedPlaces: [SavedRestaurant] @connection(name: "User_SavedRestaurant")
}

type SavedRestaurant @model(queries: null) {
	id: ID!
	user: User! @connection(name:"User_SavedRestaurant")
	restaurant: Restaurant! @connection
}

type Restaurant 
	@model 
	@searchable
	@versioned
{
	id: ID!
	name: String!
	googleId: String
	location: String
}

amplify status:
<img width="1131" alt="Screen Shot 2019-06-12 at 8 33 22 PM" src="https://user-images.githubusercontent.com/42987147/59377132-dfe98000-8d51-11e9-992d-222a55b96264.png">

After:
type User 
	@model 
	@searchable 
	@versioned
{
	id: ID!
[...]

PR #1717, opened 2019-06-24
fix(graphql-transformer-core) - Fixing stack mapping.

mikeparisstuff:
*Issues*
#1581
#1652 - This behavior pattern is now fixed. If you have previously deployed an API with this issue, the new utility method updates `transform.conf.json` such that any subsequent pushes with the fixed implementation do not try to move existing resources out of their existing stacks.

*Description of changes:*

Fixing an issue with an internal mechanism called the stack mapping. The issue materialized when you tried to create an API with types that share name parts (e.g. Post & UserPost & User). When using names that share parts, the order of the types in the input document could impact which stack a resource got mapped to. This fix removes using regular expressions and replaces the stack map with a full mapping where each resource is explicitly mapped to a stack.

*Testing:*

This change includes new tests that cover all directives that create independent resources and strictly inspect the stack outputs.

### DIFF
--- a/packages/graphql-transformer-core/src/__tests__/TransformFormatter.test.ts
+++ b/packages/graphql-transformer-core/src/__tests__/TransformFormatter.test.ts
@@ -75,7 +75,6 @@ test('Test getTemplateReferences', () => {
     context.mapResourceToStack('PostModel', 'PostTableOutput');
     context.template = template;
     const deploymentResources = formatter.format(context)
-    console.log(deploymentResources);
     expect(Object.keys(deploymentResources.stacks.PostModel.Resources)).toHaveLength(4)
     expect(Object.keys(deploymentResources.rootStack.Resources)).toHaveLength(3)
     expect(Object.keys(deploymentResources.stacks.PostModel.Outputs)).toHaveLength(1);
--- a/packages/graphql-transformer-core/src/__tests__/TransformFormatter.test.ts
+++ b/packages/graphql-transformer-core/src/__tests__/TransformFormatter.test.ts
@@ -66,15 +66,16 @@ const template: Template = {
 
 
 test('Test getTemplateReferences', () => {
-    const stackRules = new Map<string, string>();
-    stackRules.set('.*PostResolver', 'PostModel');
-    stackRules.set('^PostTable.*', 'PostModel');
-    const formatter = new TransformFormatter({
-        stackRules: stackRules,
-    });
+    const formatter = new TransformFormatter();
     const context = new TransformerContext('type Post @model { id: ID! title: String }')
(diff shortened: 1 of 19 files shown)


packages/graphql-transformer-core/src/__tests__/TransformFormatter.test.ts
```diff
@@ -66,15 +66,16 @@ const template: Template = {
...
+    const formatter = new TransformFormatter();
     const context = new TransformerContext('type Post @model { id: ID! title: String }')
+    context.mapResourceToStack('PostModel', 'CreatePostResolver');
+    context.mapResourceToStack('PostModel', 'UpdatePostResolver');
+    context.mapResourceToStack('PostModel', 'PostTableDataSource');
+    context.mapResourceToStack('PostModel', 'PostTable');
+    context.mapResourceToStack('PostModel', 'PostTableOutput');
     context.template = template;
     const deploymentResources = formatter.format(context)
+    console.log(deploymentResources);
```
attilah: nit: leftover or intentional?

packages/graphql-transformer-core/src/util/amplifyUtils.ts
```diff
@@ -82,6 +90,66 @@ function adjustBuildForMigration(resources: DeploymentResources, migrationConfig
     return resources;
 }
 
+
+export type FindMissingStackMappingConfig = ProjectOptions & { currentCloudBackendDirectory: string }
+/**
+ * Provided a build configuration & current-cloud-backend directory, calculate
+ * any missing stack mappings that might have been caused by the stack mapping
+ * bug in June 2019. This allows APIs that were deployed with the bug to continue
```
attilah: nit: adding link to issue for history?

packages/graphql-transformer-core/src/util/amplifyUtils.ts
```diff
@@ -82,6 +90,66 @@ function adjustBuildForMigration(resources: DeploymentResources, migrationConfig
...
+    const copyOfCloudBackend = await readFromPath(currentCloudBackendDirectory);
+    const stackMapping = transformOutput.stackMapping;
+    const stackNames = Object.keys(copyOfCloudBackend.build.stacks);
+
+    // We walk through each of the stacks that were deployed in the most recent deployment.
+    // If we find a resource that was deployed into a different stack than it should have
+    // we make a note of it and include it in the missing stack mapping.
+    for (const stackFileName of stackNames) {
+        const stackName = stackFileName.slice(0, stackFileName.length - path.extname(stackFileName).length);
+        // const stack = transformOutput.stacks[stackName];
```
attilah: nit: leftover commented line.

PR #1782, opened 2019-07-05
Triggers auth

haverchuck:
*Description of changes:*

### DIFF
--- a/packages/amplify-category-auth/provider-utils/awscloudformation/service-walkthroughs/auth-questions.js
+++ b/packages/amplify-category-auth/provider-utils/awscloudformation/service-walkthroughs/auth-questions.js
@@ -49,7 +49,7 @@ async function serviceWalkthrough(
     // ASK QUESTION
     const answer = await inquirer.prompt(q);
 
-    if (answer.triggers && answer.triggers.length > 0) {
+    if (answer.triggers) {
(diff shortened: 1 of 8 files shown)


packages/amplify-category-auth/provider-utils/awscloudformation/service-walkthroughs/auth-questions.js
```diff
@@ -49,7 +49,7 @@ async function serviceWalkthrough(
     // ASK QUESTION
     const answer = await inquirer.prompt(q);
 
-    if (answer.triggers && answer.triggers.length > 0) {
+    if (answer.triggers) {
```
kaustavghosh06: Wouldn't you want this check out here as well -> && props.triggers !== '{}'

packages/amplify-category-function/provider-utils/awscloudformation/index.js
```diff
@@ -245,6 +245,19 @@ async function updateResource(context, category, service, parameters, resourceTo
 
   if (answers.parameters) {
     createParametersFile(context, answers.parameters, answers.resourceName);
+  } else if (answers.trigger) {
```
kaustavghosh06: I think this should be a standalone if - not an else if?

Issue #1806, opened 2019-07-10
Cant create a new ENV after update to v1.8.0  - no such file or directory ... function-parameters.json


davidbiller:
**Describe the bug**
Cant create a new ENV after update to v1.8.0 

**To Reproduce**
Do you want to use an existing environment? (n)
Enter a name for the environment: develop
Do you want to use an AWS profile? (Y) - default
...
✔ Successfully created initial AWS cloud resources for deployments.
✔ Initialized provider successfully.
Do you want to configure Lambda Triggers for Cognito? (n)

`✖ There was an error initializing your environment.
init failed
{ Error: ENOENT: no such file or directory, open '/Users/davidbiller/Desktop/Developer/aivy-rn-app/amplify/backend/function/aivy/function-parameters.json'
    at Object.openSync (fs.js:451:3)
    at Object.readFileSync (fs.js:351:35)
    at Object.readJsonFile (/usr/local/lib/node_modules/@aws-amplify/cli/src/extensions/amplify-helpers/read-json-file.js:11:33)
    at updateConfigOnEnvInit (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-category-function/provider-utils/awscloudformation/index.js:427:42)
    at /usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-category-function/index.js:115:28
    at /usr/local/lib/node_modules/@aws-amplify/cli/node_modules/promise-sequential/index.js:16:18
    at process.internalTickCallback (internal/process/next_tick.js:77:7)
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path:
   '/Users/davidbiller/Desktop/Developer/aivy-rn-app/amplify/backend/function/aivy/function-parameters.json' }`

If i choose Y or N the same result. 

**Additional context**

amplify --v
1.8.0

//Seems there is no aivy/function-parameters.json
Note: aivy is not a function, it is the project name.
My functions are direct in the functions folder.

dbramwell:
I'm suffering from this issue as well. It seems in previous versions of amplify the `function-parameters.json` wasn't generated when adding a function. In my project some functions have this `function-parameters.json` and some don't.

A simple fix could be checking if the file exists. Not sure if this is the best solution though. I'll make a pull request.

davidbiller:
[...]

PR #1808, opened 2019-07-10
fix: check that function-parameters.json exists before trying to read it

dbramwell:
*Issue #, if available:* [1806](https://github.com/aws-amplify/amplify-cli/issues/1806)

*Description of changes:*
Functions in projects created with an older version of the Amplify cli seem not to have a `function-parameters.json` file. We should check if this file exists before trying to read it.

### DIFF
--- a/packages/amplify-category-function/provider-utils/awscloudformation/index.js
+++ b/packages/amplify-category-function/provider-utils/awscloudformation/index.js
@@ -421,10 +421,14 @@ function getHeadlessParams(context, service) {
 
 
 async function updateConfigOnEnvInit(context, category, service) {
-  const srvcMetaData = context.amplify.readJsonFile(`${__dirname}/../supported-services.json`)
-    .Lambda;
+  const srvcMetaData = context.amplify.readJsonFile(`${__dirname}/../supported-services.json`).Lambda;
   const providerPlugin = context.amplify.getPluginInstance(context, srvcMetaData.provider);
-  const resourceParams = context.amplify.readJsonFile(`${context.amplify.pathManager.getBackendDirPath()}/function/${service}/function-parameters.json`);
+  const functionParametersPath = `${context.amplify.pathManager.getBackendDirPath()}/function/${service}/function-parameters.json`;
+  let resourceParams = {};
+  const functionParametersExists = await fs.exists(functionParametersPath);
+  if (functionParametersExists) {
+    resourceParams = context.amplify.readJsonFile(functionParametersPath);
+  }
   let envParams = {};
 
   // headless mode


jonmifsud:
Not sure, but maybe it would be smarter to create if they don't exist? Ensuring future compatibility rather than silently handling? 

It tends to be awkwardly confusing whenever files in between versions appear magically (sometimes handled sometimes not) and only discovered by creating new environments.

davidbiller:
i tried it, but get the same error.

`const functionParametersPath = `${context.amplify.pathManager.getBackendDirPath()}/function/${service}/function-parameters.json`;`

the ${service} is not a functionname, its the project name?!
No such file or directory: /amplify/backend/function/aivy/function-parameters.json
So, aivy is not a function, it is the project name.

dbramwell:
@jonmifsud Yeah, might be better to create it.

@davidbiller Strange, it fixed the issue for me. So are you saying that `amplify/backend/function/aivy` doesn't actually exist for you?
When you run `amplify configure function` does a function called `aivy` appear in the list?

davidbiller:
> @jonmifsud Yeah, might be better to create it.
> 
> @davidbiller Strange, it fixed the issue for me. So are you saying that `amplify/backend/function/aivy` doesn't actually exist for you?
> When you run `amplify configure function` does a function called `aivy` appear in the list?

nope. aivy is not a function. 
my functions folder looks like this:
`amplify/backend/function/estest`
is yours like this?:
`amplify/backend/function/projectName/estest`

davidbiller:
fixed this problem, yes.

dbramwell:
@davidbiller my functions are also in the same place as yours:
```
amplify/backend/function/<function 1 name>
amplify/backend/function/<function 2 name>
```
I'm not sure what your issue is. Can you run `amplify configure function` to see if a function called aivy appears in the list?

davidbiller:
> @davidbiller my functions are also in the same place as yours:
> 
> ```
> amplify/backend/function/<function 1 name>
> amplify/backend/function/<function 2 name>
> ```
> 
> I'm not sure what your issue is. Can you run `amplify configure function` to see if a function called aivy appears in the list?

you fix is resolving my error too, but i dont know why its searching in the functions/projectname/ folder.
All my functions has the  function-parameter.json, so this is not the failure, its becuase the wrong path i guess. so your fix is fixing this too, for sure, because it is not exist. Do you understand?

dbramwell:
@davidbiller got it. Do you have auth set up? Is that called `aivy` ? The only other place that `updateConfigOnEnvInit` is called is [here](https://github.com/aws-amplify/amplify-cli/blob/c6fc83834ae70f3e0f5e1c8810a56de76ba36d41/packages/amplify-category-auth/index.js#L214).

@jonmifsud this means we would need to check if `category === 'function'` before creating the function-parameter.json right?

haverchuck:
@dbramwell Thanks for your fix!

PR #1815, opened 2019-07-12
Sanity check

mikeparisstuff:
*Issue #, if available:*

#922 & #1406

*Description of changes:*

This change adds:
- Support for the "Retain" deletion policy on DynamoDB tables created by @model via a new config option in `transform.conf.json`.
- Adds a sanity check component that protects against invalid migrations of a variety of types.
- Adds a breadth of new integration tests that cover both valid and invalid migration flows.

See `packages/amplify-e2e-tests/__tests__/migration` for examples of what is covered by the sanity check.


packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -242,11 +241,14 @@ place .graphql files in a directory at ${schemaDirPath}`);
   fs.writeFileSync(parametersFilePath, jsonString, 'utf8');
 }
 
-function getModelConfig(project) {
-  if (project && project.config && project.config.Model && project.config.Model.BillingMode) {
-    return {
-      BillingMode: project.config.Model.BillingMode,
```
kaustavghosh06: Why were we just passing BillingMode previously and what would this chnage enable us to do?

packages/graphql-elasticsearch-transformer/src/resources.ts
```diff
@@ -45,13 +45,6 @@ export class ResourceFactory {
                 Description: 'The number of instances to launch into the Elasticsearch domain.',
                 Default: 1
             }),
-            [ResourceConstants.PARAMETERS.ElasticsearchDomainName]: new StringParameter({
```
kaustavghosh06: Why are we removing this?

packages/graphql-transformer-core/src/util/transformConfig.ts
```diff
@@ -28,7 +28,18 @@ export interface TransformConfig {
...
         [resourceId: string]: string
-    }
+    },
+
+    /**
+     * Provide build time options to GraphQL Transformer constructor functions.
+     * Certain options cannot be configured via CloudFormation parameters and
+     * need to be set at build time. E.G. DeletionPolicies cannot depend on parameters.
+     */
+    TransformerOptions?: {
```
kaustavghosh06: What are the list of options that could be passed out here?

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
mikeparisstuff: This was actually dead code as there has never been a value passed to the config at the path model.BillingMode. I believe this was meant to be deleted but was missed in a previous PR since this variable is actually exposed via parameters.json and not via the transform.conf.json file.

packages/graphql-elasticsearch-transformer/src/resources.ts
mikeparisstuff: This is no longer used and is not documented so I am removing it.

packages/graphql-transformer-core/src/util/transformConfig.ts
mikeparisstuff: The intention is to have each transformer define its own set of options that can be passed. Right now the only acceptable option is like:

```
{
  "@model": {
    "EnableDeletionProtection": true/false
  }
}
```

packages/amplify-e2e-tests/__tests__/api.test.ts
```diff
@@ -15,21 +18,70 @@ describe('amplify add api', () => {
...
+
     await expect(GraphQLAPIIdOutput).toBeDefined()
     await expect(GraphQLAPIEndpointOutput).toBeDefined()
     await expect(GraphQLAPIKeyOutput).toBeDefined()
   });
+
+  it('inits a project with a simple model and then migrates the api', async () => {
+    const projectName = 'retaintables';
+    const initialSchema = 'simple_model.graphql';
+    console.log(projRoot);
```
kaustavghosh06: Remove this?

packages/amplify-category-api/commands/api/push.js
```diff
@@ -9,8 +9,7 @@ module.exports = {
     context.amplify.constructExeInfo(context);
     return amplify.pushResources(context, category, resourceName)
       .catch((err) => {
-        context.print.info(err.stack);
-        context.print.error('There was an error pushing the API resource');
+        context.print.error(err.toString());
```
kaustavghosh06: wouldn't the error message be more verbose and unformatted if we just do a toString()?
 Also, we should retain this error message - context.print.error('There was an error pushing the API resource');

packages/amplify-category-api/commands/api/gql-compile.js
```diff
@@ -6,8 +6,7 @@ module.exports = {
     try {
       await context.amplify.executeProviderUtils(context, 'awscloudformation', 'compileSchema', { noConfig: true, forceCompile: true });
     } catch (err) {
-      context.print.info(err.stack);
-      context.print.error(err.message);
+      context.print.error(err.toString());
```
attilah: It would be nice to have an easily identifiable, customer friendly message in every catch before the technical things like stacktrace, here is like:
```context.print.error('There was an error while compiling the schema.');```

packages/amplify-category-api/commands/api/push.js
mikeparisstuff: It makes the error a lot more readable in my opinion. The stack trace is only marginally helpful to the user and now you get much cleaner, more helpful error messages. I have added back the original error text before the new text but in general I think its better to allow the origin of the error to define what type of error it is instead of having the catch all text hard coded into the handler. In this instance, the transformer has not tried to push yet so the text "There was an error pushing the API resource" isn't as focused as it could be. Ideally each error type would know how to represent itself and then the downstream code can throw one of those errors and the it will bubble up to the top level where it can be nicely handled (by switching on an enum for example).

PR #1872, opened 2019-07-20
fix: remove grunt-lambda dependency for local function testing

kaustavghosh06:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-category-function/provider-utils/awscloudformation/utils/invoke.js
+++ b/packages/amplify-category-function/provider-utils/awscloudformation/utils/invoke.js
@@ -1,15 +1,15 @@
 const path = require('path');
-const fs = require('fs');
 
 function loadFunction(fileName) {
   return require(path.resolve(fileName));
 }
 /*
 var options = this.options({
-    'packageFolder': './',
-    'handler': 'handler',
-    'fileName': 'index.js',
-    'event': 'event.json'
+    'packageFolder': './', //required
+    'handler': 'handler', // required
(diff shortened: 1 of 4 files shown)


packages/amplify-category-function/provider-utils/awscloudformation/utils/invoke.js
```diff
@@ -0,0 +1,67 @@
...
+    awsRequestId: 'LAMBDA_INVOKE',
+    logStreamName: 'LAMBDA_INVOKE',
+  };
+
+  const callback = (error, object) => {
+    context.done(error, object);
+  };
+
+  const lambda = loadFunction(options.fileName);
+  const event = JSON.parse(fs.readFileSync(path.resolve(options.event), 'utf8'));
```
yuth: Could we make the invoke function just accept this as a parameter instead of reading this from the file

packages/amplify-category-function/provider-utils/awscloudformation/utils/invoke.js
```diff
@@ -0,0 +1,67 @@
...
+*/
+function invokeFunction(options) {
+  console.log('Testing function locally');
+  let cwd;
+  if (options.packageFolder) {
+    cwd = process.cwd();
+    process.chdir(path.resolve(options.packageFolder));
+  }
+
+  const context = {
```
yuth: Could we allow context to be passed as part of options to allow custom implementation of context functions.

Issue #1897, opened 2019-07-25
Adding auth fail to push


artecoop:
**Describe the bug**
I'm creating a new project with api and Cognito auth

**To Reproduce**
- When requested, check email as method for Cognito authentication
- add confirmation link (enter whatever you want as subject and message)
- all other value default
- amplify push

**Desktop (please complete the following information):**
 - OS: Fedora 30
 - Node 12.7.0
 - npm 6.10.2

**Error**
Following resources failed

Resource Name: UserPool (AWS::Cognito::UserPool)
Event Type: create
Reason: 1 validation error detected: Value 'functionvdp3dcf8667CustomMessageArn' at 'lambdaConfig.customMessage' failed to satisfy constraint: Member must satisfy regular expression pattern: arn:[\w+=/,.@-]+:[\w+=/,.@-]+:([\w+=/,.@-]*)?:[0-9]+:[\w+=/,.@-]+(:[\w+=/,.@-]+)?(:[\w+=/,.@-]+)? (Service: AWSCognitoIdentityProviderService; Status Code: 400; Error Code: InvalidParameterException; Request ID: 8facc4dd-fe6a-4ecc-ad54-7df4319f5ac7)

Obviously it rollback

haverchuck:
@artecoop In the cloudformation console, can you check the Parameters section in the auth stack, and  post the value listed for `functionvdp3dcf8667CustomMessageArn`?

JarrydTrokis:
I'm also having the same problem. Interestingly, this only happened with the updated version of the cli. Previously, I had been running on 1.7 — since running on 1.8, I started seeing the bug.

@artecoop I did find that by creating **auth** first, then creating **api**, I managed to get things up and running.

artecoop:
Thanks for the comment and input.
First of all, I want to apologize for posting in the wrong repo.

I can confirm, as @JarrydTrokis suggested, that the flow:
- amplify add auth
- amplify push
- amplify add api
- amplify push 

did the trick.

You can close the issue when needed.

kleingtm:
I see this bug on both 1.8.5 and 1.8.6 (macOS, npm 6.9.0).

For what it's worth, it doesn't happen if I use a default(ish) setup. However, if you choose to manually edit a variety of options during `amplify add auth`, it'll fail.

kaustavghosh06:
We have a PR out for this - #1897 and we'll publish the fix to npm as soon as we merge it.

[...]

PR #1898, opened 2019-07-25
fixes(#1897)

haverchuck:
*Issue #, if available:*
Fixes(#1897)

*Description of changes:*
Adds dependsOn block for externally added auth (ie. when auth added via API).
Adds integration test to catch this failure point in the future.


packages/amplify-category-auth/index.js
```diff
@@ -42,6 +42,9 @@ async function add(context) {
         service: resultMetadata.service,
         providerPlugin: resultMetadata.providerName,
       };
+      if (context.amplify.auth && context.amplify.auth.dependsOn) {
```
kaustavghosh06: How do we add context.amplify.auth? Is this the amplifyMeta?

haverchuck: I can't recall why we were adding this to the context as opposed to reading the parameters file, but I think it may be because we were previously updating the parameters file differently.  At this point there is no need for it, so I refactored to read the auth params file directly.

packages/amplify-category-auth/commands/auth/enable.js
```diff
@@ -31,8 +31,10 @@ module.exports = {
         return providerController.addResource(context, category, result.service);
       })
       .then((resourceName) => {
-        if (context.amplify.auth && context.amplify.auth.dependsOn) {
-          options.dependsOn = context.amplify.auth.dependsOn;
+        const authParameters = amplify.readJsonFile(`${amplify.pathManager.getBackendDirPath()}/auth/${resourceName}/parameters.json`);
```
kaustavghosh06: This won't work on windows. please use path.join() instead.

packages/amplify-category-auth/commands/auth/update.js
```diff
@@ -49,8 +49,9 @@ module.exports = {
         return providerController.updateResource(context, category, options);
       })
       .then((resourceName) => { // eslint-disable-line no-shadow
-        if (context.amplify.auth && context.amplify.auth.dependsOn) {
-          amplify.updateamplifyMetaAfterResourceUpdate(category, resourceName, 'dependsOn', context.amplify.auth.dependsOn);
+        const authParameters = amplify.readJsonFile(`${amplify.pathManager.getBackendDirPath()}/auth/${resourceName}/parameters.json`);
```
kaustavghosh06: This won't work on windows. please use path.join() instead.

packages/amplify-category-auth/index.js
```diff
@@ -42,6 +42,12 @@ async function add(context) {
         service: resultMetadata.service,
         providerPlugin: resultMetadata.providerName,
       };
+
+      const authParameters = amplify.readJsonFile(`${amplify.pathManager.getBackendDirPath()}/auth/${resourceName}/parameters.json`);
```
kaustavghosh06: same comment as above.

packages/amplify-category-auth/index.js
```diff
@@ -135,6 +141,11 @@ async function externalAuthEnable(context, externalCategory, resourceName, requi
         service: 'Cognito',
         providerPlugin: 'awscloudformation',
       };
+      const authParameters = amplify.readJsonFile(`${amplify.pathManager.getBackendDirPath()}/auth/${resourceName}/parameters.json`);
```
kaustavghosh06: same comment as above.

packages/amplify-category-auth/index.js
```diff
@@ -164,9 +187,13 @@ async function checkRequirements(requirements, context) {
   let authParameters;
 
   if (existingAuth && Object.keys(existingAuth).length > 0) {
-    authParameters = amplify.readJsonFile(`${amplify.pathManager.getBackendDirPath()}/auth/${
-      Object.keys(existingAuth)[0]
-    }/parameters.json`);
+    const resourceDirPath = path.join(
+      amplify.pathManager.getBackendDirPath(),
+      '/auth/',
```
kaustavghosh06: Does getBackendDir take care of the /?

PR #1900, opened 2019-07-25
 feat: allow for individual resource/category push

kaustavghosh06:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/resource-status.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/resource-status.js
@@ -56,7 +56,7 @@ function capitalize(str) {
   return str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
 }
 
-function getCommonResources(resources, filteredResources) {
+function filterResources(resources, filteredResources) {
   if (!filteredResources) {
     return resources;
   }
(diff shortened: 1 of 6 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/resource-status.js
```diff
@@ -56,7 +56,30 @@ function capitalize(str) {
...
-function getAllResources(amplifyMeta, category, resourceName) {
+function getCommonResources(resources, filteredResources) {
+  if (!filteredResources) {
+    return resources;
+  }
+
+  resources = resources.filter((resource) => {
+    let common = false;
+    for (let i = 0; i < filteredResources.length; i += 1) {
+      if (filteredResources[i].category === resource.category
```
yuth: We could make (Array.Find)[https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find] to make this more readable

packages/amplify-cli/src/extensions/amplify-helpers/resource-status.js
```diff
@@ -56,7 +56,30 @@ function capitalize(str) {
   return str.charAt(0).toUpperCase() + str.slice(1).toLowerCase();
 }
 
-function getAllResources(amplifyMeta, category, resourceName) {
+function getCommonResources(resources, filteredResources) {
```
yuth: Can we name this differently. Some thing like `filterResources` or something as getCommonResource doesn't convey the meaning

Issue #1486, opened 2019-05-20
Ideas for implementation of ignoring self-signed certs.


willkara:
**Is your feature request related to a problem? Please describe.**
We need the ability to ignore self-signed certs. This is mainly found when companies have a self-signed proxy.

**Describe the solution you'd like**
Have the option to ignore self-signed certs. What we will need to discuss is how to configure that option.

- Use a CLI flag. Something like `--ignore-self-signed` or `--no-verify-ssl`

- Add a config flag in the `$HOME/.aws/credentials` file

- Add it to each individual project

**Describe alternatives you've considered**
None

**Additional context**

I already have a bug report/request at @1452. 

This fix will likely have to wait until PR @869 is finished.

My thought is to pass the config'd option to the AWS object config within the lib.

```

// We'd have to decide on the best way to configure it for each project/command/environment
const noVerifySsl = <<somehow grab the config option>>

if (httpProxy) {
  const proxyAgent = require('proxy-agent');
  aws.config.update({
    httpOptions: {
      rejectUnauthorized: noVerifySsl , // This is where we would set the correct option
      agent: proxyAgent(httpProxy)
    },
  });
}

```

willkara:
@kaustavghosh06 I can take a look at implementing this, we'd just need to settle on how to manage the command/config.

kaustavghosh06:
@willkara You can go ahead and submit a proposal and we're happy to discuss it within our team. Thank you!

willkara:
@kaustavghosh06 sounds good, I'm likely to go the route of adding a cli flag like --no-verify-ssl. I'll let you know when the PR is in.

willkara:
@kaustavghosh06 after taking in the changes from the proxy-support PR, my `rejectUnauthorized: false` code no longer works. I had it working before and only needed to add CLI tool parsing.

@AlessandroVol23 any idea what may have changed?

I'm trying something like this and no luck

```
  if (httpProxy) {
    aws.config.update({
      httpOptions: {
        agent: proxyAgent(httpProxy)
      }
    });
  }

  aws.config.update({
    httpOptions: {
      rejectUnauthorized: false
    }
  });
```

Of course the code would need some logic and tweaking, but at this point I'm mainly focusing on the mechanism itself.

AlessandroVol23:
[...]

PR #1908, opened 2019-07-26
Allow self-signed certs.

willkara:
Issue #1486 

*Description of changes:*

I added logic to search for the `--ignore-ssl` argument in the CLI tool and then set a variable in the context that'll change the http options.

It sets the rejectUnauthorized flag in httpOptions to false to allow self-signed certs.

The next plan is to add in logic to:

- Specify a location for a cert.
- Add support for profiles so a user doesn't need to use command-line arguments.

This contribution has been made by [Lockheed Martin](https://www.lockheedmartin.com/en-us/index.html)

### DIFF
--- a/packages/amplify-cli/src/cli.js
+++ b/packages/amplify-cli/src/cli.js
@@ -28,6 +28,8 @@ async function run(argv) {
 
   const context = await cli.run(argv);
 
+  context.ignoreSSL = !(argv.includes('--ignoressl') || argv.includes('--ignore-ssl'));
+
   // send it back (for testing, mostly)
   return context;
 }
--- a/packages/amplify-provider-awscloudformation/lib/system-config-manager.js
+++ b/packages/amplify-provider-awscloudformation/lib/system-config-manager.js
@@ -62,6 +62,7 @@ function setProfile(awsConfig, profileName) {
 async function getProfiledAwsConfig(context, profileName, isRoleSourceProfile) {
   let awsConfig;
   const httpProxy = process.env.HTTP_PROXY || process.env.HTTPS_PROXY;
+
   const profileConfig = getProfileConfig(profileName);
   if (profileConfig) {
     if (!isRoleSourceProfile && profileConfig.role_arn) {
@@ -86,7 +87,10 @@ async function getProfiledAwsConfig(context, profileName, isRoleSourceProfile) {
   if (httpProxy) {
     awsConfig = {
       ...awsConfig,
-      httpOptions: { agent: proxyAgent(httpProxy) },
+      httpOptions: {
+        agent: proxyAgent(httpProxy),
+        rejectUnauthorized: context.ignoreSSL,
+      },
     };
   }
 
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-pinpoint.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-pinpoint.js
@@ -43,11 +43,13 @@ async function getConfiguredPinpointClient(context, category, action, options = 
   };
 
   if (httpProxy) {
-    aws.config.update({
+    awsConfig = {
+      ...awsConfig,
       httpOptions: {
         agent: proxyAgent(httpProxy),
+        rejectUnauthorized: context.ignoreSSL,
       },
-    });
+    };
   }
 
   return new aws.Pinpoint({ ...cred, ...defaultOptions, ...options });
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws.js
@@ -5,17 +5,20 @@ const configurationManager = require('../../lib/configuration-manager');
 
 aws.configureWithCreds = async (context) => {
   const httpProxy = process.env.HTTP_PROXY || process.env.HTTPS_PROXY;
+
   const config = await configurationManager.loadConfiguration(context, aws);
   if (config) {
     aws.config.update(config);
   }
 
   if (httpProxy) {
-    aws.config.update({
+    awsConfig = {
+      ...awsConfig,
       httpOptions: {
         agent: proxyAgent(httpProxy),
+        rejectUnauthorized: context.ignoreSSL,
       },
-    });
+    };
   }
 
   return aws;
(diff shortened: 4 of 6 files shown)


willkara:
For some reason, I'm unable to get the linter to run on my machine. I'll have to take a look at my npm setup later.

I've confirmed the tests only fail the linter for changes outside of my commit.

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -86,7 +87,10 @@ async function getProfiledAwsConfig(context, profileName, isRoleSourceProfile) {
   if (httpProxy) {
     awsConfig = {
       ...awsConfig,
```
willkara: The linter seems to fail here.

kaustavghosh06: @willkara here's the lint failures:

$ eslint .

/home/circleci/repo/packages/amplify-provider-awscloudformation/src/aws-utils/aws-pinpoint.js
  46:5   error  'awsConfig' is not defined  no-undef
  47:10  error  'awsConfig' is not defined  no-undef

/home/circleci/repo/packages/amplify-provider-awscloudformation/src/aws-utils/aws.js
  15:5   error  'awsConfig' is not defined  no-undef
  16:10  error  'awsConfig' is not defined  no-undef

To fix them, from the root of your amplify cli repo - you can run run `yarn lint-fix` or `npm run lint-fix`

willkara: Gotcha, I'll try that on my Linux machine as Windows was giving me issues for some reason.

willkara: Everytime I try running it I get an error:
```

willkara@NORMANDY:~/amplify-cli$ eslint .

Oops! Something went wrong! :(

ESLint: 6.1.0.

ESLint couldn't find a configuration file. To set up a configuration file for this project, please run:

    eslint --init

ESLint looked for configuration files in /home/willkara/amplify-cli and its ancestors. If it found none, it then looked in your home directory.

If you think you already have a configuration file or if you need more help, please stop by the ESLint chat room: https://gitter.im/eslint/eslint
```

This is from a fresh clone and install.

```
willkara@NORMANDY:~/amplify-cli$ yarn lint-fix
yarn run v1.13.0
$ lerna run lint-fix
lerna notice cli v3.10.8
........
.........
$ eslint . --fix
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

lerna ERR! yarn run lint-fix stderr:

Oops! Something went wrong! :(

ESLint: 6.1.0.

ESLint couldn't find the config "airbnb-base" to extend from. Please check that the name of the config is correct.

The config "airbnb-base" was referenced from the config file in "/home/willkara/amplify-cli/packages/amplify-category-api/.eslintrc.json".

If you still have problems, please stop by https://gitter.im/eslint/eslint to chat with the team.

error Command failed with exit code 2.
```

willkara: So I had to init eslint but now it's giving me TONS of errors. Is there not a way to point to the packages eslint config?

willkara: Still trying but every-time I run the linter, it's given me tons of other line-ending and other errors, even though I didn't change anything else.

@UnleashedMind / @kaustavghosh06  any ideas?

willkara: @UnleashedMind / @kaustavghosh06 , hey guys any ideas on this? I've tried multiple linters/runs of it but it's giving me the same errors. I can upload a file of them if you'd like.

packages/amplify-cli/src/cli.js
```diff
@@ -28,6 +28,8 @@ async function run(argv) {
 
   const context = await cli.run(argv);
 
+  context.ignoreSSL = !(argv.includes('--ignoressl') || argv.includes('--ignore-ssl'));
+
```
UnleashedMind: This won't work, the cli.run(argv) won't carries out the actual execution of the command logic. 
Adding something to the context right before the return won't have any effect on the command execution. 
And why is there a `!` ?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-pinpoint.js
```diff
@@ -43,11 +43,13 @@ async function getConfiguredPinpointClient(context, category, action, options =
   };
 
   if (httpProxy) {
-    aws.config.update({
+    awsConfig = {
+      ...awsConfig,
       httpOptions: {
         agent: proxyAgent(httpProxy),
+        rejectUnauthorized: context.ignoreSSL,
```
UnleashedMind: the rejectUnauthorized field should be in the agent, instead of the httpOptions

packages/amplify-cli/src/cli.js
willkara: Wouldn't this grab the flag from the cli and set it into the context? As for the !, I use that to translate between the **true** the user states and the **false** required for rejectUnauthorized.

Make it easier for the user to set the options.

packages/amplify-provider-awscloudformation/src/aws-utils/aws-pinpoint.js
willkara: Unless I'm misreading the API-Docs, there's no option for that in the agent constructor. You can only set them in the options themselves?

https://nodejs.org/api/https.html#https_https_request_url_options_callback

packages/amplify-provider-awscloudformation/src/aws-utils/aws.js
```diff
@@ -5,17 +5,20 @@ const configurationManager = require('../../lib/configuration-manager');
...
 aws.configureWithCreds = async (context) => {
   const httpProxy = process.env.HTTP_PROXY || process.env.HTTPS_PROXY;
+
   const config = await configurationManager.loadConfiguration(context, aws);
   if (config) {
     aws.config.update(config);
   }
 
   if (httpProxy) {
-    aws.config.update({
```
UnleashedMind: this line shouldn't be removed, otherwise how is the aws configuration updated with the proxy?

packages/amplify-cli/src/cli.js
UnleashedMind: if you want to access the input options such as '--ignoressl' in your code, try to use
`context.parameters.options`

again, please test your implementation and make sure it works end-to-end.

willkara:
Hey sorry for the late reply.

I've tested this on my company's internal network where it works and also on regular connections where it also works.

willkara:
If you want I can re-implement using the context parameters.options method.

UnleashedMind:
Close this PR because the CLI went through a major platform change since the PR was originally created.
If interested, please send a new PR for this.

Issue #1043, opened 2019-03-14
RFC - @auth directive improvements


mikeparisstuff:
# RFC - Transform Auth Improvements

This document will outline designs for a number of new features relating to authentication & authorization within the GraphQL Transform. The goal of these features is too fill holes and introduce new mechanisms that make protecting your valuable information easier.

# Proposal 1: Replace 'queries' and 'mutations' arguments with 'operations'

Currently an @auth directive like this:

```graphql
type Task @model @auth(rules: [{allow: owner}]) {
    id: ID!
    title: String
    owner: String
}
```

causes these changes to the following resolvers:

1. *Query.getTask* - Returns the post only if the logged in user is the post's owner.
2. *Query.listTasks* - Filter items such that only owned posts are returned.
3. *Mutation.createTask* - If an owner is provided via $ctx.args.input.owner and matches the identity of the logged in user, succeed. If no owner is provided, set logged in user as the owner, else fail.
4. *Mutation.updateTask* - Append a conditional expression that will only update the record if the logged in user is its owner.
5. *Mutation.deleteTask* - Append a conditional expression that will only delete the record if the logged in user is its owner.

In other words, the @auth directive currently protects the root level query & mutation fields that are generated for an @model type.

### Problem: The 'queries' and 'mutations' arguments imply top level protection

GraphQL APIs are a graph and we need to be able to define access rules on any field, not just the top level fields.

### Solution

I suggest replacing the *queries* and *mutations* arguments on the @auth directive with a single *operations* argument. This would be the new @auth directive definition.

```graphql
directive @auth(rules: [AuthRule!]!) on OBJECT
input AuthRule {
    allow: AuthStrategy!
    ownerField: String # defaults to "owner"
    identityField: String # defaults to "cognito:username"
    groupsField: String
    groups: [String]

    # The new argument
    operations: [ModelOperation]

    # Old arguments
    queries: [ModelQuery] @deprecated(reason: "The 'queries' argument will be deprecated in the future. Please replace this argument with the 'operations' argument.")
    mutations: [ModelMutation] @deprecated(reason: "The 'mutations' argument will be deprecated in the future. Please replace this argument with the 'mutations' argument.")
}
enum AuthStrategy { owner groups }

# The new enum
enum ModelOperation { create update delete read }

# The old enums
enum ModelQuery { get list }
enum ModelMutation { create update delete }
```

[...]

PR #1916, opened 2019-07-28
feat: implement multi-auth functionality

attilah:
Issue #, if available:
- Various proposals of #1043
- Enable #1485

Description of changes:
- Enable the usage of new AppSync multi-auth directives in the GraphQL schema directly.
- Enhance CodeGen to support multi-auth enabled AppSync APIs.
- Update AppSync API walkthrough to support multi-auth configuration as advanced settings.
- Extend ```@auth``` directive definition with new ```AuthStrategy``` members: ```public``` and ```private```.
- Extend ```@auth``` directive with ```provider``` property to enable default provider overrides. Providers are: ```userPools```, ```apiKey```, ```oidc```, ```iam```.
- Extend ```@auth``` with support for ```owner``` use with OpenID Connect.
- Introduce ```CreateAPIKey``` parameter to automatically control API key creation and deprecate - but still handle - ```APIKeyExpirationEpoch```.


packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -294,58 +443,105 @@ Static group authorization should perform as expected.`
...
-            // are done before calling PutItem.
-            const dynamicGroupAuthorizationExpression = this.resources.dynamicGroupAuthorizationExpressionForCreateOperationsByField(
-                dynamicGroupAuthorizationRules,
-                field.name.value
-            )
-            const fieldIsList = (fieldName: string) => {
-                const field = parent.fields.find(field => field.name.value === fieldName);
-                if (field) {
-                    return isListType(field.type);
+                const throwIfUnauthorizedExpression = this.resources.throwIfUnauthorized()
```
mikeparisstuff: Can you please add some comments throughout this section that give more details and context about what each section is doing? This is getting pretty difficult to decipher and any future devs will appreciate it.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -721,73 +1050,110 @@ All @auth directives used on field definitions are performed when the field is r
...
-                    field && ifCondition ?
-                        iff(ifCondition, authorizationLogic) :
-                        authorizationLogic
-                ),
-                resolver.Properties.RequestMappingTemplate
-            ]
-            resolver.Properties.RequestMappingTemplate = templateParts.join('\n\n')
-            ctx.setResource(resolverResourceId, resolver)
+                if (authModesToCheck.size > 0) {
+                    expressions.push (this.resources.getAuthModeDeterminationExpression(authModesToCheck));
```
mikeparisstuff: Do you mean to be calling `getAuthModeDeterminationExpression` twice here and again on line 1143?

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -884,6 +1256,221 @@ All @auth directives used on field definitions are performed when the field is r
...
+
+        //
+        // We only add a directive if it is not the default auth or
+        // if it is the default one, but there are other rules for a
+        // different provider.
+        // For fields we don't add the default, since it would open up
+        // the access rights.
+        //
+
+        if ((this.configuredAuthProviders.default !== 'apiKey' &&
```
mikeparisstuff: I see duplicated code. Can we refactor this and parameterize over the expected auth provider string. You could define a helper closure inline.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -689,4 +775,117 @@ identityField: "${rule.identityField || DEFAULT_IDENTITY_FIELD}" }`
...
+        if (!authProviders || authProviders.size === 0) {
+            return comment(`No authentication mode determination needed`);
+        }
+
+        const expressions = [];
+
+        for (const authProvider of authProviders) {
+
+            if (authProvider === 'userPools') {
+                const userPoolsExpression = iff(
```
mikeparisstuff: This looks unnecessarily complicated to me. At the resolver layer do we need to do anything different when handling OIDC & Cognito? Both of  those auth mechanisms will have similar things and the current @auth owner & group rules use values in `identity.claims()` which is included in both auth modes. In other words, I don't think you need to do anything different at the resolver layer to authorize OIDC vs User Pools. When I use `{ allow: owner, identityField: "sub", provider: OIDC }`, the resolver logic is no different than if I did `{ allow: owner, identityField: "sub", provider: userPools }`. If you needed to you could probably just look for the existence of "sub" & "issuer" to determine that there was a valid JWT provided was provided.

packages/graphql-auth-transformer/tsconfig.json
```diff
@@ -1,11 +1,10 @@
 {
     "compilerOptions": {
-        "target": "es5",
+        "target": "es6",
```
mikeparisstuff: Why are you making this change? I am not saying it won't work but we have customers that run old versions of node locally and some es6 features may not be supported in those environments. Unless we have a compelling reason lets not make these changes without really thinking through it.

packages/graphql-auth-transformer/src/graphQlApi.ts
```diff
@@ -0,0 +1,90 @@
...
+    OpenIDConnectConfig?: OpenIDConnectConfig
+    UserPoolConfig?: UserPoolConfig
+    Tags?: Tags
+    Name: Value<string>
+    AuthenticationType: Value<string>
+    LogConfig?: LogConfig
+    AdditionalAuthenticationProviders?: AdditionalAuthenticationProviders
+}
+
+export default class GraphQLApi extends ResourceBase {
```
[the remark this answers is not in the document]
attilah: At the top of the file I added a TODO for this reason to track the removal of the file. The latest stable release of cloudform does not have support for the new AppSync multi-auth properties, so cannot be used and 4.0 has structural changes which would require significant changes and if we sync all the packages to cloudform 4.0 that should be a separate PR, so out of scope for this. I had to create a hybrid version of this file it is conforming to the latest stable cloudform one, but taking in the new properties for AppSync.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -33,6 +30,25 @@ export class ResourceFactory {
...
+            }),
+            [ResourceConstants.PARAMETERS.APIKeyExpirationEpoch]: new NumberParameter({
+                Description: 'The epoch time in seconds when the API Key should expire.' +
+                    ' Setting this to 0 will default to 180 days from the deployment date.' +
+                    ' Setting this to -1 will not create an API Key.',
+                Default: 0,
+                MinValue: -1
+            }),
+            [ResourceConstants.PARAMETERS.CreateAPIKey]: new NumberParameter({
+                Description: 'The boolean value to control if an API Key will be created or not.' +
```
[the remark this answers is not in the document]
attilah: IMHO using the epoch for two different things, specifying absolute expiration and disabling creation of API key is confusing for customers, and since now we've a way of configuring API key expiration during the service walkthrough and since API key creation depends on the configured auth providers I added this specific property to control this and also marked with a warning ```APIKeyExpirationEpoch``` deprecated, but we're translating it into CreateAPIKey when possible.

packages/amplify-category-api/commands/api/add-datasource.js
```diff
@@ -139,7 +139,36 @@ module.exports = {
         return datasource;
       })
       .then((datasourceName) => {
-        context.amplify.executeProviderUtils(context, 'awscloudformation', 'compileSchema', { noConfig: true, forceCompile: true });
+        const amplifyMetaFilePath = context.amplify.pathManager.getAmplifyMetaFilePath();
+        const amplifyMeta = context.amplify.readJsonFile(amplifyMetaFilePath);
```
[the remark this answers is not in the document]
attilah: Good catch, moved to ```transformGraphQLSchema```.

packages/amplify-provider-awscloudformation/lib/upload-appsync-files.js
```diff
@@ -43,16 +43,76 @@ async function uploadAppSyncFiles(context, resourcesToUpdate, allResources, opti
     return deploymentRootKey;
   };
 
+  const getApiKeyConfigured = () => {
+    const projectDetails = context.amplify.getProjectDetails();
```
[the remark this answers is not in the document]
attilah: The duplicated code was eliminated since it was moved to a method where no ```projectDetails``` digging needed since we already have the AppSync API available filtered, so this snippet became a non-duplicate.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
attilah: Refactored.

packages/graphql-auth-transformer/src/resources.ts
attilah: For this check I exactly checking what the AppSync service is doing to determine the authentication type. By raising this question I was wondering if the existing code supports this:
```
{ allow owner, identityField: "sub", provider: oidc, ownerField: "oidcOwner"  }
{ allow owner, identityField: "sub", provider: userPools}
```
Because it became a possible scenario and it do requires to differentiate between ```userPools``` and ```oidc``` since unique ```sub```s are not guaranteed between different identity providers.

packages/graphql-auth-transformer/tsconfig.json
attilah: We're transpiling to ES5 with Babel and the new ```cloudform``` will throw: ```TypeError: Class constructor ResourceBase cannot be invoked without 'new'``` errors. Since Node 8 - which is the minimum version to run the CLI - supports ES6 natively and it would simplify the CLI code by removing unneeded polyfills from the released code.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
attilah: No, on line 1143 I'm calling ```getAuthModeCheckWrappedExpression```. The new resolver logic has 2 parts:
- determine the authMode from the request
- wrap the existing owner/group etc checks into an ```if``` block if authMode is not matching the one we need compare against.

in 1131 I'm adding the authMode check to the start of the resolver, later the ```if``` block.

packages/graphql-auth-transformer/src/__tests__/MultiAuth.test.ts
```diff
@@ -0,0 +1,282 @@
...
+            ownerAuthDirective,
+            userPoolsDefaultConfig
+        );
+    });
+
+    test(`When provider is not the same as default directive is added`, () => {
+        transformTest(
+            ownerAuthDirective,
+            withAuthModes(apiKeyDefaultConfig, ['AMAZON_COGNITO_USER_POOLS']),
+            [userPoolsDirectiveName, apiKeyDirectiveName]
```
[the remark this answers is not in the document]
attilah: Yes, [this test](https://github.com/aws-amplify/amplify-cli/pull/1916/files/edd72107df356784937a686719aef3ec1c00ce83#diff-b54b4ab3b559862d3fb2a05514489d1fR241) does exactly that.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
attilah: Done.

j-r-t:
any updates on when this is going to be merged/released? this seems like a really important PR

iwaduarte:
And this is a year old issue. :(
https://github.com/aws-amplify/amplify-js/issues/711

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -20,13 +20,6 @@ const parametersFileName = 'parameters.json';
 const schemaFileName = 'schema.graphql';
 const schemaDirName = 'schema';
 
-function failOnInvalidAuthType(usedDirectives, opts) {
-  if (usedDirectives.includes('auth') && !opts.isUserPoolEnabled) {
```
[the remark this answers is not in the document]
attilah: We've a more sophisticated validation because of the multi flags in another method.

packages/amplify-frontend-ios/lib/frontend-config-creator.js
```diff
@@ -33,7 +33,7 @@ function getAWSConfigObject(amplifyResources) {
   const { serviceResourceMapping } = amplifyResources;
   const configOutput = {
     UserAgent: 'aws-amplify/cli',
-    Version: '0.1.0',
+    Version: '1.0',
```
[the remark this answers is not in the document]
attilah: I think that it would be nice to send the correct version so perhaps metrics from service side could reflect the used CLI version and help with potential troubleshooting, but I reverse it to 0.1.0.

aej11a:
This looks great, currently have projects waiting on these feature. Will be a **huge** step forward for Amplify when it gets merged.

thedgbrt:
Anxiously awaiting

ohsik:
Any updates on when is this branch going to be merged? :)

ahansson89:
Did this make it into 3.20?

CodySwannGT:
@ahansson89 [looks to be that way](https://aws-amplify.github.io/docs/cli-toolchain/graphql?sdk=js)

packages/amplify-frontend-ios/lib/frontend-config-creator.js
```diff
@@ -261,32 +261,45 @@ function getAppSyncConfig(appsyncResources, projectRegion) {
...
   };
+
   if (testMode) {
     result.AppSync.Default.DangerouslyConnectToHTTPEndpointForTesting = true;
   }
 
-  const additionalAuths = appsyncResource.output.additionalAuthenticationProviders || [];
-  additionalAuths.forEach((authType) => {
-    const apiName = `${appsyncResource.resourceName}_${authType}`;
+  const additionalAuths = appsyncResource.output.authConfig.additionalAuthenticationProviders || [];
```
rb365: so I'm trying to initialize environment with new cli version and it fails on line 293, can anyone help? why is this not backward compatible?

$ amplify env add
Note: It is recommended to run this command from the root of your app directory
? Do you want to use an existing environment? Yes
? Choose the environment you would like to use: production
Using default provider  awscloudformation
✔ Initialized provider successfully.
✖ There was an error initializing your environment.
init failed
TypeError: Cannot read property 'additionalAuthenticationProviders' of undefined
    at getAppSyncConfig (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-frontend-ios/lib/frontend-config-creator.js:293:61)

rb365: @attilah

dtelaroli:
@attilah

This feature it's not working with big schema model, because it adds all permissions to the same policy document.

Any workaround?

UPDATE_FAILED   AuthRolePolicy AWS::IAM::Policy     Fri Nov 29 2019 12:58:40 GMT-0200 (Brasilia Summer Time) Maximum policy size of 10240 bytes exceeded for role amplify-appsync-dev-authRole (Service: AmazonIdentityManagement; Status Code: 409; Error Code: LimitExceeded; Request ID: cc7b0f1c-c470-42a9-98cc-e515db4a1f29)

attilah:
@dtelaroli this is something I'm actively working on, we're moving to ManagedPolicies, which has a size restriction of 6144 bytes, but 10 of these can be attached to a role, and that limit can be raised to 20 by support. This enables fine grained access control up to ~120kb of policy size.

Issue #659, opened 2019-01-04
support TypeScript for backend functions


troygoode:
**Is your feature request related to a problem? Please describe.**

I use the TypeScript generation features for my front-end client, but am frustrated that any of the Lambda functions I create via `amplify add api` or `amplify add function` must be written in vanilla JavaScript. This also makes it very difficult to share code between the backend and frontend (note that many of the auto-generated files created by `amplify cli` use syntax like `export default blah` that is invalid for use in the backend files).

**Describe the solution you'd like**

Switch to using `tsc` to compile the files found in `amplify/backend/function/*/src/` so that we can opt-in to type-safety by changing our file extension from `.js` to `.ts` or `.tsx`, similar to how Create React App has handled this in version 2.0.0+.

**Describe alternatives you've considered**

Alternatively, add (or document) how we can hook into the build process for `amplify/backend/function/*` so we can inject a TypeScript transpilation step ourselves.

awjreynolds:
Really want this feature. I know I'm not adding much to the conversation but there needs to be options within the cli to scaffold typescript outputs for all aspects of an amplify solution.

ajhool:
Have you considered using the "postinstall" hook in npm? I have not tried using it but am about to give it a shot.

Amplify uses a "amplify function build" command and the docs say it calls `npm install` under the hood. The npm docs say that `npm run postinstall` is called automatically after `npm install`, so it looks promising. This might require installing typescript + build tools on the build machine without utilizing the function's package.json, but that's not too bad.

ajhool:
This looks like a promising approach but I have a stumbling block that I haven't been able to figure out.

This might look like a long procedure, but it should take about 2 minutes and once it is configured, it should be easy to maintain.

**WARNING: THIS WILL OVERWRITE CODE IN YOUR FUNCTION'S "src" directory (eg. `amplify/backend/function/myFunction/src`)**

Strategy:

1. Create a sister directory to `amplify/backend/function/myFunction/src` called `function/myFunction/tsc-src`
2. Create a file at `function/myFunction/src/postinstall.sh`:

```
#!/bin/bash

cd ../tsc-src
npm run build
```
3. Make postinstall.sh executable (on my mac the command is): `chmod u+x postinstall.sh`

4. Update `function/myFunction/src/package.json` with the postinstall script:

```
{
  "name": "myFunction",
  "version": "2.0.0",
  "description": "Lambda function generated by Amplify",
  "main": "index.js",
[...]

PR #1951, opened 2019-08-01
feat(amplify-provider-awscloudformation): hooks

visusnet:
*Related issue:* #659

*Description of changes:*
You can now add a custom build script to the main package.json file which is executed before a function is deployed. This can be used to transpile code with babel/typescript or to copy files (like src/graphql/* to amplify/backend/function/mySpecialFunction/src/graphql).

The script must be defined in the root package.json file because npm which is executed by buildResource() will ignore any scripts defined in the resource's package.json. The name must follow the convention `amplify:<resource name>`.

**Example: Transpiling with Babel**

Add the hook to the `scripts` section of `package.json` (in the project root folder) and add dependencies:
```json
{
  "scripts": {
    "amplify:mySpecialFunction": "cd amplify/backend/function/mySpecialFunction; babel --root-mode upward lib -d src; cd -"
  },
  "devDependencies": {
    "babel-cli": "^6.26.0",
    "babel-preset-env": "^1.7.0"
  }
}
```

Add a `.babelrc` to `amplify/backend/function/mySpecialFunction`:

```json
{
  "presets": [
    [
      "env",
      {
        "exclude": ["transform-regenerator"],
        "targets": {
          "node": "8.10"
        }
      }
    ]
  ],
  "plugins": [
    "transform-async-to-generator",
    "transform-exponentiation-operator",
    "transform-object-rest-spread"
  ]
}
```
That's basically it. Just add the code which should be transpiled to `amplify/backend/function/mySpecialFunction/lib`. The next `amplify push` will automatically execute the `amplify:mySpecialFunction`-script which will in turn transpile the code before it is uploaded to AWS Lambda.

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/build-resources.js
+++ b/packages/amplify-provider-awscloudformation/lib/build-resources.js
@@ -2,6 +2,7 @@ const fs = require('fs-extra');
 const path = require('path');
 const archiver = require('archiver');
 const { hashElement } = require('folder-hash');
+const childProcess = require('child_process');
 
 async function run(context, category, resourceName) {
   const { allResources } = await context.amplify.getResourceStatus(category, resourceName);
@@ -16,6 +17,7 @@ async function run(context, category, resourceName) {
 async function buildResource(context, resource) {
   const { category, resourceName } = resource;
   const backEndDir = context.amplify.pathManager.getBackendDirPath();
+  const projectRoot = context.amplify.pathManager.searchProjectRootPath();
   const resourceDir = path.normalize(path.join(backEndDir, category, resourceName, 'src'));
   const packageJsonPath = path.normalize(path.join(backEndDir, category, resourceName, 'src', 'package.json'));
   const packageJsonMeta = fs.statSync(packageJsonPath);
@@ -28,10 +30,11 @@ async function buildResource(context, resource) {
     !resource.lastBuildTimeStamp ||
     new Date(packageJsonMeta.mtime) > new Date(resource.lastBuildTimeStamp)
   ) {
-    const npm = /^win/.test(process.platform) ? 'npm.cmd' : 'npm';
-    require('child_process').spawnSync(npm, ['install'], { cwd: resourceDir });
+    installDependencies(resourceDir);
     context.amplify.updateamplifyMetaAfterBuild(resource);
   }
+
+  runBuildScriptHook(resourceName, projectRoot);
 
   if (
     !resource.lastPackageTimeStamp ||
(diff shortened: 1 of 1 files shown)


kaustavghosh06:
@visusnet Could you please rebase your branch with the master? Thank you.

visusnet:
@kaustavghosh06 Done! :)

undefobj:
@visusnet @kaustavghosh06 can we add an update to the docs under "usage" for this functionality: https://aws-amplify.github.io/docs/cli-toolchain/usage?sdk=js

visusnet:
@undefobj Sure, where exactly on that page would you expect it? Would you want a general usage description (i.e. you can add `amplify:<resource name>` to your package.json and it will be executed) or a more specific example (like the Babel example above)?

undefobj:
> @undefobj Sure, where exactly on that page would you expect it? Would you want a general usage description (i.e. you can add `amplify:<resource name>` to your package.json and it will be executed) or a more specific example (like the Babel example above)?

Lets create a new top-level section right before "Architecture" called "Build options" and then a sub section in that called "Functions" with this information. That way if we introduce new build & deploy tooling for other categories we can introduce it there.

visusnet:
@undefobj https://github.com/aws-amplify/docs/pull/823

undefobj:
@kaustavghosh06 @yuth @UnleashedMind @SwaySway @attilah Lets please look at this PR for our next bug bash and/or release. Docs are ready to go if you think this PR can be merged.

kanerogers:
Super excited about this one! Thanks @visusnet!

packages/amplify-provider-awscloudformation/lib/build-resources.js
```diff
@@ -28,11 +33,14 @@ async function buildResource(context, resource) {
     !resource.lastBuildTimeStamp ||
     new Date(packageJsonMeta.mtime) > new Date(resource.lastBuildTimeStamp)
   ) {
-    const npm = /^win/.test(process.platform) ? 'npm.cmd' : 'npm';
-    require('child_process').spawnSync(npm, ['install'], { cwd: resourceDir });
+    childProcess.spawnSync(npm, ['install'], { cwd: resourceDir });
```
attilah: It would be nice to encapsulate the process spawn into a function and handle non-success results of running . NPM. In case of an error I think the command should fail and should not update the meta file and run a script. The same function could be reused below.

packages/amplify-provider-awscloudformation/lib/build-resources.js
```diff
@@ -28,11 +33,14 @@ async function buildResource(context, resource) {
...
     context.amplify.updateamplifyMetaAfterBuild(resource);
   }
 
+  if (mainPackageJsonContents.scripts && mainPackageJsonContents.scripts[scriptName]) {
+    childProcess.spawnSync(npm, ['run-script', scriptName], { cwd: projectRoot });
```
attilah: Same as above.

packages/amplify-provider-awscloudformation/lib/build-resources.js
RossWilliams: A nice improvement would be to support yarn here. Bad one-liner below.
```javascript
fs.existsSync('./yarn.lock') ? (/^win/.test(process.platform) ? 'yarn.cmd' : 'yarn')  : npm
```

attilah: I agree, that would be a nice addition to this PR. @visusnet, please add it to the list of requested changes.

visusnet: Done.

visusnet:
@attilah @RossWilliams I have updated the branch with the requested changes:
- It detects whether to use yarn or npm.
- It throws an error if the called script fails. The promise returned by buildResource will automatically reject in this case and the error message will be logged into the console.

markwhitfeld:
Was just wrestling with this today and found this PR! Can't wait.

markwhitfeld:
I was just thinking, wouldn't it be good to also add the execution of a `build` npm/yarn script in each resources' base folder?
Then, for resources that have a `package.json`, their `build` script stays in their folder and you won't need that long `cd amplify/...`.
The root `amplify:{resourceName}` hook would still be useful for other resources that need a script to run but they haven't got their own `package.json`.

visusnet:
@markwhitfeld I was trying to implement this at first but it turned out to be more difficult to implement. I think this is a valid use case but outside of the scope of this PR. I might give it a try after this PR has been merged.

markwhitfeld:
@kaustavghosh06 Any idea on when the next npm release will be?
Really excited to try this out!

PR #1982, opened 2019-08-07
Local mock fix

yuth:
*Issue #, if available:*

*Description of changes:*


packages/amplify-velocity-template/.gitignore
```diff
@@ -7,3 +7,4 @@ build
 .idea/*
 dist/
 coverage
+src/parse/index.js
```
kaustavghosh06: Will this work if not published?

yuth: This file gets generated by gison at build time. we don't need to track this

Issue #1599, opened 2019-03-26
Amplify hosting with CloudFront and S3 still allows direct access to the S3 bucket with HTTP


maitreya1975:
**Describe the bug**
Amplify Hosting (using the CLI with 'amplify add hosting') sets up CloudFront and S3 bucket but still allows direct access to the objects in S3.

**To Reproduce**
Create an app following the getting started guide.

1. amplify add hosting
1a - choose the PROD option for CloudFront and S3 (not the DEV option)

2. amplify push

3. amplify publish 

Now you can visit the cloudfront URL displayed by the output of the publish command just fine and it is a HTTPS URL.

However you can also craft the URL for the underlying hosting S3 bucket and view objects directly, using both HTTP and HTTPS.

The reason is that the objects are uploaded with ACL Public Read and this allows direct access. This does not match best practices for using S3 with CloudFront as described here: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/using-https-cloudfront-to-s3-origin.html

Also, when S3 is used via CloudFront, static website hosting should be disabled so that all access is only allowed from CloudFront via HTTPS.

**Expected behavior**
The hosting S3 bucket is configured so that only CloudFront can access the static website.
Users trying to hit the S3 URLs directly with HTTP/HTTPS are denied access.

**Screenshots**
NA.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: firefox
 - Version: 66.0.1

**Smartphone (please complete the following information):**

**Additional context**

UnleashedMind:
#2369 addressed this.

PR #2016, opened 2019-08-12
Removed public-read acl from amplify publish method

abdsahin:
*Issue #1599 

*Description of changes: As mentioned [here](https://aws.amazon.com/blogs/security/iam-policies-and-bucket-policies-and-acls-oh-my-controlling-access-to-s3-resources/), S3 ACLs is a legacy access control mechanism that predates IAM. I don't see the value of making each individual object public through ACL since bucket policy already does that. Therefore I removed the acl public read from the upload parameters.


packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
```diff
@@ -58,7 +61,6 @@ async function uploadFile(s3Client, hostingBucketName, distributionDirPath, file
     Key: relativeFilePath,
     Body: fileStream,
     ContentType: contentType || 'text/plain',
-    ACL: 'public-read',
```
UnleashedMind: File uploader is used in both the DEV and PROD setup, in the DEV setup, there is not CloudFront, and the S3 hosting bucket itself is private, if we don't set the ACL for the uploaded items, the contents won't be accessible. 
We need to distinguish between these two setup when trying to resolve this issue.

abdsahin: I performed further updates on the repository and tested several options, add S3, turn into CloudFront, etc... The only thing that I am concerned now it the "Migrate" function. Can you give me some pointers if there are still issues.

Thanks

UnleashedMind:
This is implemented by the PR https://github.com/aws-amplify/amplify-cli/pull/2369

PR #2034, opened 2019-08-13
fix: move test package dependencies to devDependencies, fixes #2017

### LINKED ISSUE #2017: Amplify-category-auth uses Jest as normal dependency
**Describe the bug**
Amplify-category-auth project contains Jest as a normal dependency instead of it being just a devDependency.

**To Reproduce**
1. Go to https://github.com/aws-amplify/amplify-cli/blob/8bbb701317f6d6df78afed250e1e40e3c54c3b86/packages/amplify-category-auth/package.json

**Expected behavior**
Jest should be provided as a devDependency.

attilah:
*Issue #, if available:*

Fix #2017

*Description of changes:*

Move test related packages to devDependencies where needed.


packages/amplify-category-predictions/package.json
```diff
@@ -8,11 +8,9 @@
     "amplify-category-auth": "1.7.1",
     "chalk": "^2.4.1",
     "chalk-pipe": "^2.0.0",
-    "commitizen": "^3.0.2",
```
kaustavghosh06: Why re we removing this? @SwaySway  Is this not getting used?

attilah: Because the root ```package.json``` has it and it is configured there.

https://github.com/aws-amplify/amplify-cli/blob/master/package.json#L57

Issue #1766, opened 2019-07-01
I am subscribing to events and receiving data I am not authorized to see


rawadrifai:
** Which Category is your question related to? **
Amplify subscriptions

** What AWS Services are you utilizing? **
Amplify, AppSync, DynamoDB

** Provide additional details e.g. code snippets **
I am wondering what's the expected behavior of a subscription generated with codegen.
If I am subscribed to onCreateSomething, and then someone else adds a record that I am not authorized to read (based on owner field). Am I supposed to still receive the event data? I could be doing something wrong, but right now all subscribed clients receive a copy of every event authorized or not. I hope this is not the expected behavior.

licalac:
Hi there,

Just experiencing this security breach myself with a very simple app.

I have to add that @searchable directive is also concerned, results from searchQuery are not @auth complients
This securities breaches are for me as critical as surprising in AWS cloud, because I remember the AWS's CTO was saying that every AWS developer is at first a security engineer ... I thought it was true :(
So disappointing and looking forward for a fast and serious response.

Best regards
Lionel

kaustavghosh06:
@rawadrifai @licalac What does your annotated schema look like?

licalac:
Hi,

Many thanks for your reply.

Mine is as follow:

type Deck @model @auth(rules: [{ allow: owner }]) {  
  id: ID!
  name: String!
  cards: [Card]! @connection(name: "DeckCards")
}

type Card @model @searchable @auth(rules: [{ allow: owner }]) {  
  id: ID!
  question: String!
  answer: String
  deck: Deck! @connection(name: "DeckCards")
}

When I make a subscription OnCreateCard and then receive all new card created by whatever user.

Lionel

mikeparisstuff:
This behavior has been outlined in the #1043. You have the ability to fine protect subscriptions yourself but it requires custom logic. You have the option to a) turn off subscriptions via the `@model(subscriptions: null)` argument or b) add custom authorization logic yourself using custom resources. To do so you add a resolver to the subscription field that implements your authorization check. The resolver will be run at connect time (i.e. when you first issue the `subscription` query) and allows you to implement fine-grained authorization.

rawadrifai:
@kaustavghosh06 the schema can be anything, I don't think it's specific to me.
Take this for instance:
```
type Something @model @auth(rules: [
  { allow: owner }
]) {
  id: ID!
  owner: String
}
```

[...]

Issue #868, opened 2019-02-13
Feature Request: Add ability for custom stacks to reference each other's output


lookitsatravis:
**Is your feature request related to a problem? Please describe.**

The issue that I'm facing is that I'd like to be able to do the following:

1. Using a stack configuration, create a custom data source
2. Using another stack configuration, create a custom resolver which uses data source defined in stack 1.
3. Continues on for as many stacks might need to reference stack 1's resources.

In order to do this, though, subsequent stacks would need to wait on the first stack to finish so that the resource exists. It is possible to work around this limitation by adding multiple data sources pointing to the same underlying resource, but it would be fantastic to be able to create one data source (or role or whatever) and share it amongst many other stacks. For my particular use case, I have a Lambda function which does some data validation not possible in VTL templates. I'd like to just create that Data Source once.

**Describe the solution you'd like**

Normally this is what `DependsOn` would be used for, but I don't believe it works across stack templates (or at least I couldn't get it to). I'd like to be able to Output a value from a custom stack, and then add a Parameter to a sibling custom stack which references the output value. According to option 2 of [this answer](https://stackoverflow.com/a/49183967/1096391) on Stack Overflow, an implicit dependency would be created and order of operations would work correctly. In order for this to work properly, the parent stack would need to be able to supply the child stack with the proper reference as a Parameter value.

Example:

```js
// custom stack 1

{
    "Resources": {
        "ThisThing": {
            "Name": "FooThing"
        }
    },
    "Outputs": {
        "FooName": {
            "Value": {
                "Fn::GetAtt": [
                    "ThisThing",
                    "Name"
                ]
            }
        }
    }
}
```

```js
// custom stack 2
{
    "Parameters": {
        "FooThingName" : {
            "Type" : "String",
            "Description" : "The name of the foo thing"
        }
    }
}
```

```js
// parent stack (auto-generated)

{
    "Resources": {
        "ChildStack2": {
            "Type" : "AWS::CloudFormation::Stack",
            "Properties" : {
                "Parameters" : {
                    "FooThingName" : { "Fn::GetAtt" : "Outputs.FooThingName" }
                }
            }
        }
    }
}
```

**Describe alternatives you've considered**

There really isn't a work around to cross-stack dependencies without the parent stack passing data between children as far as I've found.

**Additional context**
N/A

kaustavghosh06:
@lookitsatravis So are these stacks used as a custom stack as a part of your AppSync API?

[...]

PR #2068, opened 2019-08-16
Auth sub

SwaySway:
*Issue #, if available:*
re #1766 #1043

*Description of changes:*
adding protection on subscriptions

BREAKING CHANGE: If an owner is included in the auth directive it will either be a requirement if it's the only rule or an optional input if used with other rules


packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -276,6 +275,69 @@ groupsField: "${rule.groupsField || DEFAULT_GROUPS_FIELD}" }`
...
+        variableToCheck: string = 'ctx.args',
+        variableToSet: string = ResourceConstants.SNIPPETS.IsOwnerAuthorizedVariable,
+        formatComment?: (rule: AuthRule) => string,
+    ) {
+        let ownershipAuthorizationExpressions = []
+        let ruleNumber = 0;
+        for (const rule of rules) {
+            const ownerAttribute = rule.ownerField || DEFAULT_OWNER_FIELD
+            const rawUsername = rule.identityField || DEFAULT_IDENTITY_FIELD
+            const isUsern = isUsername(rawUsername)
```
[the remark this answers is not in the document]
SwaySway: using the abbreviation for isUsern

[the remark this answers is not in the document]
attilah: Please do not.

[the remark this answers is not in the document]
SwaySway: Changed the naming convention

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -172,6 +176,7 @@ export class ModelAuthTransformer extends Transformer {
         if (!modelDirective) {
             throw new InvalidDirectiveError('Types annotated with @auth must also be annotated with @model.')
         }
+        MODEL_DIRECTIVE_STACK = def.name.value;
```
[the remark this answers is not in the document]
SwaySway: good point - I can refer to the model directive stack name with def.name.value. So I've made this change in the code where it no longer has a need to ref modelDirectiveStack

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -689,4 +768,34 @@ identityField: "${rule.identityField || DEFAULT_IDENTITY_FIELD}" }`
...
+                        "payload": {}
+                    }`)
+            ),
+            ResponseMappingTemplate: print(
+                raw(`{ "ok": true }`)
+            )
+        });
+    }
+
+    public subscriptionHas(subscription: any): Object {
```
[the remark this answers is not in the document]
SwaySway: Changed in the new revision to getDirectiveArgs()

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -187,6 +192,31 @@ export class ModelAuthTransformer extends Transformer {
         this.protectListQuery(ctx, ResolverResourceIDs.DynamoDBListResolverResourceID(def.name.value), queryRules.list)
         this.protectConnections(ctx, def, operationRules.read)
         this.protectQueries(ctx, def, operationRules.read)
+
+        // Check if subscriptions is enabled
+        const subscription = modelDirective.arguments.find((arg) => arg.name.value === 'subscriptions');
+        if (subscription && subscription.value.kind !== "NullValue") {
```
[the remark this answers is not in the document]
SwaySway: Changed in the new revision

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -872,6 +902,191 @@ All @auth directives used on field definitions are performed when the field is r
         }
     }
 
+    private protectCreateSubcriptions(ctx: TransformerContext, rules: AuthRule[],  parent: ObjectTypeDefinitionNode, nameOverride?: string) {
```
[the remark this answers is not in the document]
SwaySway: changed in the new revision

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -689,4 +768,34 @@ identityField: "${rule.identityField || DEFAULT_IDENTITY_FIELD}" }`
     public setUserGroups(): SetNode {
         return set(ref('userGroups'), raw('$util.defaultIfNull($ctx.identity.claims.get("cognito:groups"), [])'));
     }
+
+    public generateSubscriptionResolver(type: string, fieldName: string, subscriptionTypeName: string = 'Subscription') {
+        return new AppSync.Resolver({
+            ApiId: Fn.GetAtt(ResourceConstants.RESOURCES.GraphQLAPILogicalID, 'ApiId'),
+            DataSourceName: LOCAL_DATASOURCE,
```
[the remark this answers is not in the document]
SwaySway: Added the nonedatasource for the subscription resolvers

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -872,6 +1012,128 @@ All @auth directives used on field definitions are performed when the field is r
...
+    }
+
+    // generate subscrption based on operation
+    private protectSubscription(ctx: TransformerContext, rules: AuthRule[],
+        parent: ObjectTypeDefinitionNode, fieldName: string) {
+        const resolverResourceId = ResolverResourceIDs.ResolverResourceID("Subscription", fieldName);
+        const resolver = this.resources.generateSubscriptionResolver(fieldName);
+
+        // creates the none data source
+        const noneDS = ctx.getResource(ResourceConstants.RESOURCES.NoneDataSource)
```
yuth: Do we want to create a data source if its not being used? Should you set this only if there is `rules.length > 0`

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -478,10 +602,22 @@ Static group authorization should perform as expected.`
...
+                @auth groupClaim is only for 'allow: groups'`);
+            }
+            if ( groupsField && groups) {
+                throw new InvalidDirectiveError("This rule has groupsField and groups - use one or the other!")
+            }
         }
     }
 
     private validateFieldRules(rules: AuthRule[]) {
+        // check if the rules in the field are a subset of the rules in authrule
```
yuth: Is this relevant anymore

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -537,15 +673,19 @@ All @auth directives used on field definitions are performed when the field is r
...
             staticGroupAuthorizationExpression,
             newline(),
             dynamicGroupAuthorizationExpression,
             newline(),
             ownerAuthorizationExpression,
             newline(),
             throwIfUnauthorizedExpression
-        ])
+        ]
+        expressions.push(
```
yuth: Do you need this?

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -119,6 +132,12 @@ export class DynamoDBModelTransformer extends Transformer {
             }
         )
 
+        // check if auth if enabled
```
yuth: cruft?

packages/graphql-transformer-core/src/util/transformConfig.ts
```diff
@@ -50,6 +50,12 @@ export interface TransformConfig {
      * generalizes this behavior.
      */
     Migration?: TransformMigrationConfig;
+
+    /**
```
yuth: Change the comment since the flag has been changed to version number

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -162,8 +171,35 @@ export class ResourceFactory {
                 ])
             ])
         ])
+
+        const customGroupClaim: Expression = block('Static Group Authorization Checks with Custom Groups', [
```
RossWilliams: There is a lot of code duplication here that could be avoided. You are doing three extra steps: adding a comment when a custom claim is used, calling a different method to set userGroups, and handling the claim field being either a string or an array. You could remove this duplication (and source of future mis-alignment / bugs) by updating setCustomClaim function to convert any string claim to an array. Further, if you change setUserGroups to take a claim name as the first parameter with a default value of 'cognito:groups' you can completely remove this method, and remove the setCustomClaim method.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -685,4 +828,43 @@ identityField: "${rule.identityField || DEFAULT_IDENTITY_FIELD}" }`
     public setUserGroups(): SetNode {
         return set(ref('userGroups'), raw('$util.defaultIfNull($ctx.identity.claims.get("cognito:groups"), [])'));
     }
-}
+
+    public setCustomClaim(customGroups: string): SetNode {
```
RossWilliams: This method is problematic as you are setting userGroups to an unknown type, while setUserGroups uses cognito:groups which is known to be an array type. userGroups should have a stable type. Ive left a comment above on how this method can be removed completely.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -217,7 +253,15 @@ groupsField: "${rule.groupsField || DEFAULT_GROUPS_FIELD}" }`
         formatComment?: (rule: AuthRule) => string,
     ) {
         let groupAuthorizationExpressions = []
+        let customClaim: string;
         for (const rule of rules) {
+            if (rule.groupClaim) {
+                if (customClaim) {
+                    throw new InvalidDirectiveError('@auth directive currently only supports one source for groupClaim!')
```
RossWilliams: This restriction isn't necessary if you set the userGroups as the first item (or after the comments) of the groupAuthorizationExpressions.
```typescript
groupAuthorizationExpressions = groupAuthorizationExpressions.concat(
  formatComment ?
    comment(formatComment(rule)) :
    comment(`Authorization rule: { allow: ${rule.allow}, groupsField: "${groupsAttribute}" }`),
  this.setUserGroups(rule.groupClaim),
  set(
     ref(variableToSet),
    raw(`$util.defaultIfNull($${variableToSet}, false)`)
  ),
```

assumes you adopt the setUserGroups overload, otherwise you can use a ternary to choose the right function to call.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -405,7 +520,14 @@ identityField: "${rule.identityField || DEFAULT_IDENTITY_FIELD}" }`
 
         let groupAuthorizationExpressions = []
         let ruleNumber = 0
+        let customClaim: string;
         for (const rule of rules) {
+            if (rule.groupClaim) {
+                if (customClaim) {
+                    throw new InvalidDirectiveError('@auth directive currently only supports one source for groupClaim!')
```
RossWilliams: See above solution to remove this restriction on the create operation.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -504,7 +627,14 @@ identityField: "${rule.identityField || DEFAULT_IDENTITY_FIELD}" }`
             return comment(`No Dynamic Group Authorization Rules`)
         }
         let groupAuthorizationExpressions = [];
+        let customClaim: string;
         for (const rule of rules) {
+            if (rule.groupClaim) {
+                if (customClaim) {
+                    throw new InvalidDirectiveError('@auth directive currently only supports one source for groupClaim!')
```
RossWilliams: See above comment to remove this restriction.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -138,21 +139,29 @@ export class ResourceFactory {
...
                 }
             }
+            if (rule.groupClaim) {
+                if (customClaim) {
+                    throw new InvalidDirectiveError('@auth directive currently only supports one source for groupClaim!')
+                }
+                customClaim = rule.groupClaim;
+            }
         }
         // TODO: Enhance cognito:groups to work with non cognito based auth.
```
RossWilliams: Comment can be removed with this PR

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -267,6 +352,13 @@ Static group authorization should perform as expected.`
                 resolver = this.resources.blankResolver(typeName, fieldName)
             }
             const authExpression = this.authorizationExpressionOnSingleObject(rules, 'ctx.source')
+            if (protectPrivateFields) {
```
RossWilliams: The modified tests suggest this isn't correct.
context.result.operation is set to query in setOperationExpression, but then context.source.operation is checked for query in operationCheckExpression, fails, and returns null.

packages/graphql-transformers-e2e-tests/src/__tests__/ConnectionsWithAuthTests.e2e.test.ts
```diff
@@ -276,7 +276,7 @@ test('Testing reading an owner protected field as a non owner', async () => {
     console.log(response1);
     expect(response1.data.createFieldProtected.id).toEqual("1")
     expect(response1.data.createFieldProtected.owner).toEqual(USERNAME1)
-    expect(response1.data.createFieldProtected.ownerOnly).toEqual("owner-protected")
+    expect(response1.data.createFieldProtected.ownerOnly).toBeNull()
```
RossWilliams: I don't understand why this test should be updated. The field is protected to only allow the owner, but USERNAME1 is creating the object and USERNAME1 is set as the owner field. ownerOnly should not be null for the owner. Possible source of issue left as comment above in the protectPrivateFields check block.

packages/graphql-transformers-e2e-tests/src/__tests__/ConnectionsWithAuthTests.e2e.test.ts
```diff
@@ -372,7 +372,7 @@ test('Test that owners cannot set the field of a FieldProtected object unless au
     console.log(JSON.stringify(response1));
     expect(response1.data.createFieldProtected.id).toEqual("2")
     expect(response1.data.createFieldProtected.owner).toEqual(USERNAME1)
-    expect(response1.data.createFieldProtected.ownerOnly).toEqual("owner-protected")
+    expect(response1.data.createFieldProtected.ownerOnly).toBeNull()
```
RossWilliams: This test should not have been updated.

packages/graphql-transformers-e2e-tests/src/__tests__/ConnectionsWithAuthTests.e2e.test.ts
```diff
@@ -414,7 +414,7 @@ test('Test that owners cannot update the field of a FieldProtected object unless
     console.log(JSON.stringify(response1));
     expect(response1.data.createFieldProtected.id).not.toBeNull()
     expect(response1.data.createFieldProtected.owner).toEqual(USERNAME1)
-    expect(response1.data.createFieldProtected.ownerOnly).toEqual("owner-protected")
+    expect(response1.data.createFieldProtected.ownerOnly).toBeNull()
```
RossWilliams: This test should not have been updated. See above comment.

packages/graphql-transformers-e2e-tests/src/__tests__/ConnectionsWithAuthTests.e2e.test.ts
```diff
@@ -439,7 +439,7 @@ test('Test that owners cannot update the field of a FieldProtected object unless
     console.log(response3);
     expect(response3.data.updateFieldProtected.id).toEqual(response1.data.createFieldProtected.id)
     expect(response3.data.updateFieldProtected.owner).toEqual(USERNAME1)
-    expect(response3.data.updateFieldProtected.ownerOnly).toEqual("updated")
+    expect(response3.data.updateFieldProtected.ownerOnly).toBeNull()
```
RossWilliams: This test does not have a reason to change. See above comment.

packages/graphql-transformers-e2e-tests/src/__tests__/ConnectionsWithAuthTests.e2e.test.ts
RossWilliams: @yuth You may want to review these test changes as you have marked this PR as approved.

undefobj:
@SwaySway per chat we need to respond with `null` and not `{ "ok": true }` with an authorized request on subscriptions otherwise the GraphQL schema will give a type error and fail the subscription request from a client

undefobj:
Also note that for existing projects the check:

```
? The default behavoir for @auth has changed in the latest version of Amplify
Read here for details: https://aws-amplify.github.io/docs/cli-toolchain/graphql#authorizing-subscriptions 
 Do you wish to continue? Yes
authsubstestt $
```

This is not working correctly, it always will do a `process.exit()` which blocks an `amplify push` no matter what you select. We need to look at this logic and see what the bug is. @SwaySway @kaustavghosh06

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
SwaySway: Creating a none data source if the subscription resolvers are being created.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -221,6 +301,8 @@ Static group authorization should perform as expected.`
         // Get and validate the auth rules.
         const rules = getArg('rules', []) as AuthRule[]
         this.validateFieldRules(rules)
+        // if the object has auth enabled and the field rules are a superset of the object rules
```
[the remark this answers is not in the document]
SwaySway: Removed Comment

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
SwaySway: Removed Comment

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
SwaySway: That was from a prev change removed in latest commit

packages/graphql-transformer-core/src/util/transformConfig.ts
SwaySway: Changed comment

packages/graphql-auth-transformer/src/resources.ts
SwaySway: Good call out - I've edited setUserGroups to allow an optional param for a custom group claim.

packages/graphql-auth-transformer/src/resources.ts
SwaySway: Added this check in setUsergroups where it'll reassign the the string into an array

packages/graphql-transformers-e2e-tests/src/__tests__/ConnectionsWithAuthTests.e2e.test.ts
SwaySway: The behavior of per field auth has changed - the PR for docs elaborates on this more
[Docs PR #868 L941](https://github.com/aws-amplify/docs/pull/868/files#diff-8076c2db248c895d2ade586451fa9b55R944#L941)

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -22,11 +26,55 @@ const schemaDirName = 'schema';
...
+
+  if (unAuthModelTypes.length) {
+    context.print.warning('\nThe following types do not have \'@auth\' enabled. Consider using @auth with @model');
+    context.print.warning(unAuthModelTypes.map(type => `\t - ${type}`).join('\n'));
+    context.print.info('Learn more about @auth here: https://aws-amplify.github.io/docs/cli-toolchain/graphql#auth');
+  }
+}
+
+/**
+ * @TODO change authWarning to version number to compare
```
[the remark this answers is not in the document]
SwaySway: Changed @TODO to add a version mapping

packages/graphql-auth-transformer/src/resources.ts
SwaySway: Ah I see each rule would have it's own respective claim. Though I'm not sure of the use case since the groups could exist in the same groupClaim. This might warrant a separate pr where we can discuss a use case for this.

SwaySway:
> @SwaySway per chat we need to respond with `null` and not `{ "ok": true }` with an authorized request on subscriptions otherwise the GraphQL schema will give a type error and fail the subscription request from a client

I've added a change on the subscription resolver to return null
> This is not working correctly, it always will do a `process.exit()` which blocks an `amplify push` no matter what you select. We need to look at this logic and see what the bug is. @SwaySway @kaustavghosh06

Corrected in the latest commit

SwaySway:
Added an e2e test in `PerFieldAuthTests.e2e.test.ts` to confirm new `@auth` behavior

packages/graphql-auth-transformer/src/resources.ts
RossWilliams: I already use this on my Amplify fork, and do have a use case. I have an auth rule based on the TenantId claim, and another auth rule based on a TeamId, and another AuthId based on the AppGroupId.
Only requires moving the setUserGroups to inside the forEach loop, so each rules evaluates against its own claimGroup. And then you can delete ~15 lines of code.

undefobj: @RossWilliams it's less about the code and more related to complexity. We're already adding new functionality here which is pretty flexible, and adding multiple claims that could overlap permissions is complexity on already a complex concept which could have downstream impact to the authorization rules which needs evaluation. At this time leaving this as-is seems to be a more prudent choice which we could re-evaluate as a separate feature in the future.

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -22,11 +26,61 @@ const schemaDirName = 'schema';
 
 function failOnInvalidAuthType(usedDirectives, opts) {
   if (usedDirectives.includes('auth') && !opts.isUserPoolEnabled) {
-    throw new Error(`You are trying to use the @auth directive without enabling Amazon Cognito user pools for your API.
-Run \`amplify update api\` and choose "Amazon Cognito User Pool" as the authorization type for the API.`);
+    throw new Error(`You are trying to
+use the @auth directive without enabling Amazon Cognito user pools for your API.
+Run \`amplify update api\` and choose
+"Amazon Cognito User Pool" as the authorization type for the API.`);
```
attilah: In the light of multi-auth being merged, we should rephrase this to 'enable' or 'add' Cognito, because this wording could suggest that it works if *only* Cognito enabled.

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -22,11 +26,61 @@ const schemaDirName = 'schema';
...
+  if (!transformerConfig.Version && usedDirectives.includes('auth')
+  && resources.length > 0) {
+    if (context.exeInfo &&
+      context.exeInfo.inputParams && context.exeInfo.inputParams.yes) {
+      context.print.warning(`\n${versionChangeMessage}\n`);
+    } else {
+      const response = await inquirer.prompt({
+        name: 'transformerConfig',
+        type: 'confirm',
+        message: `${versionChangeMessage} \n Do you wish to continue?`,
```
attilah: We're printing this message anyway, but in one case as part of a prompt, otherwise as a warning, why not print it as warning anyway before the if, then have the prompt if needed?

nit: I'd not add newlines to messages as prefix/suffix, since once we'll add logging it will be hard to cleanup all these 'additions'.

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -22,11 +26,61 @@ const schemaDirName = 'schema';
...
+        default: false,
+      });
+      if (!response.transformerConfig) {
+        process.exit(0);
```
attilah: Why is the hard process exit, can't we throw and let SEH handle the exit from the process?

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -22,11 +26,61 @@ const schemaDirName = 'schema';
...
+      }
+    }
+  }
+  transformerConfig.Version = 3.0;
```
attilah: 3.0 as number? I think for versions in general we'd need something that can be parsed AND compared with ```semver``` lib to make differentiation later much easier.

packages/graphql-auth-transformer/package.json
```diff
@@ -7,7 +7,8 @@
     "test": "jest",
     "test-ci": "jest --ci -i",
     "build": "tsc",
-    "clean": "rm -rf ./lib"
+    "clean": "rm -rf ./lib",
```
attilah: nit: rm is *nix only, we should use rimraf everywhere, since it will break on windows. Yes I know we're currently breaking dev on windows, but it would be nice to no to add new 'problems'.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -88,12 +121,16 @@ export class ModelAuthTransformer extends Transformer {
                 # Specifies the auth rule's strategy. Allowed values are 'owner' and 'groups'.
                 allow: AuthStrategy!
 
+                # Legacy name for identifyClaim
```
attilah: typo: identiFyClaim, F instead of T multiple times.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1,26 +1,58 @@
...
+    query?: string;
+}
+
+interface MutationNameMap {
+    create?: string;
+    update?: string;
+    delete?: string;
+}
+
+enum ModelSubscriptionLevel {
```
attilah: Why assigning string literals to enum value? if we need string, then use type like for auth modes. That gives us value safety with typescript.

Why the ```ON``` value is not represented in the enum?

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -103,6 +140,10 @@ export class ModelAuthTransformer extends Transformer {
                 # Specifies the field of type String or [String] that contains group(s) that can access the object.
                 groupsField: String
```
attilah: Why don't we deprecate ```groupsField``` isn't it the same what we do with ```identityField``` ?

packages/graphql-auth-transformer/src/AuthRule.ts
```diff
@@ -5,9 +5,11 @@ export interface AuthRule {
     allow: 'owner' | 'groups';
     ownerField?: string;
     identityField?: string;
+    identityClaim?: string;
     groupsField?: string;
     groups?: string[];
     operations?: ModelOperation[]
     queries?: ModelQuery[]
     mutations?: ModelMutation[]
+    groupClaim?: string;
```
attilah: Please move it under ```groupsField``` for maintainability.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1,26 +1,58 @@
...
+    delete?: string;
+}
+
+enum ModelSubscriptionLevel {
+    OFF = "OFF",
+    PUBLIC = "PUBLIC",
+}
+
+interface SubscriptionNameMap {
+    onCreate?: string[];
```
attilah: With this structure you cannot configure level per operation. You can't make ```onCreate``` ```PUBLIC``` and protect only ```onUpdate``` and ```onDelete```. Are we ok with it?

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -187,6 +228,34 @@ export class ModelAuthTransformer extends Transformer {
         this.protectListQuery(ctx, ResolverResourceIDs.DynamoDBListResolverResourceID(def.name.value), queryRules.list)
         this.protectConnections(ctx, def, operationRules.read)
         this.protectQueries(ctx, def, operationRules.read)
+
+        // Check if subscriptions is enabled
+        const directiveArguments: ModelDirectiveArgs = getDirectiveArguments(modelDirective);
+        if ("subscriptions" in directiveArguments) {
+            const subscription = directiveArguments.subscriptions;
+            // check that subscription is not set to off
+            if (subscription && !subscription.level) {
```
attilah: Should not we validate the value of ```level``` to ```OFF``` here?

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -187,6 +228,34 @@ export class ModelAuthTransformer extends Transformer {
...
+                    subscription.onUpdate.forEach( (fieldName: string) => {
+                        this.protectOnUpdateSubcription(ctx, operationRules.update, def, fieldName);
+                    })
+                }
+                if (subscription.onDelete) {
+                    subscription.onDelete.forEach( (fieldName: string) => {
+                        this.protectOnDeleteSubcription(ctx, operationRules.delete, def, fieldName);
+                    })
+                }
+            } else { return; }
```
attilah: nit: break into multiple lines for readability.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -205,7 +274,17 @@ export class ModelAuthTransformer extends Transformer {
...
             return argument ? valueFromASTUntyped(argument.value) : dflt
         }
-
+        let protectPrivateFields = true;
+
+        // get model args
+        const modelDirective = parent.directives.find((dir) => dir.name.value === 'model')
+        const parentModelArgs: ModelDirectiveArgs = modelDirective ? getDirectiveArguments(modelDirective) : {};
+        //  check if subscriptions is explicity disabled or if it's set to public
+        if ('subscriptions' in parentModelArgs &&
```
attilah: The checks this way are not really explicit, based on a lot of JS behavior. What about making these checks more explicit? I can't see how are we checking for ```PUBLIC``` here? if subscriptions are explicitly disabled then it should have a ```null``` value.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -236,24 +315,27 @@ Static group authorization should perform as expected.`
...
+            updateRules)
 
         // Delete operations are only protected by @auth directives on objects.
         const deleteRules = rules.filter((rule: AuthRule) => isDeleteRule(rule))
         this.protectDeleteForField(ctx, parent, definition, deleteRules)
     }
 
-    private protectField(ctx: TransformerContext, typeName: string, fieldName: string, rules: AuthRule[]) {
+    private protectField(ctx: TransformerContext, typeName: string,
+        field: FieldDefinitionNode, rules: AuthRule[], protectPrivateFields?: boolean) {
```
attilah: Why is protectPrivateFields optional? make it mandatory with a default value, then code would be more explicit inside IMHO.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -264,9 +346,20 @@ Static group authorization should perform as expected.`
...
+                resolver = this.resources.blankResolver(typeName, field.name.value)
             }
             const authExpression = this.authorizationExpressionOnSingleObject(rules, 'ctx.source')
+            if (protectPrivateFields) {
+                if (field.type.kind === Kind.NON_NULL_TYPE) {
+                    throw new InvalidDirectiveError(`\nPer-field auth on the required field ${field.name.value} is not supported with subscriptions.
+Either make the field optional, set auth on the object and not the field, or disable subscriptions for the object (setting level to OFF or PUBLIC)\n`)
+                }
+                // add operation to queryField
+                this.protectMutations(ctx, typeName, 'mutation')
```
attilah: Don't use ```'mutation'``` literal here, use ```ctx.getMutationTypeName```. What other value this parameter can have? Because the called function name suggests only mutations are supported. The comment above suggests queryField, which is not correct.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -478,6 +597,17 @@ Static group authorization should perform as expected.`
                 console.warn(
                     `It is not recommended to use 'queries' and 'operations'. The 'operations' argument will be used.`)
             }
+            if ( allow === 'groups' && (identityClaim || identityField)) {
+                throw new InvalidDirectiveError(`
+                @auth identifyField/Claim is only for 'allow: owner'`)
```
attilah: Friendlier message: "can only be used" or something similar

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -478,6 +597,17 @@ Static group authorization should perform as expected.`
...
+            }
+            if (allow === 'owner' && groupClaim) {
+                throw new InvalidDirectiveError(`
+                @auth groupClaim is only for 'allow: groups'`);
```
attilah: Same as above.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -478,6 +597,17 @@ Static group authorization should perform as expected.`
...
+            }
+            if ( groupsField && groups) {
```
attilah: I think this validation is incorrect, since ```groups``` defines the list of allowed groups, ```groupsField``` specifies which field to look for the group values in the type.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -469,7 +586,9 @@ Static group authorization should perform as expected.`
 
     private validateRules(rules: AuthRule[]) {
         for (const rule of rules) {
-            const { queries, mutations, operations } = rule;
+            const { queries, mutations, operations,
```
attilah: Missing validations:
- groupField and groupClaim is mutually exclusive
- identityField and identityClaim is mutually exclusive

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -872,6 +1001,128 @@ All @auth directives used on field definitions are performed when the field is r
         }
     }
 
+    // OnCreate Subscription
+    private protectOnCreateSubcription(ctx: TransformerContext, rules: AuthRule[],
```
attilah: Why have 3 extra methods if the graphqlName could be just a parameter for ```protectSubscription```?

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -872,6 +1001,128 @@ All @auth directives used on field definitions are performed when the field is r
...
+            resolver.Properties.ResponseMappingTemplate = templateParts.join('\n\n');
+            ctx.setResource(resolverResourceId, resolver);
+            ctx.mapResourceToStack(parent.name.value, resolverResourceId);
+        }
+        // check if owner is enabled in auth
+        const hasOwner = rules.find( rule => rule.allow === DEFAULT_OWNER_FIELD);
+        if (hasOwner) {
+            this.addOwner(ctx, parent.name.value);
+        }
+        if (rules.length === 1 && hasOwner ) {
```
attilah: Code identation is wrong, some comments about why are we adding rules would be nice for future developers, consequent if could be an ```else if```?

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -138,21 +139,29 @@ export class ResourceFactory {
...
         for (const rule of rules) {
             const groups = rule.groups;
             for (const group of groups) {
                 if (group) {
                     allowedGroups.push(group);
                 }
             }
+            if (rule.groupClaim) {
+                if (customClaim) {
+                    throw new InvalidDirectiveError('@auth directive currently only supports one source for groupClaim!')
```
attilah: format: no need for exclamation mark, also we should output what type, what directive, what claim we found, what extra claim we found, that helps the developers to locate the error easier.

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -28,16 +28,22 @@ interface MutationNameMap {
     delete?: string;
 }
 
+enum ModelSubscriptionLevel {
```
attilah: Duplicate from ```AuthTransformer```, can't we share these types somehow? It is also missing the ```ON``` value, see my comment above at ```AuthTransformer```

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -79,6 +85,12 @@ export class DynamoDBModelTransformer extends Transformer {
                 onCreate: [String]
                 onUpdate: [String]
                 onDelete: [String]
+                level: ModelSubscriptionLevel
+            }
+            enum ModelSubscriptionLevel {
```
attilah: We standardized on lowercase enum values for our directives, see ```AuthStrategy``` et al.

packages/graphql-transformer-core/src/util/sanity-check.ts
```diff
@@ -261,6 +262,10 @@ export function cantHaveMoreThan200Resources(diffs: Diff[], currentBuild: Diffab
     }
 }
 
+export function authSubscriptions(diff: Diff[]) {
+    // check if the subscription
```
attilah: Unfinished function, TODO?

PR #2070, opened 2019-08-16
chore: update CI script to freeze lock file and update yarn.lock

attilah:
*Description of changes:*

After last merge, release task was failing because of differences in ```yarn.lock```, this PR updates a regenerated one and also updates the CI build script to ensure CI build is running with frozen ```yarn.lock``` file.


.circleci/config.yml
```diff
@@ -13,9 +13,9 @@ jobs:
     <<: *defaults
     steps:
       - checkout
-      - run: yarn cache clean --force
-      - run: yarn
       - run: yarn config set workspaces-experimental true
+      - run: yarn cache clean --force
+      - run: yarn --frozen-lockfile
```
yuth: We don't need this as `yarn production-build` does this

.circleci/config.yml
```diff
@@ -71,7 +71,7 @@ jobs:
             TERM: dumb
       steps:
         - checkout
-        - run: yarn install
+        - run: yarn --frozen-lockfile
```
yuth: probably we can get rid of both checkout and yarn --forzen-lockfile

.circleci/config.yml
attilah: Removed

.circleci/config.yml
```diff
@@ -92,14 +92,14 @@ jobs:
               cd ..
               git clone $AUTH_CLONE_URL
               cd aws-amplify-cypress-auth
-              yarn install
+              yarn install --frozen-lockfile
```
yuth: I don't think this is needed. This is a test repo with UI code that we use to run our cypress tests

attilah: Removed install, checkout needs to stay as we discussed.

.circleci/config.yml
attilah: Removed install, checkout needs to stay as we discussed.

Issue #1056, opened 2019-03-18
Show friendlier error message when required runtime CLI files are absent


attilah:
**Describe the bug**
After creating an AppSync project with environment support, git clean -fdx will remove .config/local-env-info.json and amplify/backend/amplify-meta.json as well which results in runtime errors for various amplify commands like amplify push, amplify status

**To Reproduce**
1. git init
2. amplify init (default settings are OK, env name: dev)
3. git add --all
4. amplify status
5. git clean -fdx
6. amplify status

**Expected behavior**
The required files should not be in gitignore, to keep the project functional

**Desktop (please complete the following information):**
 - OS: Irrelevant
 - Amplify CLI Version 1.1.7

Output for step 6:
```
Current Environment: undefined

~/.npm-global/lib/node_modules/@aws-amplify/cli/node_modules/gluegun/build/index.js:13
    throw up;
    ^

Error: ENOENT: no such file or directory, open '/projects/foobar2/amplify/backend/amplify-meta.json'
    at Object.openSync (fs.js:439:3)
    at Object.readFileSync (fs.js:344:35)
    at getResourceStatus (~/.npm-global/lib/node_modules/@aws-amplify/cli/src/extensions/amplify-helpers/resource-status.js:207:37)
    at Object.showResourceTable (~/.npm-global/lib/node_modules/@aws-amplify/cli/src/extensions/amplify-helpers/resource-status.js:268:13)
    at Command.run (~/.npm-global/lib/node_modules/@aws-amplify/cli/src/commands/status.js:5:27)
    at Runtime.<anonymous> (~/.npm-global/lib/node_modules/@aws-amplify/cli/node_modules/gluegun/build/runtime/run.js:90:58)
    at step (~/.npm-global/lib/node_modules/@aws-amplify/cli/node_modules/gluegun/build/runtime/run.js:32:23)
    at Object.next (~/.npm-global/lib/node_modules/@aws-amplify/cli/node_modules/gluegun/build/runtime/run.js:13:53)
    at ~/.npm-global/lib/node_modules/@aws-amplify/cli/node_modules/gluegun/build/runtime/run.js:7:71
    at new Promise (<anonymous>)
```

Output for 'amplify push':

```
Current Environment: undefined
```

Output for 'amplify checkout dev':
```
~/.npm-global/lib/node_modules/@aws-amplify/cli/node_modules/gluegun/build/index.js:13
    throw up;
    ^

[...]

PR #2088, opened 2019-08-19
fix: #1056, dedup environment file reading

attilah:
*Issue #, if available:*

#1056

*Description of changes:*

- Add uncaught error handler to CLI for friendlier error messages, due to the lack of GlueGun capability for this
- Throw friendly error message when environment seems to be uninitialized for various commands
- Deduplicate the code when reading the environment file
- Fix some typos

### DIFF
--- a/packages/amplify-cli/src/cli.js
+++ b/packages/amplify-cli/src/cli.js
@@ -6,6 +6,17 @@ const globalPrefix = require('./lib/global-prefix');
 
 const MIGRATE = 'migrate';
 
+// Gluegun has no hooks for errors happening inside executing the commands,
+// so we need a process level handler, to log the errors like environment
+// not initialized and such errors in a friendly way without technical data
+// like stacktrace which does not add value in this case.
+process.on('uncaughtException', handleError);
+
+function handleError(error) {
+  console.error(error.message);
+  process.exit(1);
+}
+
 async function run(argv) {
   const localNodeModulesDirPath = path.normalize(path.join(__dirname, '../', 'node_modules'));
   const globalNodeModulesDirPath = globalPrefix.getGlobalNodeModuleDirPath();
--- a/packages/amplify-cli/src/extensions/amplify-helpers/get-env-info.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/get-env-info.js
@@ -7,6 +7,9 @@ function getEnvInfo() {
   let envInfo = {};
   if (fs.existsSync(envFilePath)) {
     envInfo = readJsonFile(envFilePath);
+  } else {
+    // EnvInfo is required by all the callers so we can safely throw here
+    throw new Error('Current environment cannot be determined\nUse \'amplify init\' in the root of your app directory to initialize your project with Amplify');
   }
 
   return envInfo;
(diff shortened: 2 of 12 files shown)


packages/amplify-cli/src/cli.js
```diff
@@ -6,6 +6,17 @@ const globalPrefix = require('./lib/global-prefix');
 
 const MIGRATE = 'migrate';
 
+// Gluegun has no hooks for errors happening inside executing the commands,
+// so we need a process level handler, to log the errors like environment
+// not initialized and such errors in a friendly way without technical data
+// like stacktrace which does not add value in this case.
+process.on('uncaughtException', handleError);
+
+function handleError(error) {
```
kaustavghosh06: Will this show some messages twice?

packages/amplify-cli/src/extensions/amplify-helpers/get-env-info.js
```diff
@@ -7,6 +7,9 @@ function getEnvInfo() {
   let envInfo = {};
   if (fs.existsSync(envFilePath)) {
     envInfo = readJsonFile(envFilePath);
```
kaustavghosh06: We can use the getEnvInfo() out here right as well right?

packages/amplify-cli/src/extensions/amplify-helpers/get-project-details.js
```diff
@@ -12,11 +13,7 @@ function getProjectDetails() {
     amplifyMeta = readJsonFile(amplifyMetaFilePath);
   }
 
-  let localEnvInfo = {};
-  const envFilePath = pathManager.getLocalEnvFilePath();
-  if (fs.existsSync(envFilePath)) {
-    localEnvInfo = readJsonFile(envFilePath);
-  }
+  const localEnvInfo = getEnvInfo();
```
kaustavghosh06: What if the file doesn't exist? Will it throw an error right? 
I think in the previous logic we wern't throwing an error when file doesn't exist and just checking for its existence.

packages/amplify-cli/src/cli.js
attilah: It is only for uncaught errors, generally speaking: no, but we do have a few places in our ```catch``` blocks where we ```console.log``` and ```throw``` afterwards, but cannot be determined if there is a higher catch for those blocks or not.

IMHO refactoring the error handling to be true SEH is a different issue, that should be visited in a separate PR for the new platform.

packages/amplify-cli/src/extensions/amplify-helpers/get-env-info.js
attilah: What do you mean, this IS ```getEnvInfo``` itself.

packages/amplify-cli/src/extensions/amplify-helpers/get-project-details.js
attilah: I visited all the places where we were reading this files and we're consuming straight ahead what was returned, without checking it, that's why I added the throw to ```getEnvInfo``` itself. LMK if you find a place where it is not the case, I could missed that, but tried to be thorough.

Issue #1720, opened 2019-06-25
Invalid GraphQL type names are generated for @key when name contains dashes or other characters


dantonijevic:
**Which Category is your question related to?**

GraphQL API

**What AWS Services are you utilizing?**

AppSync

**Provide additional details e.g. code snippets**

 I am using this newly introduced @key directive when creating my scheme. Currently I have a scheme with primary and one GSI. It looks like this

```
type Item @model
  @key(fields: ["id", "version"])
  @key(name: "type-subtype", fields: ["type", "subtype"])
{
  id: ID!
  version: String!
  type: String!
  subtype: String!
}
```

I want to change GSI defined here to be:

```
@key(name: "version-type-subtype", fields: ["version", "type", "subtype"])
```
 but I am getting `Schema Creation Status is FAILED with details: Failed to parse schema document - ensure it's a valid SDL-formatted document.` error when trying to push.

Can someone help? Is there any constraint using sort key from primary index as hash in GSI that I am not aware of?

Thanks

attilah:
@dantonijevic just to confirm, do you've ```fileds``` or ```fields``` in your schema?, since that typo could make the schema invalid.

dantonijevic:
@attilah Yeah, typo is just in a post, not in actual code.

attilah:
@dantonijevic thanks for confirming it, we will look into it then.

mikeparisstuff:
@dantonijevic It is not possible to edit an existing GSI. To make this work, you should create a new `@key` on the updated fields and then remove the existing `@key`. There is a note about this here https://aws-amplify.github.io/docs/cli/graphql#evolving-apis-with-key

dantonijevic:
@mikeparisstuff Thank you for the response.

I actually tried with a fresh AppSync project and tried to define a GSI with 3 fields which got me the same error as above. I created a repo to demonstrate - [here](https://github.com/dantonijevic/amplify-GSI-test/blob/master/amplify/backend/api/keys/schema.graphql) is the schema I am trying to push where the current DynamoDB table deployed has `type-subtype` GSI defined together with the primary index `(id, version)`. 

<img width="1266" alt="Screenshot 2019-07-01 at 20 26 40" src="https://user-images.githubusercontent.com/3730543/60459138-9c8f7c80-9c40-11e9-83b2-056bd0cb5f7e.png">

So, with this push I was expecting this third index `version-type-subtype` to be deployed but no luck.

[...]

PR #2093, opened 2019-08-20
fix: #1720 - fix GraphQL name generation for @key

attilah:
*Issue #, if available:*

#1720

*Description of changes:*

Make sure that for 2+ part @key names valid GraphQL type names will be generated, with proper pascal casing.

### DIFF
--- a/packages/graphql-key-transformer/src/KeyTransformer.ts
+++ b/packages/graphql-key-transformer/src/KeyTransformer.ts
@@ -20,6 +20,7 @@ import {
 } from 'graphql';
 import { AppSync, IAM, Fn, DynamoDB, Refs } from 'cloudform-types'
 import { Projection, GlobalSecondaryIndex, LocalSecondaryIndex } from 'cloudform-types/types/dynamoDb/table';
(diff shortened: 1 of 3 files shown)


packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -160,7 +161,7 @@ export default class FunctionTransformer extends Transformer {
         if (args.fields.length > 2) {
             const compositeKeyFieldNames = args.fields.slice(1);
             const compositeKeyFields = definition.fields.filter(field => Boolean(compositeKeyFieldNames.find(k => k === field.name.value)));
-            const keyName = args.name || 'Primary';
+            const keyName = Case.pascal(args.name || 'Primary');
```
kaustavghosh06: Is this a breaking change?

attilah: This keyName variable is only used for the GraphQL input and condition types, nothing that's directly referenced like DynamoDB GSI name, so I don't think it is a breaking change.

PR #2114, opened 2019-08-22
fix: #1581 - Fix stack enumeration bug

attilah:
*Issue #, if available:*

#1581 

*Description of changes:*

Fix stack enumeration so transform.conf.json will be generated

### DIFF
--- a/packages/graphql-transformer-core/tsconfig.json
+++ b/packages/graphql-transformer-core/tsconfig.json
@@ -4,7 +4,6 @@
         "module": "commonjs",
         "sourceMap": true,
         "outDir": "lib",
-        "sourceRoot": "src",
         "lib": [
             "es2015",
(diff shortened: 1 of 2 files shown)


packages/graphql-transformer-core/tsconfig.json
```diff
@@ -4,7 +4,6 @@
         "module": "commonjs",
         "sourceMap": true,
         "outDir": "lib",
-        "sourceRoot": "src",
```
kaustavghosh06: Why did we remove this?

attilah: It is causing a problem when debugging in VSCode, no need for that. With that setting you cannot debug the TS files, breakpoints were not hit, just the compiled JS files.

Issue #2071, opened 2019-08-16
RFC - Extending @connection to Work With @key


AmanJaveriKadri:
# Extending @connection to Work With @key 

The existing directive that is used to create relationships between
objects is the @connection directive:  

    directive @connection(name:      String,
                          keyField:  String,
                          sortField: String) on FIELD_DEFINITION

Although useful to create connections that fit common use-cases, the
@connection directive lacks flexibility. Connections are created by
automatically configuring GSI’s as needed for each connection without
giving users control over what specific GSI’s are used for the
connections. This leads to users sometimes having duplicate GSI’s, since
they are also separately creating an equivalent GSI with @key, and users
generally having less control over the kinds of connections that they
need for specific use-cases.  
  
This RFC documents a proposed second parameterization of the @connection
directive that lets users use GSIs created with @key to run queries for
connected objects. The @key directive is better at configuring indices
than the @connection directive is. @connection does not support
composite sort keys, for example, whereas @key makes it easy to choose
the partition and sort key while abstracting away the process of making
a composite sort key. The @key directive gives the user clarity on what
the primary key of a table will be, @connection, on the other hand,
makes it difficult see what is actually being done behind the scenes.
Hence, being able to leverage GSI’s setup by @key to create @connection
would likely be useful.  

## Proposed Solution

Add a new parameterization of @connection such that the directive is
given a key to query and a list of fields to query by. It returns a list
of objects (or in certain cases an object) that are connected to the
object it is called on.   
  
Existing connections do **not** need to be changed. All changes made to
the @connection directive are backwards compatible and the one can still
choose to parameterize @connection with the original parameters and the
old behavior. (See Appendix 3)  
  
<span class="underline">**Directive Definition:**</span>  

    directive @connection(keyName: String,
                          fields: [String],
                          model:   String) on FIELD_DEFINITION

The directive takes a list of fields to use in the query and the name of
a key (created with the @key directive) to run this query on. If no key
name is provided, the query will be run on the default table. The
‘model’ parameter is optional and only needs to be used for a
particular case where one wants to query a table that is not the table
of the child model. This is particularly useful if one wants to use a
single-table design. (See the second Many-to-Many example.)  

<div data-section-style="5" style="">

  - <span id="UZV9CAf1dPX">keyName: Name of key created with @key
    directive that one wants to query to find connected objects.
    (Optional. Default table is used if no key name is provided.)   
    </span>
[...]

PR #2118, opened 2019-08-22
New Parameterization of @connection to work with @key

AmanJaveriKadri:
*Issue #, if available:*
#2071 

*Description of changes:*
Added a new parameterization of @connection:
    directive @connection(keyName: String, fields: [String]) on FIELD_DEFINITION

This runs a query on the table of another model (inferred by field type) with the primary key being the fields passed in to connection. This is used to get a list of objects connected to the object in question and returned through a field. Details of the directive's functionality can be found in the RFC (Issue #2071).

There are basic unit tests in packages/graphql-connection-transformer/src/__tests__/NewConnectionTransformer.test.ts
Integration tests for the transformer itself in packages/graphql-transformers-e2e-tests/src/__tests__/NewConnectionTransformer.e2e.test.ts
And integration tests for @auth compatibility in packages/graphql-transformers-e2e-tests/src/__tests__/NewConnectionWithAuth.e2e.test.ts


alex-chew:
LGTM

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -26,6 +26,23 @@ import Resource from 'cloudform-types/types/resource';
 
 const CONNECTION_STACK_NAME = 'ConnectionStack'
 
+interface KeySchema {
```
mikeparisstuff: Are these importable from cloudform instead of redefining them?

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -46,6 +63,74 @@ function validateKeyField(field: FieldDefinitionNode): void {
...
+        return
+    }
+    const isNonNull = isNonNullType(field.type);
+    const isAList = isListType(field.type)
+
+    // The only valid key fields are single non-null fields.
+    if (!isAList && isNonNull) {
+        return;
+    }
+    throw new InvalidDirectiveError(`All fields used for a connection cannot be lists and must be of Non-Null types.`)
```
mikeparisstuff: Can we clean up this error msg to be more active. I.E. `All fields provided to a @connection must be non-null scalar fields.`

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -104,6 +195,11 @@ export class ModelConnectionTransformer extends Transformer {
             throw new InvalidDirectiveError(`Object type ${relatedTypeName} must be annotated with @model.`)
         }
 
+        if (getDirectiveArgument(directive)("fields")) {
+            this.newParameterization(parent, field, directive, ctx)
```
mikeparisstuff: Nice clean way of diving into the new logic. Can you please place a comment here explaining why and that the behavior is changing based on the fields presence.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -325,6 +421,144 @@ export class ModelConnectionTransformer extends Transformer {
...
+                    throw new InvalidDirectiveError(item + ' is not a field in ' + parentTypeName)
+                }
+
+                validateKeyFieldNewConnection(inputFields[fieldsArrayLength]);
+            })
+
+        let index : Index = undefined;
+        // If no index is provided use the default index for the related model type and
+        // check that the query fields match the PK/SK of the table. Else confirm that index exists.
+        if (!args.keyName || args.keyName === 'default' || args.keyName === 'Default') {
```
mikeparisstuff: What is this default behavior? Do we need it? I would prefer not to have this as it means that I cannot name an index "default" without this logic treating it special.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -325,6 +421,144 @@ export class ModelConnectionTransformer extends Transformer {
...
+                throw new InvalidDirectiveError('Key ' + args.keyName + ' does not exist for model ' + relatedTypeName)
+            }
+
+            // Confirm that types of query fields match types of PK/SK of the index being queried.
+            checkFieldsAgainstIndex(parent.fields, relatedType.fields, args.fields, index.KeySchema);
+        }
+
+        // If the related type is not a list, the index has to be the default index and the fields provided must match the PK/SK of the index.
+        if (!isListType(field.type)) {
+            if (args.keyName !== 'default') {
```
mikeparisstuff: ditto to above.

packages/graphql-transformers-e2e-tests/src/__tests__/NewConnectionTransformer.e2e.test.ts
```diff
@@ -0,0 +1,367 @@
...
+          authorID
+          authorName
+          authorSurname
+          postContents
+        }
+      }`, {})
+    expect(createPostModel.data.createPostModel.id).toBeDefined()
+    expect(createPostModel.data.createPostModel.authorID).toEqual(createUser.data.createUser.id)
+    expect(createPostModel.data.createPostModel.authorName).toEqual(createUser.data.createUser.name)
+    expect(createPostModel.data.createPostModel.authorSurname).toEqual(createUser.data.createUser.surname)
```
mikeparisstuff: strange formatting here?

packages/graphql-transformers-e2e-tests/src/__tests__/NewConnectionTransformer.e2e.test.ts
```diff
@@ -0,0 +1,367 @@
...
+                    id
+                    name
+                    surname
+                }
+            }
+        }
+    }`, {})
+    expect(queryPostModel.data.getPostModel).toBeDefined()
+    const items = queryPostModel.data.getPostModel.authors.items
+    expect(items.length).toEqual(1)
```
mikeparisstuff: If I have more than one value with the same id, name, surname combo will they appear here as expected?

packages/graphql-connection-transformer/src/resources.ts
```diff
@@ -189,4 +193,214 @@ export class ResourceFactory {
...
+            [partitionKeyName] :
+                ref(`util.dynamodb.toDynamoDBJson($util.defaultIfNullOrBlank($ctx.source.${connectionAttributes[0]}, "${NONE_VALUE}"))`)
+            })
+
+        // Add sort key if there is one.
+        if (connectionAttributes.length > 2) {
+            const rangeKeyFields = connectionAttributes.slice(1);
+            const sortKeyName = keySchema[1].AttributeName as string
+            const condensedSortKeyValue = this.condenseRangeKey(
+                rangeKeyFields.map(keyField => `\${ctx.args.${keyField}}`)
```
mikeparisstuff: Won't these values come off the `$ctx.source`? Can we make sure there is a test case that tests this GetOne scenario.

packages/graphql-transformers-e2e-tests/src/__tests__/NewConnectionTransformer.e2e.test.ts
```diff
@@ -0,0 +1,367 @@
...
+	@model
+	@key(fields: ["id", "name"])
+{
+	id: ID!
+	name: String!
+
+	parents: [Parent] @connection(keyName: "byChild", fields: ["id"])
+}
+
+type Parent
```
mikeparisstuff: It would be great if we could get a reflective `User -> Friendship -> User` scenario covered in the tests since it is an oft-asked for example.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -46,6 +63,74 @@ function validateKeyField(field: FieldDefinitionNode): void {
...
+    const queryPKType = parentFields.find(f => f.name.value === inputFieldNames[0]).type;
+    const numFields = inputFieldNames.length;
+
+    if (getBaseType(tablePKType) !== getBaseType(queryPKType)) {
+        throw new InvalidDirectiveError(inputFieldNames[0] + ' field is not of type ' + getBaseType(tablePKType))
+    }
+    if (numFields > keySchema.length && keySchema.length !== 2) {
+        throw new InvalidDirectiveError('Too many fields passed in to @connection directive.')
+    }
+    if (numFields === 2) {
```
mikeparisstuff: What about if numFields === 1?

ambientlight:
@AmanJaveriKadri, @mikeparisstuff: this is huge, amazing job! can't wait to see this merged

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
AmanJaveriKadri: Then there is only a partition key provided and that case is already checked with the first if-condition on line 101.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
AmanJaveriKadri: That makes sense, I'll get rid of the 'default' option

packages/graphql-connection-transformer/src/resources.ts
AmanJaveriKadri: You're right there was a missing test and error here, fixed now.

packages/graphql-transformers-e2e-tests/src/__tests__/NewConnectionTransformer.e2e.test.ts
AmanJaveriKadri: Yes, added a test for this.

packages/graphql-transformers-e2e-tests/src/__tests__/NewConnectionTransformer.e2e.test.ts
AmanJaveriKadri: Added a test for this.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -46,6 +52,74 @@ function validateKeyField(field: FieldDefinitionNode): void {
     throw new InvalidDirectiveError(`If you define a field and specify it as a 'keyField', it must be of type 'ID' or 'String'.`)
 }
 
+/**
+ * Ensure that the field passed in is compatible to be a key field
+ * (Not a list and of type ID or String)
+ * @param field: the field to be checked.
+ */
+function validateKeyFieldNewConnection(field: FieldDefinitionNode): void {
+    if (!field) {
```
attilah: It's a private function, this could not happen, right, feels to be code smell.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -46,6 +52,74 @@ function validateKeyField(field: FieldDefinitionNode): void {
...
+        return
+    }
+    const isNonNull = isNonNullType(field.type);
+    const isAList = isListType(field.type)
+
+    // The only valid key fields are single non-null fields.
+    if (!isAList && isNonNull) {
+        return;
+    }
+    throw new InvalidDirectiveError(`All fields provided to an @connection must be non-null scalar fields.`)
```
attilah: Should we validate if the field type is scalar, since the error message suggest that we validate it.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -46,6 +52,74 @@ function validateKeyField(field: FieldDefinitionNode): void {
...
+}
+
+/**
+ * Checks that the fields being used to query match the expected key types for the index being used.
+ * @param parentFields: All fields of the parent object.
+ * @param realtedTypeFields: All fields of the related object.
```
attilah: typo: related

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -46,6 +52,74 @@ function validateKeyField(field: FieldDefinitionNode): void {
...
+ * @param inputFieldNames: The fields passed in to the @connection directive.
+ * @param keySchema: The key schema for the index being used.
+ */
+function checkFieldsAgainstIndex(parentFields: ReadonlyArray<FieldDefinitionNode>,
+                                 relatedTypeFields: ReadonlyArray<FieldDefinitionNode>,
+                                 inputFieldNames: string[],
+                                 keySchema: KeySchema[]): void {
+    const hashAttributeName = keySchema[0].AttributeName;
+    const tablePKType = relatedTypeFields.find(f => f.name.value === hashAttributeName).type;
```
attilah: Can we've a situation when ```hashAttributeName``` cannot be found within the fields? because there is no check for the result of ```find```, just the result used straight ahead.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -46,6 +52,74 @@ function validateKeyField(field: FieldDefinitionNode): void {
...
+    if (numFields > keySchema.length && keySchema.length !== 2) {
+        throw new InvalidDirectiveError('Too many fields passed in to @connection directive.')
+    }
+    if (numFields === 2) {
+        const sortAttributeName = keySchema[1].AttributeName;
+        const tableSKType = relatedTypeFields.find(f => f.name.value === sortAttributeName).type;
+        const querySKType = parentFields.find(f => f.name.value === inputFieldNames[1]).type;
+
+        if (getBaseType(tableSKType) !== getBaseType(querySKType)) {
+            throw new InvalidDirectiveError(inputFieldNames[1] + ' field is not of type ' + getBaseType(tableSKType))
```
attilah: Use string interpolation please

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -46,6 +52,74 @@ function validateKeyField(field: FieldDefinitionNode): void {
...
+        const tableSortKeyTypes = tableSortFields.map(name => relatedTypeFields.find(f => f.name.value === name).type);
+        const querySortFields = inputFieldNames.slice(1);
+        const querySortKeyTypes = querySortFields.map(name => parentFields.find(f => f.name.value === name).type);
+
+        // Check that types of each attribute match types of the fields that make up the composite sort key for the
+        // table or index being queried.
+        querySortKeyTypes.forEach((fieldType, index) => {
+            if (getBaseType(fieldType) !== getBaseType(tableSortKeyTypes[index])) {
+                throw new InvalidDirectiveError(
+                            querySortFields[index] + ' field is not of type ' + getBaseType(tableSortKeyTypes[index])
```
attilah: Use string interpolation please

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -59,7 +133,11 @@ export class ModelConnectionTransformer extends Transformer {
     constructor() {
         super(
             'ModelConnectionTransformer',
-            gql`directive @connection(name: String, keyField: String, sortField: String) on FIELD_DEFINITION`
+            gql`directive @connection(name: String,
+                                      keyField: String,
+                                      sortField: String,
+                                      keyName: String,
+                                      fields: [String]) on FIELD_DEFINITION`
```
attilah: ```fields: [String!]``` seems to be a better fit

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -325,6 +412,141 @@ export class ModelConnectionTransformer extends Transformer {
...
+        const relatedTypeName = getBaseType(field.type)
+        const relatedType = ctx.inputDocument.definitions.find(
+            d => d.kind === Kind.OBJECT_TYPE_DEFINITION && d.name.value === relatedTypeName
+        ) as ObjectTypeDefinitionNode | undefined
+
+        // Get Child object's table.
+        const tableLogicalID = ModelResourceIDs.ModelTableResourceID(relatedType.name.value);
+        const tableResource = ctx.getResource(tableLogicalID) as Table;
+
+        // Ensure that there is at least one field provided.
```
attilah: Since it is a hard failure this check should be moved to the top of the method.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -325,6 +412,141 @@ export class ModelConnectionTransformer extends Transformer {
...
+        }
+
+        // Check that each field provided exists in the parent model and that it is a valid key type (single non-null).
+        let inputFields : FieldDefinitionNode[] = [];
+        args.fields.forEach(
+            item => {
+                const fieldsArrayLength = inputFields.length;
+                inputFields[fieldsArrayLength] = parent.fields.find(f => f.name.value === item);
+                if (!inputFields[fieldsArrayLength]) {
+                    throw new InvalidDirectiveError(item + ' is not a field in ' + parentTypeName)
```
attilah: Use string interpolation everywhere please

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -325,6 +412,141 @@ export class ModelConnectionTransformer extends Transformer {
...
+            }
+
+            // Confirm that types of query fields match types of PK/SK of the index being queried.
+            checkFieldsAgainstIndex(parent.fields, relatedType.fields, args.fields, <KeySchema[]>index.KeySchema);
+        }
+
+        // If the related type is not a list, the index has to be the default index and the fields provided must match the PK/SK of the index.
+        if (!isListType(field.type)) {
+            if (args.keyName) {
+                throw new InvalidDirectiveError('Connection is to a single object but the keyName provided does not reference the default table.')
```
attilah: Add the ```keyName``` to the error message to make it easier for the developer to locate the error.

packages/graphql-connection-transformer/src/resources.ts
```diff
@@ -189,4 +193,217 @@ export class ResourceFactory {
...
+
+    /**
+     * Create a get item resolver for singular connections.
+     * @param type The parent type name.
+     * @param field The connection field name.
+     * @param relatedType The name of the related type to fetch from.
+     * @param connectionAttributes The names of the underlying attributes containing the fields to query by.
+     * @param keySchema Key schema of the index or table being queried.
+     */
+    public makeGetItemNewConnectionResolver(type: string,
```
attilah: Naming came up in other comments by @yuth, perhaps using the term ParameterizedConnection would be a good name for this?

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -405,7 +627,8 @@ export class ModelConnectionTransformer extends Transformer {
 
         // Create sort key condition inputs for valid sort key types
         // We only create the KeyConditionInput if it is being used.
-        if (sortKeyInfo) {
+        // Don't create a key condition input for composite sort keys since it already done by @key.
+        if (sortKeyInfo && (sortKeyInfo.typeName !== "Composite")) {
```
attilah: We're using single quotes for strings in the codebase. (missing Lint rule?)

packages/graphql-dynamodb-transformer/src/definitions.ts
```diff
@@ -539,8 +544,15 @@ export function makeModelConnectionField(fieldName: string, returnTypeName: stri
         makeInputValueDefinition('nextToken', makeNamedType('String'))
     ];
     if (sortKeyInfo) {
+        let namedType : NamedTypeNode;
+        if (sortKeyInfo.typeName === 'Composite') {
```
attilah: Can we create a ```type SortKeyInfo = 'Composite' | string``` Type to constrain the valid values of string literals because we're using ```Composite``` at multiple places and it would give some TypeScript safety and would say more about the type that it is a special value.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -325,6 +412,141 @@ export class ModelConnectionTransformer extends Transformer {
...
+                }
+
+                validateKeyFieldNewConnection(inputFields[fieldsArrayLength]);
+            })
+
+        let index : GlobalSecondaryIndex = undefined;
+        // If no index is provided use the default index for the related model type and
+        // check that the query fields match the PK/SK of the table. Else confirm that index exists.
+        if (!args.keyName) {
+            checkFieldsAgainstIndex(parent.fields, relatedType.fields, args.fields, tableResource.Properties.KeySchema);
```
attilah: If we've a ```keyName``` specified, then shouldn't we validate that the a ```@key``` exists on the connected type?

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
RossWilliams: *scalar or enum.

packages/graphql-dynamodb-transformer/src/definitions.ts
```diff
@@ -526,16 +526,19 @@ export function makeSubscriptionField(fieldName: string, returnTypeName: string,
     )
 }
 
+export type typeName = 'Composite' | string;
```
attilah: Types should by PascalCased, and since it is an export ```typeName``` is too general, I'd suggest ```SortKeyFieldInfoTypeName```, that shows well what the type is used for.

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -485,11 +494,12 @@ export class ModelConnectionTransformer extends Transformer {
         // If the related type is not a list, the index has to be the default index and the fields provided must match the PK/SK of the index.
         if (!isListType(field.type)) {
             if (args.keyName) {
-                throw new InvalidDirectiveError('Connection is to a single object but the keyName provided does not reference the default table.')
+                throw new InvalidDirectiveError(`Connection is to a single object but the keyName ${args.keyName} ` +
```
attilah: why are we concatenating still? if because of linter, there is a nicer way to fix it ;-)

packages/graphql-transformer-core/src/TransformerContext.ts
```diff
@@ -181,6 +182,26 @@ export default class TransformerContext {
...
+    /**
+     * Scans through the context nodeMap and returns all type definition nodes
+     * that are of the given kind.
+     * @param kind Kind value of type definition nodes expected.
+     */
+    public getTypeDefinitionsOfKind(kind: string) {
+        const typeDefs: TypeDefinitionNode[] = [];
+        for (const key of Object.keys(this.nodeMap)) {
+            const definition = this.nodeMap[key];
+            switch (definition.kind) {
```
attilah: ```if``` would be a better fit for a single check.

PR #2175, opened 2019-08-30
Codegen decouple

nikhname:
*Issue #, if available:*

*Description of changes:*
Allows for codegen without amplify init, provided a schema.json file (introspection schema) is present in the parent directory of the project.


packages/amplify-codegen/src/commands/add.js
```diff
@@ -20,72 +20,99 @@ const addWalkThrough = require('../walkthrough/add');
...
-      codeGenTarget: answer.target || '',
-      generatedFileName: answer.generatedFileName || '',
-      docsFilePath: answer.docsFilePath,
-      region,
-    },
-  };
+  let newProject;
+  if (!context.withoutInit) {
+    newProject = {
+      projectName: apiDetails.name,
```
yuth: Can we use ternary operator to switch to different name based on with or without init. Everything else looks the same in both the casses

packages/amplify-codegen/src/commands/configure.js
```diff
@@ -3,6 +3,11 @@ const configureProjectWalkThrough = require('../walkthrough/configure');
 const add = require('./add');
 
 async function configure(context) {
+  try {
+    context.amplify.getProjectMeta();
+  } catch (e) {
+    context.withoutInit = true;
```
yuth: Could we move this to a separate helper function instead. I am not a fan of adding more things to context

packages/amplify-codegen/src/utils/downloadIntrospectionSchema.js
```diff
@@ -21,7 +21,10 @@ async function downloadIntrospectionSchema(context, apiId, downloadLocation, reg
       const introspectionDir = dirname(downloadLocation);
       fs.ensureDirSync(introspectionDir);
       fs.writeFileSync(downloadLocation, schema, 'utf8');
-      return relative(amplify.getEnvInfo().projectPath, downloadLocation);
+      if (!context.withoutInit) {
```
yuth: Do we need this check here? It looks like we call this helper function only for the project which have been initialized

packages/amplify-codegen/src/walkthrough/questions/selectProfile.js
```diff
@@ -0,0 +1,15 @@
+const inquirer = require('inquirer');
```
yuth: Can you remove this. I don't see this being used anywhere

nikhname: Removing it

packages/amplify-codegen/src/utils/downloadIntrospectionSchema.js
nikhname: True check here is unnecessary. I will remove my changes to this file since we don't download the schema in this workflow

packages/amplify-codegen/src/commands/configure.js
nikhname: Moving this as well as context.frontend and context.framework to parameters in the functions called from here

packages/amplify-codegen/commands/codegen/add.js
```diff
@@ -1,6 +1,13 @@
 const codeGen = require('../../src/index');
 const constants = require('../../src/constants');
+const path = require('path');
```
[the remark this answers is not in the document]
nikhname: moving

packages/amplify-codegen/src/commands/add.js
nikhname: Changing to ternary operator

nikhname: Changing to ternary

PR #2180, opened 2019-08-30
Allow more than one groupClaim in an auth ruleset.

RossWilliams:
*Description of changes:*
#2068 introduced groupClaim, but restricted this parameter to be set on a single rule in the set. This allows a groupClaim to be set on multiple rules in a set. It does this by setting the userGroups property inside the rules loop, rather than outside.


packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -138,39 +137,31 @@ export class ResourceFactory {
...
-            if (rule.groupClaim) {
-                if (customClaim) {
-                    throw new InvalidDirectiveError(`@auth directive currently only supports one source for groupClaim.
-  - Identified '${customClaim}' and '${rule.groupClaim}'`)
-                }
-                customClaim = rule.groupClaim;
+           
+            if (groups) {
+                groupAuthorizationExpressions = groupAuthorizationExpressions.concat(
+                    comment(`Authorization rule: { allow: groups, groups: "${JSON.stringify(groups)}" }`),
```
attilah: We could add the checked claim to the comment for clarity.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -138,39 +137,31 @@ export class ResourceFactory {
...
+                    this.setUserGroups(rule.groupClaim),
+                    set(ref('allowedGroups'), list(groups.map(s => str(s)))),
+                    forEach(ref('userGroup'), ref('userGroups'), [
+                        iff(
+                            raw(`$allowedGroups.contains($userGroup)`),
+                            set(ref(variableToSet), raw('true'))
```
attilah: We should add a ```break``` statement after the first match to exit the loop.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -229,20 +220,14 @@ groupsField: "${rule.groupsField || DEFAULT_GROUPS_FIELD}" }`
...
-                    throw new InvalidDirectiveError('@auth directive currently only supports one source for groupClaim!')
-                }
-                customClaim = rule.groupClaim;
-            }
             // for loop do check of rules here
             const groupsAttribute = rule.groupsField || DEFAULT_GROUPS_FIELD
             groupAuthorizationExpressions = groupAuthorizationExpressions.concat(
                 formatComment ?
                     comment(formatComment(rule)) :
                     comment(`Authorization rule: { allow: ${rule.allow}, groupsField: "${groupsAttribute}" }`),
```
attilah: Same as above.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -496,20 +477,14 @@ identityClaim: "${rule.identityField || rule.identityClaim || DEFAULT_IDENTITY_F
...
-                if (customClaim) {
-                    throw new InvalidDirectiveError('@auth directive currently only supports one source for groupClaim!')
-                }
-                customClaim = rule.groupClaim;
-            }
             const groupsAttribute = rule.groupsField || DEFAULT_GROUPS_FIELD
             const groupsAttributeName = `groupsAttribute${ruleNumber}`
             const groupName = `group${ruleNumber}`
             groupAuthorizationExpressions = groupAuthorizationExpressions.concat(
                 comment(`Authorization rule${fieldMention}: { allow: ${rule.allow}, groupsField: "${groupsAttribute}" }`),
```
attilah: Same as above.

RossWilliams:
Updates: Found that "userGroup" was used instead of "userGroups". Found Cognito pre-token lambdas do not allow setting claims as array, so try to detect a stringified array.

PR #2201, opened 2019-09-04
es default req res

SwaySway:
*Issue #, if available:*

*Description of changes:*
 - when querying against elasticsearch the resolver will use default sort values

### DIFF
--- a/packages/graphql-elasticsearch-transformer/package.json
+++ b/packages/graphql-elasticsearch-transformer/package.json
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "build": "tsc && cd streaming-lambda && bestzip ../lib/streaming-lambda.zip python_streaming_function.py",
+    "watch": "tsc -w && cd streaming-lambda && bestzip ../lib/streaming-lambda.zip python_streaming_function.py",
     "clean": "rm -rf ./lib"
   },
   "keywords": [
--- a/packages/graphql-elasticsearch-transformer/src/SearchableModelTransformer.ts
+++ b/packages/graphql-elasticsearch-transformer/src/SearchableModelTransformer.ts
@@ -1,6 +1,6 @@
 import {
     Transformer, TransformerContext, getDirectiveArguments,
-    gql } from "graphql-transformer-core";
+    gql, InvalidDirectiveError } from "graphql-transformer-core";
 import {
     DirectiveNode,
     ObjectTypeDefinitionNode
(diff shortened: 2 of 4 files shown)


packages/graphql-elasticsearch-transformer/package.json
```diff
@@ -5,6 +5,7 @@
   "main": "lib/index.js",
   "scripts": {
     "build": "tsc && cd streaming-lambda && bestzip ../lib/streaming-lambda.zip python_streaming_function.py",
+    "watch": "tsc -w && cd streaming-lambda && bestzip ../lib/streaming-lambda.zip python_streaming_function.py",
```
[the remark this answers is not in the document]
attilah: @kaustavghosh06: This is to watch for changes during development so your 'amplify-dev' will be usable without manual compile between changes.

packages/graphql-transformer-common/src/ResolverResourceIDs.ts
```diff
@@ -1,6 +1,9 @@
 import { graphqlName, toUpper } from "./util";
 
 export class ResolverResourceIDs {
+    static DynamoDBTableResourceID(typeName: string): string {
```
attilah: Duplicate code, we already have a function for this: ```ModelResourceIDs.ModelTableResourceID```, please use that.

packages/graphql-elasticsearch-transformer/src/resources.ts
```diff
@@ -382,7 +382,8 @@ export class ResourceFactory {
     /**
      * Create the Elasticsearch search resolver.
      */
-    public makeSearchResolver(type: string, nonKeywordFields: Expression[], nameOverride?: string, queryTypeName: string = 'Query') {
+    public makeSearchResolver(type: string, nonKeywordFields: Expression[],
+        primaryKey: string, nameOverride?: string, queryTypeName: string = 'Query') {
```
attilah: I think we can remove the ```queryTypeName``` parameter, I only found a single caller to this method and for the ```Query``` type name we've a utility function to get it.

packages/graphql-elasticsearch-transformer/src/SearchableModelTransformer.ts
```diff
@@ -27,12 +27,12 @@ import path = require('path');
 const STACK_NAME = 'SearchableStack';
 const nonKeywordTypes = ["Int", "Float", "Boolean", "AWSTimestamp", "AWSDate", "AWSDateTime"];
```
attilah: Question: Do we have all the scalar types here, what we need?https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html

SwaySway: This is list is a subset of the scalar types which are not serialized as strings and types which may be serialized as strings in ddb though they are set to a different type in ES. (For example AWSDate is serialized as a string though it is picked up as a date type in ES)

packages/graphql-elasticsearch-transformer/src/resources.ts
SwaySway: changed in the latest commit

packages/graphql-transformer-common/src/ResolverResourceIDs.ts
SwaySway: changed in the latest commit

PR #2225, opened 2019-09-09
feat: add amplify prefix to every cfn stack provisioned via the CLI

kaustavghosh06:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/initializer.js
+++ b/packages/amplify-provider-awscloudformation/lib/initializer.js
@@ -14,9 +14,10 @@ async function run(context) {
   await configurationManager.init(context);
   if (!context.exeInfo || context.exeInfo.isNewEnv) {
     const initTemplateFilePath = path.join(__dirname, 'rootStackTemplate.json');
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/lib/initializer.js
```diff
@@ -14,9 +14,10 @@ async function run(context) {
   await configurationManager.init(context);
   if (!context.exeInfo || context.exeInfo.isNewEnv) {
     const initTemplateFilePath = path.join(__dirname, 'rootStackTemplate.json');
-    const timeStamp = `${moment().format('YYYYMMDDHHmmss')}`;
+    let timeStamp = `${moment().format('YYYYMMDDHHmmss')}`;
+    timeStamp = timeStamp.slice(timeStamp.length - 5);
```
UnleashedMind: This leads to the timestamp to be 'Hmmss', I don't think that's what we wanted.  If you want to reduce the length, you can directly do it on the variable definition `${moment().format('.....')}`

jwarchol:
This prefix is problematic in environments where stack prefixes are enforced by policy.

Issue #2196, opened 2019-09-02
Amplify CLI Migration fails


fabianjj:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[Amplify Console GitHub Issue Tracker](https://github.com/aws-amplify/amplify-console/issues)

**Describe the bug**
Amplify CLI fails to migrate to newer version. 

**To Reproduce**
1. I have a project configured with an older version of the CLI.
2. Update the CLI to 3.0.0.
3. Try almost any amplify command (this triggers a request to migrate).
4. Try migrating.
5.

During the migration the following appears a bit in:

```
⠴ Migrating your project
The following types do not have '@auth' enabled. Consider using @auth with @model
         - TripComponent
         - PackItem
         - TripComponentConnection
         - TripPackItemExclude
         - TripPackItemInclude
         - Trip
         - User
         - TripParticipation
         - PackAssign
         - PackList
         - PackListEntry
Learn more about @auth here: https://aws-amplify.github.io/docs/cli-toolchain/graphql#auth 

⠙ Migrating your projectReverting API migration.
⠋ Migrating your project
```

And at the end, all I get is the following error:

```
⠋ Migrating your projectAPI successfully reverted.
✖ There was an error migrating your project.
migration operations are rolledback.
Cannot convert undefined or null to object
```

**Expected behavior**
I expect the automatic migration command to work. I haven't really changed anything in the API other than a simple graphql schema. 

**Desktop (please complete the following information):**
 Using WSL in Visual Studio Code.

yuth:
@fabianjj could you share a sample schema where we can reproduce this issue

houmark:
Oh, migration is necessary?

On a related note, using ANY command in Amplify CLI with a new version (for example `amplify status`), Amplify CLI should detect that the project is running configuration based on an older and clearly with a WARNING text in red color explain how to migrate or have an option to migrate all services/categories in with one acceptance by the user.

I want to test this out, but considering this bug report, I think I will hold off a bit :)

houmark:
Feature request related to my comment opened in #2202.

fabianjj:
@houmark: Using any command for me triggers the request for migration, including a red warning text. It does not allow any command to be performed unless a migration is made. Declining to migrate cancels the command.

fabianjj:
@yuth, here is the schema I'm using:

`type TripComponent @model {
  id: ID!
  name: String!
[...]

PR #2245, opened 2019-09-11
fix(graphql-transformer-core): fix migration errors

yuth:
Updated stack sanity check and stack mapping code to work with older version of project when
migrating project

partial fix to #2196

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-transformer-core/src/util/amplifyUtils.ts
+++ b/packages/graphql-transformer-core/src/util/amplifyUtils.ts
@@ -118,7 +118,7 @@ export async function ensureMissingStackMappings(config: ProjectOptions) {
         const transformOutput = await _buildProject(config);
         const copyOfCloudBackend = await readFromPath(currentCloudBackendDirectory);
         const stackMapping = transformOutput.stackMapping;
-        if (copyOfCloudBackend && copyOfCloudBackend.build) {
+        if (copyOfCloudBackend && copyOfCloudBackend.build && copyOfCloudBackend.build.stacks) {
             // leave the custom stack alone. Don't split them into separate stacks
             const customStacks = Object.keys(copyOfCloudBackend.stacks || {});
             const stackNames = Object.keys(copyOfCloudBackend.build.stacks).filter(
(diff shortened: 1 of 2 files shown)


sacrampton:
My ConnectionStack.json that is generated by amplify has 200 resources - so I have implemented stack mapping with the transform.conf.json file as follows:

{
    "StackMapping": {
        "AssetVisitAttributecrowdResultsResolver": "ConnectionStack2",
        "CrowdResultscrowdSourcingJobResolver": "ConnectionStack2",
        "CrowdResultsuomResolver": "ConnectionStack2"
   }
}

Everything worked great - with the 3 additional resources going into ConnectionStack2.json and successfully pushed.

THEN I updated the Amplify CLI to the latest version a few days ago and now Stack Mapping no longer works. I'm commenting here because this is a change to Stack Mapping that has happened in the same time frame.

I add an additional few resources to my transform.conf.json - which I have done successfully several times - 

{
    "StackMapping": {
        "AssetVisitAttributecrowdResultsResolver": "ConnectionStack2",
        "CrowdResultscrowdSourcingJobResolver": "ConnectionStack2",
        "CrowdResultsuomResolver": "ConnectionStack2",
        "FieldListCrowdResults": "ConnectionStack2",
        "FieldListcrowdResultsResolver": "ConnectionStack2",
        "CrowdResultsfieldListResolver": "ConnectionStack2",
        "CharacteristicValueCrowdResults": "ConnectionStack2"
   }
}

When I compile this (amplify api gql-compile) it fails because it ignores this stack mapping and pushes the excess resources into the ConnectionStack.json file and overwrites the transform.conf.json file with the previous version and adds a "Version" item at the end.

{
    "StackMapping": {
        "AssetVisitAttributecrowdResultsResolver": "ConnectionStack2",
        "CrowdResultscrowdSourcingJobResolver": "ConnectionStack2",
        "CrowdResultsuomResolver": "ConnectionStack2"
   },
 "Version": 4
}

Like I said - this stack mapping was previously working perfectly - but about 5 days ago it stopped working. Logging this here because in my searches this change is the only stack mapping change I can see.

PR #2270, opened 2019-09-13
refactor(cli): enhance the plugin add command to take input argument

UnleashedMind:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-cli/src/commands/plugin/add.ts
+++ b/packages/amplify-cli/src/commands/plugin/add.ts
@@ -13,8 +13,8 @@ import inquirer, { InquirerOption, EXPAND } from '../../domain/inquirer-helper';
 import { AddPluginError } from '../../domain/add-plugin-result';
 import { normalizePluginDirectory } from '../../plugin-helpers/scan-plugin-platform';
 
-const NEWPLUGINPACKAGE = 'A new plugin package';
-const CANCAL = 'cancel';
+const NEW_PLUGIN_PACKAGE = 'A new plugin package';
+const CANCEL = 'cancel';
 
 export async function run(context: Context) {
   if (context.input.subCommands && context.input.subCommands.length > 1) {
@@ -49,6 +49,10 @@ async function resolvePluginPathAndAdd(context: Context, inputPath: string) {
 
 async function resolvePluginPackagePath(context: Context, inputPath: string):
 Promise<string | undefined> {
+  if (path.isAbsolute(inputPath)) {
+    return inputPath;
+  }
+
   let result;
 
   const { pluginPlatform } = context;
@@ -77,7 +81,7 @@ Promise<string | undefined> {
     const { confirmed } = await inquirer.prompt({
       type: 'confirm',
       name: 'confirmed',
-      message: `Please confirm to add the plugin package to your Amplify CLI.`,
+      message: `Confirm to add the plugin package to your Amplify CLI.`,
       default: true,
     });
     if (confirmed) {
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/commands/plugin/add.ts
```diff
@@ -1,19 +1,110 @@
...
+  if (candicatePluginDirPaths.length === 0) {
+    context.print.error('Can not locate the plugin package.');
+    result = await promptForPluginPath();
+  } else if (candicatePluginDirPaths.length === 1) {
+    context.print.green('Plugin package found.');
+    context.print.blue(candicatePluginDirPaths[0]);
+    const { confirmed } = await inquirer.prompt({
+      type: 'confirm',
+      name: 'confirmed',
+      message: `Please confirm to add the plugin package to your Amplify CLI.`,
```
attilah: Based on our guide we should not use "Please" and similar

packages/amplify-cli/src/commands/plugin/add.ts
```diff
@@ -1,19 +1,110 @@
...
+      default: true,
+    });
+    if (confirmed) {
+      result = candicatePluginDirPaths[0];
+    }
+  } else if (candicatePluginDirPaths.length > 1) {
+    context.print.warning('Multiple plugins with the package name are found.');
```
attilah: We should output what packages is a duplicate: "Multiple plugins found with the same package name: {packageName}".

packages/amplify-cli/src/commands/plugin/add.ts
```diff
@@ -1,19 +1,110 @@
...
 import {
     addUserPluginPackage,
     addExcludedPluginPackage as addFromExcluded,
     confirmAndScan,
 } from '../../plugin-manager';
 import inquirer, { InquirerOption, EXPAND } from '../../domain/inquirer-helper';
 import { AddPluginError } from '../../domain/add-plugin-result';
+import { normalizePluginDirectory } from '../../plugin-helpers/scan-plugin-platform';
 
 const NEWPLUGINPACKAGE = 'A new plugin package';
```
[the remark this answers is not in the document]
UnleashedMind: check

packages/amplify-cli/src/commands/plugin/add.ts
```diff
@@ -1,19 +1,110 @@
...
-
+const CANCAL = 'cancel';
```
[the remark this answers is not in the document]
UnleashedMind: check

packages/amplify-cli/src/commands/plugin/add.ts
UnleashedMind: check

packages/amplify-cli/src/commands/plugin/add.ts
```diff
@@ -1,19 +1,110 @@
...
 
 export async function run(context: Context) {
+  if (context.input.subCommands && context.input.subCommands.length > 1) {
+    const input = context.input.subCommands[1];
+    const { excluded } = context.pluginPlatform;
+    if (excluded[input] && excluded[input].length > 0) {
+      const { confirmed } = await inquirer.prompt({
+        type: 'confirm',
+        name: 'confirmed',
+        message: `Add from previously removed ${input} plugin`,
```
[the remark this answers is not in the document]
UnleashedMind: the command that the user ran was actually `amplify plugin remove`, so for the user it's `remove`, and we only put it inside the excluded array if the package is under the scan radar.

packages/amplify-cli/src/commands/plugin/add.ts
UnleashedMind: the package name is the user's input. and below this line is the selection with all the package paths prompted to the user

Issue #2273, opened 2019-09-13
Multi-Auth (public,iam) doesn't work with @keys


GeorgeBellTMH:
Main Get and List query's are working with multi-auth (public, iam)

Get queries for @key items are resulting in Unauthorized 

type Video 
 @model 
 @searchable
 @key(fields:["id"])
 @key(name: "ByYoutubeIdent", fields: ["YoutubeIdent"], queryField:"getVideoByYoutubeIdent")
 @key(name: "ByVideoTypes",fields: ["videoTypes","publishedDate"], queryField:"getVideoByVideoType")
 @auth(rules:[{allow:public, provider:iam, operations:[read]}])
{.....}

PR #2276, opened 2019-09-13
fix: #2272, #2273 - create correct policies when IAM is the default auth

### LINKED ISSUE #2272: Policy statement must contain resources after upgrade to 3.2 and add @auth
1. Upgraded from 3.0 to 3.2
2. Push failed (defaultAuthorization not set)
3. Ran update api
4. Push succeeds
5. Added @auth(rules:[{allow:public, provider:iam, operations:[read]}])
6. Push fails with:

Resource Name: theme-Unau-ES6W4PH9WDDZ (AWS::IAM::Policy)
Event Type: create
Reason: Policy statement must contain resources. (Service: AmazonIdentityManagement; Status Code: 400; Error Code: MalformedPolicyDocument; Request ID: 0dbcd5ff-d5ca-11e9-93c6-3da30b79754d)

7. Removed aim from above to @auth(rules:[{allow:public, operations:[read]}])
8. Push succeeds...but is not what I need...

attilah:
*Issue #, if available:*

Fixes #2272, #2273

*Description of changes:*

- When IAM is the default authentication mode, an empty policy was generated, because @aws_iam directives were not added, so it did not add the policy resource.
- @key generated queries were not properly protected, they missed the directives and the IAM policy resource entries.

### DIFF
--- a/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
+++ b/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
@@ -325,7 +325,7 @@ export class ModelAuthTransformer extends Transformer {
         // Retrieve the configuration options for the related @model directive
         const modelConfiguration = new ModelDirectiveConfiguration (modelDirective, def);
         // Get the directives we need to add to the GraphQL nodes
-        const directives = this.getDirectivesForRules(rules);
+        const directives = this.getDirectivesForRules(rules, false);
 
         // Add the directives to the Type node itself
         if (directives.length > 0) {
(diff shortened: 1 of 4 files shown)


packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -431,28 +442,33 @@ Static group authorization should perform as expected.`
...
                     }
                 }
 
                 addDirectivesForOperation('get');
                 addDirectivesForOperation('list');
             }
 
+            if (operationName) {
+                this.addFieldToResourceReferences(ctx.getQueryTypeName(), operationName, rules);
+                this.addFieldToResourceReferences(ctx.getQueryTypeName(), operationName, rules);
```
kaustavghosh06: Why do we have the same statement twice?

kaustavghosh06:
Could you also please add an e2e test with @key.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
attilah: Already refactored, that piece was not correct there, victim of copy and paste.

attilah:
> Could you also please add an e2e test with @key.

I plan to add tests as well, to make sure we're covering the grounds.

PR #2282, opened 2019-09-13
chore: add precommit hook to run prettier

yuth:
Amplify CLI codebase does not follow consistent style in the code base. One way to easily get consistent style is to use prettier to format the code. 

Updated the repository to run precommit lint which runs prettier on changed files before doing a git commit


attilah:
@yuth please update and merge it finally! ;-)

Issue #2241, opened 2019-09-11
Field @auth transformer combinaisons can lead to unexpected authorized mutation


dz0l038:
Hi!

I am facing a strange behavior for a few days now on the "field @auth transformer". Authorization to perform some mutations are granted when it shouldn't, and the fields order seems to have an impact on mutation execution while I would have think that it is not the expected behavior.
I am still unsure if I misunderstood this transformer or if it is an actual bug, but it is odd enough to me to open it as an issue. Hopefully an Amplify guru will see that and enlighten me if I didn't get something.

Amplify version: 1.12.0

## How the model should behave:
Only two users have access to the model, user1 and user2. At creation, `user1Sub` and `user2Sub` fields are populated so those two users will have read access. The user2 can read any fields but cannot update anything. User1 can read any fields and can update the `content` and `updatedAt` field but CANNOT update `user1Sub` and `user2Sub`.

So the schema I first came up with was this one:

**schema nbr1**
```
type Agreement
  @model(subscriptions: null)
  @key(fields: ["id"])
  @key(name: "ByUser1", fields: ["user1Sub", "updatedAt"], queryField: "agreementsByUser1")
  @key(name: "ByUser2", fields: ["user2Sub", "updatedAt"], queryField: "agreementsByUser2")
  
  @auth(rules: [{ allow: groups, groups: ["Admin"], operations: [create, read, update, delete] },
                { allow: owner, ownerField: "user1Sub", operations: [read, create] },
                { allow: owner, ownerField: "user2Sub", operations: [read] }])
{
  id: ID!
  user1Sub: ID!
  user2Sub: ID!
  updatedAt: AWSDateTime! @auth(rules: [{ allow: owner, ownerField: "user1Sub", operations: [update] }])
  content: String! @auth(rules: [{ allow: owner, ownerField: "user1Sub", operations: [update] }])
}
```

From my understanding of the doc, the model above should behave as expected. The global @auth prevents any update, but fields @auth allow it for user1 on `updatedAt` and `content`.

When I run (as user1):
**Mutation nbr1**: only `user1Sub`
```
mutation UpdateAgreement {
  updateAgreement(input: {
    id: "d28eb0e-7db3-a61f-bde2-132c131e54",
    user1Sub: "50ae7808-aed8-4bf3-9078-5ab653f0bc36"
  }) {
    id
    updatedAt
    user1Sub
    user2Sub
    content
  }
}
```
[...]

PR #2316, opened 2019-09-17
fix(graphql-auth-transformer): verify multiple static group auth rules

SwaySway:
*Issue #, if available:*
re #2241
*Description of changes:*
ensure multiple static group rules are verified per each field

### DIFF
--- a/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
+++ b/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
@@ -584,7 +584,7 @@ Either make the field optional, set auth on the object and not the field, or dis
                 // Generate the expressions to validate each strategy.
                 const staticGroupAuthorizationExpression = this.resources.staticGroupAuthorizationExpression(
                     staticGroupAuthorizationRules,
-                    field.name.value)
+                    field)
 
                 // In create mutations, the dynamic group and ownership authorization checks
                 // are done before calling PutItem.
@@ -1268,7 +1268,7 @@ All @auth directives used on field definitions are performed when the field is r
                 // Generate the expressions to validate each strategy.
                 const staticGroupAuthorizationExpression = this.resources.staticGroupAuthorizationExpression(
                     staticGroupAuthorizationRules,
-                    field ? field.name.value : undefined)
+                    field)
 
                 // In create mutations, the dynamic group and ownership authorization checks
                 // are done before calling PutItem.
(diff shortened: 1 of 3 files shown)


packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -813,11 +816,13 @@ identityClaim: "${rule.identityField || rule.identityClaim || DEFAULT_IDENTITY_F
     // A = IsStaticallyAuthed
     // B = AuthConditionIsNotNull
     // ! (A OR B) == (!A AND !B)
-    public throwIfNotStaticGroupAuthorizedOrAuthConditionIsEmpty(): Expression {
+    public throwIfNotStaticGroupAuthorizedOrAuthConditionIsEmpty(fieldName?: string): Expression {
+        const staticGroupAuthorizedVariable = fieldName ? `${fieldName}_${ResourceConstants.SNIPPETS.IsStaticGroupAuthorizedVariable}` :
```
yuth: can we move this to a helper function

SwaySway: added a helper function in latest commit

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -1022,6 +1018,11 @@ identityClaim: "${rule.identityField || rule.identityClaim || DEFAULT_IDENTITY_F
         return block("Determine request authentication mode", expressions);
     }
 
+    public setStaticAuthVariable(field: FieldDefinitionNode): string {
```
yuth: this should be called `getStaticAuthVariableName` as we are not setting the variable itself in this method

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1287,8 +1291,10 @@ All @auth directives used on field definitions are performed when the field is r
                 )
 
                 const collectAuthCondition = this.resources.collectAuthCondition()
+                const staticGroupAuthorizedVariable = field ? `${field.name.value}_${ResourceConstants.SNIPPETS.IsStaticGroupAuthorizedVariable}` :
```
yuth: leftover? can we use the helper method here?

SwaySway: removed leftover

packages/graphql-auth-transformer/src/resources.ts
SwaySway: changed to set

Issue #2299, opened 2019-09-16
Local mocking does not appendError or validate correctly


chrisco255:
**Describe the bug**

**To Reproduce**
1. Edit any request / response VTL template
2. Add a line to the top of your template: `$util.appendError("something went wrong")`
3. Mock API with `amplify mock`
4. Run the query/mutation template you edited
5. Appended errors are ignored

**Expected behavior**
Appended errors should be returned along with the regular data from the query.

In addition, it doesn't look like the overloaded versions of `$util.error` nor `$util.validate` are being respected in the mocked API utils.

Issue #2211, opened 2019-09-05
Mock resolvers do not adhere to velocity template version


nateiler:
I noticed a discrepancy in the execution of a response resolver between the Mock and Cloud instances.

```
query GetPreferences {
  getPreferences {
    id
  }
}
```

req.vtl
```
{
  "version": "2017-02-28",
  "operation": "GetItem",
  "key": #if( $modelObjectKey ) $util.toJson($modelObjectKey) #else {
    "id": $util.dynamodb.toDynamoDBJson($ctx.identity.claims.get("cognito:username"))
  } #end
}
```

res.vtl
```
#if( $util.isNullOrEmpty($context.result) )
    #set ($result = {})
    $util.qr($result.put("id", $ctx.identity.claims.get("cognito:username")))
    $util.toJson($result)
#else
    $util.toJson($ctx.result)
#end
```

When mocked, if a result is empty the response template is still executed.  Per [upgrade docs](https://docs.aws.amazon.com/appsync/latest/devguide/resolver-mapping-template-changelog.html#aws-appsync-resolver-mapping-template-version-2018-05-29) I changed the request template version to "2018-05-29" and both Mock and Cloud produced the same results.

I wasn't aware of the impact regarding different template versions, but made the change to 'req.vtl' to:
```
{
  "version": "2018-05-29",
  "operation": "GetItem",
  "key": #if( $modelObjectKey ) $util.toJson($modelObjectKey) #else {
    "id": $util.dynamodb.toDynamoDBJson($ctx.identity.claims.get("cognito:username"))
  } #end
}
```

visusnet:
Seems to be related to https://github.com/aws-amplify/amplify-cli/issues/2134

PR #2329, opened 2019-09-18
fix(amplify-appsync-simulator): add support for AppSync template version

### LINKED ISSUE #2134: AppSync simulator does not work with @auth for empty results
Currently, request templates are generated with version 2017-02-28. The AppSync simulator automatically assumes that the version is at least 2018-05-29. The change that is causing problems has been described here: https://docs.aws.amazon.com/appsync/latest/devguide/resolver-mapping-template-changelog.html#aws-appsync-resolver-mapping-template-version-2018-05-29
The AppSync simulator calls the response resolver template even though the 2017 version wouldn't.
As long as you don't have any `@auth` directives in your schema, you won't have any problems because the response template will only have `$util.toJson($context.result)`. Schemas that use `@auth` directives cannot be mocked correctly at the moment. They will raise an authentication error if there was no result. 

To my mind, there are two possible approaches to resolve this:
1. Either generate resolver templates for 2018-05-29 which check whether there was a result, i.e. something like this:
```
#if(!$util.isNull($ctx.result))
    ## The auto-generated code that ends with...
    $util.toJson($context.result)
#end
```
2. or check for the version in `packages/amplify-appsync-simulator/src/resolvers/function.ts` and render the `responseMappingTemplate` conditionally.

Alternatively, one could create custom resolver templates for everything manually...

**Additional context**
Something like this might already work (for solution 2):
change

yuth:
Mock server did have support for Resolver versions. Updated mock implementation to support template
version and surface errors raised using $util.appendError

fix #2134 #2211 and #2299

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-appsync-simulator/src/resolvers/pipeline-resolver.ts
+++ b/packages/amplify-appsync-simulator/src/resolvers/pipeline-resolver.ts
@@ -33,7 +33,7 @@ export class AppSyncPipelineResolver {
     let stash = {};
     let templateErrors;
 
-    // Pipe line request mapping template
+    // Pipeline request mapping template
     ({ result, stash, errors: templateErrors } = requestMappingTemplate.render(
       { source, arguments: args, stash },
       context,
@@ -41,7 +41,7 @@ export class AppSyncPipelineResolver {
     ));
     context.appsyncErrors = [...context.appsyncErrors, ...templateErrors];
 
-    // Pipe line functions
+    // Pipeline functions
     await this.config.functions
       .reduce((chain, fn) => {
         const fnResolver = this.simulatorContext.getFunction(fn);
--- a/packages/amplify-appsync-simulator/src/resolvers/pipeline-resolver.ts
+++ b/packages/amplify-appsync-simulator/src/resolvers/pipeline-resolver.ts
@@ -31,13 +31,17 @@ export class AppSyncPipelineResolver {
 
     let result = {};
     let stash = {};
+    let templateErrors;
 
-    ({ result, stash } = requestMappingTemplate.render(
+    // Pipe line request mapping template
+    ({ result, stash, errors: templateErrors } = requestMappingTemplate.render(
       { source, arguments: args, stash },
       context,
       info
     ));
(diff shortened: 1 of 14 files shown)


packages/amplify-appsync-simulator/src/resolvers/pipeline-resolver.ts
```diff
@@ -31,13 +31,17 @@ export class AppSyncPipelineResolver {
 
     let result = {};
     let stash = {};
+    let templateErrors;
 
-    ({ result, stash } = requestMappingTemplate.render(
+    // Pipe line request mapping template
```
attilah: nit: "Pipeline", at some other places as well.

packages/amplify-appsync-simulator/src/server/operations.ts
```diff
@@ -123,12 +123,9 @@ export class OperationServer {
...
             operationName
           );
-          if (Object.keys(context.appsyncErrors).length) {
-            results.errors = JSON.stringify(context.appsyncErrors);
-          }
-          // Make extensions available at the root level
-          if (results.errors) {
-            results.errors = exposeGraphQLErrors(results.errors);
+          const errors = [...(results.errors || []), ...context.appsyncErrors];
+          if (errors.length) {
```
attilah: Don't we need a >0 check since the ```errors``` will always be an initialized array and will have a ```length``` property defined.

packages/amplify-appsync-simulator/src/velocity/util/errors.ts
```diff
@@ -1,24 +1,47 @@
...
+    let path = info.path;
+    const pathArray = []
+    do {
+      pathArray.splice(0, 0, path.key);
+      path = path.prev
+    } while(path);
+
+    const fieldNode = info.fieldNodes.find(f => f.name.value === fieldName);
+    const filedLocation = fieldNode && fieldNode.loc.startToken || null;
+    Object.assign(this, { message, errorType, data, errorInfo });
```
attilah: Instead of this can't we could use [Typescript constructor assignment](https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/)?

PR #2344, opened 2019-09-19
fix(amplify-provider-awscloudformation): fix amplify configure

UnleashedMind:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws.js
@@ -1,5 +1,15 @@
-//process.env.AWS_SDK_LOAD_CONFIG = true;
-const aws = require('aws-sdk');
+let aws;
+
+try {
+  delete require.cache[require.resolve('aws-sdk')];
+  process.env.AWS_SDK_LOAD_CONFIG = true;
+  aws = require('aws-sdk');
+} catch (e) {
+  delete require.cache[require.resolve('aws-sdk')];
+  delete process.env.AWS_SDK_LOAD_CONFIG;
+  aws = require('aws-sdk');
+}
+
 const proxyAgent = require('proxy-agent');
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/aws.js
```diff
@@ -1,5 +1,15 @@
-process.env.AWS_SDK_LOAD_CONFIG = true;
-const aws = require('aws-sdk');
+let aws;
+
+try {
+  delete require.cache[require.resolve('aws-sdk')];
```
yuth: why are we deleting this even before requiring this?

UnleashedMind: To ensure consistent behavior, if aws-sdk is previously required, it'll be in the cache, and the new require won't really kick off, then the process.env.AWS_SDK_LOAD_CONFIG = true; setting is nullified.

yuth: How could someone require `aws-sdk`? Don't we use this codepath to be executed all the time?

UnleashedMind: this is a safeguard, because we have a plugin platform, other plugins can require the aws-sdk before the amplify-provider-awscloudformation plugin is loaded.

yuth: If aws-sdk is loaded beforehand no harm no foul. I don't like us removing stuff from cache if its not necessary

UnleashedMind: Then it's inconsistent behavior, if the the project is configured to use 'general' config level, the default profile won't be read, aws access will be denied.

PR #2348, opened 2019-09-19
fix: #2335 - change the transform.conf.json version flag check logic

### LINKED ISSUE #2335: Stack Mapping Broken by recent CLI Upgrade
**Describe the bug**
CloudFormation has a limit of 200 Resources in a stack and my ConnectionStack.json file has exactly 200 Resources in it. I needed to add more connections so created a transform.conf.json file to direct resources to an overflow (ie. ConnectionStack2.json). This worked perfectly - compiled and pushed perfectly with the resources defined in transform.conf.json into ConnectionStack2.json.

However, I upgraded my amplify cli about a week ago and it stopped working. Now if I add any more resources to transform.conf.json this file is overwritten with the previous version and it has a "Version" label appended to the bottom of the file.

**To Reproduce**
In a version prior to 3.2.0 I had the following transform.conf.json

{
    "StackMapping": {
        "AssetVisitAttributecrowdResultsResolver": "ConnectionStack2",
        "CrowdResultscrowdSourcingJobResolver": "ConnectionStack2",
        "CrowdResultsuomResolver": "ConnectionStack2"
      }
}

This compiled and deployed correctly with these resources in the ConnectionStack2.json file.

Now after upgrading to 3.2.0 about a week ago, I edit the transform.conf.json as follows:

{
    "StackMapping": {
        "AssetVisitAttributecrowdResultsResolver": "ConnectionStack2",
        "CrowdResultscrowdSourcingJobResolver": "ConnectionStack2",
        "CrowdResultsuomResolver": "ConnectionStack2",
        "FieldListCrowdResults": "ConnectionStack2",
        "FieldListcrowdResultsResolver": "ConnectionStack2",
        "CrowdResultsfieldListResolver": "ConnectionStack2"
       }
}

Prior to amplify 3.2.0 this would have added 6 resources to ConnectionStack2. But what happens now is that my transform.conf.json reverts to the previous version with a version label as follows:

{
    "StackMapping": {
        "AssetVisitAttributecrowdResultsResolver": "ConnectionStack2",
        "CrowdResultscrowdSourcingJobResolver": "ConnectionStack2",
        "CrowdResultsuomResolver": "ConnectionStack2"
      },
     "Version": 4
}

And I end up with 203 resources in ConnectionStack.json and 3 resources in ConnectionStack2.json

**Expected behavior**
I expect 200 resources in ConnectionStack.json and 6 resources in ConnectionStack2.json and for them to be successfully deployed with push.

**Additional context**
The @connection directive in Amplify Transform is great - but, it generates 3 resources for each one. So not being able to do this stack mapping means I'm limited in my ability to define my data model.

attilah:
*Issue #, if available:*

Fixes #2335 

*Description of changes:*

The ```Version``` property check in ```transform.conf.json``` is changed to not to overwrite local file with cloud deployed version under any circumstances.

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
+++ b/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
@@ -41,18 +41,24 @@ async function transformerVersionCheck(
   updatedResources, usedDirectives,
 ) {
   const versionChangeMessage = 'The default behaviour for @auth has changed in the latest version of Amplify\nRead here for details: https://aws-amplify.github.io/docs/cli-toolchain/graphql#authorizing-subscriptions';
-  let transformerConfig;
+  const checkVersionExist = config => (config && config.Version);
+
   // this is where we check if there is a prev version of the transformer being used
   // by using the transformer.conf.json file
-  try {
-    transformerConfig = await readTransformerConfiguration(cloudBackendDirectory);
-  } catch (err) {
-    // check local resource if the question has been answered before
-    transformerConfig = await readTransformerConfiguration(resourceDir);
-  }
+  const cloudTransformerConfig = await readTransformerConfiguration(cloudBackendDirectory);
+  const cloudVersionExist = checkVersionExist(cloudTransformerConfig);
+
+  // check local resource if the question has been answered before
+  const localTransformerConfig = await readTransformerConfiguration(resourceDir);
+  const localVersionExist = checkVersionExist(localTransformerConfig);
+
+  // if we already asked the confirmation question before at a previous push
+  // or during current operations we should not ask again.
+  const showPrompt = !(cloudVersionExist || localVersionExist);
+
   const resources = updatedResources.filter(resource => resource.service === 'AppSync');
-  if (!transformerConfig.Version && usedDirectives.includes('auth')
-  && resources.length > 0) {
+
+  if (showPrompt && usedDirectives.includes('auth') && resources.length > 0) {
     if (context.exeInfo &&
       context.exeInfo.inputParams && context.exeInfo.inputParams.yes) {
       context.print.warning(`\n${versionChangeMessage}\n`);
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -41,35 +41,45 @@ async function transformerVersionCheck(
...
       }
     }
   }
-  transformerConfig.Version = 4.0;
-  await writeTransformerConfiguration(resourceDir, transformerConfig);
+
+  // Only touch the file if it misses the Version property
+  if (!localTransformerConfig.Version) {
+    localTransformerConfig.Version = 4.0;
+    await writeTransformerConfiguration(resourceDir, localTransformerConfig);
```
SwaySway: If a cloud config is only present, would this overwrite the cloud config as it is writing to local?

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -41,35 +41,45 @@ async function transformerVersionCheck(
...
+
   // this is where we check if there is a prev version of the transformer being used
   // by using the transformer.conf.json file
-  try {
-    transformerConfig = await readTransformerConfiguration(cloudBackendDirectory);
-  } catch (err) {
-    // check local resource if the question has been answered before
-    transformerConfig = await readTransformerConfiguration(resourceDir);
-  }
+  const cloudTransformerConfig = await readTransformerConfiguration(cloudBackendDirectory);
```
SwaySway: this might still throw an error if this is done on the first attempt, since there is not cloudBackendDirectory yet.

attilah: I think that ```readTransformerConfiguration``` covers it, because if the file does not exist it returns an empty object, that's the concept.

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -41,35 +41,45 @@ async function transformerVersionCheck(
...
+  const cloudVersionExist = checkVersionExist(cloudTransformerConfig);
+
+  // check local resource if the question has been answered before
+  const localTransformerConfig = await readTransformerConfiguration(resourceDir);
```
SwaySway: this would fail when adding api for the first time - if we want to catch this at the level of transformConfig I've added a pr [2354](https://github.com/aws-amplify/amplify-cli/pull/2354) to address that

PR #2357, opened 2019-09-20
fix(amplify-util-mock): use lambda fn name instead of resource name

### LINKED ISSUE #2280: Amplify mock cannot find function

**Describe the bug**
I had an amplify project with appsync, function & storage. I upgraded to amplify 3 and i tried to use `amplify mock` when it tries to start it complains that it cannot find the function `Failed to start API Mock endpoint Error: Lambda function command-processor does not exist in your project.`. The function does exist because when i run `amplify push` it is in the output
```
| Category | Resource name | Operation | Provider plugin   |
| -------- | ------------- | --------- | ----------------- |
| Function | bookings      | Update    | awscloudformation |
| Api      | bookings      | No Change | awscloudformation |
| Storage  | eventstore    | No Change | awscloudformation |
```
and when sending a request to appsync the lambda is invoked (its broken but it gets invoked).

**To Reproduce**
I don't know what the prereq is to get into this state but once you are
1. run `amplify mock`

**Expected behavior**
Mock instance is started successfully

Is there any way for me to get amplify cli to log more details?

yuth:
Mock implementation used lambda function resource name instead of the function name when mocking API. Update the implementation to parse lambda cloudformation to get function name, handler and enviornment variables and use them for mock invocation
    
fix #2280

### DIFF
--- a/packages/amplify-util-mock/src/api/api.ts
+++ b/packages/amplify-util-mock/src/api/api.ts
@@ -12,6 +12,7 @@ import { ResolverOverrides } from './resolver-overrides';
 import { ConfigOverrideManager } from '../utils/config-override';
 import { configureDDBDataSource, ensureDynamoDBTables } from '../utils/ddb-utils';
 import { invoke } from '../utils/lambda/invoke';
+import { getAllLambdaFunctions } from '../utils/lambda/load';
 
 export class APITest {
   private apiName: string;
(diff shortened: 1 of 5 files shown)


packages/amplify-util-mock/src/api/api.ts
```diff
@@ -80,7 +81,7 @@ export class APITest {
     let config: any = processAppSyncResources(stack, transformerOutput);
     await this.ensureDDBTables(config);
     this.transformerResult = this.configureDDBDataSource(config);
-    this.transformerResult = this.configureLambdaDataSource(config);
+    this.transformerResult = this.configureLambdaDataSource(context, config);
```
attilah: line 83 also assign to ```this.transformerResult```, is it intentional?

yuth: It works, but I think it is mutating the same is being assigned multiple times. In a follow up PR I am going to clean this up and make input config in helper function immuteable

PR #2358, opened 2019-09-20
feat(cli): Support for samples with --app

nikhname:
*Issue #, if available:*

*Description of changes:*
Add samples using amplify init --app <github url> in an empty directory.


packages/amplify-cli/src/lib/init-steps/preInitSetup.js
```diff
@@ -0,0 +1,121 @@
...
+ *
+ * @param packageManager either npm or yarn
+ */
+async function installPackage(packageManager) {
+  if (packageManager === 'yarn') {
+    await execSync('yarn install', { stdio: 'inherit' });
+  } else {
+    await execSync('npm install', { stdio: 'inherit' });
+  }
+}
```
UnleashedMind: on windows os, the commands become `npm.cmd` and `yarn.cmd`. you need to do a normalization based on the os. check [here](https://github.com/aws-amplify/amplify-cli/blob/f33c21f44b344b14bc9617eb97c3aa0f8330b138/packages/amplify-frontend-javascript/lib/builder.js#L36)

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Clones repo from url to current directory (must be empty)
+ *
+ * @param repoUrl the url to be cloned
+ */
+async function cloneRepo(repoUrl) {
+  execSync(`git clone ${repoUrl} .`, { stdio: 'inherit' });
```
UnleashedMind: I got this error: 

`fatal: destination path '.' already exists and is not an empty directory.`

You need to remove the `.`

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
```diff
@@ -0,0 +1,121 @@
...
+/**
+ * Generate local-env-info.json file from context
+ *
+ * @param context
+ */
+async function generateLocalEnvInfoFile(context) {
+  const { projectPath } = context.exeInfo.localEnvInfo;
+  const jsonString = JSON.stringify(context.exeInfo.localEnvInfo, null, 4);
+  const localEnvFilePath = context.amplify.pathManager.getLocalEnvFilePath(projectPath);
+  fs.writeFileSync(localEnvFilePath, jsonString, 'utf8');
```
UnleashedMind: I got this error: 

`Error: ENOENT: no such file or directory, open '/Users/zhoweimi/workspace/ExperimentProjects/amplify/.config/local-env-info.json'`

see the above comment for reason

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
UnleashedMind: And after clone the repo, you need to call process.chdir() to change the process.cwd to the repo's root directory, otherwise you will get a bunch of errors.

UnleashedMind: tested in an empty folder, and it's working end to end. 
I think this is the PM's decision if to allow the user to just run the command anywhere, or the user has to create an empty folder first and step into it and then run the command.

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
nikhname: Made this change

packages/amplify-cli/src/lib/init-steps/postInitSetup.js
```diff
@@ -0,0 +1,44 @@
...
+async function run(context) {
+  if (context.parameters.options.app) {
+    // Pushing a sample app
+    try {
+      context.parameters.options.app = true;
+      context.parameters.options.y = true;
+      context.amplify.constructExeInfo(context);
+      await context.amplify.pushResources(context);
+      const packageManager = await getPackageManager();
+      await runPackage(packageManager);
```
SwaySway: `runPackage()` has no arguments  is this the byproduct of a merge? Is the behaviour to take an argument or get packageManager within the function?

nikhname: getPackageManager should be inside the runPackage function only, removed it

packages/amplify-cli/src/lib/packageManagerHelpers.js
```diff
@@ -0,0 +1,44 @@
+const path = require('path');
+const fs = require('fs');
+
+/**
+ * Determine the package manager of the current project
+ *
+ * @return {string} 'yarn' if yarn.lock exists, 'npm' if package.json exists, undefined otherwise
+*/
+async function getPackageManager() {
+    const yarnLock = './yarn.lock';
```
yuth: Should the sample app decide what package manager to be used. Its possible that the sample app used yarn to install the dependencies, but the customer who is cloning this project does not even have yarn installed. We should also check if the customer has this package manage installed before running it

packages/amplify-cli/src/lib/init-steps/postInitSetup.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Install package using the correct package manager if package handling file exists
+ *
+ * @param packageManager either npm or yarn
+ */
+async function runPackage() {
+  const packageManager = getPackageManager();
+  const normalizedPackageManager = normalizePackageManagerForOS(packageManager);
+  if (packageManager === 'yarn') {
+    await execSync(`${normalizedPackageManager} start`, { stdio: 'inherit' });
```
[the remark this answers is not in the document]
nikhname: Do you mean the runPackage function or the execSync line? The execSync is not reused...yarn/npm start happens only once

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
```diff
@@ -0,0 +1,112 @@
...
+}
+
+/**
+ * Checks whether a url is a valid address
+ *
+ * @param str the url to validated
+ * @return true if str is a valid address, false otherwise
+ */
+async function isURL(str) {
+  const pattern = new RegExp('^(https?:\\/\\/)?' + // protocol
```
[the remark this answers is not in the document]
nikhname: I can give that a try

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
nikhname: Removed the stack trace

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
```diff
@@ -0,0 +1,112 @@
...
+  }
+}
+
+/**
+ * Set the default environment and editor for the local env
+ *
+ * @param context
+ */
+async function setLocalEnvDefaults(context) {
+  const projectPath = process.cwd();
```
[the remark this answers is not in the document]
nikhname: We have to avoid asking any more questions, including the option to use defaults. But if you think this could be breaking I can make the change

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
```diff
@@ -0,0 +1,112 @@
...
+  await generateLocalEnvInfoFile(context);
+}
+
+/**
+ * Generate local-env-info.json file from context
+ *
+ * @param context
+ */
+async function generateLocalEnvInfoFile(context) {
+  const { projectPath } = context.exeInfo.localEnvInfo;
```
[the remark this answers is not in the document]
nikhname: yes, I will export this function.

packages/amplify-cli/src/lib/packageManagerHelpers.js
nikhname: Added a check to see if the user has yarn installed

packages/amplify-cli/src/lib/packageManagerHelpers.js
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Determines the OS and returns the corresponding command given a package manager
+ *
+ * @param {string} packageManager the type of package manager detected
+ * @return {string} the package manager command for the correct OS
+ */
+async function normalizePackageManagerForOS(packageManager) {
+const isOnWindows = /^win/.test(process.platform);
+if (isOnWindows) {
+    if (packageManager === 'yarn') {
```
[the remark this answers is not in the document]
nikhname: Yep :( fixed it

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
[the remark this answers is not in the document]
nikhname: switched to this api

packages/amplify-cli/src/commands/init.js
```diff
@@ -9,11 +11,13 @@ module.exports = {
   name: 'init',
   run: async (context) => {
     constructExeInfo(context);
-    await analyzeProject.run(context)
+    await preInitSetup.run(context)
+      .then(analyzeProject.run)
       .then(initFrontendHandler.run)
       .then(initProviders.run)
       .then(onSuccess.run)
       .catch(onFailure.run);
```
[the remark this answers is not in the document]
nikhname: onFailure will exit the process so this is ok. Adding onfailure after postinit too.

packages/amplify-cli/src/lib/init-steps/postInitSetup.js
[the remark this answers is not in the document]
yuth: `execSync` we have it being used in multiple places. This might be a good candidate to have a helper function

packages/amplify-cli/src/lib/init-steps/postInitSetup.js
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Install package using the correct package manager if package handling file exists
+ *
+ * @param packageManager either npm or yarn
+ */
+async function runPackage() {
+  const packageManager = await getPackageManager();
+  const normalizedPackageManager = await normalizePackageManagerForOS(packageManager);
+  if (packageManager === 'yarn') {
```
yuth: BTW why do we need this check? we run the same statements in either case

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Clones repo from url to current directory (must be empty)
+ *
+ * @param repoUrl the url to be cloned
+ */
+async function cloneRepo(context, repoUrl) {
+  try {
+    execSync(`git clone ${repoUrl} .`, { stdio: 'inherit' });
+  } catch (e) {
```
yuth: Is this the only error that could happen when cloning the repo? I can foresee the there could be other errors, due to network issues. The reason I wanted you to ensure directory was empty was then we specifically know `git clone` is going to fail. 

May be not printing anything would also work as we are `execSync` is inheriting the stdio

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Install package using the correct package manager if package handling file exists
+ *
+ * @param packageManager either npm or yarn
+ */
+async function installPackage() {
+  const packageManager = await getPackageManager();
+  const normalizedPackageManager = await normalizePackageManagerForOS(packageManager);
+  if (packageManager === 'yarn') {
```
yuth: do we need this condition? In both the cases arguments for `execSync ` to be the same

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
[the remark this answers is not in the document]
yuth: Then we should also print message how they can change them later.

packages/amplify-cli/src/lib/packageManagerHelpers.js
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * @param {string} packageManager the type of package manager detected
+ * @return {string} the package manager command for the correct OS
+ */
+async function normalizePackageManagerForOS(packageManager) {
+  const isOnWindows = /^win/.test(process.platform);
+  if (isOnWindows) {
+    if (packageManager === 'yarn') {
+      return 'yarn.cmd';
+    } else if (packageManager === 'npm') {
```
yuth: Do we need `else`? Shouldn't we default to `npm`?

packages/amplify-cli/src/lib/init-steps/postInitSetup.js
nikhname: You're right this check is redundant - removing

packages/amplify-cli/src/lib/init-steps/postInitSetup.js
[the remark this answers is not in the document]
nikhname: Each one has different arguments though...I think I will leave it as is for this review if thats ok.

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
nikhname: Thats a good point. I'll make a check for an empty directory

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
nikhname: Removing since its redundant

packages/amplify-cli/src/lib/init-steps/preInitSetup.js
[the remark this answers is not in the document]
nikhname: Ok, adding a message for that.

packages/amplify-cli/src/lib/packageManagerHelpers.js
nikhname: Its needed because if the package manager is undefined we want to return undefined..but I will switch undefined and npm so its clear we default to npm

Issue #2361, opened 2019-09-20
Make @auth protect subscription fields fails


nookale:
**Describe the bug**
The following schema:
```gql
type Post @model @auth(rules: [{allow: owner}]) {
    id: ID!
    title: String
    owner: String
}
```

Successfully creates the following type Subscription:
```gql
type Subscription {
    onCreatePost(owner: String): Post
    onUpdatePost(owner: String): Post
    onDeletePost(owner: String): Post
}
```

While the following schema (note the combination of "allow:owner" and "ownerField"):
```gql
type Post @model @auth(rules: [{allow: owner, ownerField: "writer"}]) {
    id: ID!
    title: String
    writer: String
}
```

Will create:
```gql
type Subscription {
    onCreatePost: Post
    onUpdatePost: Post
    onDeletePost: Post
}
```

Making authorizing the subscription not possible.

**To Reproduce**
- Create react app.
- Amplify init
- Amplify add auth
- Amplify add api => graphQL
- Create schema.graphql (see code above)
- Amplify push
- Go to Appsync console and check the schema.

Is this intended? "owner" and "ownerField" must not be used together?

**Expected behavior**
Using "owner" and "ownerField" will also automatically create a type Subscription which allows authentication.

**Desktop (please complete the following information):**
 - OS: windows 10

Thanks!

attilah:
@nookale what version of the CLI are you using?

nookale:
@attilah 3.2.0

SteffeyDev:
I have the same issue as above, except for me it appears the addition of the `operations` field prevent the correct generation of the subscription type.  I need the owner to be able to read and and get subscription updates, but not create, delete, or update.  I tested, and removing the `operations` field does fix the issue.  However, this allows the user to write, which should not be allowed for my application.

Schema:
```
type VMInstance
   @model
   @key(fields: ["userId"])
   @auth(rules: [
     { allow: owner, ownerField: "userId", operations: [read] }
   ])
 {
   userId: ID!
   otherProperties: String!
}
```

Actual output:
```
type Subscription {
   onCreateVMInstance: VMInstance @aws_subscribe(mutations: ["createVMInstance"])
   onUpdateVMInstance: VMInstance @aws_subscribe(mutations: ["updateVMInstance"])
   onDeleteVMInstance: VMInstance @aws_subscribe(mutations: ["deleteVMInstance"])
 }
```

Expected output:
```
type Subscription {
[...]

PR #2368, opened 2019-09-20
fix(graphql-auth-transformer): add support for ownerfield

SwaySway:
add support for ownerField in auth subscriptions

*Issue #, if available:*
- re #2361

*Description of changes:*

- If there are no static group rules and only owner rules then the owner rules are required int the subscription operation this also applies to owner rules with an ownerField.


SwaySway:
> Please add some tests:
> 
>     * Transformation tests to see we're generating the expected schema output
> 
>     * E2E tests to make sure we cover these grounds well

- added unit tests to verify behavior
- draft pr #2400 pending this pr has e2e tests based on this change

PR #2388, opened 2019-09-23
fix(amplify-appsync-simulator): impl missing methods in velocity utils

### LINKED ISSUE #2300: Errors with $util.time when mocking locally
**Describe the bug**
Noticed two errors when working with the `$util.time` utility functions. 
1. `$util.time.nowFormatted("yyyy-MM-dd")` is giving minutes instead of the month for the MM value. I dug into the source code a bit, and it looks like the Node.js time library that is being used for this has slightly different formatting rules than AppSync's. In `dateformat` lib, MM represents minutes. In AppSync, `mm` is supposed to represent minutes. Not sure which one is the correct UTC/ISO implementation, but this causes some confusion when testing locally.

2. `$util.time.epochMilliSecondsToISO8601()` does not work locally. In this case it will throw an error: `"Cannot read property 'first_line' of undefined"`

**To Reproduce**
1. Edit any VTL template.
2. Add a line including the above time utilities and store in a variable.
3. "Log" the variable out with `$util.error()`
4. Run `amplify mock` and hit the resolver you edited

**Expected behavior**
Time utils should work locally

yuth:
Implemented missing time releated method in velocity template utils. Updated the method to support
Java Date format string

fix #2300

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-appsync-simulator/src/velocity/util/time.ts
+++ b/packages/amplify-appsync-simulator/src/velocity/util/time.ts
@@ -9,7 +9,7 @@ declare module 'moment' {
   }
 }
 
-function parseTimeStamp(dateTime: string, format?: string, timezone?: string): moment.Moment {
+const parseTimestamp = (dateTime: string, format?: string, timezone?: string): moment.Moment => {
   if (!dateTime || !format) {
     return null;
   }
@@ -52,11 +52,11 @@ export const time = () => ({
     format: string,
     timezone?: string
   ): number | null {
-    const timestamp = parseTimeStamp(dateTime, format, timezone);
+    const timestamp = parseTimestamp(dateTime, format, timezone);
     return timestamp ? timestamp.valueOf() : null;
   },
   parseISO8601ToEpochMilliSeconds(dateTime): number | null {
-    const timestamp = parseTimeStamp(dateTime, 'YYYY-MM-DDTHH:mm:ss.SZ');
+    const timestamp = parseTimestamp(dateTime, 'YYYY-MM-DDTHH:mm:ss.SZ');
     return timestamp ? timestamp.valueOf() : null;
   },
   epochMilliSecondsToSeconds(milliseconds: number): number | null {
--- a/packages/amplify-appsync-simulator/src/velocity/util/time.ts
+++ b/packages/amplify-appsync-simulator/src/velocity/util/time.ts
@@ -1,32 +1,89 @@
-import * as dateformat from 'dateformat';
+import * as moment from 'moment';
+import 'moment-timezone';
+import 'moment-jdateformatparser';
 
-export const time = now => ({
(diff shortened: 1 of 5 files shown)


packages/amplify-appsync-simulator/src/velocity/util/time.ts
```diff
@@ -1,32 +1,89 @@
...
-  nowISO8601() {
-    return now.toISOString();
+declare module 'moment' {
+  export interface Moment {
+    toMomentFormatString: (format: string) => string;
+    formatWithJDF: (format: string) => string;
+  }
+}
+
+function parseTimeStamp(dateTime: string, format?: string, timezone?: string): moment.Moment {
```
attilah: typo: "parseTimestamp", lowercase s in stamp.

TS specific:
```
const parseTimestamp = (dateTime: string, format?: string, timezone?: string): moment.Moment => {
...
}
```

instead of ```function```

yarn.lock
```diff
@@ -6618,7 +6625,7 @@ date-now@^0.1.4:
   resolved "https://registry.yarnpkg.com/date-now/-/date-now-0.1.4.tgz#eaf439fd4d4848ad74e5cc7dbef200672b9e345b"
   integrity sha1-6vQ5/U1ISK105cx9vvIAZyueNFs=
 
-dateformat@^3.0.0, dateformat@^3.0.3:
+dateformat@^3.0.0:
```
attilah: What else is using dateformat? :-O

packages/amplify-appsync-simulator/src/velocity/util/time.ts
yuth: Is there any benefit in converting this to an arrow function instead of regular old function

yarn.lock
yuth:
```
 yarn why dateformat           
yarn why v1.17.3
[1/4] 🤔  Why do we have the module "dateformat"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "dateformat@3.0.3"
info Reasons this module exists
   - "_project_#lerna#@lerna#version#@lerna#conventional-commits#conventional-changelog-core" depends on it
   - Hoisted from "_project_#lerna#@lerna#version#@lerna#conventional-commits#conventional-changelog-core#dateformat"
   - Hoisted from "_project_#lerna#@lerna#version#@lerna#conventional-commits#conventional-changelog-core#conventional-changelog-writer#dateformat"
info Disk size without dependencies: "28KB"
info Disk size with unique dependencies: "28KB"
info Disk size with transitive dependencies: "28KB"
info Number of shared dependencies: 0
✨  Done in 2.21s.
```

Issue #2386, opened 2019-09-23
[SEVERE]: Updates to default Botocore Version in AWS Lambda for Python 3.6 BREAKS Amplify ElasticSearch Lambda Streaming Function


sollipse:
**Describe the bug**
Root cause of the symptoms discussed in https://github.com/aws-amplify/amplify-js/issues/4058

I have performed a local test of Python 3.6 Lambda Execution environment to confirm:

printing BOTOCORE version available by default in lambda
according to AWS Docs, this should be 1.10.74 for Python 3.6
ACTUAL BOTOCORE VERSION:
1.12.221

In this botocore version, attempts to import the following module from botocore.endpoint fail:
from botocore.endpoint import BotocoreHTTPSession

This in turn cause every write to ElasticSearch from DDB Stream functions to fail, for all tables. 

**To Reproduce**
In any current amplify environment, visit the logs for the ES Streaming function and confirm that they indicate an error when importing BotocoreHTTPSession

**High Severity**
This change is affecting your customers *without a deployment or amplify push*. Real production customers using your stack are experiencing total failures to write to ES, with no deployment to root cause. For applications that use Amplify ES Search as a core functionality, this is the equivalent of no new records being written to their primary query layer. 

### ⚠️Please treat this as a high severity issue. This is not a drill. ⚠️

kaustavghosh06:
@sollipse We're looking into it. I'll update on this as soon as we have a fix for it. As a workaround please take a look at - https://github.com/aws-amplify/amplify-cli/issues/2353#issuecomment-533558801 for a workaround until we push the fix.

kaustavghosh06:
@sollipse As discussed offline, this issue should be fixed from the Lambda service side as they rolled-back their changes. Please feel free to comment on this thread if you're still seeing this issue.

kaustavghosh06:
@sollipse We dug into the issue and after talking to the Lambda team, we understood the reason for the change and errors. We're working on a fix in the CLI that will enable you to re-deploy your function as a part of the ampliy push command to avoid seeing the Lambda errors. For the time being, here's an explanation and the workaround provided by the Lambda team to prevent your app from seeing errors:

[...]

Issue #2385, opened 2019-09-21
Unable to stream models to ElasticSearch


usmansbk:
**What is the current behavior?**
When I create a mutation, the `lambda` function that streams `dynamodb` `model`s to `elasticsearch` fails and logs `Unable to import module 'python_streaming_function': cannot import name 'BotocoreHTTPSession'`.
I haven't edited any of the code as it is automatically generated by the `GraphQL Transform`

**What is the expected behavior?**
I should be able to search my models in `AWS elasticsearch`

**Which versions and which environment (browser, react-native, nodejs) / OS are affected by this issue? Did this work in previous versions?**

```
    "aws-amplify": "^1.1.40",
    "aws-amplify-react-native": "^2.1.19",
    "aws-appsync": "^2.0.0",
    "aws-appsync-react": "^2.0.0",
    "react": "^16.9.0",
    "react-apollo": "2.5.8",
    "react-native": "^0.60.5",

     aws-cli/1.16.243 Python/3.6.0 Windows/10 botocore/1.12.233
```

kaustavghosh06:
We're looking into it. I'll update on this as soon as we have a fix for it. As a workaround please take a look at - #2353 (comment) for a workaround until we push the fix.

kaustavghosh06:
This issue should be fixed from the Lambda service side as they rolled-back their changes - causing this behavior. Please feel free to comment on this thread if you're still seeing this issue.

kaustavghosh06:
@usmansbk  We dug into the issue and after talking to the Lambda team, we understood the reason for the change and errors. We're working on a fix in the CLI that will enable you to re-deploy your function as a part of the ampliy push command to avoid seeing the Lambda errors. For the time being, here's an explanation and the workaround provided by the Lambda team to prevent your app from seeing errors:

_AWS Lambda is updating the default boto3 and botocore SDKs for the Python 2.7, Python 3.6 and Python 3.7 runtimes. This update changes the version of boto3 from 1.7.74 to 1.9.221 for the Python 2.7 and Python 3.6 runtimes and from 1.9.42 to 1.9.221 for the Python 3.7 runtime. The update contains SDK enhancements and updated APIs. One notable example of an internal class that underwent change is the BotocoreHTTPSessions class which is no longer supported starting with botocore version 1.11.0._

Since the CLI used the `BotocoreHTTPSessions class`, that's the reason why you're seeing these errors.

[...]

PR #2393, opened 2019-09-24
fix: bundle up dependencies  for DdbToESFn

nikhname:
*Issue #, if available:*
#2386 
#2385 
#2353 
*Description of changes:*
Dependencies to fix botocoreHTTPSession which is deprecated with the latest Lambda runtime update for Python 3.6 - cause Lambda invocation failures.


packages/graphql-elasticsearch-transformer/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "build": "tsc && cd streaming-lambda && bestzip ../lib/streaming-lambda.zip python_streaming_function.py",
```
SwaySway: Can we add the zip command in here as well? and watch
`"tsc $$ zip"`

PR #2394, opened 2019-09-24
fix(cli): add cli core aliases, and two minor fixes

UnleashedMind:
some minor fixes: 
1. add back cli core aliases: "h", "ls", "serve"

2. relax input params requirements to allow plain string
instead of just Json object. 

3. fix a bug when awscloudformation configure is run directly instead of invoked by the core confgiure command

*Issue #, if available:*

*Description of changes:*


packages/amplify-cli/src/lib/input-params-manager.js
```diff
@@ -26,7 +26,12 @@ function normalizeKey(key) {
 }
 
 function normalizeValue(key, value) {
-  const normalizedValue = JSON.parse(value);
+  let normalizedValue = value;
+  try {
+    normalizedValue = JSON.parse(value);
+  } catch (e) {
+    // do nothing
```
kaustavghosh06: This change seems unrelated to the PR. What's this for?

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
```diff
@@ -30,6 +30,7 @@ async function init(context) {
 }
 
 async function configure(context) {
+  context.exeInfo = context.exeInfo || context.amplify.getProjectDetails();
```
kaustavghosh06: This change seems unrelated to the PR. What's this for?

packages/amplify-cli/src/lib/input-params-manager.js
UnleashedMind: It's in the description of the PR. 
It's too restrictive to only allow Json string as parameters. 
This update allows plain string as parameter, which is used for things such as sample's '--app' option.

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
UnleashedMind: The provider plugin command: `amplify awscloudformation configure` is broken for a while, but I guess it's not directly used by the users, but invoked through "amplify configure project", which works,  so it's not  discovered. This fixes it the direct invocation of the configuration of the provider plugin.

UnleashedMind:
This PR contains three separate updates because they are all small fixes. But it can also be separated into 3 individual PR's depending on the team's opinions.

packages/amplify-cli/src/lib/input-params-manager.js
```diff
@@ -26,10 +26,12 @@ function normalizeKey(key) {
 }
 
 function normalizeValue(key, value) {
-  if (key === 'app') {
```
kaustavghosh06: Why are we removing this? This is a part of the latest release for the sample apps I believe. @nikhname could you please review this?

packages/amplify-cli/src/lib/input-params-manager.js
nikhname: I'm not sure this is the best way to allow a plain string parameter...do we want to return the value if an invalid JSON is passed?

packages/amplify-cli/src/lib/input-params-manager.js
nikhname: Yes, I don't think this should be removed. The workaround below is a general way to do this but it ignores errors in JSON.parse

packages/amplify-cli/src/lib/input-params-manager.js
UnleashedMind: If the value can not be parsed as JSON, then it's passed as plain string.

PR #2399, opened 2019-09-25
fix(graphql-elasticsearch-transformer): add layer based on region

### LINKED ISSUE #2386: [SEVERE]: Updates to default Botocore Version in AWS Lambda for Python 3.6 BREAKS Amplify ElasticSearch Lambda Streaming Function

nikhname:
Using a map to determine the correct arn layer from the region of the user

fix #2386

*Issue #, if available:*
#2386

*Description of changes:*
Add a layer that contains the previous version of the SDK to the lambda resource based on the region of the user

### DIFF
--- a/packages/graphql-elasticsearch-transformer/src/resources.ts
+++ b/packages/graphql-elasticsearch-transformer/src/resources.ts
@@ -10,14 +10,7 @@ import {
     int, Expression
 } from 'graphql-mapping-template'
 import { toUpper, plurality, graphqlName, ResourceConstants, ModelResourceIDs } from 'graphql-transformer-common'
-
-interface Mappings {
-    [key: string]: {
-        [key: string]: {
-            [key: string]: string | number | string[];
-        };
-    };
-};
+import { MappingParameters } from 'graphql-transformer-core/src/TransformerContext'
 
 export class ResourceFactory {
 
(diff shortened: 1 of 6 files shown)


packages/graphql-elasticsearch-transformer/src/resources.ts
```diff
@@ -11,6 +11,14 @@ import {
 } from 'graphql-mapping-template'
 import { toUpper, plurality, graphqlName, ResourceConstants, ModelResourceIDs } from 'graphql-transformer-common'
 
+interface Mappings {
```
SwaySway: nit: the mapping interface in here and in transformer context might want to be the same in terms of naming convention though I think it's only best defined in the transformer context

nikhname: Made this change, interface is now MappingParameters and is imported from transformer context

Issue #2405, opened 2019-09-25
Joint access to DynamoDB record with AppSync using Dynamic Group Authorization


kevinold:
** Which Category is your question related to? **
Authentication and Authorization

** What AWS Services are you utilizing? **
Amplify CLI 3.9.0, AppSync, Cognito, DynamoDB

** Provide additional details e.g. code snippets **

I have a requirement of having 2 users with shared access to a DynamoDB Item.

I cannot use a Cognito Group for the users, as I wish to have 2 users that sign up via social authentication "connect" to each other, so the "group" would be a dynamic one, in which the "PeopleCanAccess" list is perfect.

I understand via https://docs.aws.amazon.com/appsync/latest/devguide/security-authorization-use-cases.html#reading-data in the "Use Case: Multiple People Can Read" how my resolver should be for the read case.

"Use Case: Multiple Owners" under "Writing Data" (https://docs.aws.amazon.com/appsync/latest/devguide/security-authorization-use-cases.html#writing-data) says it is for the PeopleCanAccess, but I'm confused with this expectedOwner example.

Using the common Todo's app, here's the scenario for what I need to accomplish in this "shared todos" version of the app:

1. UserA registered with the service
2. UserA can do the usual CRUD for a list of todo's
3. UserA can "invite" UserB to their Todo List (thinking a Lambda to do this with an invite code of some sort"
4. UserB registers via invite code
5. UserB has access to Todo's
6. When UserA or UserB CRUD's Todo's, PeopleCanAccess is updated accordingly.
7. Thinking through this, using the invite mechanism I described in step 3, I'd have an attribute on UserA's record to holder UserB's ID. When a Todo item is CRUD'd by UserA, I'd envision the resolver to write UserB to the PeopleCanAccess list.

I've looked into Dynamic Group Authorization in the GraphQL Transform: https://aws-amplify.github.io/docs/cli-toolchain/graphql#dynamic-group-authorization and tried to combine them with Custom Claims: https://aws-amplify.github.io/docs/cli-toolchain/graphql#custom-claims

I've setup a test project (https://github.com/kevinold/amplify-notes-test) which:

- Limited a Task type to use the Dynamic Groups https://github.com/kevinold/amplify-notes-test/blob/master/amplify/backend/api/amplifynotes/schema.graphql#L5

[...]

PR #2411, opened 2019-09-25
fix(graphql-auth-transformer): fixed usergroups for groupclaim

SwaySway:
*Issue #, if available:*
re #2405
*Description of changes:*
fixed userGroups in resolver when using groupClaim and adding unit test to verify this behavior


RossWilliams:
btw, this fix is also in #2180

PR #2428, opened 2019-09-27
feat: move away from mosca to forked mqtt server

yuth:
Amplify CLI mock uses MQTT for AppSync subscription. The MQTT server used was mosca which has provides feature which are not used by mock and brings in outdated dependencies and causes some optional dependency installation failures . Forking Mosca and slimming down the dependencies. 

Fix https://github.com/aws-amplify/amplify-cli/issues/2440


packages/amplify-mqtt-server/src/server.ts
```diff
@@ -0,0 +1,290 @@
...
+    this.closed = true;
+
+    Object.keys(this.clients).forEach(i => {
+      stuffToClose.push(this.clients[i]);
+    });
+
+    steed.each(
+      stuffToClose,
+      (toClose, cb) => {
+        toClose.close(cb, 'server closed');
```
attilah: What do you think about putting the ```close``` invocation into a try catch to make sure that if 1 close fails, then we can still call the others and not bailing out from the loop.

packages/amplify-mqtt-server/src/server.ts
```diff
@@ -0,0 +1,290 @@
...
+        this.listener.close(() => {
+          this.logger.info('server closed');
+          this.emit('closed');
+          callback();
+        });
+      }
+    );
+  }
+
+  generateUniqueId() {
```
attilah: Why put it into a separate function if invoked from 1 place?

packages/amplify-mqtt-server/src/client.ts
```diff
@@ -0,0 +1,440 @@
...
+  private id;
+  private will;
+  private clean;
+  constructor(private connection: Connection, private server: MQTTServer) {
+    this.logger = server.logger;
+    this.setup();
+  }
+
+  private setup() {
+    const client = this.connection;
```
attilah: Please think of renaming, calling something totally different then its type can be misleading when reading the code, because a question will be raised, how a connection become a client? IMHO it is either always a connection or always a client, but not both. It degrades the maintainability index, and in this case we've a ```client``` member in a ```Client``` class ;-)

packages/amplify-mqtt-server/src/client.ts
```diff
@@ -0,0 +1,440 @@
...
+  inflightCounter: number = 0;
+  private keepAlive: number = 0;
+  private _lastDedupId: number = -1;
+  private _closed: boolean = false;
+  private _closing: boolean = false;
+  private timer;
+  private id;
+  private will;
+  private clean;
+  constructor(private connection: Connection, private server: MQTTServer) {
```
attilah: Based on how the ```Client``` is getting instantiated, it will own the ```Connection``` instance, is everything cleaned up upon the destruction of the ```Client``` instance?

yuth: We have cleanup code running which do its job

packages/amplify-mqtt-server/src/server.ts
```diff
@@ -0,0 +1,290 @@
...
+import * as ws from 'websocket-stream';
+import * as steed from 'steed';
+import * as pino from 'pino';
+import * as extend from 'extend';
+import * as nanoid from 'nanoid';
+import { TrieListener } from './trie-listener';
+import { Client } from './client';
+import { EventEmitter } from 'events';
+
+export type MQTTServerOptions = {
```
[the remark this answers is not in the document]
yuth: Can you elobarate

Issue #2296, opened 2019-09-15
No key generated for gsi using @connection


tw1t611:
**Describe the bug**
Hi, I want to connect two types using the following schema:
The first type uses email as key field.
This seems to be a very common use case, therefore I don't really know, if I am missing something.

```gql
type Webshop
  @model(subscriptions: null)
  @key(fields: ["email"])
  @auth(
    rules: [
      { allow: owner }
      { allow: groups, groups: ["Users"], queries: [get], mutations: null }
    ]
  ) {
  email: String!
  products: [Product] @connection(name: "WebshopProducts")
}

type Product
  @model(subscriptions: null)
  @auth(
    rules: [
      { allow: owner }
      {
        allow: groups
        groups: ["Users"]
        queries: [get, list]
        mutations: null
      }
    ]
  ) {
  id: ID!
  title: String!
  description: String!
  price: Int
  images: [Image] @connection
  webshop: Webshop @connection(name: "WebshopProducts")
}
```

GSI:
```
Webshop.json: none
Product.json:
"AttributeName": "productWebshopId",
"KeyType": "HASH
```

**To Reproduce**
1. Generate the schema.
2. Mock the api.
3. Create a Webshop and a related Product.
4. Query the webshop.
5. Get this error:
```
{
  "data": {
    "__typename": "Query",
    "getProduct": {
      "webshop": null,
      "id": "test"
    }
  },
  "errors": [
    {
      "message": "One of the required keys was not given a value",
      "locations": [
        {
          "line": 8,
          "column": 5
        }
      ],
      "path": [
        "getProduct",
        "webshop"
      ],
      "errorType": "DynamoDB:ValidationException"
    }
  ]
}
```

**Expected behavior**
As a response to the query, I want to get the webshop with it's related products.

**Desktop (please complete the following information):**
 - latest Debian
 - latest Chromium

chrisco255:
Getting the same error when mocking locally. Documentation claims the `@connection` transformer is covered by local mocking, but it doesn't seem to be working.

attilah:
@tw1t611 @chrisco255  could you please tell me which version of the CLI are you using? That would help the troubleshooting.

I can confirm that we've some issues around custom primary keys and connection fields, but the investigation is in progress.

The problem seems to be unrelated to @auth

RossWilliams:
[...]

Issue #2304, opened 2019-09-16
connectionAttribute holds default value, not the renamed key value


Delannor:
When setting keyField values for connections the connectionAttribute in _type_._field_.req.vtl file holds a default generated value and not the renamed keyField value.

I have the following graphql schema:
```graphql
type TestUser @model(subscriptions: null) {
  id: ID!
  followees(followerId: String): [TestUserRelation!] @connection(name: "TestUserFollowees")
  followers: [TestUserRelation!] @connection(name: "TestUserFollowers")
}

type TestUserRelation @model(queries: null, subscriptions: null) @key(fields: ["followerId", "followeeId"]) {
  followerId: String!
  follower: TestUser @connection(name: "TestUserFollowees", keyField: "followerId")
  followeeId: String!
  followee: TestUser @connection(name: "TestUserFollowers", keyField: "followeeId")
}
```
I push it to the cloud using amplify version 3.2.0
Then I create some test data:
```graphql
mutation createA{createTestUser(input:{id:"A@test.com"}){id}}

mutation createB{createTestUser(input:{id: "B@test.com"}){id}}

mutation BfollowsA{createTestUserRelation(input:{followerId:"B@test.com", followeeId:"A@test.com"}){followerId followeeId}}
```
When trying to execute the following query:
```graphql
query getBFollowees{getTestUser(id:"B@test.com"){id followees{items{followee{id}}}}}
```
I get this error:
```
Query condition missed key schema element: followerId
```
I can see that the connectionAttribute in the generated resolver file, TestUser.followees.req.vtl, is set to "testUserRelationFollowerId" and likewise TestUser.followers.req.vtl has connectionAttribute set to "testUserRelationFolloweeId".
```vtl
...
#set( $query = {
  "expression": "#connectionAttribute = :connectionAttribute",
  "expressionNames": {
      "#connectionAttribute": "testUserRelationFollowerId"
  },
  "expressionValues": {
      ":connectionAttribute": {
          "S": "$context.source.id"
    }
  }
} )
...
```

If I override the generated resolver files and replace the connectionAttribute names with followerId and followeeId respectively then I get the expected result from the query.
```
...
      "#connectionAttribute": "followerId"
...
```
I would have expected the auto-generated code to use the explicit naming and not default to toCamelCase([type, field, 'id'])

RossWilliams:
[...]

PR #2439, opened 2019-09-27
fix: #2296 #2304

attilah:
*Issue #, if available:*

Fixes: #2296, #2304

### DIFF
--- a/packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
+++ b/packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
@@ -196,12 +196,12 @@ export class ModelConnectionTransformer extends Transformer {
 
         // Checks if "fields" argument is provided which indicates use of the new parameterization
         // hence dive straight to new logic and return.
-        if (getDirectiveArgument(directive)("fields")) {
+        if (getDirectiveArgument(directive, 'fields')) {
             this.connectionWithKey(parent, field, directive, ctx)
             return
         }
 
-        let connectionName = getDirectiveArgument(directive)("name")
+        let connectionName = getDirectiveArgument(directive, 'name');
         let associatedSortFieldName = null
         let sortType = null
         // Find the associated connection field if one exists.
@@ -213,9 +213,9 @@ export class ModelConnectionTransformer extends Transformer {
                 }
                 const relatedDirective = f.directives.find((dir: DirectiveNode) => dir.name.value === 'connection');
                 if (relatedDirective) {
-                    const relatedDirectiveName = getDirectiveArgument(relatedDirective)("name")
+                    const relatedDirectiveName = getDirectiveArgument(relatedDirective, 'name');
                     if (connectionName && relatedDirectiveName && relatedDirectiveName === connectionName) {
-                        associatedSortFieldName = getDirectiveArgument(relatedDirective)('sortField')
+                        associatedSortFieldName = getDirectiveArgument(relatedDirective, 'sortField');
                         return true
                     }
                 }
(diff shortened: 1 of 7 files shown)


packages/graphql-connection-transformer/src/resources.ts
```diff
@@ -105,17 +105,37 @@ export class ResourceFactory {
      * @param relatedType The name of the related type to fetch from.
      * @param connectionAttribute The name of the underlying attribute containing the id.
      */
-    public makeGetItemConnectionResolver(type: string, field: string, relatedType: string, connectionAttribute: string): Resolver {
-        return new Resolver({
+    public makeGetItemConnectionResolver(type: string, field: string, relatedType: string, connectionAttribute: string,
```
yuth: Can you update the doc strings to include the new arguments

packages/graphql-transformer-common/src/util.ts
```diff
@@ -28,4 +28,5 @@ export function toPascalCase(words: string[]): string {
     return formatted.join('')
 }
 
-export const NONE_VALUE = '___xamznone____'
\ No newline at end of file
+export const NONE_VALUE = '___xamznone____'
+export const NONE_INT_VALUE = -2147483648;
```
yuth: What does this magic number mean

packages/graphql-transformers-e2e-tests/src/__tests__/ModelConnectionWithKeyTransformer.e2e.test.ts
```diff
@@ -0,0 +1,452 @@
+import { ResourceConstants } from 'graphql-transformer-common'
```
yuth: It would be great if you could add this test on mock-e2e as well

packages/graphql-transformers-e2e-tests/src/__tests__/ModelConnectionWithKeyTransformer.e2e.test.ts
```diff
@@ -0,0 +1,452 @@
...
+    expect(items[0].teams.items.length).toEqual(2)
+    expect(items[0].teams.items[0].teamId).toEqual('T1')
+    expect(items[0].teams.items[0].project).toBeDefined()
+    expect(items[0].teams.items[0].project.projectId).toEqual('P1')
+    expect(items[0].teams.items[1].teamId).toEqual('T2')
+    expect(items[0].teams.items[1].project).toBeDefined()
+    expect(items[0].teams.items[1].project.projectId).toEqual('P1')
+})
+
+test('Unnamed connection with sortField parameter only #2100', async () => {
```
yuth: Can you add listModel test and see if the items are sorted as expected as well

packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -651,4 +725,45 @@ export class ModelConnectionTransformer extends Transformer {
             }
         }
     }
+
```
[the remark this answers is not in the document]
attilah: The reason for that is other properties of the fieldNode - like field type - will be needed by future changes.

packages/graphql-transformers-e2e-tests/src/__tests__/ModelConnectionWithKeyTransformer.e2e.test.ts
attilah: That case is already covered by @key tests, what is new here is connection via sort field.

packages/graphql-transformer-common/src/util.ts
attilah: Int32.MinValue, we cannot use 0 as a "none value" for numeric default-if-null default values.

Issue #766, opened 2019-01-24
RFC: Auth Enhancements - Admin Tasks


undefobj:
**Overview**: Currently the Amplify CLI supports provisioning Cognito User Pools and Cognito Identity Pools with a handful of options such as:
-	Creation of User Pools and Federation with Identity Pools
o	Default “best practice” settings for connecting to common services, as well as the ability to overwrite defaults
-	Social Provider Federation with Cognito Identity
-	IAM role scoping for unauthenticated Read/Write access

Customer feedback and interviews, as well as triaging common issues for building apps we have found that there are many common "administrative tasks" that we can automate and setup with enhancements to the CLI auth category, reducing the need to manually connect things within the AWS console. This RFC is a proposal for feedback on workflow and requirements as part of this effort.
 
**Common asks/current feature plans:**
Below are the common asks we have heard from the community, please reply with a +1 or detailed comment on a feature if you have specific thoughts around how it should work:
 

- Cognito User Pools Hosted UI support

- Federation with the Cognito User Pools Hosted UI (this will be the default Federation path for Social Providers, and the returned OIDC token from User Pools will be used to get AWS credentials from Cognito Identity Pools).
 
- Triggers
  - Ability to select from common “Templates” and add them as User Pool Triggers. These are Lambda functions and will be written in NodeJS. The functions will be saved in your project directory so that you can edit them and push to the cloud.
  - Current proposed templates:
    - MFA w/ Captcha or QR code similar to this: https://aws.amazon.com/blogs/mobile/customizing-your-user-pool-authentication-flow/ 
    - Device tracking registration similar to this: https://aws.amazon.com/blogs/mobile/tracking-and-remembering-devices-using-amazon-cognito-your-user-pools/ 
    - Sign-Up email filtering (e.g. whitelist/blacklist email domains)
    - Email confirmation link instead of code: https://github.com/aws-amplify/amplify-js/issues/612
    - Auto-add to Group upon post registration (potentially using SAML claims)
 
- Fine-grained access control by dynamically assigning roles based on Group membership or rules/token claims. How would you like this to work in your developer workflow? If you’re unfamiliar with this capability you can read more here: https://aws.amazon.com/blogs/aws/new-amazon-cognito-groups-and-fine-grained-role-based-access-control-2/
  - Essentially this would allow you to dynamically control what groups/users could access specific GraphQL/REST APIs, S3 resources, chatbots, etc.
 
[...]

PR #2443, opened 2019-09-27
feat: User Pool Groups, Admin Auth Support, Custom Group Role Policies

undefobj:
*Issue #, if available:*
#766 

*Description of changes:*
User Pool Groups, Admin Lambda Functions


packages/amplify-category-auth/provider-utils/awscloudformation/index.js
```diff
@@ -607,6 +723,145 @@ function removeDeprecatedProps(props) {
   return props;
 }
 
+async function addAdminAuth(context, authResourceName, operation, adminGroup, functionName) {
```
haverchuck: We don't have to do this right now, but we should start breaking up this index file.

Issue #2451, opened 2019-09-28
Type definitions are missing for most modules


alexmcmanus:
Module `graphql-transformer-core` has typescript definitions - but in order to use it standalone, you need a bunch of other modules that do not. E.g. `graphql-dynamodb-transformer`. Given the source code is Typescript, there doesn't seem to be any reason why you wouldn't also generate type definitions for all of the modules (excepting a few modules where `allowJs` is set: amplify-cli, amplify-graphql-explorer, amplify-graphql-types-generator, amplify-ui-tests).

When you alter the `tsconfig.json` to generate type definitions, a few minor type errors emerge in the `amplify-cli` build (PR to follow).

```
import { DynamoDBModelTransformer } from 'graphql-dynamodb-transformer'
```
results in an `Could not find a declaration file for module` error on compilation.

**Expected behavior**
Published NPM packages to include type definitions, except where `allowJs` setting makes that impossible.

renebrandel:
Closing as the related PR is merged.

PR #2452, opened 2019-09-28
fix: export Typescript definitions and fix resulting type errors

alexmcmanus:
Few of the modules export their Typescript definitions, despite being written in Typescript. This
makes them more difficult to use standalone from Typescript code. As a result of exporting the
definitions, some minor type errors emerge.

#2451

*Issue #, if available:*

*Description of changes:*


packages/amplify-util-mock/src/CFNParser/appsync-resource-processor.ts
```diff
@@ -257,16 +259,21 @@ export function processResources(
...
+    },
     resolvers: [],
     functions: [],
     dataSources: [],
     mappingTemplates: [],
     tables: [],
     appSync: {
       name: '',
-      defaultAuthenticationType: {},
+      defaultAuthenticationType: {
```
yuth: Why OIDC as default auth type? I would prefer this to be APIKey or cognito

packages/amplify-util-mock/src/__e2e__/versioned-model-transformer.e2e.test.ts
```diff
@@ -29,14 +29,7 @@ beforeAll(async () => {
...
         new DynamoDBModelTransformer(),
-        new ModelAuthTransformer({
-          authConfig: {
-            defaultAuthentication: {
-              authenticationType: 'API_KEY',
-            },
-            additionalAuthenticationProviders: [],
-          },
-        }),
+        new ModelAuthTransformer(),
```
yuth: why was this removed?

packages/amplify-util-mock/src/CFNParser/appsync-resource-processor.ts
alexmcmanus: Changed to API_KEY.

packages/amplify-util-mock/src/__e2e__/versioned-model-transformer.e2e.test.ts
alexmcmanus: This is an interesting one. Following back to the import, you see:
```ts
import {
  default as ModelAuthTransformer,
  default as VersionedModelTransformer,
} from 'graphql-versioned-transformer';
```
So, `ModelAuthTransformer` is in fact `VersionedModelTransformer` and the TS parser correctly rejects the `authConfig` parameters. I've no idea whether it makes sense to have a `VersionedModelTransformer` masquerading as a `ModelAuthTransformer`, so otherwise I left it alone. 

This does highlight why it's useful to get the type checking working across packages.

alexmcmanus: If I make that change, there will be no import of `ModelAuthTransformer`. Just so I'm clear, are you asking me to drop `ModelAuthTransformer` from the list of transformers?

attilah: That import statement imports whatever is the default with an alias, basically importing the VersionTransformer class with 2 names, which is not good, but did not make harm. Since in the e2e test no ```@auth``` is used I think we don't need the 'real' ModelAuthTransformer imported.

alexmcmanus: Done.

PR #2462, opened 2019-09-30
ci: upload test run reports from e2e tests

yuth:
Update e2e test configuration to generate test results in JUnit format and updated circle ci config
to upload those results for analysis


packages/amplify-ui-tests/package.json
```diff
@@ -27,7 +27,8 @@
     "rimraf": "2.6.2",
     "ts-jest": "22.4.6",
     "tslint": "5.10.0",
-    "typescript": "2.8.3"
+    "typescript": "2.8.3",
```
attilah: typescript version is 3.5.3 what we use all over.

PR #2484, opened 2019-10-03
auth on @searchable

SwaySway:
*Issue #, if available:*
re #1043
*Description of changes:*
added auth logic for `@searchable`

### DIFF
--- a/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
+++ b/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
@@ -115,7 +115,8 @@ const validateAuthModes = (authConfig: AppSyncAuthConfiguration) => {
 }
 
 export type ModelAuthTransformerConfig = {
-    authConfig?: AppSyncAuthConfiguration
+    authConfig?: AppSyncAuthConfiguration,
+    searchableFlag?: boolean
 };
 
 export type ConfiguredAuthProviders = {
@@ -310,6 +311,10 @@ export class ModelAuthTransformer extends Transformer {
             throw new InvalidDirectiveError('Types annotated with @auth must also be annotated with @model.')
         }
 
+        // check if searchable is enabled on the type
+        const searchableDirective = def.directives.find((dir) => dir.name.value === 'searchable')
+        this.config.searchableFlag =  searchableDirective ? true : false
+
         // Get and validate the auth rules.
         const rules = this.getAuthRulesFromDirective(directive);
         // Assign default providers to rules where no provider was explicitly defined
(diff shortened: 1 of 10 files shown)


packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -1063,4 +1063,40 @@ identityClaim: "${rule.identityField || rule.identityClaim || DEFAULT_IDENTITY_F
...
+     * ES EXPRESSIONS
+     */
+
+    public makeESItemsExpression() {
+        // generate es expresion to appsync
+        return compoundExpression([
+            set(ref('es_items'), list([])),
+            forEach(
+                ref('entry'),
+                ref('context.result.hits.hits'),
```
nikhname: Is context.results.hits.hits intentional?

packages/graphql-transformers-e2e-tests/src/cognitoUtils.ts
```diff
@@ -163,3 +176,212 @@ export async function createUserPoolClient(client: CognitoClient, userPoolId: st
...
+    const identityPoolRoleMap = new cfnCognito.IdentityPoolRoleAttachment({
+        IdentityPoolId: { Ref: 'IdentityPool' } as unknown as string,
+        Roles: {
+            'unauthenticated': { 'Fn::GetAtt': [ 'UnauthRole', 'Arn' ] },
+            'authenticated': { 'Fn::GetAtt': [ 'AuthRole', 'Arn' ] }
+        }
+    });
+
+    const userPoolClientWeb = new cfnCognito.UserPoolClient({
+        ClientName: USER_POOL_CLIENTWEB_NAME,
```
nikhname: Why don't we generate a secret for userPoolClientWeb?

packages/graphql-auth-transformer/src/resources.ts
SwaySway: Yup that's the response given back from elasticsearch
A hits object which contains list of hits.

packages/graphql-transformers-e2e-tests/src/cognitoUtils.ts
SwaySway: I don't think this is needed as the clientID is this e2e is testing auth resolver logic against Elasticsearch responses.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1469,6 +1480,33 @@ All @auth directives used on field definitions are performed when the field is r
         }
     }
 
+    private protectSearchQuery(ctx: TransformerContext, def: ObjectTypeDefinitionNode,
+        resolverResourceId: string, rules: AuthRule[]) {
+            const resolver = ctx.getResource(resolverResourceId);
+            if (!rules || rules.length === 0 || !resolver) {
```
yuth: When woukd `resolver` be falsy? Should we not protect if resolver is `falsy`

SwaySway: The `resolver` would be falsy if say if the respective query for searchable is set to null.
```gql
@searchable(queries: null)
```

PR #2497, opened 2019-10-04
fix(amplify-appsync-simulator): support early return from resolvers

### LINKED ISSUE #2427: API Mock error when returning within request resolver.
When mocking the API, if you use `#return` within the request velocity template, an error is thrown.  When deployed to the cloud, it works as expected.

Example.req.vtl
```
#return("Foo Bar")
```

Here is an example error message:
```json
{
  "data": {
    "__typename": "Mutation",
    "createCompanyEntity": {
      "properties": null,
      "name": "Foo Bar",
      "id": "6a4a5500-f1a2-4612-b22b-49661678b73a"
    }
  },
  "errors": [
    {
      "message": "Unable to convert   {mapper=function map(value) {\n    if (value instanceof map_1.JavaMap)\n        return value;\n    if (value instanceof array_1.JavaArray)\n        return value;\n    if (Array.isArray(value)) {\n        return new array_1.JavaArray(value.map(function (x) { return map(x); }), map);\n    }\n    if (lodash_1.isPlainObject(value)) {\n        return map_1.createMapProxy(new map_1.JavaMap(Object.entries(value).reduce(function (sum, _a) {\n            var _b;\n            var k = _a[0], v = _a[1];\n            return __assign(__assign({}, sum), (_b = {}, _b[k] = map(v), _b));\n        }, {}), map));\n    }\n    // eslint-disable-next-line\n    if (typeof value === 'string' && !(value instanceof string_1.JavaString)) {\n        // eslint-disable-next-line\n        return new string_1.JavaString(value);\n    }\n    // for now we don't handle number.\n    return value;\n}, map=[object Map]} to class com.amazonaws.deepdish.transform.model.lambda.LambdaVersionedConfig.",
      "errorType": "MappingTemplate",
      "data": null,
      "errorInfo": null,
      "path": [
        "createCompanyEntity",
        "properties"
      ],
      "locations": [
        {
          "line": 4,
          "column": 5,
          "sourceName": "GraphQL request"
        }
      ]
    }
  ]
}
```

yuth:
AppSync has support to return early from the resolver, if the template has #return statement. If the
#return statement is executed in request mapping template, niether the invoker or response template
is executed. If the #return statemnt is executed in an AppSync function, the function returns
immediately with out executing the remaining template or data invoker. Update the mock
impletentation to support this behavior

fix #2427

*Issue #, if available:*

*Description of changes:*


packages/amplify-velocity-template/src/compile/compile.js
```diff
@@ -96,8 +96,16 @@ module.exports = function(Velocity, utils) {
 
             case 'return':
               this._state.stop = true;
+              this._state.return = true;
               if (ast.value) {
-                str += this.format(this.getExpression(ast.value, true));
+                const result = this.getExpression(ast.value, true);
+                try {
+                  str = JSON.stringify(result);
```
SwaySway: in this switch you are appending the result to `str` is that not expected in this switch case?

yuth:
> Please add some tests where we can verify that the execution was finished by the return statement in request and in response resolver as well, otherwise LGTM.

I will add the tests in up coming PR

marcoreni:
@attialh @yuth do we need the tests in this PR for it to be merged? Can we contribute somehow? 

We use return heavily in our VTLs so this is blocking for us.

Issue #1984, opened 2019-08-07
AWS Amplify removes "id" field from generated input types


mattjennings:
** Which Category is your question related to? **
AWS Amplify codegen

** What AWS Services are you utilizing? **
AWS Amplify GraphQL API

** Provide additional details e.g. code snippets **

This is similar to #1499, except I am trying to use an object type that contains an "id" field inside of a @model type

This is my schema.graphql

```graphql
type Post @model {
  id: ID!
  tweet: Tweet
}

type Tweet {
  id: String!
  text: String!
}
```

`tweet` would be a nested object specific to `Post`. However, when I upload the schema via `amplify push`, it removes the `id` field from the generated `TweetInput`.

The generated schema:

```graphql
input CreatePostInput {
    id: ID
    tweet: TweetInput
}

type Tweet {
    id: String!
    text: String!
}

input TweetInput {
    text: String!
}
```

Is there a rule against using `id` for object types? I can't seem to find any documentation about it.

RossWilliams:
@mattjennings There is a line in the code that filters id properties for non-model input types [here](https://github.com/aws-amplify/amplify-cli/blob/8bbb701317f6d6df78afed250e1e40e3c54c3b86/packages/graphql-dynamodb-transformer/src/definitions.ts#L63).

There are no references I can find for why properties named 'id' are excluded. I'll be curious to hear the explanation.

kaustavghosh06:
@mattjennings I've assigned this as a bug. "id" shouldn't be removed for types which are not annotated with @model.

SwaySway:
Closing as the fix is now in the latest release.

PR #2530, opened 2019-10-08
fix(graphql-dynamodb-transformer): allow id for non model objects

SwaySway:
*Issue #, if available:*
ref #1984

*Description of changes:*
- allow id as a type for non model objects
- lint fixes 


SwaySway:
> Please redo the changes without prettier formatting, since it is impossible to review based on diff.

Redid changes and added a unit test

PR #2538, opened 2019-10-09
fix: #2389

### LINKED ISSUE #2389: Missing model when @model(queries: null) but with @key queryField defined
**Describe the bug**
When I set `queries: null` on `@model` but I define a `@key` with a `queryField`, the `ModelContentCategoryConnection` is not compiled.

```
type ContentCategory @model(queries: null, mutations: { create: "addContentToCategory", delete: "deleteContentFromCategory"})
@key(name: "ContentByCategory", fields: ["category", "type", "language", "datetime"], queryField: "listContentByCategory")
{
	id: ID!
	category: Int!
	datetime: String!
  	type: ContentType!
  	language: Language!
	content: Content @connection
}
```

This is compiled:
```
type Query {
  listContentByCategory(category: Int, typeLanguageDatetime: ModelContentCategoryContentByCategoryCompositeKeyConditionInput, sortDirection: ModelSortDirection, filter: ModelContentCategoryFilterInput, limit: Int, nextToken: String): ModelContentCategoryConnection
}
```

However, the `ModelContentCategoryConnection` input is missing in the compiled schema.

**Workaround**
```
type ContentCategory @model(queries: { list: "listContentCategories"}, mutations: { create: "addContentToCategory", delete: "deleteContentFromCategory"})
@key(name: "ContentByCategory", fields: ["category", "type", "language", "datetime"], queryField: "listContentByCategory")
```

Adding a query (or removing the null of course) fixes the generation of the model.

**Expected behavior**
A model to be generated if the only query added is via `@key`.

attilah:
*Issue #, if available:*

Fixes #2389


SwaySway:
Can we get a unit test that verifies the fix in the schema?

Issue #2092, opened 2019-08-20
Best way access table name from custom resolvers


iparrabb:
>**Note**: If your question is regarding the AWS Amplify Console service, please log it in the 
[official AWS Amplify Console forum](https://forums.aws.amazon.com/forum.jspa?forumID=314&start=0)

** Which Category is your question related to? **
AppSync DynamoDB

** What AWS Services are you utilizing? **
AppSync DynamoDB

** Provide additional details e.g. code snippets **
I have a project with multiple environments and need access to DynamoDB tables from custom resolvers with AppSyncApiId and env variables.

CustomResources.json
```js
"AddCustomItemResolver": {
	"Type": "AWS::AppSync::Resolver",
	"Properties": {
		"ApiId": {
			"Ref": "AppSyncApiId"
		},
		"DataSourceName": "ItemTable",
		"TypeName": "Mutation",
		"FieldName": "addCustomItem",
		"RequestMappingTemplateS3Location": {
			"Fn::Sub": [
				"s3://${S3DeploymentBucket}/${S3DeploymentRootKey}/resolvers/Mutation.addItems.req.vtl",
				{
					"S3DeploymentBucket": {
						"Ref": "S3DeploymentBucket"
					},
					"S3DeploymentRootKey": {
						"Ref": "S3DeploymentRootKey"
					}
				}
			]
		},
		"ResponseMappingTemplateS3Location": {
			"Fn::Sub": [
				"s3://${S3DeploymentBucket}/${S3DeploymentRootKey}/resolvers/Mutation.addItems.res.vtl",
				{
					"S3DeploymentBucket": {
						"Ref": "S3DeploymentBucket"
					},
					"S3DeploymentRootKey": {
						"Ref": "S3DeploymentRootKey"
					}
				}
			]
		}
	}
}
```

Mutation.addItems.req.vtl
```js
#set($items = [])
#foreach($item in ${ctx.args.items})
  $util.qr($item.put("__typename", "Item"))
  $util.qr($item.put("id", $util.defaultIfNullOrBlank($item.id, $util.autoId())))
  $util.qr($items.add($util.dynamodb.toMapValues($item)))
#end
{
  "version" : "2018-05-29",
  "operation" : "BatchPutItem",
  "tables" : {
    "ItemTable-{AppSyncApiId}-{env}": $utils.toJson($items)
  }
}
```

How I can set ItemTable-{AppSyncApiId}-{env} dynamically for each environment?

Thanks!

[...]

Issue #1946, opened 2019-06-20
How Do I get amplify environment variables into custom resolvers mapping template.


damianrr:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I need to hardcode tablenames in mapping templates, this happens often with batch operations that require the tablename, for example:
```
#set($ids = [])
#foreach($id in ${ctx.args.ids})                                                                  
  #set($map = {})
  $util.qr($map.put("id", $util.dynamodb.toString($id)))              
  $util.qr($ids.add($map))
#end

{                                                                                                                 
  "version" : "2018-05-29",
  "operation" : "BatchGetItem",
  "tables" : {
    "MyTablename-GraphQLIDRandomThing-Env": {
      "keys": $util.toJson($ids),
      "consistentRead": true
    }
  }
}
```
 

**Describe the solution you'd like**

If I could have access to env name and GraphqlID I'd be able to construct the table name instead of hardcoding it.

Thanks!!

ericweissmann:
A less desirable alternative would be a $util function that returns the tablename of the associated datasource. It is less desirable because it only solves the problem for single table batch operations -- but it may be easier to implement.

ericweissmann:
OK, here's the solution for getting an environment variable into an Appsync resolver (facepalm) in an Amplify project.

Create a function in a custom cloudformation stack that sticks the desired environment variable(s) in $ctx.stash. Then you can refer to it whenever you want.  Here's my function resource declaration:

`    "getEnvironmentFunction": {
      "Type": "AWS::AppSync::FunctionConfiguration",
      "Properties": {
        "ApiId": {
          "Ref": "AppSyncApiId"
        },
        "DataSourceName": {
          "Fn::GetAtt": [
            "NoneDS",
            "Name"
          ]
        },
        "Description": "Sets $ctx.stash.environment equal to the Amplify development environment pushed",
        "FunctionVersion": "2018-05-29",
        "Name": "getEnvironment",
        "RequestMappingTemplate": "{\n          \"version\": \"2017-02-28\",\n          \"payload\": {}\n        }",
        "ResponseMappingTemplate": {
          "Fn::Join": [
            "",
            [
              "$util.qr($context.stash.put(\"environment\", \"",
              { "Ref" : "env" },
              "\"))\n$util.toJson($ctx.prev.result)"
            ]
          ]
        }
      }
    }`

Note that you need a "none" datasource (here it is called 'NoneDS').

[...]

PR #2553, opened 2019-10-10
Add variable support for user-defined resolver templates.

kevcam4891:
*Issue #, if available:*
Add variable support for user-defined resolver templates.
Pertains to #2092, #1946, maybe others.

*Description of changes:*
Certain resolver files depend on environment-specific variables, especially for
DynamoDB operations like BatchGetItem, where table references are required.
Since table names are dependent on the environment, user may now use 3 tokens
within the file that will resolve to their environment-specific values:

`{{env}}`: Environment name (Ex: develop, master)
`{{apiId}}`: GraphQLApiIdOutput (ex: shxxxvihabcj5hswrhxxxxxxxx)
`{{apiIdEnv}}`: Convenience method matching DynamoDB table name convention (ex: shxxxvihabcj5hswrhxxxxxxxx-develop)

With these tokens, you can now do something like:

ExampleObject.fieldName.req.vtl

```
#set($ids = [])
#foreach($id in ${ctx.source.userIds})
#set($map = {})
$util.qr($map.put("id", $util.dynamodb.toString($id)))
$util.qr($ids.add($map))
#end
{
"version" : "2018-05-29",
"operation" : "BatchGetItem",
"tables" : {
“ExampleTableName-{{apiIdEnv}}“: {
"keys": $util.toJson($ids),
"consistentRead": true
}
}
}
```

It would be nice to open this up to more replacement possibilities in the future, like being able
to use Cloudformation parameters, etc.

I welcome feedback if this is not an appropriate way to solve this challenge, but it seems useful to my workflow in that my resolvers adapt to the environment I'm in. Thanks!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -59,6 +59,7 @@ The Amplify CLI supports the commands shown in the following table.
 | amplify add `<category>` | Adds cloud features to your app. |
 | amplify update `<category>` | Updates existing cloud features in your app. |
 | amplify push [--no-gql-override]| Provisions cloud resources with the latest local developments. The 'no-gql-override' flag does not automatically compile your annotated GraphQL schema and will override your local AppSync resolvers and templates. |
+| amplify pull | Fetch upstream backend environment definition changes from the cloud and updates the local environment to match that definition. |
 | amplify publish | Runs `amplify push`, publishes a static assets to Amazon S3 and Amazon CloudFront (*hosting category is required). |
 | amplify status | Displays the state of local resources that haven't been pushed to the cloud (Create/Update/Delete). |
 | amplify serve | Runs `amplify push`, and then executes the project's start command to test run the client-side application. |
--- a/packages/amplify-cli/src/extensions/amplify-helpers/show-all-help.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/show-all-help.js
@@ -21,6 +21,11 @@ function showAllHelp(context) {
       description: 'Provisions cloud resources with the latest local developments.',
     },
     {
+      name: 'pull',
+      description:
+        'Fetch upstream backend environment definition changes from the cloud and updates the local environment to match that definition.',
+    },
+    {
       name: 'publish',
       description: 'Executes amplify push, and then builds and publishes client-side application for hosting.',
     },


kevcam4891:
Hi @attilah, any thoughts on this PR? I hope the intentions come across well in the example, but understand if maybe there is a better way long-term, to achieve the same effect. Let me know. Thanks!

attilah:
@kevcam4891 sorry, the team is swamped right now, I'll raise this PR this week's bugbash to discuss it.

kevcam4891:
@attilah my force push in my own repo caused this to close. I'll reopen it, but I'm curious, does this thing have legs? I'd really like to be able to use resolvers with Batch* DDB queries, and this seems to be the best way to do it. Especially when you're creating/cloning backends using the AWS console and you aren't pushing from your CLI where you can do all sorts of shell scripting to translate these values before a push. It'd need to be part of the amplify-cli that get downloaded by the instance that builds when you push to github and the console builds it out.

tfendt:
@kevcam4891 Did this ever get merged? I too need a solution like this.

kevcam4891:
@tfendt It never ended up getting merged, unfortunately. I never ended up getting feedback on this and since Amplify uses Amplify CLI to deploy, it was a blocker for me until it got merged. However, do not dispair!

A better solution in the end (for me at least) is to create a bash/node script that runs just prior to `amplify push`. I've got this script doing two things: 

- Converting YAML cloudformation files in my `amplify/backend/[api-name]/stacks-yaml` directory to `amplify/backend/[api-name]/stacks` using `cfn-flip` (https://github.com/awslabs/aws-cfn-template-flip). This has made making CFN files sooooo much easier, since amplify-cli does not yet support YAML files. Big game changer.

- Pertinent to this ticket, I've got a script that swaps all instances of {{apiIdEnv}} to its proper value, using the values in the `amplify/backend/amplify-meta.json` file.

The beauty with this setup, is this future proofs me against any sort of extra customizations I want to effect on my cloudformation templates before they get run. This is likely far better than trying to  do string replacements during the run itself.

Best practice is to create a bash alias that runs your pre-push script then follows it up with the push, so you just have to run:
```
apush="./prePushScript.sh && amplify push -y"
```

While you're at it, if you want to get extra fancy, make it play a system sound (mac example below) when the process finishes. For me this is great, because it takes 5-10 minutes for CFN to finish, and I want to start testing things right away -- however, I usually get distracted reading reddit or something while its running and 30 minutes later I remember to start working again. The sound should jolt you back to work :)

```
apush="./prePushScript.sh && amplify push -y && afplay /System/Library/Sounds/Ping.aiff"
```

You might need to tweak the && to || for sound to play regardless of whether the push succeeds or fails.

My YAML conversion script (some data may be redacted):
```
for i in `ls ./amplify/backend/api/[name]/stacks-yaml`;
    do
        cfn-flip -j ./amplify/backend/api/[name]/stacks-yaml/$i ./amplify/backend/api/[name]/stacks/${i%.*}.json;
    done
```

My resolver replacement script:
```
let fs = require("fs");
let amplifyMetaFile = "./amplify/backend/amplify-meta.json";
let localInfoFile = "./amplify/.config/local-env-info.json";
let resolverRawPath = "./amplify/backend/api/[name]/resolvers-raw";
let resolverPath = "./amplify/backend/api/[name]/resolvers";
let gqlID, env;

if (fs.existsSync(amplifyMetaFile) && fs.existsSync(localInfoFile)) {
  let fileJson = JSON.parse(fs.readFileSync(amplifyMetaFile));
  let localJson = JSON.parse(fs.readFileSync(localInfoFile));
  gqlID = fileJson.api.[name].output.GraphQLAPIIdOutput;
  env = localJson.envName;
}

let files = fs.readdirSync(resolverRawPath);
if (files.length > 0) {
  for (let file of files) {
    let read = fs.readFileSync(`${resolverRawPath}/${file}`).toString();

    // replace only if the variables exists and are available for use.
    if (gqlID && env) {
      read = read.replace("{{apiIdEnv}}", `${gqlID}-${env}`);
    }

    fs.writeFileSync(`${resolverPath}/${file}`, read);
  }
}
```

*Caveat:* The variables will not get replaced if you're creating a new backend from scratch. This is because you won't know the AppSync API ID or other details until it actually gets created. So when you first build a backend, you'll need to run this twice.

Hope this helps you and anyone needing some guidance on how to run batch queries in DynamoDB, but hopefully this offered some other workflow insights that are helpful as well.

Cheers.

karrettgelley:
@kevcam4891 This is a hacky solution, but I love it. Work with what you got. Any word on whether or not this is still the accepted workaround?

Issue #2600, opened 2019-10-17
how can I include the "total" count in SearchableConnection?


monaye:
if I specify @searchable directive, Amplify generates SearchableConnection type but it doesn't have a total int defined but response resolver has it. 
How can I include a total in the schema? or any workaround??

houmark:
There seems to be no workaround at this time. I'm also hoping this will be wired in all the way soon as it provides useful information for pagination etc. when using ES for listing/filtering.

UnleashedMind:
There's a PR out for this, #2602.

SwaySway:
PR #2602 has been merged marking as pending-release

houmark:
@SwaySway I commented on the PR right now as I believe your solution is not correct as it will just count the amount of results returning while the amount we are looking for is the *total* amount for the query ignoring `limit`. Please see my line comment on the PR.

kaustavghosh06:
We're tracking this as an enhancement as a part of #1865

PR #2602, opened 2019-10-18
feat(graphql-elasticsearch-transformer): add total in es response

SwaySway:
add total field in SearchableXObjectConnection and add auth logic to protect that response

#2600

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-auth-transformer/src/resources.ts
+++ b/packages/graphql-auth-transformer/src/resources.ts
@@ -1037,10 +1037,12 @@ identityClaim: "${rule.identityField || rule.identityClaim || DEFAULT_IDENTITY_F
         ref('es_response'),
         obj({
           items: ref('es_items'),
-          total: ref('ctx.result.hits.total'),
         })
       ),
-      iff(raw('$es_items.size() > 0'), qref('$es_response.put("nextToken", $nextToken)')),
+      iff(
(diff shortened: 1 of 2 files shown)


monaye:
@SwaySway 
I don't think we want to have total count of item array, but the total number that is coming from Elasticsearch. 

```
         compoundExpression([qref('$es_response.put("nextToken", $nextToken)'), qref('$es_response.put("total", $es_items.size())')])
```

I believe the original code should stay instead
```
 total: ref('ctx.result.hits.total'),
```

SwaySway:
> @SwaySway
> I don't think we want to have total count of item array, but the total number that is coming from Elasticsearch.
> 
> ```
> compoundExpression([qref('$es_response.put("nextToken", $nextToken)'), qref('$es_response.put("total", $es_items.size())')])
> ```
> 
> I believe the original code should stay instead
> 
> ```
>  total: ref('ctx.result.hits.total'),
> ```

@monaye The reason this might want to stay as the size of the list being returned is because the list is being filtered by the `@auth`.

If there are more hits than what was returned in the list than this would leak information. This only happens if you use `@auth` with `@searchable`

monaye:
@SwaySway 
My bad 😅  Didn't know @seachable support the auth already and assumed the changes are for regular search. Thank you for the clarification 😄

Issue #2591, opened 2019-10-16
@auth for static group authorization on @function field in Query without @model transformer


jamesw6811:
>**Note**: If your question is regarding the AWS Amplify Console service, please log it in the 
[official AWS Amplify Console forum](https://forums.aws.amazon.com/forum.jspa?forumID=314&start=0)

** Which Category is your question related to? **
GraphQL Transformer

** What AWS Services are you utilizing? **
Amplify, Cognito, Lambda

** Provide additional details e.g. code snippets **
I was previously using static group authorization on a custom query for lambda function access control, as shown below.

```gql
type Query {
  calculate(submission: CalculatorSubmission): CalculatorResponse @function(name: "calculate")
  @auth(rules: [{allow: groups, groups: ["CalculatorAccess"]}]) 
}
```

Recently, this was broken by the following code change to the transformer, and I receive the thrown InvalidDirectiveError
```js
    const modelDirective = parent.directives.find(dir => dir.name.value === 'model');
    if (!modelDirective) {
      throw new InvalidDirectiveError('Types annotated with @auth must also be annotated with @model.');
    }
```

As shown in the following warning message that is still in the transformer, authorization logic should work with static group authorization, which initially informed my design. However, there is no way for this code to run now, right?

```js
    if (
      parent.name.value === ctx.getQueryTypeName() ||
      parent.name.value === ctx.getMutationTypeName() ||
      parent.name.value === ctx.getSubscriptionTypeName()
    ) {
      console.warn(
        `Be careful when using @auth directives on a field in a root type. @auth directives on field definitions use the source \
object to perform authorization logic and the source will be an empty object for fields on root types. \
Static group authorization should perform as expected.`
      );
    }
```

Is there any workaround to secure lambda functions with static group authorization using the @auth transformer? I don't need an RDS object.

jamesw6811:
thanks!

PR #2607, opened 2019-10-18
fix(graphql-auth-transformer): remove enforce model check for field

SwaySway:
*Issue #, if available:*
#2591 
*Description of changes:*
removing unnecessary enforce check for @model in when adding @auth on a field


attilah:
@SwaySway isn't this code will break when you've auth without model directive and you get ```undefined``` for ```find``` function? https://github.com/aws-amplify/amplify-cli/blob/18f04d1fc0a1e3c630160d1d21f20f7c61e3c108/packages/graphql-transformer-core/src/util/getDirectiveArguments.ts#L6

What do you think about instead of removing the ```@model``` validation, make it more specific and check if it is required or not for the given ```@auth``` configuration, based on the rules. If we don't enforce it, can't we run into situations, where we open up the model because no ```@model``` was supplied, so not a meaningful ```ModelDirectiveConfiguration``` was constructed?

Issue #2579, opened 2019-10-14
amplify env remove does not remove all resources


rfpedrosa:
**Describe the bug**
`amplify env remove` does not remove all resources previously created with amplify cli. Resources that are not removed are:
- Analytics category / Pinpoint (`amplify add analytics`);
- S3 buckets used to deploy;

In the long term, AWS account gets messy with unused resources and it makes the creation of new environments with the same name to fail.

**To Reproduce**
1. amplify env add (let say that new environment is called 'newenv')
2. amplify push
2. amplify add analytics
3. amplify env remove newenv

**Expected behavior**
AWS Pinpoint as well S3 buckets used to deploy are removed.

**Additional context**
NA

UnleashedMind:
@rfpedrosa 
This is a known behavior. 
Most resource names are marked with time stamps to avoid name clashes. 
I marked this as a enhancement request.

jkeys-ecg-nmsu:
@UnleashedMind is there a reason that the DeploymentBucket resource defaults to Retain?

UnleashedMind:
Because it might contain useful data for our customers.

jkeys-ecg-nmsu:
@UnleashedMind ah I see. So then why doesn't the API stack default to retain?

ammarkarachi:
Closing this issue the PR has been merged

PR #2615, opened 2019-10-21
feat: Delete all

ammarkarachi:
*Issue #, if available:*
#2579
*Description of changes:*
remove env --force &&
delete --all 

deletes all resources including s3


packages/amplify-cli/src/extensions/amplify-helpers/remove-env-from-cloud.js
```diff
@@ -28,6 +29,10 @@ async function removeEnvFromCloud(context, envName) {
     const notificationsModule = require(categoryPlugins.notifications);
     await notificationsModule.deletePinpointAppForEnv(context, envName);
   }
+  const { frontend } = context.amplify.getProjectConfig();
```
kaustavghosh06: I wouldn't remove codegen files on `env remove` -> Would remove it only when running `delete`

packages/amplify-e2e-tests/__tests__/delete.test.ts
```diff
@@ -1,27 +1,69 @@
...
+  afterEach(() => {
     deleteProjectDir(projRoot);
   });
 
-  it('should delete resources', async () => {
-    const meta = getProjectMeta(projRoot).providers.awscloudformation;
-    expect(meta.Region).toBeDefined();
-    const { AuthRoleName, UnauthRoleName } = meta;
+  it('should delete resources javascript', async () => {
+    await initProjectWithProfile(projRoot, {});
```
kaustavghosh06: initJSProjectWithProfile?

packages/amplify-frontend-android/lib/frontend-config-creator.js
```diff
@@ -1,6 +1,48 @@
...
+
+const fileNames = ['queries', 'mutations', 'subscriptions'];
+
+function deleteAmplifyConfig(context) {
+  const { amplify } = context;
+  const projectPath = context.exeInfo ? context.exeInfo.localEnvInfo.projectPath : amplify.getEnvInfo().projectPath;
+  const projectConfig = context.exeInfo ? context.exeInfo.projectConfig[constants.Label] : amplify.getProjectConfig()[constants.Label];
+  const frontendConfig = projectConfig.config;
+  const srcDirPath = path.join(projectPath, frontendConfig.ResDir, 'raw');
+  const targetFilePath = path.join(srcDirPath, constants.amplifyConfigFilename);
```
kaustavghosh06: I wonder if we could move this to a helper function since its a copy past from the function below as is

packages/amplify-frontend-android/lib/frontend-config-creator.js
```diff
@@ -1,6 +1,48 @@
...
+  if (fs.existsSync(targetFilePath)) {
+    fs.removeSync(targetFilePath);
+  }
+
+  const awsConfigFile = path.join(srcDirPath, constants.awsConfigFilename);
```
kaustavghosh06: targetFilePath is the same as awsConfigFile..why do we need to this twice?

packages/amplify-frontend-android/lib/frontend-config-creator.js
```diff
@@ -1,6 +1,48 @@
...
+
+  if (!fs.existsSync(path.join(projectPath, '.graphqlconfig.yml'))) return;
+
+  const gqlConfig = graphQLConfig.getGraphQLConfig(projectPath);
+  if (gqlConfig && gqlConfig.config) {
+    const { projects } = gqlConfig.config;
+    Object.keys(projects).forEach(project => {
+      const { codeGenTarget, docsFilePath } = projects[project].extensions.amplify;
+      fileNames.forEach(filename => {
+        const file = path.join(projectPath, docsFilePath, `${filename}.${FILE_EXTENSION_MAP[codeGenTarget] || 'graphql'}`);
```
kaustavghosh06: @yuth Can you please verify this?

kaustavghosh06: const file = path.join(projectPath, docsFilePath, `${filename}.${FILE_EXTENSION_MAP[codeGenTarget] || 'graphql'}`); 

Not sure if this logic is correct..can't we have both the graphql directory and a swift or TS file?

packages/amplify-provider-awscloudformation/lib/constants.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   ProviderName: 'awscloudformation',
   Label: 'awscloudformation',
   CacheFileName: 'cache.json',
+  ZipFileName: '#current-cloud-backend.zip',
```
kaustavghosh06: Can we name this differently? ZipFileName is too generic.

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -383,6 +385,39 @@ class CloudFormation {
           reject(err);
         }
       });
+      if (deleteS3) {
+        new S3(this.context, {}).then(s3 => {
+          const amplifyDir = this.context.amplify.pathManager.getAmplifyDirPath();
+          const tempDir = `${amplifyDir}-${envName}/.temp`;
```
kaustavghosh06: I think this tempDir should be created inside the amplify directory. Also please use path.join() instead of using `/` explicitly, as this would fail in windows.

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -383,6 +385,39 @@ class CloudFormation {
...
+          downloadZip(s3, tempDir, ZipFileName, envName).then((file, err) => {
```
kaustavghosh06: `file` out here is the zip file?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.js
```diff
@@ -71,6 +68,50 @@ class S3 {
     });
   }
 
+  deleteObjects(bucketName) {
```
kaustavghosh06: this could be renamed to deleteAllObjects based on it's functionality

packages/amplify-frontend-android/lib/frontend-config-creator.js
ammarkarachi: Where do you think the helper function should go?

packages/amplify-frontend-android/lib/frontend-config-creator.js
ammarkarachi: The codeGenTarget doesn't have any files so I default it to 'graphql'

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
ammarkarachi: Yeah it is, rename it?

PR #2640, opened 2019-10-29
fix: build break, chore: typescript, lerna update

attilah:
*Description of changes:*

Fix build break caused by a recent merge.

### DIFF
--- a/packages/amplify-appsync-simulator/package.json
+++ b/packages/amplify-appsync-simulator/package.json
@@ -65,7 +65,7 @@
     "jest": "^24.8.0",
     "ts-jest": "^24.0.2",
     "tslint": "^5.18.0",
-    "typescript": "^3.5.3"
+    "typescript": "^3.6.4"
   },
   "jest": {
     "transform": {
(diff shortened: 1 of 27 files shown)


packages/graphql-transformers-e2e-tests/src/__tests__/KeyTransformerLocal.e2e.test.ts
```diff
@@ -346,7 +346,7 @@ test('Test that connection type is generated for custom query when queries is se
     `;
 
   const transformer = new GraphQLTransform({
-    transformers: [new ModelTransformer(), new KeyTransformer()],
+    transformers: [new DynamoDBModelTransformer(), new KeyTransformer()],
```
kaustavghosh06: Did we change the name of the Transformer? There are many folks out here who use the transformer lib directly..so would this be a breaking change for them?

packages/amplify-appsync-simulator/package.json
```diff
@@ -65,7 +65,7 @@
     "jest": "^24.8.0",
     "ts-jest": "^24.0.2",
     "tslint": "^5.18.0",
-    "typescript": "^3.5.3"
+    "typescript": "^3.6.4"
```
kaustavghosh06: Any specific reason to upgrade the version explicitly today?

PR #2684, opened 2019-11-05
fix(amplify-util-mock): fixed the content check for gitignore

ammarkarachi:
*Issue #, if available:*
2624

*Description of changes:*
Check for amplify/mock-data in windows was being generated with '\' as the directory separator.
GitIgnore only uses '/' more info at https://git-scm.com/docs/gitignore#_pattern_format


packages/amplify-util-mock/src/utils/git-ignore.ts
```diff
@@ -1,16 +1,21 @@
...
-      gitIgnoreContent += '\n' + mockDataDirectory;
-      fs.writeFileSync(gitIgnoreFilePath, gitIgnoreContent);
+    const gitIgnoreContent = fs
+      .readFileSync(gitIgnoreFilePath)
+      .toString()
+      .split(os.EOL);
+    const gitIgnoreMap = gitIgnoreContent.reduce((map, val) => {
+      map.set(val, true);
+      return map;
+    }, new Map<string, boolean>());
```
yuth: Why do you need a Map? Can't you use a Set since you don't really care about the values

packages/amplify-util-mock/src/utils/git-ignore.ts
```diff
@@ -1,16 +1,21 @@
...
+    const amplifyMockData = 'amplify/mock-data';
```
yuth: Does it work as expected on windows?

ammarkarachi: I don't think windows/linux is relevant, git does require usage of forward slashes the only time  backward slashes are to be used is to escape characters.

ammarkarachi:
@yuth Can we revisit this, there is also another [pull request](https://github.com/aws-amplify/amplify-cli/pull/2964) with the same solution

ammarkarachi:
Closing this pull request already merged users much better pull request

Issue #2741, opened 2019-11-14
apiKeyConfig ignored when API Key is additional authentication provider


chadkouse:
**Describe the bug**
Setting `description` or `apiKeyExpirationDays` seems to have no effect and a default config of an empty description and expiration days of 7 is used.  This may only be the case when API_KEY is listed in `additionalAuthenticationProviders`

**Amplify CLI Version**
3.16.0

**To Reproduce**
configure your API to use API_KEY as an additional auth provider - give it a name and an expiration days != 7 to see it in action

**Expected behavior**
The API key is created in AppSync with the specified name and set to expire the specified days away.

**Desktop (please complete the following information):**
Arch Linux
node v11.10.1

PR #2744, opened 2019-11-14
fix: ignored ApiKeyConfig for additional providers (#2741)

chadkouse:
* fix: ignored ApiKeyConfig for additional providers

* fix bug with use of Array.concat()


chadkouse:
The circle ci error seems to be a timeout or something, should just be rerun.

attilah:
@chadkouse good catch! LGTM.

PR #2756, opened 2019-11-15
Codegen - Models support

yuth:
*Issue #, if available:*

*Description of changes:*


packages/amplify-codegen-model-plugin/src/languages/swift-declaration-block.ts
```diff
@@ -0,0 +1,302 @@
...
+
+    return this;
+  }
+
+  final(isFinal: boolean = true): SwiftDeclarationBlock {
+    this._flags.final = isFinal;
+    return this;
+  }
+
+  withComment(comment: string | StringValueNode | null): SwiftDeclarationBlock {
```
attilah: Generic question: shouldn't we guard against immutability while building the object? What I mean is check if ```withComment``` or ```withName``` was previously invoked.

packages/amplify-codegen-model-plugin/src/languages/swift-declaration-block.ts
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    this._enumValues[name] = value || name;
+    return this;
+  }
+
+  addClassMethod(
+    name: string,
+    returnType: string | null,
+    impl: string,
```
attilah: Is there a reason why it is abbreviated?

packages/amplify-codegen-model-plugin/src/languages/typescript-declaration-block.ts
```diff
@@ -0,0 +1,267 @@
+import { indent, indentMultiline } from '@graphql-codegen/visitor-plugin-common';
+import { access } from 'fs';
```
attilah: We standardized on ```fs-extra``` and not using ```fs``` anywhere

packages/amplify-codegen-model-plugin/src/languages/typescript-declaration-block.ts
```diff
@@ -0,0 +1,267 @@
...
+import { StringValueNode } from 'graphql';
+
+function isStringValueNode(node: any): node is StringValueNode {
```
attilah: Should we move the common functions into a shared file?

packages/amplify-codegen-model-plugin/src/languages/typescript-declaration-block.ts
```diff
@@ -0,0 +1,267 @@
...
+
+      if (this._flags.isDeclaration) {
+        return `${methodHeaderStr};`;
+      }
+      return [`${methodHeaderStr}`, '{', method.implmentation ? indentMultiline(method.implmentation) : '', '}'].join('\n');
+    });
+    return methods.join('\n');
+  }
+
+  protected generateInterface(): string {
```
attilah: Not implemented "yet"? or we don't support interfaces? Based on the answer we should update the text here.

packages/amplify-codegen-model-plugin/src/plugin.ts
```diff
@@ -0,0 +1,51 @@
+import { PluginFunction, Types } from '@graphql-codegen/plugin-helpers';
+import { GraphQLSchema, parse, visit } from 'graphql';
+import { printSchemaWithDirectives } from 'graphql-toolkit';
+import { AppSyncSwiftVisitor } from './visitors/appsync-swift-visitor';
+import { RawAppSyncLocalConfig } from './visitors/appsync-visitor';
+import { AppSyncJSONVisitor } from './visitors/appsync-json-metadata-visitor';
+import { AppSyncJavaVisitor } from './visitors/appsync-java-visitor';
+import { AppSyncLocalTypeScriptVisitor } from './visitors/appsync-typescript-visitor';
```
attilah: Are we keeping "AppSyncLocal" naming around?

packages/amplify-codegen-model-plugin/src/visitors/appsync-visitor.ts
```diff
@@ -0,0 +1,379 @@
...
+    typeArr.sort(sortFields);
+    return crypto
+      .createHash('MD5')
+      .update(JSON.stringify(typeArr))
+      .digest()
+      .toString('hex');
+  }
+
+  protected ensureIdField(model: CodeGenModel) {
+    const idField = model.fields.find(field => field.name === 'id');
```
attilah: This will break when ```@key``` fixes will be implemented as ```@key``` can make any field as ```id``` field.

packages/amplify-provider-awscloudformation/lib/utility-functions.js
```diff
@@ -6,22 +6,34 @@ const AppSync = require('../src/aws-utils/aws-appsync');
...
+        handleMigration: resources.length ? opts => updateStackForAPIMigration(context, category, resources[0], opts) : undefined,
+      };
+    }
+
     return transformGraphQLSchema(context, optionsWithUpdateHandler);
   },
+  getTransformerDirectives: async (context, options) => {
+    const { resourceDir } = options;
+    if (!resourceDir) {
+      throw new Error('missing resource directory');
```
attilah: "Missing"

packages/amplify-codegen-model-plugin/yarn.lock
```diff
@@ -0,0 +1,2636 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
attilah: Why do we've a yarn.lock in a package folder, I think we should remove it.

packages/amplify-cli/src/extensions/amplify-helpers/get-cloud-init-status.js
```diff
@@ -0,0 +1,25 @@
+const fs = require('fs-extra');
+const pathManager = require('./path-manager');
+
+const CLOUD_INITIALIZED = 'CLOUD_INITIALIZED';
+const CLOUD_NOT_INITIALIZED = 'CLOUD_NOT_INITIALIZED';
+const NON_AMPLIFY_PROJECT = 'NON_AMPLIFY_PROJECT';
```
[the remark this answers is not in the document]
yuth: Yes

packages/amplify-codegen-model-plugin/src/languages/swift-declaration-block.ts
```diff
@@ -0,0 +1,302 @@
+import { indent, indentMultiline } from '@graphql-codegen/visitor-plugin-common';
+import { NameNode, StringValueNode } from 'graphql';
+
+function isStringValueNode(node: any): node is StringValueNode {
+  return node && typeof node === 'object' && node.kind === 'StringValue';
+}
+
+function transformComment(comment: string | StringValueNode, indentLevel = 0): string {
+  if (!comment || comment === '') {
```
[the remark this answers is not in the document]
yuth: I doubt this would be possible when using the GraphQL JS library. For good measure I moved the empty string check after the `stringValueNode` check

packages/amplify-codegen-model-plugin/src/languages/swift-declaration-block.ts
yuth: I don't think its worth it much check this.

packages/amplify-codegen-model-plugin/src/languages/swift-declaration-block.ts
yuth: not really. Wanted to type less

packages/amplify-codegen-model-plugin/src/languages/typescript-declaration-block.ts
yuth: This is the side effect of vscode being too helpful. It auto imported this, and I had wrong thing pushed into an array in https://github.com/aws-amplify/amplify-cli/pull/2756/files#diff-d8a19b27e070f63d0ceec21e7b643dceR232

packages/amplify-codegen-model-plugin/src/plugin.ts
yuth: good catch

packages/amplify-codegen-model-plugin/src/visitors/appsync-visitor.ts
yuth: This was an explicit decision. Data Store wants to maintain the hash key and won't let customers change the type or name or value of it. This is needed for conflict resolution

packages/amplify-cli/src/extensions/amplify-helpers/resource-status.js
```diff
@@ -219,11 +220,18 @@ async function asyncForEach(array, callback) {
 }
 
 async function getResourceStatus(category, resourceName, providerName, filteredResources) {
-  const amplifyMetaFilePath = pathManager.getAmplifyMetaFilePath();
-  const amplifyMeta = readJsonFile(amplifyMetaFilePath);
-
-  const currentamplifyMetaFilePath = pathManager.getCurentAmplifyMetaFilePath();
-  const currentamplifyMeta = readJsonFile(currentamplifyMetaFilePath);
+  const amplifyProjectInitStatus = getCloudInitStatus();
+  let amplifyMeta;
```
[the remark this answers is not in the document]
yuth: I have this for the sake of completeness

PR #2763, opened 2019-11-16
Walkthrough update

yuth:
*Issue #, if available:*

*Description of changes:*


packages/amplify-category-api/provider-utils/awscloudformation/sync-conflict-handler-assets/sync-conflict-handler-package.json.ejs
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "<%= props.functionName %>",
+  "version": "1.0.0",
+  "description": "Lambda function generated by Amplify for conflict handling",
+  "main": "index.js",
+  "license": "Apache-2.0"
```
attilah: Do we want to enforce a license on generated code on the customer's behalf? Can it be empty instead?

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -170,6 +170,10 @@ export class DynamoDBModelTransformer extends Transformer {
     }
   };
 
+  // makeInputValueDefinition('_deleted', makeNamedType('Boolean')),
```
attilah: Leftover?

packages/amplify-category-api/provider-utils/awscloudformation/sync-conflict-handler-assets/sync-conflict-handler-index.js.ejs
```diff
@@ -0,0 +1,35 @@
+exports.handler = async (event, context, callback) => {
```
attilah: This code is generic and only applies to a default schema. After the walkthrough customers can end up with a non-working code, because they've no ```Post``` type or other things. I think we should add a comment that it is sample code and they must edit it to fit their own schema, at a later time we could look into if we can generate this from their schema.

Also I think that the walkthrough should write a message/warning to the console to notify the customers to make sure they take a look at this file.

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -566,7 +570,7 @@ export class DynamoDBModelTransformer extends Transformer {
     if (resolverConfig && resolverConfig.project) {
       syncConfig = resolverConfig.project;
     }
-    if (resolverConfig && resolverConfig.models[typeName]) {
+    if (resolverConfig && resolverConfig.models && resolverConfig.models[typeName]) {
```
attilah: Shouldn't we have to check if ```typeName``` has a value?

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -250,10 +284,62 @@ async function serviceWalkthrough(context, defaultValuesFilename, serviceMetadat
...
+  localTransformerConfig.ResolverConfig = syncConfig;
+  await writeTransformerConfiguration(resourceDir, localTransformerConfig);
+}
+
+async function createSyncFunction(context) {
+  const targetDir = context.amplify.pathManager.getBackendDirPath();
+  const pluginDir = __dirname;
+  const [shortId] = uuid().split('-');
+
+  const functionName = `synConflictHandler${shortId}`;
```
attilah: missing 'c' from function name after 'syn'

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -372,8 +447,165 @@ async function askSecurityQuestions(context, parameters) {
   const advancedSettingsAnswer = await inquirer.prompt([advancedSettingsQuestion]);
 
   if (advancedSettingsAnswer.advancedSettings) {
+    authConfig = await askAdditionalAuthQuestions(context, parameters, authConfig, defaultAuthType);
+    if(process.env.SYNC === "true") {
```
attilah: I think a more unique amplify specific name for the env variable would be better.

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -372,8 +447,165 @@ async function askSecurityQuestions(context, parameters) {
...
+      resolverConfig = await askResolverConflictQuestion(context, parameters, modelTypes);
+    }
+  }
+
+  return { authConfig, resolverConfig };
+}
+
+async function askResolverConflictQuestion(context, parameters, modelTypes) {
```
attilah: Why are we collecting ```modelTypes``` earlier in the process? Are we using them even if the answer is 'N' to the question?

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -372,8 +447,165 @@ async function askSecurityQuestions(context, parameters) {
...
+  let resolverConfig = {};
+
+  if (await context.prompt.confirm('Configure conflict detection?')) {
+    const askConflictResolutionStrategy = async msg => {
+      let conflictResolutionStrategy;
+
+      do {
+        if (conflictResolutionStrategy === 'Learn More') {
+          context.print.info('');
+          context.print.info('DataStore help text');
```
attilah: Add //TODO comment to replace this

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -372,8 +447,165 @@ async function askSecurityQuestions(context, parameters) {
...
+      }
+
+      return syncConfig;
+    };
+
+    resolverConfig.project = await askConflictResolutionStrategy('Select the default resolution strategy');
+
+    // Ask for per-model resolver override setting
+
+    if (modelTypes && modelTypes.length > 0) {
```
attilah: What about having a "Select All" and "Select None" option as well (if possible)?

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -372,8 +447,165 @@ async function askSecurityQuestions(context, parameters) {
...
+    type: 'list',
+    name: 'syncLambdaAnswer',
+    message: 'Select from the options below',
+    choices: [
+      {
+        name: 'Create a new Lambda Function',
+        value: 'NEW',
+      },
+      {
+        name: 'Pre-existing Lambda Function',
```
attilah: Wording: wouldn't "Existing" be better?

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -372,8 +447,165 @@ async function askSecurityQuestions(context, parameters) {
...
+      type: 'input',
+      name: 'lambdaFunctionName',
+      message: 'Enter lambda function name',
+      validate: val => !!val,
+    };
+    ({ lambdaFunctionName } = await inquirer.prompt([syncLambdaNameQuestion]));
+  }
+
+  if (!lambdaFunctionName) {
+    throw new Error('Error in configuring Lambda function');
```
attilah: The error message should be more specific that it is about the name not defined.

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
yuth: The enquirer plugin supports that out of the box

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
yuth: I don't think we need this check at all as validation function won't let customer go to next question if they don't enter anything

PR #2774, opened 2019-11-18
Android setup

nikhname:
*Issue #, if available:*

*Description of changes:*
Add amplify plugin to tools/android. Creates config files to support this plugin


packages/amplify-app/src/index.js
```diff
@@ -329,10 +329,25 @@ async function createJSHelperFiles() {
   });
 }
 
+async function createAndroidHelperFiles() {
+  const configJsonData = '{"profile":"default", "envName":"amplify"}';
+  const configJsonObj = JSON.parse(configJsonData);
+  const configJsonStr = JSON.stringify(configJsonObj);
```
yuth: why parsing first. Can't we use just an Object and stringify it

packages/amplify-app/src/index.js
```diff
@@ -329,10 +329,25 @@ async function createJSHelperFiles() {
...
+  const configDir = path.join(process.cwd(), './amplify-gradle-config.json');
```
yuth: name this config file instead

packages/amplify-app/src/index.js
```diff
@@ -329,10 +329,25 @@ async function createJSHelperFiles() {
...
+  if (await !fs.existsSync(configDir)) {
```
yuth: why await in existsSync

packages/amplify-app/src/index.js
nikhname: I'm parsing so that the object is formatted properly before we stringify it. Otherwise there will be backslashes in the json file

packages/amplify-app/src/index.js
nikhname: removed await

packages/amplify-app/src/index.js
nikhname: changed

packages/amplify-app/src/index.js
nikhname: changed to json object

PR #2802, opened 2019-11-21
arrow function support

alikadir:
added arrow function support in nodejs function template

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-category-function/provider-utils/awscloudformation/function-template-dir/index.js.ejs
+++ b/packages/amplify-category-function/provider-utils/awscloudformation/function-template-dir/index.js.ejs
@@ -1,8 +1,10 @@
 <%= props.topLevelComment %>
 
-exports.handler = function (event, context) { //eslint-disable-line
-  console.log(`value1 = ${event.key1}`);
-  console.log(`value2 = ${event.key2}`);
-  console.log(`value3 = ${event.key3}`);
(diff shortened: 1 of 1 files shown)


packages/amplify-category-function/provider-utils/awscloudformation/function-template-dir/index.js.ejs
```diff
@@ -1,8 +1,10 @@
 <%= props.topLevelComment %>
 
-exports.handler = function (event, context) { //eslint-disable-line
-  console.log(`value1 = ${event.key1}`);
-  console.log(`value2 = ${event.key2}`);
-  console.log(`value3 = ${event.key3}`);
-  context.done(null, 'Hello World'); // SUCCESS with message
```
kaustavghosh06: Won't be still need context.done even if we use arrow functions? or does returning the response replaces the need for having context.done?

onlybakam:
`context` is optional for `async` functions. i dont think it's the recommended approadh.

see: https://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html#nodejs-handler-async

> For async functions, you return a response, error, or promise to the runtime instead of using callback.

The change was introduced with the Node.js 4.3.2 Runtime
https://aws.amazon.com/blogs/compute/node-js-4-3-2-runtime-now-available-on-lambda/

PR #2806, opened 2019-11-22
fix: quickstart change and profile selection

nikhname:
*Issue #, if available:*

*Description of changes:*
Remove auth from quickstart and change api name to amplifyDatastore
Ask question for profile if default doesn't exist


packages/amplify-app/src/scripts/amplify-push.js
```diff
@@ -24,14 +25,25 @@ function getNamedProfiles() {
   return namedProfiles;
 }
 
-function checkIfProfileExists(profileToUse) {
+async function checkIfProfileExists(profileToUse) {
```
yuth: functin name should be changed. Better yet split this into 2 functions. One for checking and other for asking the user which profile they want to use if selected profile is missing in the computer

packages/amplify-app/src/scripts/amplify-push.js
```diff
@@ -24,14 +25,25 @@ function getNamedProfiles() {
...
-
-  if (!namedProfiles || !namedProfiles[profileToUse]) {
-    console.log(`Profile with name ${profileToUse} not found. Please create one`);
-    return false;
+  if (Object.keys(namedProfiles).length) {
+    if (namedProfiles[profileToUse]) {
+      return profileToUse;
+    } else if (Object.keys(namedProfiles).length === 1) {
+      return Object.keys(namedProfiles)[0];
+    }
```
yuth: Do we want to still ask the customer if they want to use the only profile available?

nikhname: They are implicitly choosing that profile if they are going through this flow and that is the only profile that exists

nikhname: Changing to ask even if is just one profile

packages/amplify-app/src/scripts/amplify-push.js
nikhname: Separating into two functions

packages/amplify-app/src/xcodeHelpers.js
```diff
@@ -5,7 +5,7 @@ const fs = require('fs-extra');
 async function getXcodeProjectDir() {
   const EXTENSION = '.xcodeproj';
   const files = fs.readdirSync(process.cwd());
-  const targetFiles = files.filter(function(file) {
+  const targetFiles = files.filter(function extenstionFilter(file) {
```
yuth: use arrow functions for these things when possible, so this does not get rebound

PR #2811, opened 2019-11-22
(fix) Use semver package

nikhname:
*Issue #, if available:*

*Description of changes:*
Use the semver package to validate node and cli versions, updating cli version if necessary

### DIFF
--- a/packages/amplify-app/src/index.js
+++ b/packages/amplify-app/src/index.js
@@ -8,8 +8,12 @@ const frameworkConfigMapping = require('./framework-config-mapping');
 const args = require('yargs').argv;
 const { addFileToXcodeProj } = require('./xcodeHelpers');
 const ini = require('ini');
+const semver = require('semver');
+const stripAnsi = require('strip-ansi');
 
 const amplifyCliPackageName = '@aws-amplify/cli@canary';
+const minCLIVersion = '>=3.17.1';
+const minNodeVersion = '>=8.12.0';
 
 function run() {
   const projpath = args.path;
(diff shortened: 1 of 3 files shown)


packages/amplify-app/src/index.js
```diff
@@ -8,8 +8,12 @@ const frameworkConfigMapping = require('./framework-config-mapping');
 const args = require('yargs').argv;
 const { addFileToXcodeProj } = require('./xcodeHelpers');
 const ini = require('ini');
+const semver = require('semver');
+const stripAnsi = require('strip-ansi');
 
 const amplifyCliPackageName = '@aws-amplify/cli@canary';
```
attilah: Please remove @canary, as we don't want to have +1 PR just to remove it.

We should do version check the NPM way, put the minimum versions into ```packages.json``` and read it from there:
```
  "engines": {
    "node": ">=8.12.0",
    "@aws-amplify/cli": ">=3.17.1"
  },
```

Also the version we should check should be 4.0.0 not 3.17.1

packages/amplify-app/src/index.js
```diff
@@ -38,11 +42,7 @@ function run() {
...
   const currentNodeVersion = process.versions.node;
-  const semver = currentNodeVersion.split('.');
-  const major = semver[0];
-  const minor = semver[1];
-
-  if (major < 8 || (major === 8 && minor < 12)) {
+  if (semver.satisfies(currentNodeVersion, minNodeVersion)) {
     console.error(
       `You are running Node ${currentNodeVersion}.\n` +
         `Amplify CLI requires Node 8.12.0 or higher. \n` +
```
attilah: String interpolate the minimum version from the version we read above instead make if part of the string.

packages/amplify-app/src/index.js
```diff
@@ -54,9 +54,10 @@ async function checkNodeVersion() {
...
   const amplifyCLIVersionCheck = spawnSync('amplify', ['-v']);
+  const amplifyCLIVersion = semver.coerce(stripAnsi(amplifyCLIVersionCheck.stdout.toString()));
 
-  if (amplifyCLIVersionCheck.stderr !== null) {
-    console.log(`${emoji.get('white_check_mark')} Found Amplify CLI v${amplifyCLIVersionCheck.stdout.toString()}`);
+  if (amplifyCLIVersionCheck.stderr !== null && semver.satisfies(amplifyCLIVersion, minCLIVersion)) {
+    console.log(`${emoji.get('white_check_mark')} Found Amplify CLI v${amplifyCLIVersion}`);
   } else {
     // Install the CLI
     console.log(`${emoji.get('worried')} Amplify CLI not found on your system.`);
```
attilah: "was not found"

PR #2838, opened 2019-11-26
try/catch around reading the metadata

SwaySway:
Added a try catch around the s3resourcecheck

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
+++ b/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
@@ -558,7 +558,10 @@ function s3ResourceAlreadyExists(context) {
     }
     return resourceName;
   } catch (error) {
-    return undefined;
+    if (error.name === 'UndeterminedEnvironmentError') {
+      return undefined;
+    }
+    throw error;
   }
 }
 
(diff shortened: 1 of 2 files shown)


packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -540,21 +540,26 @@ async function getDirectiveDefinitions(context, resourceDir) {
...
+    if (amplifyMeta[storageCategory]) {
+      const categoryResources = amplifyMeta[storageCategory];
+      Object.keys(categoryResources).forEach(resource => {
+        if (categoryResources[resource].service === s3ServiceName) {
+          resourceName = resource;
+        }
+      });
+    }
+    return resourceName;
+  } catch (error) {
```
yuth: Can you catch the specific exception. Swallowing every error is a bad idea

attilah: It must be handled thoroughly as other places in the code. Are we expecting to have it configured or not, that is the question.

SwaySway: added a check to allow a specific error

Issue #2703, opened 2019-11-07
IAM policy size exceeded


oliverandersencox:
I just created a fresh amplify project.

1. Amplify init
2. amplify add api
3. amplify push
4. amplify add api
5. amplify push / amplify push api < errors here

> CREATE_FAILED      AuthRolePolicy   AWS::IAM::Policy         Thu Nov 07 2019 20:31:40 GMT+0200 (South Africa Standard Time) Maximum policy size of 10240 bytes exceeded for role amplify-culture-app-dev-163322-authRole (Service: AmazonIdentityManagement; Status Code: 409; Error Code: LimitExceeded; Request ID: e116276c-a03f-469a-b0f5-e0ac1fc43460)  
> CREATE_FAILED      UnauthRolePolicy AWS::IAM::Policy         Thu Nov 07 2019 20:31:39 GMT+0200 (South Africa Standard Time) Maximum policy size of 10240 bytes exceeded for role amplify-culture-app-dev-163322-unauthRole (Service: AmazonIdentityManagement; Status Code: 409; Error Code: LimitExceeded; Request ID: 93255697-fe0d-4f4f-a085-69a63574d832)

I simply copied the schema from my working project and added it into the new one. I can't event see the policy amplify creates to see why it is too long.

Most models look similar to this:
```
`type City @model 
@key(name: "cityByName", fields: ["title"], queryField: "findCityByName")
  @auth(rules: [
    { allow: groups, groups: ["admin"] },
    { allow: groups, groups: ["users"], operations: [read] },
    { allow: public, provider: iam, operations: [read] }
  ])
@searchable {
  id: ID!
  title: String!
  description: String
  country: String
  continent: String
  location: Location 
  timezoneOffset: Float
  currency: String
  images: [String]
  createdAt: Float!
  updatedAt: Float!
}
`
```

any ideas as this is a blocker... again.

kaustavghosh06:
@oliverandersencox Do you just have one AppSync API in your project and you're seeing the IAM policy limit getting exceeded?

oliverandersencox:
yes, I just created the one api after a fresh init of a new amplify project

kaustavghosh06:
Okay, i'm trying to reproduce this right now. Is the schema size correct or do you have more types?

oliverandersencox:
my schema is very large, with over 30 types.

kaustavghosh06:
[...]

PR #2883, opened 2019-11-30
fix: use managedpolicies and slice them, fixes: #2703

attilah:
*Issue #, if available:*

fixed: #2703

*Description of changes:*

Use managed policies instead of policies for GraphQL API policy generation for auth and unauth roles. Also take into account the policy size and create multiple policies of needed. This means that 6144 bytes long can be 1 policy and 10 Managed Policies can be attached to a single role and that 10 can be raised to 20 by AWS Support, which raises the maximum policy size to ~120kb, which is 10 times the currently supported size.

This PR does not solve the policy size issue for API Gateway (#2084), but since sizes are adding up, perhaps customers can be unblocked by this change.

### DIFF
--- a/packages/graphql-auth-transformer/src/resources.ts
+++ b/packages/graphql-auth-transformer/src/resources.ts
@@ -1,5 +1,4 @@
 import Template from 'cloudform-types/types/template';
-import Policy from 'cloudform-types/types/iam/policy';
 import { AppSync, Fn, StringParameter, Refs, NumberParameter, IAM, Value } from 'cloudform-types';
 import { AuthRule, AuthProvider } from './AuthRule';
 import {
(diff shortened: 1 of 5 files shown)


packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -973,9 +973,34 @@ identityClaim: "${rule.identityField || rule.identityClaim || DEFAULT_IDENTITY_F
...
+    const policies = new Array<ManagedPolicy>();
     const authPiece = isAuthPolicy ? 'auth' : 'unauth';
-    const policyResources: object[] = [];
+    let policyResources: object[] = [];
+    let resourceSize = 0;
+
+    // 6144 bytes is the maximum policy payload size, but there is structural overhead, hence the 6000 bytes
+    const MAX_BUILT_SIZE_BYTES = 6000;
+    // The overhead is the amount of static policy arn contents like region, accountid, etc.
+    const RESOURCE_OVERHEAD = 90;
```
yuth: can you add the math how we came up with the size

PR #2885, opened 2019-11-30
ci(cli): add plugin scan to postinstall

UnleashedMind:
*Issue #, if available:*

*Description of changes:*
1.  add `plugin scan` to postinall
2. add plugin platform health check as the last step to `plugin scan`, in which all the official plugin packages are verified for existence and correct version.
3. add logic to trigger `plugin scan` if the cli version is different than what's in the `plugins.json` file. 


packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
```diff
@@ -0,0 +1,190 @@
+import path from 'path';
+import chalk from 'chalk';
+import { PluginInfo } from '../domain/plugin-info';
+import { PluginPlatform } from '../domain/plugin-platform';
+import { readJsonFileSync } from '../utils/readJsonFile';
+
+type PluginDescription = {
+  name: string;
+  type: string;
```
attilah: Why do we need ```type```?
Since these are our own plugins the type member does not carry any additional value in the comparison, when matching dependencies and plugins found by the plugin platform.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
```diff
@@ -0,0 +1,190 @@
...
+  packageName: string;
+  packageVersion?: string;
+};
+
+const officialPlugins: { [key: string]: PluginDescription } = {
```
attilah: Can we move it from code the the CLI's ```package.json```, since the other half of the information is coming from there as well. We could have an ```amplify``` section with ```officialPlugins``` member and this information. The maintenance cost would be lower since when we add a new package to the CLI it is obvious that ```officialPlugins``` needs an update as well.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
```diff
@@ -0,0 +1,190 @@
...
+  core: {
```
attilah: We could simplify it to be an array or ```PluginDescription```, no need for the nested objects as we don't use the key values for anything, just to iteration.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  const dependencies: { [key: string]: string } = packageJson.dependencies;
+
+  Object.keys(officialPlugins).forEach((plugin: string) => {
+    const { packageName } = officialPlugins[plugin];
+    if (dependencies[packageName]) {
+      const version = dependencies[packageName];
+      officialPlugins[plugin].packageVersion = version;
+    } else {
+      delete officialPlugins[plugin].packageVersion;
```
attilah: How can we have an official plugin which is not a direct dependency in the CLI?

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
```diff
@@ -0,0 +1,190 @@
...
+  if (missingOfficialPlugins.length > 0) {
+    console.log(chalk.yellow('The following official plugins are missing or inactive:'));
+    missingOfficialPlugins.forEach((pluginDescription: PluginDescription) => {
+      console.log('    ' + pluginDescription.name);
+    });
+  }
+
+  if (mismatchedOfficialPlugins.length > 0) {
+    console.log(chalk.yellow('The following official plugins have incorrect versions:'));
+    mismatchedOfficialPlugins.forEach((pluginDescription: PluginDescription) => {
```
attilah: If we've a version mismatch it would be a better DX if we print what version did we expected and what version was found by the plugin platform. That would help the troubleshooting.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
```diff
@@ -0,0 +1,190 @@
...
+      console.log('    ' + pluginDescription.name);
+    });
+  }
+
+  return missingOfficialPlugins.length === 0 && mismatchedOfficialPlugins.length === 0;
+}
+
+function isMatching(pluginDescription: PluginDescription, pluginInfo: PluginInfo) {
+  let result = pluginDescription.packageName === pluginInfo.packageName && pluginDescription.type === pluginInfo.manifest.type;
```
attilah: This whole function could be combined into a single expression.

packages/amplify-cli/src/plugin-helpers/scan-plugin-platform.ts
```diff
@@ -58,13 +60,21 @@ export async function scanPluginPlatform(pluginPlatform?: PluginPlatform): Promi
   pluginPlatform!.lastScanTime = new Date();
   await writePluginsJsonFile(pluginPlatform!);
 
+  await checkPlatformHealth(pluginPlatform);
```
attilah: We don't use the result of the function for anything. What happens if we fail to find core plugins, how will we exit the process to make sure not running into further errors down the execution path because of the missing plugins?

packages/amplify-cli/src/plugin-manager.ts
```diff
@@ -42,8 +47,10 @@ export async function getPluginPlatform(): Promise<PluginPlatform> {
 function isCoreMatching(pluginPlatform: PluginPlatform): boolean {
   try {
     const currentCorePluginDirPath = getCorePluginDirPath();
```
attilah: This could be combined into a single expression.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
yuth: Can we automate the generation of official plugin for every build. I don't like the prospect of maintaining one more list

packages/amplify-cli/scripts/post-install.js
```diff
@@ -1,4 +1,7 @@
+#!/usr/bin/env node --no-warnings
+require = require('esm')(module);
 const chalk = require('chalk');
+const { scan } = require('../lib/plugin-manager');
 
 console.log(chalk.bold('Ignore installation errors for optional dependencies: gyp, fs-ext'));
```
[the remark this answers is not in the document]
UnleashedMind: Maybe Yathi can send another PR for it? it not related to what this PR is for.

packages/amplify-cli/src/plugin-helpers/access-plugins-file.ts
```diff
@@ -63,7 +63,7 @@ function getPluginsJsonFileName(): string {
   let result = constants.PLUGINS_FILE_NAME;
   const amplifyExecutableName = path.basename(process.argv[1]);
 
-  if (amplifyExecutableName !== constants.Amplify) {
+  if (amplifyExecutableName !== constants.Amplify && amplifyExecutableName.includes(constants.Amplify)) {
     result = amplifyExecutableName + '-' + constants.PLUGINS_FILE_NAME;
```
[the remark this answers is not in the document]
UnleashedMind: Not sure what you meant, will chat.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: This for logic completeness, plugin type is an essential part of a plugin identity, adding it should not affect the performance that much, but makes the check more reliable and stable going forward.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: @attilah that sounds like a good idea. I will give it a try. 
@yuth I'm not sure about auto generation, a little concerned about having too many moving parts in the plugin platform.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: check

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: The platform allows that, but currently no such plugins.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: I know, but that will need to go through API review, in the original design doc, those diagnostic info will be in the plugin platform's logs, but that's not implemented yet.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: I moved it to a function for readability. I prefer coding style that's easy to read than concise and "smart". because that's easy to maintain and check for updates.

packages/amplify-cli/src/plugin-helpers/scan-plugin-platform.ts
UnleashedMind: The additional health check here is only meant to check on it, not for fixing problems. The messages printed out will help the customer and us to identify and fix the issues.

packages/amplify-cli/src/plugin-manager.ts
UnleashedMind: same reason as above, I think it's easier to read and maintain to explicitly lay out the code like this.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: move that it to the package.json in the latest update.

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: I kept the key values in the update, because it's much easier to  pinpoint plugins like `officialPlugins.core` (this is used in the codebase) and `officialPlugins.awscloudformation` etc. For future updates, we are going to log the plugin scan result, it will be much eaiser to iterate through the keys which are the just plugin names.

packages/amplify-cli/package.json
```diff
@@ -99,5 +99,94 @@
       "node"
     ],
     "collectCoverage": true
+  },
+  "amplify": {
```
kaustavghosh06: Why do we need to add this to our main package.json? Each time we add a plugin, we need to add a blog here?

packages/amplify-cli/src/plugin-helpers/access-plugins-file.ts
```diff
@@ -63,8 +63,8 @@ function getPluginsJsonFileName(): string {
   let result = constants.PLUGINS_FILE_NAME;
   const amplifyExecutableName = path.basename(process.argv[1]);
 
-  if (amplifyExecutableName !== constants.Amplify) {
-    result = amplifyExecutableName + '-' + constants.PLUGINS_FILE_NAME;
+  if (amplifyExecutableName === 'amplify-dev') {
```
kaustavghosh06: do we need to change this back? or is this intentional?

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
kaustavghosh06: did we ppush the change for this?

packages/amplify-cli/src/plugin-helpers/scan-plugin-platform.ts
kaustavghosh06: do we need a try catch here - to exit the process?

packages/amplify-cli/src/plugin-helpers/platform-health-check.ts
UnleashedMind: yes, it's in line 52 of platform-health-check.ts

packages/amplify-cli/src/plugin-helpers/scan-plugin-platform.ts
UnleashedMind: health check normally does not throw errors, it only prints out abnormal things that it found in the plugin platform

packages/amplify-cli/package.json
UnleashedMind: It was originally put inside the health check ts file, but Attila and Yathi suggested to put it here, and I do agree with them because it is easier to check for consistency. If we add an official plugin, the CLI core will reference it as a dependency, and also put its information inside the amplify field in the package.json.

packages/amplify-cli/src/plugin-helpers/access-plugins-file.ts
UnleashedMind: This is done intentionally. 
We don't want to produce too many of those plugins.json files.

kaustavghosh06: So shouldn't it be `amplify-dev || amplify`

packages/amplify-cli/package.json
kaustavghosh06: We need to document this behavior in some docs otherwise it's super hard to remember the existence of new blocks and requirements like these.

kaustavghosh06: Please add a section to the plugins doc explaining this section. Thanks!

PR #2898, opened 2019-12-02
fix: fix lint warnings

hirochachacha:


kaustavghosh06:
What benefit does this change provide?

hirochachacha:
Sorry I forgot to reply. This change just suppress lint warnings.

https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments

The following directive is useless.
```
while (true) {
// eslint-disable-line
```
If we want to suppress a lint warning from `while (true)` line, we should write
```
while (true) { // eslint-disable-line
```
or
```
// eslint-disable-next-line
while (true) {
```
Theses fixes are very trivial. You can fix it by yourself within 5 minutes.

Issue #1657, opened 2019-06-13
using createdAt as a sortKey in @key directive results mandatory id and createdAt fields in CreateItemInput and CreateItem mutation in GraphQL Transform


iselcuk:
In GraphQL Transform, transforming a schema that includes `@key` directive like: `@key(fields: ["id", "createdAt"])` forces to make key fields non-nullable and that results with mandatory `id` and `createdAt` fields in `CreateItemInput`. Since `CreateItemInput` has these mandatory fields, it is a must to pass `id` and `createdAt` to `createItem` mutation.  I want `createdAt` to be handled by automatically like mentioned here if it is possible: [Timestamps for GraphQL/AppSync models](https://github.com/aws-amplify/docs/issues/355#issue-396466935)
 
Example: 
```
type Item @model @key(fields: ["id", "createdAt"]){
    id: ID!
    createdAt: String!
}
```
Result: 
```
input CreateItemInput {
    id: ID!
    createdAt: String!
}
```
Expected Result:
```
input CreateItemInput {
    id: ID
    createdAt: String
}
```

iselcuk:
Also it is not limited to createdAt. Using @key directive with custom fields like @key(fields: ["id", "sortKey"] results with non-nullable id and sortKey fields in CreateItemInput and UpdateItemInput.

Crisp3333:
I am in favor of this. A scenario where this is critical is a real time application where the date field must be controlled by one global system/server that sets the time. If the time is set by the application code, then there could be lapses and inconsistencies reflected in time of multiple events. If device A creates a mutation before device B, there is a strong possibility device B's time record in the database is later than the record of device A (This is due to the system time of each device). This is a BIG problem if device A and B are communicating.

A solution in the mean time would be to send the records to a AWS lambda function to deal with the mutation, that is, the AWS lambda function handling the `createdAt` time would be global for all devices.

Crisp3333:
I did some research and a quick solution would be is to use an api to get a uniform time. If it is very important for the time timestamp to be consistent (which it should) you can use `truetime` library. I am using the `truetime` library so regardless of any system my application is running on, I will at least have consistent times for my database entries.

dseeker:
bump, just encountered this limitation. 
Key directive must be able to accept non-nullables to allow autoid, server createdAt and updatedAt
related to https://github.com/aws-amplify/amplify-cli/pull/356

ajhool:
[...]

PR #2899, opened 2019-12-02
fix(graphql-connection-transformer): make validation permissive

hirochachacha:
Current implementation prohibits to use nullable fields as connection fields.
However if these field values are guaranteed to be assigned by @auth directives,
nullabilities of field types don't matter.

### DIFF
--- a/packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
+++ b/packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
@@ -67,15 +67,123 @@ function validateKeyField(field: FieldDefinitionNode): void {
  * (Not a list and of type ID or String)
  * @param field: the field to be checked.
  */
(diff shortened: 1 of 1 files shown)


attilah:
@hirochachacha thanks for the contribution. I see why this is useful, but we don't like to couple the behavior of directives together, so I will not approve this PR for now. However this functionality could be put into a custom transformer - which will run after all the transformers but before @auth - so this cross directive check and update of given fields would be doable there. Would you be interested to give it a try and implement this transformer? I had a builder session at re:Invent 2019 about this topic, you can find the session content here: https://github.com/aws-samples/aws-reinvent-2019-mobile-workshops/tree/master/MOB405

hirochachacha:
Thank you for your advice. I've created https://github.com/hirochachacha/graphql-auto-transformer which should solve https://github.com/aws-amplify/amplify-cli/issues/1657 too.

PR #2907, opened 2019-12-02
fix(cli): print correct message if no provider plugins are active

UnleashedMind:
This reverts commit 374937fe071a531ca506da42d37037e5c1aca010.

Issue #, if available:

Description of changes:

Current behavior of the CLI plugin platform: If there are not active provider plugins, in the init or the pull workflows, it prompts the customer to select from an empty list.
This PR address this bug by printing more relevant error message when it occurs, and then throws an error to properly end the commands' executions.


packages/amplify-cli/src/lib/attach-backend-steps/a10-queryProvider.js
```diff
@@ -17,6 +17,12 @@ async function getProvider(context, providerPlugins) {
   let result;
   const providers = [];
   const providerPluginList = Object.keys(providerPlugins);
+
+  if (providerPluginList.length === 0) {
+    const errorMessage = 'Found no provider plugins';
+    throw new Error(errorMessage);
```
kaustavghosh06: We should give customers some next steps to mitigate this issue like - "Run `amplify plugin scan` for scanning provider plugins in your system".

packages/amplify-cli/src/lib/init-steps/s2-initProviders.js
```diff
@@ -22,6 +22,12 @@ async function run(context) {
 async function getProviders(context, providerPlugins) {
   let providers = [];
   const providerPluginList = Object.keys(providerPlugins);
+
+  if (providerPluginList.length === 0) {
+    const errorMessage = 'Found no provider plugins';
```
kaustavghosh06: Same as above.

PR #2932, opened 2019-12-06
fix: update cli if it doesn't satisify minimum version

nikhname:
*Issue #, if available:*

*Description of changes
Install cli if it doesn't satisfy the min version

### DIFF
--- a/packages/amplify-app/src/index.js
+++ b/packages/amplify-app/src/index.js
@@ -82,12 +82,12 @@ async function amplifyCLIVersionCheck() {
     if (semver.satisfies(amplifyCLIVersion, minCLIVersion)) {
       console.log(`${emoji.get('white_check_mark')} Found Amplify CLI v${amplifyCLIVersion}`);
     } else {
-      console.log(`${emoji.get('worried')} Amplify CLI version ${minCLIVersion} not found.`);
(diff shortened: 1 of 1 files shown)


packages/amplify-app/src/index.js
```diff
@@ -81,6 +81,10 @@ async function amplifyCLIVersionCheck() {
     const amplifyCLIVersion = semver.coerce(stripAnsi(amplifyCLIVersionSpawn.stdout.toString()));
     if (semver.satisfies(amplifyCLIVersion, minCLIVersion)) {
       console.log(`${emoji.get('white_check_mark')} Found Amplify CLI v${amplifyCLIVersion}`);
+    } else {
+      console.log(`${emoji.get('worried')} Amplify CLI version ${minCLIVersion} not found.`);
```
attilah: I think a more developer friendly message would be something like:
- No Amplify CLI found: "Amplify CLI was not found"
- Non-satisfying CLI version found: "Amplify CLI version found was: ${amplifyCLIVersion}, the minimum required version is: ${minCLIVersion} 

This way we present more information about why are we installing a new CLI version.

nikhname: Updated

PR #2933, opened 2019-12-06
refactor: remove zip file and copy a skeleton dir

nikhname:
*Issue #, if available:*

*Description of changes:*
Allows changes to skeleton to be explicit and can be modified procedurally later


kaustavghosh06:
@nikhname Is this just required so that we can be more efficient when making changes to the code?

nikhname:
> @nikhname Is this just required so that we can be more efficient when making changes to the code?

@kaustavghosh06 yes, so that changes to the skeleton are more clear in PRs. Also it will be easier to write functions that automatically update the skeleton later

Issue #2938, opened 2019-12-06
Disabling DEBUG logging in python_streaming_function raises KeyError


PatrykMilewski:
**Describe the bug**
When I try to remove DEBUG Lambda environment variable, to stop lambda from logging debug messages, it raises KeyError exception:
```
Traceback (most recent call last):
  File "test.py", line 2, in <module>
    DEBUG = True if os.environ['DEBUG'] is not None else False
  File "/usr/local/Cellar/python@2/2.7.17/Frameworks/Python.framework/Versions/2.7/lib/python2.7/UserDict.py", line 40, in __getitem__
    raise KeyError(key)
```

**Amplify CLI Version**
latest

**To Reproduce**
Remove DEBUG lambda environment variable from deployed python_streaming_function or simply try to run this code without DEBUG env variable:
`DEBUG = True if os.environ['DEBUG'] is not None else False`

**Expected behavior**
When I unset DEBUG env variable, it shouldn't raise KeyError and stop logging debug logs.

I will fix it myself and create Pull Request.

PatrykMilewski:
Oh wait, I can't create a new Pull Request? Anyways, it should be fixed like this:
`DEBUG = True if os.environ.get('DEBUG') is not None else False`

ammarkarachi:
@PatrykMilewski thanks for submitting this issue. As for the pull request you can create one what you would have to do is.

- Fork amplify-cli
- Make changes and push on a branch from your fork
- Create a pull request using the branch on your fork

SwaySway:
Closing this issue as the fix has been merged.

PR #2963, opened 2019-12-10
fix(graphql-dynamodb-transformer): fixed KeyError for unset DEBUG

PatrykMilewski:
#2938 

Replaced print with logger.debug in python_streaming_function.py to avoid unneeded logging.


packages/graphql-elasticsearch-transformer/streaming-lambda/python_streaming_function.py
```diff
@@ -18,16 +18,15 @@
 # The following parameters are required to configure the ES cluster
 ES_ENDPOINT = os.environ['ES_ENDPOINT']
 ES_REGION = os.environ['ES_REGION']
-DEBUG = True if os.environ['DEBUG'] is not None else False
+DEBUG = True if os.environ.get('DEBUG') not in ['0', 'False', 'false', None] else False
```
SwaySway: It might be simpler to check if DEBUG == 1 as that is the only case in which debug logs would appear.

PatrykMilewski: You are right, fixed

Issue #2624, opened 2019-10-23
Amplify mock keeps adding `amplify\mock-data` to the .gitignore file


ThijsTyZ:
**Describe the bug**
Each time I run `amplify mock` another line of `amplify\mock-data` is added to the `.gitignore` file.

**To Reproduce**
run `amplify mock`
check the .gitignore file

**Expected behavior**
If the .gitignore file already has the line `amplify\mock-data` it should not add another one

**Screenshots**

**Desktop**
 - OS: Windows 10 Pro
 - Version 1903

ammarkarachi:
@ThijsTyZ Can you let us know the Amplify the CLI version you are using?

ThijsTyZ:
3.9.0

kaustavghosh06:
Fix for this has been merged and released.

PR #2964, opened 2019-12-10
Fix for .gitignore duplicating amplify\mock-data

atmospherium:
Original bug: On Windows, path.relative returns a path with escaped backslashes. This entry is incompatible with .gitignore and results in new entries being added every time "amplify mock" is run.  

Fix: Convert escaped backslash to forward slash for detection and entry in .gitignore.

#2624 


atmospherium:
This is an alternate solution to what is proposed in PR #2684

ammarkarachi:
LGTM

Issue #2390, opened 2019-09-24
Running `amplify api gql-compile` results in 'TypeError: Cannot convert undefined or null to object'


hisham:
**Describe the bug**
I recently upgraded to 3.9.0 and now `amplify api gql-compile` results in this error: `TypeError: Cannot convert undefined or null to object`

**Additional context**
Not sure what you need to identify root cause. Happy to provide more info.

hisham:
Now I get the error again, after I changed the graphql schema and did a push.

kevcam4891:
Yes, `amplify env pull` is the solution to bring things back up to date. Also, what version of amplify-cli are you using? `amplify --version`? I am running into this issue almost never now that I'm using the latest version 3.9. I was using 3.0 when this was happening for me a lot.

hisham:
```
> amplify --version
Scanning for plugins...
Plugin scan successful
3.9.0
```

Yeah `amplify env pull` fixes but it seems to re-occur again after push with changed gql schema.

SwaySway:
@hisham are you able to share the schema as well?

hisham:
@SwaySway sure I just sent it to amplify-cli@amazon.com

SwaySway:
@hisham Recieved the schema taking a look. Could you also provide your `.graphqlconfig.yml`? Checking to see if this might be related to #2412, as you mentioned this happened when you updated the enum.

hisham:
Sure done.

kaustavghosh06:
@hisham We weren't able to reproduce this error with the latest version of the CLI? Is it still a recurring issue for you?

hisham:
@kaustavghosh06 I haven't touched the backend part of my app in a while but will be in the coming days. Currently however I am blocked with this bug after upgrading to 3.17: https://github.com/aws-amplify/amplify-cli/issues/2390

hisham:
I just got this error again. I don't have exact repro steps. I made changes to graphql, pushed. All was working fine. Then switched to another git branch with slightly different graphql schema, ran compile graphql and got the error.

kevcam4891:
`amplify env checkout` has always been my savior after switching branches
and experiencing something like this.

hisham:
I'm now getting this error consistently for one of my environments regardless what I do. Not good. Still investigating.

hisham:
I'm instrumenting the code and managed to get it to throw a stack trace, it is as follows: 

```
[...]

PR #2969, opened 2019-12-10
Fix for https://github.com/aws-amplify/amplify-cli/issues/2390

hisham:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-cli/issues/2390

*Description of changes:*
Handle case where lastDeployedStack.Outputs is undefined. Resolves `TypeError: Cannot convert undefined or null to object` error that sometimes occurs when running `amplify api gql-compile`.


packages/graphql-transformer-core/src/util/amplifyUtils.ts
```diff
@@ -145,7 +145,7 @@ async function ensureMissingStackMappings(config: ProjectOptions) {
               missingStackMappings[resourceId] = stackName;
             }
           }
-          const outputIdsInStack = Object.keys(lastDeployedStack.Outputs);
+          const outputIdsInStack = Object.keys(lastDeployedStack.Outputs || {});
```
SwaySway: Since outputs are optional it is possible the previous stack may have not outputs.
If possible could the same check be added to the root stack outputs as well?
- [L165](https://github.com/aws-amplify/amplify-cli/blob/4dc4d063a451ca783a6bcea6748f277d390d262e/packages/graphql-transformer-core/src/util/amplifyUtils.ts#L165)

hisham: Sure done!

PR #2985, opened 2019-12-12
fix: use relative path for xcode and android studio helper files

nikhname:
*Issue #, if available:*

*Description of changes:*
Use relative paths so there is not a conflict when using source control.

https://github.com/aws-amplify/amplify-ios/issues/247

### DIFF
--- a/packages/amplify-app/src/index.js
+++ b/packages/amplify-app/src/index.js
@@ -352,7 +352,7 @@ async function createAndroidHelperFiles() {
     syncEnabled: true,
   };
   const configJsonStr = JSON.stringify(configJsonObj, null, 4);
-  const configFile = './amplify-gradle-config.json';
+  const configFile = path.join(process.cwd(), './amplify-gradle-config.json');
   if (!fs.existsSync(configFile)) {
     fs.writeFileSync(configFile, configJsonStr);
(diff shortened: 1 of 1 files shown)


packages/amplify-app/src/index.js
```diff
@@ -346,19 +346,23 @@ async function createJSHelperFiles() {
...
-  const configJsonObj = { profile: 'default', envName: 'amplify', syncEnabled: true };
-  const configJsonStr = JSON.stringify(configJsonObj);
-  const configFile = path.join(process.cwd(), './amplify-gradle-config.json');
+  const configJsonObj = {
+    profile: 'default',
+    envName: 'amplify',
+    syncEnabled: true,
+  };
+  const configJsonStr = JSON.stringify(configJsonObj, null, 4);
+  const configFile = './amplify-gradle-config.json';
```
kaustavghosh06: Won't this fail for windows?

nikhname: Yes but I'm going to have all the fixes for windows support in a separate PR

nikhname: Changing this back. Absolute path is ok here since it is only for writeFileSync

Issue #2860, opened 2019-11-27
@searchable item with stringset value can't be indexed to Elasticsearch


rarira:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[Amplify Console GitHub Issue Tracker](https://github.com/aws-amplify/amplify-console/issues)

**Describe the bug**
When type with @searchable notation, all item refreshed should be indexed to elasticsearch using lambda which is deployed with amplify cli, but if the item has a value with stringset(maybe, number set also), indexing can't be done with this error (cloudwatch log).

```
f179a897-b185-42c4-98ea-07a1b7ba67b2	Traceback (most recent call last):
  File "/var/task/python_streaming_function.py", line 225, in lambda_handler
    return _lambda_handler(event, context)
  File "/var/task/python_streaming_function.py", line 195, in _lambda_handler
    doc_json = json.dumps(doc_fields)
  File "/var/lang/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/var/lang/lib/python3.6/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/var/lang/lib/python3.6/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/var/lang/lib/python3.6/json/encoder.py", line 180, in default
    o.__class__.__name__)
TypeError: Object of type 'set' is not JSON serializable
```

**Amplify CLI Version**

4.1.1 ... all version below was the same

**To Reproduce**
Steps to reproduce the behavior or terminal output if applicable
```gql
type Class
  @model
  @searchable {
  id: ID!
  title: String!
  blockedBy: [String]
}

```
fails to be indexed because of 'blockedBy' which is stringset.

**Expected behavior**
it would be best if it could be indexed with stringset value, 
but I just expect that if there is stringset, indexing should be done without that.

I could possibly achieve that  in this case  by inserting these lines to DdbToEsFn python lambda code after line 185:
```
  # Delete StringSet
        if "blockedBy" in ddb[image_name]:
            del ddb[image_name]["blockedBy"]

```

but I think it's just temporary.

SwaySway:
@rarira I tried replicating your issue from the schema given above.
I was not able to run into this issue.
Is it possible for you to send over a zip of your project?
<amplify-cli@amazon.com>

rarira:
[...]

PR #2986, opened 2019-12-12
feat(graphql-elasticsearch-transformer): support sets in es fn

SwaySway:
convert sets to lists when sending payload to es

re #2860

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformer.e2e.test.ts
+++ b/packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformer.e2e.test.ts
@@ -112,8 +112,6 @@ beforeAll(async () => {
     // change create/update to create string sets
     const out = addStringSets(transformer.transform(validSchema));
     // fs.writeFileSync('./out.json', JSON.stringify(out, null, 4))
-    // create stack with additional params
-    // const additionalParams = generateParams()
     console.log('Creating Stack ' + STACK_NAME);
     const finishedStack = await deploy(
       customS3Client,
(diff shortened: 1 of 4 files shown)


packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformer.e2e.test.ts
```diff
@@ -675,14 +685,34 @@ test('Test updatePost syncing with Elasticsearch', async () => {
...
-  return params;
-}
+// function generateParams() {
+//   const params = {
+//     [ResourceConstants.PARAMETERS.ElasticsearchAccessIAMRoleName]: 'ElasticsearchAccessIAMRoleTest',
+//     [ResourceConstants.PARAMETERS.ElasticsearchStreamingIAMRoleName]: 'ElasticsearchStreamingIAMRoleTest',
+//   };
+
+//   return params;
+// }
```
yuth: Remove this if its no longer needed instead of commenting it out

packages/graphql-transformers-e2e-tests/src/stringSetMutations.ts
```diff
@@ -0,0 +1,193 @@
+import { DeploymentResources } from 'graphql-transformer-core';
+
+/**
+ * Used with SearchableModelTransformer Tests
+ * Changes userItems list to stringsets in dynamo so the lambda function can support it
+ */
+
+const createStringSetVTL = `
```
yuth: Can we generate this instead of hard coding? How does these resolvers get generated when customers use the searchable transformer?

SwaySway: This is more of a custom implementation here since the the generated resolver in `@model` uses `$util.dynamodb.toDynamoDB($entry.value)` to intrinsically define the the type it should be in dynamodb based off the schema.

PR #2987, opened 2019-12-13
feat(cli): migrate projects that don't have Amplify AppId

### LINKED ISSUE #2972: Additional app created after upgrading amplify-cli to 4.x
**Describe the bug**
Duplicated App in Amplify Console after upgrading Amplify-CLI to 4.x (4.2 as of now). 

image

**To Reproduce**
1. Create a new App within Amplify Console
2. Install Amplify-CLI older than 4, upgrade it and push
3. You should see a duplicated app

**Expected behavior**
Single App

**Additional context**

**Sample code**
Include additional sample code or a sample repository to help us reproduce the issue. (Be sure to remove any sensitive data)

UnleashedMind:
fix #2972

*Issue #, if available:*

*Description of changes:*


packages/amplify-provider-awscloudformation/lib/initializer.js
```diff
@@ -66,6 +67,13 @@ async function run(context) {
         spinner.fail('Root stack creation failed');
         throw e;
       });
+  } else if (
```
kaustavghosh06: Can you add a comment out here as to when this code would be executed? (For readability purproses)

packages/amplify-provider-awscloudformation/lib/amplify-service-migrate.js
```diff
@@ -0,0 +1,110 @@
...
+async function run(context) {
+  const projectDetails = context.amplify.getProjectDetails();
+  const awsConfig = await configurationManager.getAwsConfig(context);
+
+  const amplifyClient = await getConfiguredAmplifyClient(context, awsConfig);
+  if (!amplifyClient) {
+    // This happens when the Amplify service is not available in the region
+    const message = `Amplify service is not available in the region ${awsConfig.region ? awsConfig.region : ''}`;
+    context.print.error(message);
+    throw new Error(message);
```
kaustavghosh06: We shouldn't throw an error for this case right? We should go ahead and provision the AWS resources without creating the Amplify app? cc @swaminator

packages/amplify-provider-awscloudformation/lib/amplify-service-migrate.js
```diff
@@ -0,0 +1,110 @@
...
+      const { backendEnvironment } = await amplifyClient.getBackendEnvironment(getEnvParams).promise();
+      if (StackName !== backendEnvironment.stackName) {
+        const message = `Stack name mistmatch for the backend environment ${envName}. Local: ${StackName}, Amplify: ${backendEnvironment.stackName}`;
+        context.print.error(message);
+        throw new Error(message);
+      }
+    }
+
+    // Add the appId to the meta data and team provider info and write the files
+    const { amplifyMeta, teamProviderInfo } = projectDetails; // context.exeInfo;
```
kaustavghosh06: What's the "// context.exeInfo;" comment?

packages/amplify-provider-awscloudformation/lib/amplify-service-migrate.js
UnleashedMind: We should, because this workflow is explicitly invoked by the customer. And we should let the customer know it's not possible to execute this workflow in the particular region, because Amplify Console service is not available in the configured region. Throwing error also stops the actions following this function, which is our intention in those regions.

packages/amplify-provider-awscloudformation/lib/amplify-service-migrate.js
UnleashedMind: removed

packages/amplify-provider-awscloudformation/lib/initializer.js
UnleashedMind: comment added

Issue #2468, opened 2019-10-01
IAM Role Names for @function directive are not unique enough.


daannijkamp:
**Describe the bug**
Automatically generated IAM Role Names for `@function` directive are not unique enough.
Will result in `AWS::IAM::Role already exists in (other) stack error` 

**To Reproduce**
1. Create a new serverless.com project
2. Give it a long service name, for example: `long-service-name-functions`
3. Create multiple functions, for example `do-something` and `get-something`
4. This will result in lambda function with the name `long-service-name-functions-prod-do-something` and `long-service-name-functions-prod-get-something`
5. Use the Lambda functions as `@function` directive
6. For example:
```
type Query {
  getSomething(email: String): String
    @function(name: "long-service-name-functions-prod-get-something")
}

type Mutation {
  doSomething(email: String): String
    @function(name: "long-service-name-functions-prod-do-something")
}

``` 
7. amplify push
8. An automatically generated IAM Role Name is created `LongServiceNameFunctionsPr-xxxxxxxxxxxxxxxxxxxxxxxxxx-production` for both functions. And will result in an `AWS::IAM::Role already exists in (other) stack` error 

**Expected behavior**
- support long prefixes in (serverless) lambda functions

danshirley1:
Does anyone have any workarounds for this in the meantime, manual steps will do?

babu-upcomer:
@danshirley1 We got the issue recently too. These kind of cloudformation limitations are becoming a huge blocker for Amplify in a greater extent. @SwaySway Is their any workaround?

daannijkamp:
@danshirley1 @babu-upcomer the +/- first 27 characters of the lambda function name is used for the IAM Role Name.
The only workaround is to use shorter Lambda function names (more likely that the first 27 characters are unique)  
Don't know whether this is a limitation of Cloudformation or Amplify.

daannijkamp:
@SwaySway Adding a hash based on the full string (for example md5) as a prefix to the simplifyName function (https://github.com/aws-amplify/amplify-cli/blob/master/packages/graphql-transformer-common/src/util.ts#L16), do you think this can work?

SwaySway:
@daannijkamp Yea it's a cloud formation limit that is 64 characters on the iam role name which is created for the lambda function.
Which is comprised of the following
- `functionName-GraphQLAPI_ID-Hash`

One option would be to remove the api id and that alone could be unique with the function name and the hash.

[...]

PR #3030, opened 2019-12-18
fix(graphql-function-transformer): add hash to function iam role name

SwaySway:
*Issue #, if available:*
re #2468
*Description of changes:*
- generated a short uuid along with the role name to keep them unique
- updated e2e tests

### DIFF
--- a/packages/graphql-transformer-common/src/FunctionResourceIDs.ts
+++ b/packages/graphql-transformer-common/src/FunctionResourceIDs.ts
@@ -1,4 +1,5 @@
 import { simplifyName } from './util';
+import * as shortid from 'shortid';
 
 export class FunctionResourceIDs {
   static FunctionDataSourceID(name: string, region?: string): string {
@@ -7,6 +8,13 @@ export class FunctionResourceIDs {
 
   static FunctionIAMRoleID(name: string, region?: string): string {
     return `${FunctionResourceIDs.FunctionDataSourceID(name, region)}Role`;
+  }
+
+  static FunctionIAMRoleName(name: string, withEnv: boolean = false): string {
+    if (withEnv) {
+      return `${simplifyName(name).slice(0, 18)}${shortid.generate()}`;
+    }
+    return `${simplifyName(name).slice(0, 28)}${shortid.generate()}`;
   }
 
   static FunctionAppSyncFunctionConfigurationID(name: string, region?: string): string {
--- a/packages/graphql-transformer-common/src/FunctionResourceIDs.ts
+++ b/packages/graphql-transformer-common/src/FunctionResourceIDs.ts
@@ -1,5 +1,5 @@
 import { simplifyName } from './util';
-import * as shortid from 'shortid';
+import * as md5 from 'md5';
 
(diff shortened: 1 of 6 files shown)


packages/graphql-transformer-common/src/FunctionResourceIDs.ts
```diff
@@ -9,6 +10,13 @@ export class FunctionResourceIDs {
     return `${FunctionResourceIDs.FunctionDataSourceID(name, region)}Role`;
   }
 
+  static FunctionIAMRoleName(name: string, withEnv: boolean = false): string {
+    if (withEnv) {
+      return `${simplifyName(name).slice(0, 18)}${shortid.generate()}`;
```
kaustavghosh06: We don't want a new UUID each time we do this. This would result in deleting an IAM role and then creating a new IAM role each time. Use the first 4 chars of the AppId instead.

SwaySway: Updated to use a hash of the name added at the end.

packages/graphql-transformer-common/src/FunctionResourceIDs.ts
```diff
@@ -9,6 +10,13 @@ export class FunctionResourceIDs {
     return `${FunctionResourceIDs.FunctionDataSourceID(name, region)}Role`;
   }
 
+  static FunctionIAMRoleName(name: string, withEnv: boolean = false): string {
+    if (withEnv) {
+      return `${simplifyName(name).slice(0, 22)}${md5(name).slice(0, 4)}`;
+    }
+    return `${simplifyName(name).slice(0, 32)}${md5(name).slice(0, 4)}`;
```
kaustavghosh06: Where is the env name added out here?

SwaySway: If the env is added then we are only accounting for the extra space of characters it will take.

kaustavghosh06: We need to add the env name - although the function name already has the env-name. If we don't add it, deployment in multi-environments would fail since IAM role names need to be globally unique in an account. 
Please test this change with multiple environments.

SwaySway: It is added in the cloudformation template.
An example of the output will look like the following.
```json
"RoleName": {
                    "Fn::If": [
                        "HasEnvironmentParameter",
                        {
                            "Fn::Join": [
                                "-",
                                [
                                    "LongServiceNameFunctiocf96",
                                    {
                                        "Ref": "GetAttGraphQLAPIApiId"
                                    },
                                    {
                                        "Ref": "env"
                                    }
                                ]
                            ]
                        },
                        {
                            "Fn::Join": [
                                "-",
                                [
                                    "LongServiceNameFunctionsProdGetScf96",
                                    {
                                        "Ref": "GetAttGraphQLAPIApiId"
                                    }
                                ]
                            ]
                        }
                    ]
                }
```

PR #3055, opened 2019-12-21
fix( amplify-codegen-appsync-model-plugin): field name errors

yuth:
Changed userClaim to identityClaim and convert username to cognito:username

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-codegen-appsync-model-plugin/src/utils/process-auth.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/utils/process-auth.ts
@@ -26,7 +26,7 @@ export enum AuthModelMutation {
 }
 
 const DEFAULT_GROUP_CLAIM = 'cognito:groups';
-const DEFAULT_USER_CLAIM = 'username';
+const DEFAULT_IDENTITY_CLAIM = 'username';
 const DEFAULT_OPERATIONS = [AuthModelOperation.create, AuthModelOperation.update, AuthModelOperation.delete];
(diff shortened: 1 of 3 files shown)


packages/amplify-codegen-appsync-model-plugin/src/utils/process-auth.ts
```diff
@@ -26,7 +26,7 @@ export enum AuthModelMutation {
 }
 
 const DEFAULT_GROUP_CLAIM = 'cognito:groups';
-const DEFAULT_USER_CLAIM = 'username';
+const DEFAULT_IDENTITY_CLAIM = 'username';
```
kaustavghosh06: Shouldn't this be "cognito:userName"?

yuth: If you mean the case sensitive, its should be `cognito:username`
https://github.com/aws-amplify/amplify-cli/blob/830ec21adbdd6365318a983aa8075b7b8802a444/packages/graphql-auth-transformer/src/resources.ts#L42

If you mean it we should default to `cognito:username` then it happens when we set the value here on line 65
```ts
identityClaim: identityClaim === 'username' ? 'cognito:username' : identityClaim,
```

kaustavghosh06: Ah, got it.

PR #3072, opened 2019-12-24
Custom auth with tests

haverchuck:
*Description of changes:*
Adds authenticationFlowType key to the awsconfiguration file for ios and android only.


desokroshan:
I have verified that : 

1. awsconfiguration.json is updated with authenticationFlowType as CUSTOM_AUTH if user enables custom auth and configures the required lambda triggers in the cli flow by running `amplify add auth`
1. awsconfiguration.json is updated with authenticationFlowType as USER_SRP_AUTH if user disables custom auth and consequently removes lambda triggers by running `amplify update auth`

Issue #2997, opened 2019-12-14
Deleting and reinitializing amplify with authentication leads to stale values in aws-exports.js


dantasfiles:
**Describe the bug**
Deleting and reinitializing amplify with authentication leads to stale values in `aws-exports.js`

**Amplify CLI Version**
4.6.0

**To Reproduce**
- Run `npx react-native init AddAuth`
- Run `amplify init`
- Run `amplify add auth`
- Run `amplify push`
- This writes the correct values for `aws_cognito_identity_pool_id` and `aws_user_pools_id` to `aws-exports.js`
- Run `amplify delete`
- Run `amplify init`
- Run `amplify add auth`
- Run `amplify push`
- ERROR: the old, stale values for the user and identity pool are in `aws-exports.js`. The new user and identity pools are created, but the `aws-exports.js` does not point to them. This leads to all sorts of weird errors when you actually try to run your app.

**Expected behavior**
The second `amplify push` should update the `aws-exports.js` file correctly

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Node v.12.9.1

**Additional context**
I've been experimenting with authentication, which is why I've been deleting and reinitializing amplify often.

yuth:
Amplify CLI does not overwrite existing values if `aws-export.js` and it is not deleteing `aws-export.js` file after running `amplify delete`.

I am marking this as a bug

Klastic:
I ran into this issue myself and found running `amplify pull` will update your `aws-export.js`.

SwaySway:
Closing this issue as the fix has been merged and was included in the 4.13+ release.

PR #3116, opened 2020-01-03
fix: point to correct file to delete for amplify config

SwaySway:
point to aws exporsts in js deleteConfig and include amplify configuration in the delete for android
and ios

re #2997

*Issue #, if available:*

*Description of changes:*


ammarkarachi:
LGTM

PR #2463, opened 2019-09-30
Ability to add custom EventSource and lambda triggers via amplify add function, Kinesis support in analytics category

ambientlight:
**Issue #, if available:**
https://github.com/aws-amplify/amplify-cli/issues/997
https://github.com/aws-amplify/amplify-cli/issues/2256

**Description of changes:**
1. Exposes `streamArnOutputId`, `datasourceOutputId`, `tableNameOutputId` from nested api category stack backing each @model dynamoDB table, so they would available inside root api category stack output and inside `amplify-meta` and can be references in other categories templates (`graphql-dynamodb-transformer`). It is needed in the context of this PR, so we can reference @model dynamoDB table streamArn in our custom eventSource.
2. A reasonable CLI support in `amplify-category-function` to confirm a lambda with a custom event source. `DynamoDBStream` / `Kinesis` event source types are supported. For DynamoDBStream, user is able to choose from api category resource `@model` dynamoDB tables, storage category existing dynamoDB table, or as last resort directly enter `streamArn`. For `Kinesis, user is able to choose from analytics category resource kinesis stream or also directly enter `streamArn`.
3. Base `lambda-cloudformation-template.json.ejs` template has been extended to optionally include `eventSourceMapping` and `LambdaTriggerPolicy`  when available as `props.triggerEventSourceMapping`.

**Motivation:**
1. A need to have a lambda trigger attached to `@model` backed dynamoDB tables. Since `graphql-transformer-core` will update CloudFormation templates on each graphql update, it is not possible to modify the templates directly like in other category resources to output nested stacks outputs
2. Existing eventSource lambda trigger functionality available in storage category is quite unintuitive, slightly opinionated, but I would argue most users will expect triggers to be available as part of function category just as it is in lambda aws console.
3. If the changes are accepted this will also allow to avoid a set of shortcuts in `amplify-category-storage` that explicitly load and manipulate resources and parameters in function CloudFormation templates, if we would go with replacing a storage category trigger creation with function category built-in implementation presented here. https://github.com/aws-amplify/amplify-cli/blob/cb8f6dddefb7f7e7f8159988563fc076f470ee79/packages/amplify-category-storage/provider-utils/awscloudformation/service-walkthroughs/dynamoDb-walkthrough.js#L606-L684
4. It would also allow users to create event sources for non-amplify managed resources.

### Caveats:
To make CLI discover `@model` backed dynamoDB table outputs, user will need to push any update to api category to have CloudFormation templates regenerated with nested stack outputs.

[...]

PR #3123, opened 2020-01-05
feat(graphql-auth-transformer): add compound auth rules

RossWilliams:
adds 'and' property on auth rules to force each auth rule with the same 'and' property to all match before allowing operations. Allows combining owner, group, and dynamic group authentication rules. Works on objects and fields for user pools.

*Issue #, if available:* #1043 #317 #1578

*Description of changes:*
The RFC for @auth directive improvements describes in option 5 the ability to join auth rules together using 'AND' and 'OR', with an arbitrary level of nesting. This PR takes a slightly different approach in only implementing AND, and not allowing nesting. This feels reasonable as @auth rules are OR'd by default and any nesting can be decomposed into a flat structure.

To implement this a new property on an auth rule is created, "and". And is assigned a string, and all rules with the same assigned and string are considered part of a group whose rules must all pass to allow the operation.

Compound rules are tracked by counting how many of the rules have passed, rejecting the operation if no compound rule sets have all their rules pass.

List operations require a deep clone of the object holding the counts via json stringify and json parse operations so that their count tracking is independent of each other.

Update and Delete operations are somewhat more challenging as their dynamic group and owner rules are evaluated on the server, while the static auth is evaluated on the client. This requires nesting condition expressions, and only executing DynamoDB calls if the static group rules have been satisfied.

### DIFF
--- a/packages/graphql-auth-transformer/src/AuthRule.ts
+++ b/packages/graphql-auth-transformer/src/AuthRule.ts
@@ -15,4 +15,5 @@ export interface AuthRule {
   operations?: ModelOperation[];
   queries?: ModelQuery[];
   mutations?: ModelMutation[];
+  and?: string;
 }
--- a/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
+++ b/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
@@ -39,7 +39,21 @@ import {
   makeField,
   ModelResourceIDs,
 } from 'graphql-transformer-common';
-import { Expression, print, raw, iff, forEach, set, ref, list, compoundExpression, newline, comment, not } from 'graphql-mapping-template';
+import {
+  Expression,
+  print,
+  raw,
+  iff,
+  forEach,
+  set,
+  ref,
+  list,
+  compoundExpression,
+  newline,
+  comment,
+  not,
+  obj,
+} from 'graphql-mapping-template';
 import { ModelDirectiveConfiguration, ModelDirectiveOperationType, ModelSubscriptionLevel } from './ModelDirectiveConfiguration';
 
 import { OWNER_AUTH_STRATEGY, GROUPS_AUTH_STRATEGY, DEFAULT_OWNER_FIELD } from './constants';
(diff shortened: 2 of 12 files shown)


davekiss:
This is a simple and elegant solution to a sorely missing feature in Amplify. What is the likelihood that this PR gets merged?

@RossWilliams My only note is `and: "ruleGroupName"` sort of mixes readability with parameterization. Not sure what sort of complexities this may introduce, but I wonder if a key/value of `ruleGroup: "ruleGroupName"` might make more sense from an implementation standpoint.

RossWilliams:
@davekiss that is an easy find/replace change once a good name is found. "And" was chosen to match #1043 Proposal 5 naming, and it reads well when it is the last parameter, but I can see it being problematic otherwise. 'ruleGroup' itself does not feel like the final solution, as the word 'group' would be an overload for the group authorisation parameters. 'ruleSet' may be more descriptive.

In terms of timeline, I'm not aware of the process, but #2463 appears to be consuming most of the review energy right now, might need to wait for that to be completed first.

attilah:
@RossWilliams thanks for this contribution. We cannot merge it as-is at this point, as it deviates the expression based syntax in the proposal, which is kind of a shared syntax when building expressions with the CLI’s directives, and the unknown performance implications what is also called out by you in the code. In addition, we’re experimenting with some auth logic change for VTL to simplify customers’ use cases and that work affects AND, OR rules as well.

RossWilliams:
@attilah 
> In addition, we’re experimenting with some auth logic change for VTL to simplify customers’ use cases and that work affects AND, OR rules as well

Do you mind sharing the auth logic changes you are playing with? I'm more than happy to help move it along and stop having to rebase these changes or another 9 months.

attilah:
@RossWilliams nothing concrete to share at this point, but sure, I agree, no need to rebase this PR until we've anything.

davekiss:
@attilah @RossWilliams wondering what the status is here. Ross, are you still using your custom fork for this?

RossWilliams:
Im still using a variant of this code with additional customisations and performance improvements.

I would not use this PR as-is, as there have been subsequent security issues found and fixed in the auth module.

ronaldocpontes:
@attilah as you close this PR... are you still working on a solution?

@RossWilliams would you be kind enough to share your updated code with additional customisations and performance improvements?

PR #3144, opened 2020-01-08
Add team-provider-info.json to gitignore appending string

hyochan:
@dabit3 confirmed that `team-provider-info.json` should be included in `.gitignore` appending string followed by the [issue](https://github.com/dabit3/amplify-auth-demo/issues/1).

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/get-git-ignore-blob.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/get-git-ignore-blob.js
@@ -12,6 +12,7 @@ function getGitIgnoreBlob() {
     'aws-exports.js',
     'awsconfiguration.json',
     'amplifyconfiguration.json',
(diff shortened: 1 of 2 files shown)


kaustavghosh06:
@hyochan If you're working in the team, `team-provider-info.json` shouldn't be in the .gitignore. Only when you want to make your infrastructure architecture public - only then you'll add it to the .gitignore file. You can find more documentation out here - https://aws-amplify.github.io/docs/cli-toolchain/quickstart#sharing-projects-outside-the-team-

karrettgelley:
@kaustavghosh06 If I'm using multiple environments (master, dev, feature_x) then wouldn't I not want team-provider.json to be updated after a merge dev <- feature_x? I've noticed that unless I remove the feature branch environments following a merge, then I end up with a team-provider with dozens of listings that are stale.

Issue #3148, opened 2020-01-08
owner subscriptions causing error: '...of type "String!" is required but not provided.'


oliverandersencox:
Framework: Angular 8
Amplify Cli - fresh install today

The responses in regards to the issue that follows have not been useful at all, and i feel maybe I am not explaining well.

I have a messaging system (realtime chat) in which a a conversation (model) links a business and a user together and messages are stored as a connection within this conversation.

On the business website I was expecting to use Amplify subscriptions to implement a real time chat system, however I have been unable to do so using many approaches. 

I have implemented authentication rules and multiple owners so as a user generated message will still be streamed into the subscription, given the user will be the owner and not the business. If the business creates the message, then there is no problem as they are the owner. Hence I then followed the docs by creating an owners array which will have owner auth rules applied.

However the only way I can get the subscription to fire is by making the subscription public, which is not ideal.

Could one of the amazon team (@mikeparisstuff @kaustavghosh06 etc) perhaps give some pointers here on how I can achieve the firing of subscriptions.

The models:

Conversation:

```gql
type ConvoLink @model(subscriptions: null) {
  id: ID!
  status: String
  createdAt: Float!
  updatedAt: Float!
  messages: [Message] @connection(name: "ConvoMsgs", sortField: "createdAt")
  user: User @connection(name: "UserConversations")
  entity: Entity @connection(name: "EntityConversations")
}
```

The message

```gql
type Message @model
  @auth(rules: [
    { allow: owner },
    { allow: owner, ownerField: "owners" }
  ])
{
  id: ID!
  messageConversationId: ID!
  conversation: ConvoLink @connection(name: "ConvoMsgs")
  content: String!
  contentType: String
  author: String!
  entity: String!
  receiver: String!
  owner: String!
  owners: [String]!
  status: String
  createdAt: Float!
  updatedAt: Float!
}
```

This completed the push operation to the API, however on graphQL codegen, it fails to create the subsections with the following errors:

> .../Ariba/src/graphql/mutations.ts: Field "onCreateMessage" argument "owners" of type "String!" is required but not provided.
> .../Ariba/src/graphql/mutations.ts: Field "onUpdateMessage" argument "owner" of type "String!" is required but not provided.
> .../Ariba/src/graphql/mutations.ts: Field "onUpdateMessage" argument "owners" of type "String!" is required but not provided.
[...]

PR #3166, opened 2020-01-11
fix(graphql-auth-transformer): add list support for ownerField in subs

SwaySway:
support list ownerFields in subscriptions

re #3148
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
+++ b/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
@@ -22,7 +22,7 @@ import {
   NamedTypeNode,
   InputObjectTypeDefinitionNode,
   TypeDefinitionNode,
-  TypeSystemDefinitionNode,
+  ListTypeNode,
 } from 'graphql';
 import {
   ResourceConstants,
@@ -31,6 +31,7 @@ import {
   getBaseType,
   makeDirective,
   makeNamedType,
+  unwrapNonNull,
   makeInputValueDefinition,
   blankObjectExtension,
   extensionWithDirectives,
@@ -347,7 +348,7 @@ export class ModelAuthTransformer extends Transformer {
     const rules = this.getAuthRulesFromDirective(directive);
     // Assign default providers to rules where no provider was explicitly defined
     this.ensureDefaultAuthProviderAssigned(rules);
-    this.validateRules(rules);
+    this.validateRules(rules, def);
     // Check the rules if we've to generate IAM policies for Unauth role or not
     this.setAuthPolicyFlag(rules);
     this.setUnauthPolicyFlag(rules);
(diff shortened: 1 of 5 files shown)


packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1784,15 +1786,28 @@ All @auth directives used on field definitions are performed when the field is r
     ctx.mapResourceToStack(parent.name.value, resolverResourceId);
   }
 
-  private addSubscriptionOwnerArgument(ctx: TransformerContext, resolver: Resolver, ownerRules: AuthRule[], makeNonNull: boolean = false) {
+  private addSubscriptionOwnerArgument(
+    ctx: TransformerContext,
+    resolver: Resolver,
+    currentFields: any,
```
yuth: Why any? From what I can see it should be of type `Record<string,string>`

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1784,15 +1786,28 @@ All @auth directives used on field definitions are performed when the field is r
...
       field => field.name.value === resolver.Properties.FieldName
     ) as FieldDefinitionNode;
-    const nameNode: any = makeNonNull ? makeNonNullType(makeNamedType('String')) : makeNamedType('String');
-    // const createArguments = [makeInputValueDefinition(DEFAULT_OWNER_FIELD, nameNode)];
     const ownerArgumentList = ownerRules.map(rule => {
-      return makeInputValueDefinition(rule.ownerField || DEFAULT_OWNER_FIELD, nameNode);
+      let argNode: any;
+      // if ownerField is defined read from fields if the value exists
+      if (rule.ownerField) {
+        argNode = currentFields[rule.ownerField] ? unwrapNonNull(currentFields[rule.ownerField]) : makeNamedType('String');
```
yuth: when would this be false?

SwaySway: This happens in the event that a ownerField rule is defined and the field is not added such as adding an ownerField named "notAllowed".

packages/graphql-transformer-core/src/util/getFieldArguments.ts
```diff
@@ -4,12 +4,12 @@ import { FieldDefinitionNode } from 'graphql';
  * Given a Type returns a plain JS map of its arguments
  * @param arguments The list of argument nodes to reduce.
  */
-export function getFieldArguments(type: any): any {
+export function getFieldArguments(type: any, withBaseType: boolean = true): any {
```
yuth: can `any` to appropriate types

PR #3170, opened 2020-01-12
feat(graphql-auth-transformer): add `source` @auth type

mdlavin:
This is a proof of concept PR that allows for allowing access to types via changed type resolution. For example, this new @auth rule could allow access to a User object only via a Post.user reference. For example:

```
type User @model @auth(rules: [{allow: source, sourceTypes: ["Post"], operations: [read]}]) {
   name: String
   posts: [Post]
       @connection(name: "PostUser", keyField: "owner")
}

type Post @model {
    id: ID!
    title: String!
    createdAt: String
    updatedAt: String
    user: User
      @connection(name: "PostUser", keyField: "owner")
}
```

When creating a schema that had many connections, I found myself wanted to grant access to a long chain of objects dynamically. Rather than try to use the `owner` feature with a dynamic list that I kept up to date, this approach seemed more flexible.

This PR is functional for my use, but probably lacks the full support for an immediate merge. I'm happy to discuss, change, fix or throw away this work. I'm hoping this is the start of a discussion about the general problem of granting access to deep chains object objects


RossWilliams:
You found a simple and elegant solution to a problem I had as well, I'm definitely adding this to my project! The main downside I see is that a custom create vtl might not set the __typename property, which could confuse users when they try to use the source feature. Renaming sourceType to sourceTypename, or just typename, might make this connection more apparent.

mdlavin:
I just rebased my changes on master and re-tested locally to make sure nothing broke.

@RossWilliams I appreciate your feedback on the implementation. I see your point about having custom vtls that don't include the __typename attribute.  As a user creating the schema, I'm thinking in types so I figured I'd call the attribute `sourceTypes`. Since the value is a string anyway, it would be horrible to call the attribute `sourceTypenames`.

@attilah do you have an opinion about the name of the attribute or anything else about this idea? I've continue to find this feature very useful and I haven't come up with a good alternative

RossWilliams:
just a follow-up, i've been using this for 6 months now and have found it very helpful to evolve my schema to include a major new feature. i've left the argument name 'sourceTypes' as is, and it reads well because i know it is related to the 'source' type.

mdlavin:
Glad it was helpful. Do you know a trick to get some attention from the
Amplify team to hear if they would even consider merging?

PR #3181, opened 2020-01-14
fix(amplify-category-analytics): reverted the hyphen and updated tests

### LINKED ISSUE #3163: `amplify add analytics` successfully with dash in resource name but fail on `amplify push`
**Describe the bug**
When I run `amplify add analytics` and then specify a name with a dash, it accepts it. This should just fail with validation logic because when I go to `amplify push`, it fails with 
```
Template format error: Resource name analyticsanalytics-with-dash is non alphanumeric.
An error occured during the push operation: Template format error: Resource name analyticsanalytics-with-dash is non alphanumeric.
```
**Amplify CLI Version**
`4.12.0`

**To Reproduce**
Steps to reproduce the behavior or terminal output if applicable
1. `amplify init`, I chose iOS
2. `amplify add analytics`
3. Provide your pinpoint resource name: `analytics-with-dash`
4. `amplify push`

**Expected behavior**
Earlier client validation on pinpoint resource name, or provide a way for developer to update the resource name after the failed `amplify push.

**Screenshots**
image

ammarkarachi:
*Issue #, if available:*
fix #3163
*Description of changes:*

The hyphens are not allowed when pushing and had to revert it back


packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
```diff
@@ -12,13 +12,15 @@ describe('amplify add analytics', () => {
 
   afterEach(async () => {
     await removeAnalytics(projRoot, {});
-    deleteProjectDir(projRoot);
+    await deleteProject(projRoot);
+    deleteProjectDir(projRoot), true;
```
kaustavghosh06: Why do we need this?

packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
```diff
@@ -1,4 +1,4 @@
-import { initJSProjectWithProfile } from '../init';
+import { initJSProjectWithProfile, amplifyPushUpdate, deleteProject } from '../init';
```
kaustavghosh06: Is this a part of this PR?

packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
ammarkarachi: Sorry I think I missed that will fix it now

packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
ammarkarachi: Yes, it is to clean up resources

packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
ammarkarachi: Nvm It should be there to clean the folders. I thought I had that in twice

packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
kaustavghosh06: That should be a part of another PR then..doesn't look like an analytics fix

ammarkarachi: It is a part of the test for Analytics

packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
kaustavghosh06: the deleteProjectDir() function is async?
deleteProjectDir(projRoot), true ; ? (What's the true for?)

packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
kaustavghosh06: Nah, deleted my comment.

packages/amplify-e2e-tests/src/__tests__/analytics.test.ts
ammarkarachi: DeleteProjectDir cleans up the folders which is not async. DeleteProject cleans up the resources on the cloud. The true is to delete the deployment bucket

kaustavghosh06: true is outside the function call :|

PR #3206, opened 2020-01-16
fix(cli): alter tsconfig to produce valid souremap locations

kevcam4891:
*Issue #, if available:*
N/A

*Description of changes:*

tsconfig.json file for amplify-cli contained a property `compilerOptions.mapRoot` = “lib”
that would cause `tsc` (version 3.7.5, latest) command to produce

{"version":3,"file":"index.js","sourceRoot":"","sources":["../index.ts"],"names":[],"mappings”:”…”}

Sources *should* read: “../src/index.ts”.

I’m not exactly sure how/why its set up this way, but all other tsconfig files in this project do not specify
a mapRoot, so this change brings this in line with all the other packages, and now I can debug the cli
package again :)

### DIFF
--- a/packages/amplify-cli/tsconfig.json
+++ b/packages/amplify-cli/tsconfig.json
@@ -3,12 +3,10 @@
     "target": "es6",
     "module": "commonjs",
     "sourceMap": true,
-    "mapRoot": "lib",
     "lib": ["es6", "es2015", "dom"],
     "declaration": false,
     "allowJs": true,
     "outDir": "lib",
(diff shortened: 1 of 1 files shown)


attilah:
@kevcam4891 thanks for the PR, we're currently doing a task to move to shared tsconfig settings, so I cannot merge it, because the fixed file will not be present in the near future. After we have the changes, please check if that looks good to you as well, if you think we missed something feel free to open an issue.

Issue #2670, opened 2019-11-03
generating custom variables for aws-exports.js


kkesley-shine:
**Is your feature request related to a problem? Please describe.**

I have a case where I have existing AWS resources and I'd like to integrate it with my React project.

The good thing about amplify is the ability to checkout an environment (e.g. `amplify env checkout dev`) and an appropriate `aws-exports.js` (in case of js project) will be automatically updated.

However, that only works when we create resources using amplify cli. It'd be really great if we can do it for existing resources. 

Example:  I have multiple Appsync API for multiple environments. It'd be great if I can save the API urls somewhere in amplify so it can be picked by my team members when they're checking out one of our amplify envs

**Describe the solution you'd like**

If there's a command to create a custom value for `aws-exports.js`, that'd be great. e.g. `amplify add variables --env dev` or something like that. and when the user executes `amplify push` the variables should appear in `aws-exports.js`

We need a solution to generate the file without pushing the backend resources as well (if no changes in backend, we can use this command). Something like `amplify push --config-only`.

The variables should not clash with existing variables generated by amplify.

**Describe alternatives you've considered**

An alternative that I can think of is to create the file manually and reference it using `process.env`. However, this is inferior compared to amplify's flexibility in swapping environment.

jkeys-ecg-nmsu:
@kkesley-shine the AppSync API endpoint is already a part of the aws-exports.js for each env. There is currently a bug where `env checkout <env>` [does not update aws-exports.js](https://github.com/aws-amplify/amplify-cli/issues/2635).

kkesley-shine:
@jkeys-ecg-nmsu but what if the AppSync API is not created by amplify itself? I have 2 existing AppSync API (made manually). How can I include the API endpoint in the `aws-exports.js` in a way that if I checkout to `dev` env, it will point to `appsync A` and if I checkout to `prod` env, it will point to `appsync B` ?

jkeys-ecg-nmsu:
@kkesley-shine you would add each AppSync environment to their respective Amplify environments with `amplify api add-graphql-datasource`. [See this documentation](https://aws-amplify.github.io/docs/cli-toolchain/graphql#how-does-this-work) for an explanation of how it works.

kkesley-shine:
@jkeys-ecg-nmsu I tried `amplify api add-graphql-datasource` but it seems to only support Aurora? Am I missing something here?

[...]

PR #3218, opened 2020-01-20
feat: support arbitary command hooks

hirochachacha:
Current implementation support hooks for only two commands - `amplify init` and `amplify push`.
It's not good enough for plugin developers.

For example, if someone want to checkout own configuration file per environment,
(like https://github.com/aws-amplify/amplify-cli/issues/2670)
they probably need to handle not only `amplify init`, but also `amplify
pull`, `amplify env add`, `amplify env checkout` and `amplify env pull`.

This CL allows to implement such features.

### DIFF
--- a/packages/amplify-cli/src/domain/amplify-event.ts
+++ b/packages/amplify-cli/src/domain/amplify-event.ts
@@ -1,20 +1,14 @@
-export enum AmplifyEvent {
+export type AmplifyEvent = string;
+
+export class AmplifyEventData {}
+
+export class AmplifyEventArgs {
+  constructor(public event: AmplifyEvent, public data?: AmplifyEventData) {}
+}
+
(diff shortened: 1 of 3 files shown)


UnleashedMind:
The Amplify CLI plugin platform's event handling workflow is not complete yet and is currently under actively developments. We are very cautious about adding events to ensure that the CLI is stable and efficient. 
If this feature is desirable now, we would recommend fork our CLI and add the changes to the fork.

hirochachacha:
I see. What I wanted to say is the current hook system isn't good enough to solve several problems.
But I'd rather wait it than make a forked version. Actually I'am maintaining a forked version for #3219, previously I have #2719 and #2899 because there're no alternatives. Those experience are just painful, I don't want more. Thank you for your response anyway.

Issue #3064, opened 2019-12-22
Not Authorized to access field on nested type (without model) using IAM


jonmifsud:
**Describe the bug**
When you have a model, that contains nested data from other types, the Authorization rules in the resolvers seem to get wonky in terms of who is authorized to view what.

Data is accessible via the default cognito access, when using Group or Private however IAM functions in Lambda get all the field values replaced by nulls on the request and recieve an error when trying to post.

**To Reproduce**

1. Create a schema which has `@auth` directives including IAM and nested types
2. Create a lambda function to query and/or mutate the model
3. Fill up some data via a cognito user.
4. Data within the nested element returns null values.

-- or --

    type Product @model
      @auth(rules: [
        { allow: private, provider:userPools, operations: [read]},
        { allow: groups, groups: ["Admin"], operations: [create, update, read, delete]}
        { allow: private, provider: iam, operations: [create, update, read, delete] }
      ])
     {
      id: ID!
      availableSlots: [AvailableSlots]
     }

    type AvailableSlots {
      date: AWSDate
      times: [AvailableHours]
    }

    type Mutation {
      bookSlot(date: AWSDateTime): String @function(name: "bookViewing-${env}")
    }

**Expected behavior**
I expect the nested types to inherit the same permissions as the other fields since I did not set different permissions on these fields.

</details>

grigull:
I seem to have run into a similar problem, but mine is even more extreme. The below example might help solve the root of the problem.

`type User @model (subscriptions: { level: off })`
  `@key(fields: ["id"])`
  `@auth (rules: [`
  `{allow: public, provider: iam, operations: [read]},`
 `{allow: private, provider: userPools, operations: [read, create, update, delete]},`
  `{allow: private, provider: iam, operations: [create, read, update, delete]}])`
  `{`
    `id: ID!`
    `userTags: [UserTags]`
  `}`

`type UserTags {`
  `id: String`
  `name: String`
`}`

With the above schema the cognito user can successfully access the userTags attribute for the user, but for iam role (lambda function in this case) they get the following error: `GraphQL error: Not Authorized to access id on type UserTags
GraphQL error: Not Authorized to access name on type UserTags`

jonmifsud:
@attilah / @kaustavghosh06 - let me know if there's any way at all I could help in getting a fix for this. We're possibly blocked from going live with this issue as it's likely going to impact us at multiple points as the solution we are working on does require quite a bit of lambda processing.

[...]

PR #3223, opened 2020-01-20
fix (graphql-auth-transformer): handle @auth propagation for non-model nested types

attilah:
*Issue #, if available:*

Fixes: #3064 for all auth providers

*Description of changes:*

This PR will handle ```@aws_*``` auth directive propagation for nested types which ones doesn't have a ```@model``` attribute in the schema. In case of a ```private``` or ```public``` auth rule for ```iam``` provider the appropriate policy entry will be added to the policy as well.


cle5eland:
Is this fix supposed to recursively add the @aws_* tag to all nested types (and their nested types)? I am only seeing one level of propogation @attilah

amcdnl:
@cle5eland - yup I'm seeing it only doing 1 level too. I would assume it would propagate down.

PR #3233, opened 2020-01-22
fix(amplify-graphql-types-generator): generate array return type

### LINKED ISSUE #2478: Angular Codegen not generating Queries that return Arrays as an Array object.
**Describe the bug**
When running codegen on a custom Query, it generates the type of the function as QueryObject, rather than Array<QueryObject>

**To Reproduce**

1. Create a new Angular project with Amplify, with api and a function.
2. Add a new Query like below:
```
type Query {
	testQuery:[QueryObject] @function(name: "testfunction-${env}") 
}

type QueryObject {
	test: String
}
```
3. Perform Angular codegen

**Expected behavior**
function would generate as
`async TestQuery(): Promise<Array<TestQueryQuery>>`

**Actual behavior**
Function generates as `async TestQuery(): Promise<TestQueryQuery>`

nikhname:
fix #2478

*Issue #, if available:*
#2478
*Description of changes:*
Add array block around return type if it is a GraphQLList


WolfWalter:
Does not work for non null lists  `arrayTest: [QueryObject]!`

PR #3235, opened 2020-01-22
chore(cli): improve error message in readJsonFile

hirochachacha:
*Issue #, if available:*

I'm tired to see useless error messages like `SyntaxError: Unexpected token } in JSON at position 6`

This CL improves error messaage like following:

```
JSONParseError: some.json:3:2: Unexpected token } in JSON at position 6
```

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/read-json-file.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/read-json-file.js
@@ -1,4 +1,5 @@
 const fs = require('fs-extra');
+const parseJson = require('parse-json');
 
 function stripBOM(content) {
   if (content.charCodeAt(0) === 0xfeff) {
@@ -8,7 +9,8 @@ function stripBOM(content) {
 }
 
 function readJsonFile(jsonFilePath, encoding = 'utf8') {
-  return JSON.parse(stripBOM(fs.readFileSync(jsonFilePath, encoding)));
+  const content = fs.readFileSync(jsonFilePath, encoding);
+  return parseJson(stripBOM(content), jsonFilePath);
 }
 
 module.exports = {
--- a/packages/amplify-cli/package.json
+++ b/packages/amplify-cli/package.json
@@ -68,11 +68,13 @@
     "lodash": "^4.17.15",
     "open": "^7.0.0",
     "ora": "^4.0.3",
+    "parse-json": "^5.0.0",
     "promise-sequential": "^1.1.1",
     "which": "^2.0.2"
   },
   "devDependencies": {
-    "@types/node": "^10.17.13"
+    "@types/node": "^10.17.13",
+    "@types/parse-json": "^4.0.0"
   },
   "jest": {
     "transform": {
--- a/packages/amplify-cli/src/utils/readJsonFile.ts
+++ b/packages/amplify-cli/src/utils/readJsonFile.ts
@@ -1,7 +1,7 @@
 import fs from 'fs-extra';
+import parseJson from 'parse-json';
 
 function stripBOM(content: string) {
-  // tslint:disable-next-line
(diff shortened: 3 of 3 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/read-json-file.js
```diff
@@ -1,4 +1,7 @@
 const fs = require('fs-extra');
+const os = require('os');
+
+const posPattern = /at position (\d+)$/;
```
SwaySway: This seems a bit hard coded here, are we sure we'll always expect this kind of pattern?

packages/amplify-cli/src/extensions/amplify-helpers/read-json-file.js
```diff
@@ -7,8 +10,37 @@ function stripBOM(content) {
   return content;
 }
 
+class JSONParseError extends Error {
```
SwaySway: I would recommend looking at an managed library as opposed to adding a custom JSONParseError class.

hirochachacha:
@SwaySway 
> View our contributing guidelines here: https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md

I wasn't aware that guide.
Probably https://github.com/aws-amplify/amplify-cli#contributing shadows that guide.

packages/amplify-cli/src/extensions/amplify-helpers/read-json-file.js
hirochachacha: Implementation level: Yes

https://github.com/nodejs/node/blob/32e7e813e93ec3d0625d315658288769e448dd99/deps/v8/src/common/message-template.h#L424-L427
https://github.com/nodejs/node/blob/32e7e813e93ec3d0625d315658288769e448dd99/deps/v8/src/json/json-parser.cc#L251-L266

Specification level: No, I guess not.

I think providing line number and column number is ideal, but providing only filename is still good enough. If we want to show line and column seriously, we need to use or create custom JSON parser like https://www.npmjs.com/package/jsonlint.

packages/amplify-cli/src/extensions/amplify-helpers/read-json-file.js
hirochachacha: Could you give me more specific suggestion?
I grepped `packages/amplify-cli` and I saw there is only one custom error class definition.
Are you suggesting that throwing generic error is good enough like `throw new Error("message")`?

SwaySway: [parse-json](https://www.npmjs.com/package/parse-json) would be a good implementation here.

hirochachacha: It looks the same idea.
https://github.com/sindresorhus/parse-json/blob/master/index.js
https://github.com/zkat/json-parse-better-errors/blob/latest/index.js
Anyway, I will.

hirochachacha: Done

hirochachacha:
By the way, how can I run test inside `packages/amplify-cli/tests` ? Perhaps, the test setting is broken.
Also I don't see defining utility functions on amplify context is the best idea. There're some places we cannot use the context. I can confirm that by `find . -name '*.js' -or -name '*.ts' | grep -v 'node_modules' | grep -v 'chunk.js' | xargs grep 'JSON.parse'`.
Furthermore, `ini` and `yaml` need equal treatment. To define new package for utility functions might be better solution.

kaustavghosh06:
@hirochachacha Could you please resolve the merge conflicts? This is ready to be merged.

hirochachacha:
@kaustavghosh06 @SwaySway PTAL

Issue #392, opened 2018-11-04
Adding validation to fields in graphql api schema


rayhaanq:
** Which Category is your question related to? **
Amplify GraphQL API

Is it possible to add validation to fields in the schema file? So min/max length, min/max value etc? If I had a field on a Person type and wanted to limit their name field to 250 characters, is that possible using the schema file only? Otherwise how can this be done?

mikeparisstuff:
We don't package any transformers that allow you to do this out of the box but have discussed use cases like this in the past. If you want to do this purely using the `schema.graphql` today, you can write your own transformer that will augment the generated resolver template code. To make this work you will also need to write your own light-weight CLI that loads and runs the transformer. There are docs on this here https://aws-amplify.github.io/docs/js/graphql#writing-custom-transformers.

The way to implement this in AppSync is to write resolver code that validates inputs during mutations. For example, if you wanted to validate that the "name" is a max of 250 characters then you would add a snippet like this to createPerson and updatePerson resolvers:

```vtl
## mutation.create/updatePerson.request.vtl **
#if( $ctx.args.input.name.length() > 250 )
    $util.error("Invalid name on Create/UpdatePersonInput.")
#end
```

If you provided a name longer than 250 characters this would cause the mutation to fail before saving the object to DynamoDB. This is how your custom transformer would work too, but you would be generating the above snippet in your transformer instead of writing it by hand.

mikeparisstuff:
What directives would you like access to? Over the long term, we can work on building a few transformers that tackle common issues but it would be helpful to hear more about the use cases you want to be solved.

For example, I would suspect these two would be useful:

**@regex**

```graphql
directive @regex(expression: String!) on FIELD_DEFINITION
```

Usage

```graphql
type Post {
  id: ID!
  title: String @regex(expression: ".*graphql.*")
}
```

**@length**

```graphql
directive @length(min: Int, max: Int) on FIELD_DEFINITION
```

```graphql
type Post {
  id: ID!
  title: String @length(min: 10, max: 255)
}
```

There are a lot of possibilities so let me know what you think and we will prioritize based on demand.

rayhaanq:
@mikeparisstuff  thanks for the code snippet. I'll take a look at this. As for the possible directives. I guess what you mentioned sounds reasonable with the addition of maybe:
- @equals - equality check which would work for all/most types like int, ID, string etc
- @notEqual - not equal to
- @in - a set of values that it can be
[...]

PR #3236, opened 2020-01-22
fix(amplify-provider-awscloudformation): custom transformer imports

hirochachacha:
Current implementation misuses createRequireFromPath API.
Unfortunately this API doesn't support directory path well.

For example,

```
createRequireFromPath("/my/project")("./node_modules/mylib")
```

isn't same as

```
require("/my/project/node_modules/mylib")
```

Actually, it's

```
require("/my/node_modules/mylib")
```

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
+++ b/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
@@ -2,6 +2,8 @@ const fs = require('fs-extra');
 const path = require('path');
 const chalk = require('chalk');
 const inquirer = require('inquirer');
+const importGlobal = require('import-global');
+const importFrom = require('import-from');
 const { DynamoDBModelTransformer } = require('graphql-dynamodb-transformer');
 const { ModelAuthTransformer } = require('graphql-auth-transformer');
 const { ModelConnectionTransformer } = require('graphql-connection-transformer');
@@ -76,6 +78,7 @@ function getTransformerFactory(context, resourceDir, authConfig) {
         // The loading of transformer can happen multiple ways in the following order:
         // - modulePath is an absolute path to an NPM package
         // - modulePath is a package name, then it will be loaded from the project's root's node_modules with createRequireFromPath.
+        // - modulePath is a name of a globally installed package
         let importedModule;
         const tempModulePath = modulePath.toString();
 
@@ -84,11 +87,19 @@ function getTransformerFactory(context, resourceDir, authConfig) {
             // Load it by absolute path
             importedModule = require(modulePath);
           } else {
-            // 'require' from project path
             const projectRootPath = context.amplify.pathManager.searchProjectRootPath();
-            const { createRequireFromPath } = require('module');
-            const projectRequire = createRequireFromPath(path.join(projectRootPath, 'noop.js')); // createRequireFromPath doesn't support directory well. we need to add noop file. See https://github.com/nodejs/node/issues/23710 for detail
-            importedModule = projectRequire(tempModulePath);
+            const projectNodeModules = path.join(projectRootPath, 'node_modules');
+
+            try {
+              importedModule = importFrom(projectNodeModules, modulePath);
+            } catch (_) {
+              // Intentionally left blank to try global
+            }
+
+            // Try global package install
+            if (!importedModule) {
+              importedModule = importGlobal(modulePath);
+            }
           }
 
           // At this point we've to have an imported module, otherwise module loader, threw an error.
--- a/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
+++ b/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
@@ -76,7 +76,6 @@ function getTransformerFactory(context, resourceDir, authConfig) {
         // The loading of transformer can happen multiple ways in the following order:
         // - modulePath is an absolute path to an NPM package
         // - modulePath is a package name, then it will be loaded from the project's root's node_modules with createRequireFromPath.
-        // - modulePath is a name of a globally installed package
         let importedModule;
         const tempModulePath = modulePath.toString();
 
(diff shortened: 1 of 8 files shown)


packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -80,33 +80,18 @@ function getTransformerFactory(context, resourceDir, authConfig) {
...
           if (path.isAbsolute(tempModulePath)) {
             // Load it by absolute path
             importedModule = require(modulePath);
           } else {
+            // 'require' from project path
             const projectRootPath = context.amplify.pathManager.searchProjectRootPath();
             const { createRequireFromPath } = require('module');
-            const projectRequire = createRequireFromPath(projectRootPath);
-
-            if (tempModulePath.startsWith('./')) {
```
hirochachacha: It seems that there is no concept like `global require` and `local require`.
`module.paths` is calculated by caller's path and `NODE_PATH`.
https://stackoverflow.com/questions/15636367/nodejs-require-a-global-module-package

hirochachacha:
I think this issue is important one. It's difficult to add custom transformers without this change

attilah:
@hirochachacha could you please elaborate on what is NOT working with the current implementation?

hirochachacha:
@attilah I tried to use my custom transformer.

1. `yarn add -D graphql-auto-transformer`
2. edit `amplify/backend/api/<YOUR_API>/transform.conf`
```
    "transformers": [
      "./graphql-auto-transformer"
    ]
```
3. `amplify-dev api gql-compile`

Then I saw 
```
Error: Cannot find module 'graphql-auto-transformer'
```

The following code is a short summary of current implementation.
```
require('module').createRequireFromPath(__dirname)("./graphql-auto-transformer")
```
You can check that this line of code cannot load `graphql-auto-transformer` from `node_modules`.

 And next one is proposed solution.
```
require('module').createRequireFromPath(`${__dirname}/noop.js`)('graphql-auto-transformer');
```

hirochachacha:
https://github.com/nodejs/node/blob/dcba12895ad58275ba5b027c2c5110461dfebf66/lib/internal/modules/cjs/loader.js#L1263-L1268

hirochachacha:
@attilah is it still unclear?

sbue:
Any update on this?

andreialecu:
@attilah it seems that you attempted to fix: https://github.com/aws-amplify/docs/pull/992#issuecomment-557298978, as per: https://github.com/aws-amplify/docs/pull/992#issuecomment-558239064.

However, it doesn't seem to work for locally installed packages.

For clarity here's what happens on `amplify-cli 4.18.0`:

**transform.conf.json**
```json
{
  "Version": 5,
  "ElasticsearchWarning": true,
  "transformers": ["graphql-auto-transformer"]
}
```

`amplify push` output:

image

There is a workaround here: https://github.com/hirochachacha/graphql-auto-transformer/issues/2

However, that adds additional maintenance complexity and comes with its own set of issues.

hirochachacha:
@attilah I still don't understand why you want to ignore this issue. Maybe you don't think this is an issue because you've already tested custom imports feature? https://github.com/amazon-archives/aws-reinvent-2019-mobile-workshops/tree/master/MOB405
If so, that's pretty wrong.

https://github.com/amazon-archives/aws-reinvent-2019-mobile-workshops/blob/master/MOB405/Starting%20from%20scratch.md#add-validator-transformer-to-list-of-custom-transformers-in-amplifybackendapivalidatortesttransformconfjson

```
"transformers": [
    "./reinvent-string-validator"
]
```
You expect that "reinvent-string-validator" module will be loaded from 
"<project_dir>/node_modules/reinvent-string-validator" right?

But in fact, it will be loaded from _"<project_dir>/../reinvent-string-validator"_ 
That's the reason you didn't see this issue from your tests. The directory structure concealed it.

andreialecu:
@hirochachacha would there be a way to provide the full path in the config?

For example:

```json
"transformers": [
    "../node_modules/graphql-auto-transformer"
]
```

My various attempts have been unsuccessful.

hirochachacha:
@andreialecu you mean, except using NODE_PATH?
```
cd <your_project>/..
ln -s <your_project>/node_modules/graphql-auto-transformer
```

It emulates directory structure of https://github.com/amazon-archives/aws-reinvent-2019-mobile-workshops/tree/master/MOB405
So it should work.

hirochachacha:
Using absolute path should work also.

andreialecu:
I got it, the following works without linking:

```
    "transformers": [
        "./projectname/node_modules/graphql-auto-transformer"
    ]
```

It's not ideal having to hardcode the root folder name in it though.

hirochachacha:
Yep, I hope these work-arounds prove that this is, obviously, an issue.

attilah:
@hirochachacha thanks for the PR and for finding this issue. I tested the proposed solution with your published package and it failed with globally installed packages (when using NVM) and if the package was referenced with ```./``` prefix in ```transform.conf.json```.

The [```resolve-from```](https://github.com/sindresorhus/resolve-from/blob/master/index.js#L27) package also uses the same "trick" for ```noop.js``` what you applied, I was not aware of that before.

I did some refactoring by leveraging ```import-global``` and ```import-from``` packages and it worked for me in all the test cases.

@hirochachacha, @andreialecu, @sbue, test my commit if it works for your use cases and to make sure this time we get it right.

@hirochachacha as previously we asked, please open an issue before opening a PR to be able to discuss the given issue and prioritize it better.

hirochachacha:
@attilah 

> @hirochachacha thanks for the PR and for finding this issue. I tested the proposed solution with your published package and it failed with globally installed packages (when using NVM) and if the package was referenced with ./ prefix in transform.conf.json.

That's probably working as intended.
Look, we don't have to distinguish local import and global import. local import covers global one.
You can confirm this by printing `module.paths`.
```
[ '/Users/hiroshiioka/work/myproject/node_modules',
  '/Users/hiroshiioka/work/node_modules',
  '/Users/hiroshiioka/node_modules',
  '/Users/node_modules',
  '/node_modules',
  '/Users/hiroshiioka/.node_modules',
  '/Users/hiroshiioka/.node_libraries',
  '/Users/hiroshiioka/.nvm/versions/node/v10.18.1/lib/node' ]
```

As you can see, it includes both local and global paths.

So, I'd like to change dot import as usual one to support project specific transformer like:

https://github.com/aws-amplify/amplify-cli/issues/392#issuecomment-458292457

Hope it's clear now.

attilah:
@hirochachacha it is environment dependent, by default you don't get the nvm one, in my setup I don't have it, here is the ```require.main.paths``` when the ```gql-compile``` running:

```javascript
[
  "<root>/amplify-cli/packages/amplify-cli/bin/node_modules",
  "<root>/amplify-cli/packages/amplify-cli/node_modules",
  "<root>/amplify-cli/packages/node_modules",
  "<root>/amplify-cli/node_modules",
  "<root>/node_modules",
  "<home>/node_modules",
  "/Users/node_modules",
  "/node_modules",
]
```

That's the reason we need the additional global import.

hirochachacha:
@attilah I see. That's interesting. I thought you're talking about special handling for './' in the previous code without reading the PR. Sorry about that. Now it looks like it's good although I didn't test it yet. Thank you for the fix.

attilah:
@hirochachacha if you encounter a problem just open an issue.

attilah:
@yuth please review my changes in this PR.

hirochachacha:
But I just wonder why do you need third party libraries? Even though https://github.com/hirochachacha/amplify-cli/commit/e1621a831447ea2eae73606c1db7ddc5779036f7 failed to load global libraries, pulling back https://github.com/hirochachacha/amplify-cli/commit/e1621a831447ea2eae73606c1db7ddc5779036f7#diff-13054d37b1e2418fc273d5352c4bcd76L105-L108 wasn't good enough?

hirochachacha:
@attilah @yuth Wait, this's still wrong. https://github.com/sindresorhus/import-from doesn't follow node's module paths, it only resolve specific path. Besides, it doesn't distinguish `graphql-auto-transformer` and `./graphql-auto-transformer`, so there're no chance to accomplish https://github.com/aws-amplify/amplify-cli/issues/392#issuecomment-458292457
Again, I'd suggest to stick with `createRequireFromPath` or its successor `createRequire` for providing maximum functionalities to users.

attilah:
@hirochachacha this change supports:
- import module/js file by absolute path
- import package from current project's node_modules (import-from), package name can be specified with './' prefix or without it
- import package from global packages

What should ./ accomplish in addition to this? if you specify ./ that was meant for current project's node_modules. What am I missing?

PR #3242, opened 2020-01-22
test(amplify-e2e-tests): use waitFor on bucket deletion check

SwaySway:
- use waitFor when checking that bucket is deleted
- removed unnecessary logging from predictions test

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/delete.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/delete.test.ts
@@ -13,7 +13,6 @@ import {
 import { addEnvironment } from '../environment/add-env';
 import { addApiWithoutSchema } from '../categories/api';
 import { addCodegen } from '../codegen/add';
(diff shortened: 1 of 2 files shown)


packages/amplify-e2e-tests/src/__tests__/delete.test.ts
```diff
@@ -86,3 +85,20 @@ async function bucketExists(bucket: string) {
...
+  const s3 = new S3();
+  const params = {
+    Bucket: bucket,
+    $waiter: { maxAttempts: 10 },
+  };
+  try {
+    await s3.waitFor('bucketNotExists', params).promise();
+    return true;
+  } catch (error) {
+    if (error.statusCode === 200) {
```
nikhname: Maybe return false after checking the bucket exists? and that the 10 tries did not fail for some other reason

PR #3306, opened 2020-01-30
fix(amplify-velocity-template): #Return directive called several times

bboure:
*Issue #, if available:*
No issue exists

*Description of changes:*
When the same resolver is called several times and a `#return` directive is in the template, it stops working the second time.
This is because the same [`Compiler`](https://github.com/aws-amplify/amplify-cli/blob/6795e783c104004a2b2576f6903b35c1c6d2ed03/packages/amplify-appsync-simulator/src/velocity/index.ts#L34) is being reused and `state.stop` is not being reset before the following `render()` is called.


attilah:
@bboure nice catch, it looks good to me, but we need @yuth's approval.

PR #3309, opened 2020-01-31
fix(amplify-provider-awscloudformation): fix a bug in headless pull

UnleashedMind:
fix ##3292

*Issue #, if available:*

*Description of changes:*
this fixes the bug for headless pull when config level is set to 'general'. 

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/attach-backend.js
+++ b/packages/amplify-provider-awscloudformation/lib/attach-backend.js
@@ -15,7 +15,7 @@ async function run(context) {
   const amplifyClient = await getConfiguredAmplifyClient(context, awsConfig);
   if (!amplifyClient) {
     // This happens when the Amplify service is not available in the region
-    const region = awsConfig && awsConfig.region ? awsConfig.region : 'unknown';
+    const region = awsConfig && awsConfig.region ? awsConfig.region : '<unknown>';
     const message = `Amplify service is not available in the region ${region}`;
     context.print.error(message);
     throw new Error(message);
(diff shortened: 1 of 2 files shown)


packages/amplify-provider-awscloudformation/lib/attach-backend.js
```diff
@@ -15,7 +15,8 @@ async function run(context) {
   const amplifyClient = await getConfiguredAmplifyClient(context, awsConfig);
   if (!amplifyClient) {
     // This happens when the Amplify service is not available in the region
-    const message = `Amplify service is not available in the region ${awsConfig.region ? awsConfig.region : ''}`;
+    const region = awsConfig && awsConfig.region ? awsConfig.region : '';
+    const message = `Amplify service is not available in the region ${region}`;
```
attilah: I think we should handle the empty region you are assigning in the previous line. Either throw if that's unexpected or use '<unknown' instead of ''.

yarn.lock
```diff
@@ -6780,7 +6780,7 @@ debug@3.1.0, debug@=3.1.0:
   dependencies:
```
attilah: Please remove the yarn.lock file from this PR as no dependencies were touched.

UnleashedMind: removed

packages/amplify-provider-awscloudformation/lib/attach-backend.js
UnleashedMind: this part of the code is only for printing out a more informative error message, error is thrown immediately afterwards.
In the "general" config level, region is not set, that's why it is missing, the aws-sdk will resolve creds and region from environment variables or other means [doc](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-region.html). So the region is not necessarily unknown, it's just the codebase here does not have knowledge of how the aws-sdk might have resolved it.

UnleashedMind: changed

Issue #3359, opened 2020-02-06
Cannot stream DDB entries with same HashKey/different SortKey to ES without overwriting existing ES entry


kevcam4891:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[Amplify Console GitHub Issue Tracker](https://github.com/aws-amplify/amplify-console/issues)

**Describe the bug**
Saving two objects that share the same HashKey lead to one overwriting the other object in ElasticSearch.

**Amplify CLI Version**
4.12.0

**To Reproduce**
Save one item sort of like:
```
mutate({
          variables: {
            input: {
              id: props.taskId,
              sk: "subtype1"
              data: "foo!"
            }
          }
})
```
This entry is successfully saved to ElasticSearch. Then try and save another object with the same HashKey, but different SortKey
```
mutate({
          variables: {
            input: {
              id: props.taskId,
              sk: "subtype2"
              data: "foo!"
            }
          }
})
```
The first saved entry is overwritten by the second saved entry. Two entries in DB, but only the second entry in ES.

**Expected behavior**
I would expect the lambda function to send over two object with disambiguated doc_id values, resulting in two entries in ElasticSearch.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: [e.g. Mac/Windows/Ubuntu]
 - Node Version. You can use `node -v` to check the node version on your system
v12.11.1

**Additional context**

kevcam4891:
It appears as though the ID is calculated as the "id" as it is supplied in the mutation, if one is supplied, and only if one doesn't exist, will it use the Hash AND Sort keys (as provided in the stream event) to create a more unique id in ES. I can't really see a reason to NOT just use the Hash/Sort keys in the stream event, as this code puts you in an un-winnable position if you need the sort key to disambiguate.

I can "fix" this if you have any recommendations for how the logic should flow here.
```
doc_id = doc_fields['id'] if 'id' in doc_fields else compute_doc_index(ddb['Keys'], ddb_deserializer)
```

kevcam4891:
If I were to change this code, I would

1. prefer ddb['Keys'] if they existed, then
2. prefer doc_fields['id'], then
3. throw exception if neither exist

Is this reasonable?

kevcam4891:
Meanwhile, I've submitted a PR to allow user to override amplify-generated functions (above).

PR #3367, opened 2020-02-06
feat(graphql-transformer-core): allow user overrides for functions

kevcam4891:
There are instances where it is helpful to be able to override a top-level api function.
Example: /amplify/backend/[category]/[resource]/functions includes
ElasticSearchStreamingLambdaFunction.zip, a helper lambda function to stream DDB
data into ElasticSearch. Issue #3359 has caused me to want to replace that function with
a similar function that contains a minor, but important change. This enhancement now
allows for this, and joins other overrides the user can affect (ex: resolvers, stacks, schema).

*Issue #, if available:*
#3359 

*Description of changes:*

### DIFF
--- a/packages/graphql-transformer-core/src/util/amplifyUtils.ts
+++ b/packages/graphql-transformer-core/src/util/amplifyUtils.ts
@@ -18,6 +18,7 @@ export interface ProjectOptions {
   currentCloudBackendDirectory: string;
   rootStackFileName?: string;
   dryRun?: boolean;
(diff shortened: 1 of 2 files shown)


kevcam4891:
@SwaySway Any chance we can get this pulled soon? I'd like this override ability to give me and potentially other developers peace of mind that we can adjust the ElasticSearchToDDDB script (or other lambdas in the future) as we see fit or if some sort of regression creeps into it.

PR #3395, opened 2020-02-10
feat: cliplugin enable and publish functionality

garyleefight:
Implemented amplify add hosting && amplify hosting publish for both CICD and manual categories.

### DIFF
--- a/packages/amplify-console-hosting/event-handlers/handle-PostInit.js
+++ b/packages/amplify-console-hosting/event-handlers/handle-PostInit.js
@@ -1,0 +1,10 @@
+const eventName = 'PostInit';
+
+async function run(context, args) {
+  // insert your code to handle the amplify cli PostInit event
+  context.print.info(`Event handler ${eventName} to be implemented.`);
+}
+
+module.exports = {
+  run,
+};
--- a/packages/amplify-console-hosting/event-handlers/handle-PostPush.js
+++ b/packages/amplify-console-hosting/event-handlers/handle-PostPush.js
@@ -1,0 +1,10 @@
+const eventName = 'PostPush';
+
+async function run(context, args) {
+  // insert your code to handle the amplify cli PostPush event
+  context.print.info(`Event handler ${eventName} to be implemented.`);
+}
+
+module.exports = {
+  run,
+};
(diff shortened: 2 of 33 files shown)


packages/amplify-console-hosting/utils/path-manager.js
```diff
@@ -0,0 +1,40 @@
+const path = require('path');
```
kaustavghosh06: Why do you need to re-mplement this file and not use `context.amplify` instead?

packages/amplify-console-hosting/package.json
```diff
@@ -0,0 +1,25 @@
+{
```
kaustavghosh06: Can you please take a look at the package json's of other packages in this repo and replicate the scripts?

packages/amplify-console-hosting/event-handlers/handle-PostInit.js
```diff
@@ -0,0 +1,10 @@
+const eventName = 'PostInit';
```
kaustavghosh06: Do you need this file?

packages/amplify-console-hosting/event-handlers/handle-PostPush.js
```diff
@@ -0,0 +1,10 @@
+const eventName = 'PostPush';
```
kaustavghosh06: Do you need this file?

kaustavghosh06:
Please also add E2E tests for this feature.

packages/amplify-console-hosting/utils/path-manager.js
garyleefight: Caused I wants template path and other paths that are not in context.amplify to be reused across packages. Just in consistence, wrap all native path-manager API I used in this utils code as well

packages/amplify-console-hosting/package.json
garyleefight: Will change

packages/amplify-console-hosting/event-handlers/handle-PostInit.js
garyleefight: No I don't use any of them. Can be deleted

packages/amplify-console-hosting/event-handlers/handle-PostPush.js
garyleefight: Same as above

PR #3406, opened 2020-02-12
fix(amplify-provider-awscloudformation): fix template not found

RossWilliams:
When pushing resources, previously the code would search for cloudformation templates in the build/ folder, if the build folder exists.
This was intended for use by the API resource, but has the unintended effect of causing templates to not be found in other resource folders,
which may contain build folders but still hold their template file in the root resource directory.

To reproduce this issue, create a function and inside the function resource directory create a build/ folder.
When pushing, the cloudformation file will not be found and not uploaded to S3.

*Issue #, if available:*

*Description of changes:*

The resolve the issue, if a cloudformation file is not found inside the build folder, continue to search for templates in the root directory.

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/push-resources.js
+++ b/packages/amplify-provider-awscloudformation/lib/push-resources.js
@@ -93,7 +93,7 @@ async function updateStackForAPIMigration(context, category, resourceName, optio
   const { resourcesToBeCreated, resourcesToBeUpdated, resourcesToBeDeleted, allResources } = await context.amplify.getResourceStatus(
     category,
     resourceName,
-    providerName
+    providerName,
   );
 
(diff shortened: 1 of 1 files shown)


goku-ssj3:
Hi, i have a similar issue. I initialized a new AWS Amplify project inside my React Native project and added just 1 category(Add Auth). However, when i do a 'Amplify push' i get following error:

"There was an error pulling the backend environment 'dev', an error occurred during the push operation. EEXIST: file already exists, mkdir '..projectDirectory/amplify/.temp/#current-cloud-backend'"

Can anyone please help me how to resolve this issue?

RossWilliams:
@goku-ssj3 i don’t think this is related. But you can delete the .temp folder and try again.

PR #3419, opened 2020-02-13
fix(graphql-connection-transformer): validate composite sortkey

yuth:
added validation around composite sort key when used with fields. The validation prevents passing of
partial sort key to connection directives field argument

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
+++ b/packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
@@ -100,7 +100,8 @@ function checkFieldsAgainstIndex(
   parentFields: ReadonlyArray<FieldDefinitionNode>,
   relatedTypeFields: ReadonlyArray<FieldDefinitionNode>,
   inputFieldNames: string[],
-  keySchema: KeySchema[]
+  keySchema: KeySchema[],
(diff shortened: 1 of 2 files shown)


packages/graphql-connection-transformer/src/ModelConnectionTransformer.ts
```diff
@@ -113,8 +114,16 @@ function checkFieldsAgainstIndex(
   if (numFields > keySchema.length && keySchema.length !== 2) {
     throw new InvalidDirectiveError('Too many fields passed in to @connection directive.');
   }
+  // when sort key is passed, ensure that the length of composite sort key matches the length of the fields passed
+  if (numFields > 1) {
+    const querySortFields = inputFieldNames.slice(1);
+    const tableSortFields = String(keySchema[1].AttributeName).split(ModelResourceIDs.ModelCompositeKeySeparator());
+    if (querySortFields.length !== tableSortFields.length) {
+      throw new InvalidDirectiveError(`Invalid @connection directive  ${field.name.value}. fields does not accept partial sort key`);
```
attilah: I think we could present a better error message here. What length was expected on which field and how many was passed in.

PR #3472, opened 2020-02-19
chore: e2e test failure for trigger tests, update verdaccio to latest

attilah:
*Description of changes:*

E2E tests were failing in circleci because after spawning Verdaccio the package registry values were not reset, so when a function was deployed and invoked ```npm install``` it failed to connect to the previously set Verdaccio repository url. This fix resets the urls after Amplify CLI install, so function deployment happens as normal.


attilah:
Requested follow up issue: https://github.com/aws-amplify/amplify-cli/issues/3473

PR #3488, opened 2020-02-20
style(amplify-appsync-simulator): more readable AppSync mock logging

visusnet:
Visually parsing lots of subscription logs from the AppSync simulator is difficult. This commit
makes logs more readable.


bboure:
Thank you @visusnet 

Probably out of scope here, but I was wondering if we could add some (opt-in?) logs arounds [here](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-appsync-simulator/src/resolvers/unit-resolver.ts#L38) in order to debug (visualize) easily requests and responses from and to the data loaders. 

`appsync-emulator-serverless` [used to do that](https://github.com/ConduitVC/aws-utils/blob/master/packages/appsync-emulator-serverless/schema.js#L309) and it was rather useful. 

cc: @yuth

visusnet:
@bboure I agree, this is out-of-scope. ... But I'm intrigued. Maybe that's something I could do in another PR.

How could/would/should opt-in work? Even though this PR improves readability, I would love to add something which gives me the option to hide/show most of the stuff.
Maybe something like this?
```bash
DEBUG_LEVEL=info amplify mock
```

bboure:
@visusnet Thank you for your feedback. I guess using a env var would do the trick.

This is in fact [how they did it](https://github.com/ConduitVC/aws-utils/blob/cc487ec5a13906da7b351337adcab9517818bd48/packages/appsync-emulator-serverless/log.js#L8) in the former emulator.
Adding a level of logs would be very useful too I guess. 

I would be happy to open a PR as well if I find some time.

PR #3495, opened 2020-02-23
feat: uplevel enabling of datastore and update of auth configs to top

kaustavghosh06:
*Description of changes:*
New amplify update api flow:

f45c89966b0d:datstore kaustavg$ amplify-dev update api
? Please select from one of the below mentioned services: GraphQL
? Select from the options below (Use arrow keys)
❯ Walkthrough all configurations 
  Update auth settings 
  Enable DataStore for entire API 

### DIFF
--- a/packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
+++ b/packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
@@ -107,11 +107,9 @@ async function serviceWalkthrough(context, defaultValuesFilename, serviceMetadat
 
   // Ask additonal questions
 
-  /* eslint-disable */
   ({ authConfig, defaultAuthType } = await askDefaultAuthQuestion(context, parameters));
   ({ authConfig, resolverConfig } = await askAdditionalQuestions(context, parameters, authConfig, defaultAuthType));
   await checkForCognitoUserPools(context, parameters, authConfig);
-  /* eslint-disable */
 
   // Ask schema file question
 
@@ -355,7 +353,9 @@ async function updateWalkthrough(context) {
   const { allResources } = await context.amplify.getResourceStatus();
   let resourceDir;
   let resourceName;
-  let authConfig, defaultAuthType, resolverConfig;
+  let authConfig;
+  let defaultAuthType;
+  let resolverConfig;
   const resources = allResources.filter(resource => resource.service === 'AppSync');
 
   // There can only be one appsync resource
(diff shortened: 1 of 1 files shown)


kaustavghosh06:
@yuth Yes, we should have a check for it but I would decouple this from control plane settings.  It should a part of the sanity check we do as a part of the gql-compile step and not when setting control plane configs - since the schema's are changed (based on which we should throw this error) and the only execution during a push is gql-compile.

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -401,35 +401,73 @@ async function updateWalkthrough(context) {
...
+  let { updateOption } = await inquirer.prompt([updateOptionQuestion]);
 
-  if (amplifyMeta[category][resourceName].output.securityType) {
-    delete amplifyMeta[category][resourceName].output.securityType;
+  if (updateOption === 'enableDatastore') {
+    resolverConfig = {
+      project: { ConflictHandler: 'AUTOMERGE', ConflictDetection: 'VERSION' },
+    };
+  } else if (updateOption === 'authUpdate') {
+    /* eslint-disable */
```
yuth: Lets avoid disabling eslint rules inside documents

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -401,35 +401,73 @@ async function updateWalkthrough(context) {
...
+    ({ authConfig, defaultAuthType } = await askDefaultAuthQuestion(context, parameters));
+    authConfig = await askAdditionalAuthQuestions(context, parameters, authConfig, defaultAuthType);
+    await checkForCognitoUserPools(context, parameters, authConfig);
+    /* eslint-disable */
```
yuth: same

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
```diff
@@ -401,35 +401,73 @@ async function updateWalkthrough(context) {
...
-  const backendConfig = context.amplify.readJsonFile(backendConfigFilePath);
+    if (amplifyMeta[category][resourceName].output.securityType) {
+      delete amplifyMeta[category][resourceName].output.securityType;
+    }
 
-  if (backendConfig[category][resourceName].output.securityType) {
-    delete backendConfig[category][resourceName].output.securityType;
-  }
+    amplifyMeta[category][resourceName].output.authConfig = authConfig;
+    let jsonString = JSON.stringify(amplifyMeta, null, '\t');
```
yuth: Are we using tabs in all the places? We should stick to one, and I see in most of places, we use 4 spaces

kaustavghosh06: Done.

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
kaustavghosh06: Removed.

packages/amplify-category-api/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.js
kaustavghosh06: Removed.

PR #3534, opened 2020-02-27
For Auth, add PreventUserExistenceErrors parameter and update NodeJS runtime version

beomseoklee:
Issue: #4214

*Description of changes:*
According to the [Amazon Cognito documentation](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pool-managing-errors.html), if ```PreventUserExistenceErrors``` parameter is enabled automatically after February 15th 2020, but it's somehow not through the AWS CloudFormation template.

For the security best practice, this option needs to be enabled, so I've added ```PreventUserExistenceErrors``` parameter.

In addition to the parameter, I've updated NodeJS runtime to the latest version for AWS Lambda functions.


packages/amplify-category-auth/provider-utils/awscloudformation/cloudformation-templates/auth-template.yml.ejs
```diff
@@ -367,7 +369,7 @@ Resources:
             - ' }'
             - '};'
       Handler: index.handler
-      Runtime: nodejs10.x
```
kaustavghosh06: Did you try pushing this template and testing it? 
I believe Cloudformation doesn't support in-line lambda's with Node 12.x runtime.

beomseoklee: CloudFormation does support Lambda Node 12.x runtime. [Here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-runtime) is the documentation.

kaustavghosh06: This is for the Lambda functions - not for the in-line custom resources which you define in your cloudformation file. Have you tried pushing this template successfully?

beomseoklee: What do you mean "for the Lambda functions"? If you mean I was able to create those Lambda functions with Node 12.x runtime, like I said, CloudFormation supports Lambda Node 12.x runtime, so the answer is yes.
image

Not for ```OpenIdLambda``` because I have not figured out how to create that one yet.

edwardfoyle (approved):

edwardfoyle:
@beomseoklee Had to revert this due to an odd error happening in one of our tests. We'll figure out what's going on with the test and merge this again.

This is the failed test: https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/2542/workflows/bbbb0bed-b0a4-4a0d-96c1-8cb6ba2f3739/jobs/36430

We need to change the expected error message [here](https://github.com/aws-amplify/amplify-cli/blob/18f04d1fc0a1e3c630160d1d21f20f7c61e3c108/cypress/integration/auth_spec.js#L13) to "Incorrect username or password" but when I tried to update the test to that, for some reason the error we're getting is "Only radix 2, 4, 8, 16, 32 are supported" likely related to [this issue](https://github.com/aws-amplify/amplify-js/issues/4430) just need to dig into why it's happening

PR #3578, opened 2020-03-04
test(cli): fix the console integration test

UnleashedMind:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-console-integration/src/pullAndInit/amplifyConsoleOperations.ts
+++ b/packages/amplify-console-integration/src/pullAndInit/amplifyConsoleOperations.ts
@@ -64,6 +64,20 @@ export async function createConsoleApp(projectName: string, amplifyClient?: any)
   return createAppResponse.app.appId;
 }
 
+export async function deleteConsoleApp(appId: string, amplifyClient?: any) {
+  if (!amplifyClient) {
+    amplifyClient = getConfiguredAmplifyClient();
+  }
+  const deleteAppParams = {
+    appId,
+  };
+  try {
+    await amplifyClient.deleteApp(deleteAppParams).promise();
+  } catch (err) {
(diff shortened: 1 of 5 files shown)


packages/amplify-console-integration/src/util.ts
```diff
@@ -22,7 +22,7 @@ export function getCLIPath() {
 }
 
 export function getProfileName() {
-  return 'amplify-console-integration-test';
+  return 'amplify-integ-test-user';
```
kaustavghosh06: Shouldn't this be "default"?

UnleashedMind: I'm just using the same profile name we are using for all the other e2e tests. 
And it does not matter which string we use, the profile will be created if it does not existing. 
Since we don't want to create another profile, I'm just using the same name that the other tests are using. 
check the amplify-e2e-tests package

nikhname: In CI amplify-integ-test-user is always the profile name configured, this is right

packages/amplify-console-integration/src/pullAndInit/amplifyConsoleOperations.ts
```diff
@@ -64,6 +64,20 @@ export async function createConsoleApp(projectName: string, amplifyClient?: any)
...
+  if (!amplifyClient) {
+    amplifyClient = getConfiguredAmplifyClient();
+  }
+  const deleteAppParams = {
+    appId,
+  };
+  try {
+    await amplifyClient.deleteApp(deleteAppParams).promise();
+  } catch (err) {
+    // Do nothing
```
nikhname: Do we want to output the error if deleteApp fails?

PR #3596, opened 2020-03-05
chore(cli): remove stale esm cache postinstall

yuth:
updated post install script to remove the stale esm cache

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-cli/scripts/post-install.js
+++ b/packages/amplify-cli/scripts/post-install.js
@@ -1,8 +1,17 @@
-require = require('esm')(module);
-const fs = require('fs-extra');
+require = require('esm')(module, { cache: false });
 const chalk = require('chalk');
+const fs = require('fs-extra');
+const path = require('path');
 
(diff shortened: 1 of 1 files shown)


packages/amplify-cli/scripts/post-install.js
```diff
@@ -1,8 +1,17 @@
...
-const fs = require('fs-extra');
+require = require('esm')(module, { cache: false });
 const chalk = require('chalk');
+const fs = require('fs-extra');
+const path = require('path');
 
-console.log(chalk.bold('Ignore installation errors for optional dependencies: gyp, fs-ext'));
+// Delete stale ESM cache
+try {
+  const cachePath = path.join(__dirname, '..', 'node_modules', '.cache', 'esm');
```
kaustavghosh06: Shouldn't it just be - /node_modules/.cache/?

yuth: esm cache is inside .esm folder. There might be other caches which I would not want to touch unless I know what it is

Issue #3509, opened 2020-02-25
S3 Storage User group List Bucket Permission Bug


wongcyrus:
The permission is not enough to list bucket.
https://stackoverflow.com/questions/38774798/accessdenied-for-listobjects-for-s3-bucket-when-permissions-are-s3
Addition permission block has to be added for list Object.
```json
      {
          "Effect": "Allow",
          "Action": [
              "s3:ListBucket"
          ],
          "Resource": [
              "arn:aws:s3:::bucketname"
          ]
      }
```
https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-category-storage/provider-utils/awscloudformation/cloudformation-templates/s3-cloudformation-template.json.ejs

      <% if (props.groupList) { %>
      <% for(var i=0; i < props.groupList.length; i++) { %>
	    "<%= props.groupList[i] %>GroupPolicy": {
	      "Type": "AWS::IAM::Policy",
	      "Properties": {
	        "PolicyName": "<%= props.groupList[i] %>-group-s3-policy",
	        "Roles": [
	        	{
                    "Fn::Join": [
                        "",
                        [
                          {
                            "Ref": "auth<%= props.authResourceName%>UserPoolId"
                          },
                          "-<%= props.groupList[i] %>GroupRole"
                        ]
                    ]
                }
	        ],
	        "PolicyDocument": {
	          "Version": "2012-10-17",
	          "Statement": [
	            {
	              "Effect": "Allow",
	              "Action": <%- JSON.stringify(props.groupPolicyMap[props.groupList[i]]) %>,
	              "Resource": [
	                  {
	                      "Fn::Join": [
	                          "",
	                          [
	                              "arn:aws:s3:::",
	                              {
	                                  "Ref": "S3Bucket"
	                              },
	                              "/*"
	                          ]
	                      ]
	                  }
	              ]
	            }
	          ]
	        }
	      }
	    },
      <% } %>
      <% } %>

SwaySway:
Hello @wongcyrus If you are referring to listing all objects in a bucket it's related to how the CLI sets up a storage. View more on file access levels here: https://aws-amplify.github.io/docs/js/storage

wongcyrus:
Hi Sway,

I know how to use the storage CLI, but the actual bug I report is under the situation.
1. Add user group 
2. Update permission for User group to access S3 Storage.
3. It adds permission to the role for the group.
[...]

PR #3612, opened 2020-03-07
fix(amplify-category-function): adds policy for list Bucket for user groups

akshbhu:
*Issue #3509 

*Description of changes:*
* added policy specific to Cognito user groups for List Bucket operation

### DIFF
--- a/packages/amplify-category-storage/provider-utils/awscloudformation/service-walkthroughs/s3-walkthrough.js
+++ b/packages/amplify-category-storage/provider-utils/awscloudformation/service-walkthroughs/s3-walkthrough.js
@@ -22,7 +22,7 @@ async function addWalkthrough(context, defaultValuesFilename, serviceMetadata, o
   while (!checkIfAuthExists(context)) {
     if (
       await context.amplify.confirmPrompt.run(
-        'You need to add auth (Amazon Cognito) to your project in order to add storage for user files. Do you want to add auth now?'
+        'You need to add auth (Amazon Cognito) to your project in order to add storage for user files. Do you want to add auth now?',
       )
     ) {
       try {
@@ -125,7 +125,7 @@ async function configure(context, defaultValuesFilename, serviceMetadata, resour
             type: 'list',
             choices: inputs[i].options,
           },
-          question
+          question,
         );
       } else if (inputs[i].type && inputs[i].type === 'multiselect') {
         question = Object.assign(
@@ -133,14 +133,14 @@ async function configure(context, defaultValuesFilename, serviceMetadata, resour
             type: 'checkbox',
             choices: inputs[i].options,
           },
-          question
+          question,
         );
       } else {
         question = Object.assign(
           {
             type: 'input',
           },
-          question
+          question,
         );
       }
       questions.push(question);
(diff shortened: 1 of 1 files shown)


packages/amplify-category-storage/provider-utils/awscloudformation/service-walkthroughs/s3-walkthrough.js
```diff
@@ -22,7 +22,7 @@ async function addWalkthrough(context, defaultValuesFilename, serviceMetadata, o
   while (!checkIfAuthExists(context)) {
     if (
       await context.amplify.confirmPrompt.run(
-        'You need to add auth (Amazon Cognito) to your project in order to add storage for user files. Do you want to add auth now?'
+        'You need to add auth (Amazon Cognito) to your project in order to add storage for user files. Do you want to add auth now?',
```
kaustavghosh06: Why do we need the extra ,?

packages/amplify-category-storage/provider-utils/awscloudformation/service-walkthroughs/s3-walkthrough.js
```diff
@@ -168,7 +168,7 @@ async function configure(context, defaultValuesFilename, serviceMetadata, resour
       if (permissionSelected === 'Learn more') {
         context.print.info('');
         context.print.info(
-          'You can restrict access using CRUD policies for Authenticated Users, Guest Users, or on individual Groups that users belong to in a User Pool. If a user logs into your application and is not a member of any group they will use policy set for “Authenticated Users”, however if they belong to a group they will only get the policy associated with that specific group.'
+          'You can restrict access using CRUD policies for Authenticated Users, Guest Users, or on individual Groups that users belong to in a User Pool. If a user logs into your application and is not a member of any group they will use policy set for “Authenticated Users”, however if they belong to a group they will only get the policy associated with that specific group.',
```
kaustavghosh06: Same as above.

packages/amplify-category-storage/provider-utils/awscloudformation/service-walkthroughs/s3-walkthrough.js
```diff
@@ -544,6 +544,28 @@ async function updateCfnTemplateWithGroups(context, oldGroupList, newGroupList,
       },
     };
   });
+  // added a new policy for the user group to make action on buckets
+  newGroupList.forEach(group => {
```
kaustavghosh06: I don't think this is the right way to go about this. Which "CRUD" operation is this IAM policy tied to? In this implementation, it seems we are adding it to all of the "CRUD" operations.

akshbhu: In the above we are adding permissions for objects not on buckets when we do bucketname/*. 
In this to use S3::listbuckets operation , we need IAM policy on bucket.

kaustavghosh06: We should associate this policy only if the the selection has "Read" as a part of the CRUD flow.

akshbhu: Do you mean this line
```javascript
NewGroupPolicyMap[group].includes('s3:ListBucket') === true
```
Here I am checking if any action is there on bucket , I am adding the policy then.

kaustavghosh06: Ah! got it, somehow I missed it.

PR #3626, opened 2020-03-09
fix: makes hosting plugin version same with CLI

garyleefight:
Passed hosting plugin integration test locally


packages/amplify-console-hosting/package.json
```diff
@@ -10,21 +10,20 @@
...
+        "open": "^7.0.0",
+        "ora": "^4.0.3",
         "request": "^2.88.0",
         "request-promise": "^4.2.4"
     },
     "devDependencies": {
-        "eslint": "^4.19.1",
+        "eslint": "^6.8.0",
         "eslint-config-airbnb-base": "^12.1.0",
-        "eslint-plugin-import": "^2.12.0"
```
UnleashedMind: "eslint-plugin-import": "^2.20.0",

PR #3641, opened 2020-03-10
feat:usage-diagnoses

ammarkarachi:
*Issue #, if available:*

*Description of changes:*


packages/amplify-cli/src/domain/telemetry.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.SessionUuid = uuid.v4();
+    this.Url = url.parse('https://aws-amplify-cli-telemetry.us-east-1.amazonaws.com/metrics');
+  }
+
+  async emit(input: Input, error: Error | null, state: String, waitForSend: boolean = false) {
+    const payload = new TelemetryPayload(this.SessionUuid, this.InstallationUuid, this.Version, input, error, state);
+    waitForSend ? await this.send(payload) : this.send(payload);
+  }
+
+  async send(payload: TelemetryPayload) {
```
attilah, line 26: This method should never fail, we've to guard against possible exceptions, it would be nice to have tests for error scenarios to make sure we are not failing normal operations when usage data reporting fails for any reason (network conditions, etc).

packages/amplify-cli/src/domain/telemetry.ts
```diff
@@ -0,0 +1,85 @@
...
+      req.on('error', () => {});
+      req.write(data);
+      req.end(() => {
+        resolve();
+      });
+    });
+  }
+}
+
+export const WorkFlowState = {
```
attilah, line 48: Please use an enum.

packages/amplify-cli/src/domain/telemetry.ts
```diff
@@ -0,0 +1,85 @@
...
+  Successfull: 'SUCCEEDED',
```
attilah, line 49: typo: `Successful`, but if we need the literal values on the backend, then call the enum members the same.

packages/amplify-cli/src/domain/telemetry.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.name = error.name;
+  }
+}
+
+class TelemetryPayload {
+  sessionUuid: String;
+  installationUuid: String;
+  amplifyCliVersion: String;
+  input: Input | null;
+  timeStamp: String;
```
attilah, line 67: `timestamp`

packages/amplify-cli/src/domain/telemetry.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export const WorkFlowState = {
+  Successfull: 'SUCCEEDED',
+  Invoke: 'INVOKED',
+  Aborted: 'ABORTED',
+  Failed: 'FAILED',
+};
+
+class SerializableError {
```
attilah, line 55: Is it intentional that we don't keep  the message part of the error? What information will we get from failures then? no stacktrace either?

packages/amplify-cli/src/domain/telemetry.ts
```diff
@@ -0,0 +1,85 @@
...
+    this.sessionUuid = sessionUuid;
+    this.installationUuid = installationUuid;
+    this.amplifyCliVersion = version;
+    this.input = redactInput(input, true);
+    this.timeStamp = new Date().toISOString();
+    this.osPlatform = os.platform();
+    this.osRelease = os.release();
+    this.nodeVersion = process.versions.node;
+    this.state = state;
+    if (error) this.error = new SerializableError(error);
```
attilah, line 83: Please use { } blocks and new lines even for single statements for readibility.

packages/amplify-cli/src/index.ts
```diff
@@ -48,11 +50,15 @@ export async function run() {
         throw new Error(verificationResult.message);
       }
     }
-
     rewireDeprecatedCommands(input);
     const context = constructContext(pluginPlatform, input);
+    await attachTelemetry(context);
+    errorHandler = boundErrorHandler.bind(context);
+    process.on('SIGINT', sigIntHandler.bind(context));
```
attilah, line 57: `SIGINT` does not work under Windows, we should follow the docs and use the `readline` module to capture `SIGINT` under Windows: https://nodejs.org/api/readline.html#readline_event_sigint

packages/amplify-cli/src/index.ts
```diff
@@ -48,11 +50,15 @@ export async function run() {
         throw new Error(verificationResult.message);
       }
     }
-
     rewireDeprecatedCommands(input);
     const context = constructContext(pluginPlatform, input);
+    await attachTelemetry(context);
+    errorHandler = boundErrorHandler.bind(context);
```
attilah, line 56: Why do we need `bind` to context instead of calling `context.telemetry.emitXXXX methods? what is the advantage?

packages/amplify-cli/src/telemetry/config.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export class Config {
+  installationUuid: String;
+  isUsageTrackingEnabled: boolean;
+  constructor() {
+    this.installationUuid = uuid.v4();
+    this.isUsageTrackingEnabled = true;
+  }
+
+  static get(path: string): Config {
```
attilah, line 12: As we agreed on that this config should be in an amplify level configuration file with a `usageData` section the reading/writing of that file should not be a telemetry concern but shared code should be used for that.

packages/amplify-cli/src/telemetry/getConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { Config } from './config';
+import os from 'os';
+import { Context } from '../domain/context';
+import * as path from 'path';
```
attilah, line 4: `import path from 'path'`

packages/amplify-cli/src/telemetry/getConfig.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function getConfig(context: Context): Config {
```
attilah, line 6: As per out discussion the file should be: `$HOME/.amplify/<executable>-configuration.json`.

packages/amplify-cli/src/telemetry/index.ts
```diff
@@ -0,0 +1 @@
+export { getConfig } from './getConfig';
```
attilah, line 1: It would be nice to export everything that is accessible from this folder in index to not to import files directly at other parts of the CLI.

packages/amplify-provider-awscloudformation/lib/configuration-manager.js
```diff
@@ -24,7 +24,7 @@ async function init(context) {
     config: defaultAWSConfig,
   };
   await newUserCheck(context);
-  printInfo(context);
+  await printInfo(context);
```
attilah, line 27: why the `await`?

packages/amplify-cli/scripts/post-install.js
```diff
@@ -12,18 +13,30 @@ try {
 } catch (e) {
   // could not delete the cache directory but don't want to fail the installation
 }
-console.log('\n');
+console.log(EOL);
```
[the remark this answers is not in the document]
ammarkarachi: I wanted to make sure it was platform independent

packages/amplify-cli/src/config-steps/c0-analyzeProject.js
```diff
@@ -44,8 +44,19 @@ async function configureProjectName(context) {
       projectName = answer.inputProjectName;
     }
   }
+  const projectConfig = { projectName };
+  if (context.parameters.options) {
+    if (context.parameters.options['usage-data-on']) {
+      Object.assign(projectConfig, { usageDataDisabled: true });
```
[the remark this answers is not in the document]
ammarkarachi: I agree I will change that flag

.circleci/config.base.yml
```diff
@@ -2,7 +2,7 @@ version: 2.1
 machine:
   environment:
     PATH: '${PATH}:${HOME}/${CIRCLE_PROJECT_REPONAME}/node_modules/.bin'
-
+    DISABLE_USAGE_TRACKING: true
```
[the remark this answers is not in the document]
ammarkarachi: I didn't want Amplify CLI to pick up the customers CI flags, that is why the it's a separate flag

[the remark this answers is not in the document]
attilah: I was referring to `AMPLIFY_CLI_ENABLE_USAGE_DATA` that we already have to override usage data behavior.

packages/amplify-cli/src/index.ts
ammarkarachi: I need the context to be present when it fails on SIGINT

packages/amplify-cli/src/context-manager.ts
```diff
@@ -2,15 +2,26 @@ import { Context } from './domain/context';
...
-
   return context;
 }
 
+export async function attachTelemetry(context: Context) {
+  const { AMPLIFY_CLI_ENABLE_USAGE_DATA } = process.env;
+  const config = getConfig(context);
+  const usageTrackingEnabled = AMPLIFY_CLI_ENABLE_USAGE_DATA ? AMPLIFY_CLI_ENABLE_USAGE_DATA === 'true' : config.isUsageTrackingEnabled;
+  if (usageTrackingEnabled) {
+    context.telemetry = new Telemetry(config.installationUuid, getVersion(context));
```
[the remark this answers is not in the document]
ammarkarachi: I agree that's a good idea

packages/amplify-cli/src/domain/telemetry.ts
ammarkarachi: Legal didn't want us to keep any identifiable information

packages/amplify-cli/src/index.ts
ammarkarachi: https://github.com/nodejs/node-v0.x-archive/issues/5054

ammarkarachi: I think this has been solved

packages/amplify-cli/src/domain/telemetry.ts
ammarkarachi: There is swagger validation on these field, Let's review the rest of it and then I can work on completing it

packages/amplify-cli/scripts/post-install.js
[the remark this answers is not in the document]
attilah: I meant that how is it different from an empty `console.log()` statement. We *want* 2 empty lines or what is the reason?

packages/amplify-cli/src/index.ts
attilah: Were you able to test it under Windows that it works?

packages/amplify-cli/src/domain/telemetry.ts
attilah: If it is capital 'S' in swagger then that needs to be changed as well.

packages/amplify-cli/scripts/post-install.js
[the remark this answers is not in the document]
ammarkarachi: Oh I am sorry I misunderstood, it's just to keep separation more obvious

packages/amplify-cli/src/domain/telemetry.ts
ammarkarachi: I updated it in Swagger too, this is resolved.

packages/amplify-cli/src/index.ts
ammarkarachi: This works on powershell
```js
process.on( "SIGINT", function() {
  console.log( "shutting down" );
  process.exit();
} );

setInterval( () => { }, 2500 );
```

packages/amplify-cli/src/__tests__/telemetry.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    scope.post(pathToUrl, () => true).reply(200, genereateLongString());
+    await checkTelemetry();
+  });
+
+  it('test delay', async () => {
+    scope.post(pathToUrl, () => true).delay(10000);
+    await checkTelemetry();
+  });
+});
+function genereateLongString() {
```
attilah, line 82: string has a method for this: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat

`'1'.repeat(140)`

packages/amplify-cli/src/app-config/config.ts
```diff
@@ -0,0 +1,52 @@
...
+  return getConfig();
+}
+
+export function getConfig() {
+  return Config.Instance;
+}
+
+export function write(context: Context, keyValues: Object) {
+  Config.Instance.setValues(keyValues);
+  fs.writeFileSync(getPath(context), JSON.stringify(Config.Instance));
```
attilah, line 24: I think this can fail if `~/.amplify` directory does not exist, not?

packages/amplify-cli/src/domain/amplify-telemetry/Telemetry.ts
```diff
@@ -0,0 +1,83 @@
...
+        port: this.url.port,
+        path: this.url.path,
+        method: 'POST',
+        headers: {
+          'content-type': 'application/json',
+          'content-length': data.length,
+        },
+      });
+      req.on('error', () => {});
+      req.setTimeout(100, () => {
```
attilah, line 67: It would be nice to move out the timeout into a const value in this class for greater visibility. The 100 will be enough for cross-region and slow connections as well?

packages/amplify-cli/src/domain/amplify-telemetry/VersionManager.ts
```diff
@@ -0,0 +1,23 @@
+import semver from 'semver';
+
+const APIVersionToPayloadVersion = new Map<string, Array<string>>([['v1.0', ['1.0.0']]]);
```
attilah, line 3: What is the design here? a given CLI version just reports 1 version of the payload what we change over time. So the CLI does not have to support multiple versions.

On the other hand it seems that the `getLatestApiVersion()` is doing the version calculation on every call which seems to be suboptimal as this code is on the hot path if telemetry is enabled, if you decide to keep the multiple versions here, can we cache it between invocations?

packages/amplify-cli/src/domain/amplify-telemetry/getTelemetryUrl.ts
```diff
@@ -0,0 +1,10 @@
+import url, { UrlWithStringQuery } from 'url';
+import { getLatestApiVersion } from './VersionManager';
+
+const version = getLatestApiVersion();
+const prodUrl = `https://aws-amplify-cli-telemetry.us-east-1.amazonaws.com/${version}/metrics`;
```
attilah, line 5: It seems that the payload version and the API version for telemetry is coupled together?

packages/amplify-cli/src/domain/amplify-telemetry/getTelemetryUrl.ts
```diff
@@ -0,0 +1,10 @@
...
+export function getUrl(): UrlWithStringQuery {
```
attilah, line 6: Are we caching the result of this function between invocations?

packages/amplify-cli/src/domain/amplify-telemetry/Telemetry.ts
```diff
@@ -0,0 +1,83 @@
...
+    this.input = new Input([]);
+  }
+
+  init(installationUuid: String, version: String, input: Input): void {
+    this.installationUuid = installationUuid;
+    this.version = version;
+    this.input = redactInput(input, true);
+  }
+
+  static get Instance(): ITelemetry {
```
attilah, line 30: Why do we need static instance variable in the `ITelemetry` implementations?

This only accessed through `context`, which already has an instance of those classes.

packages/amplify-cli/src/domain/amplify-telemetry/NoTelemetry.ts
```diff
@@ -0,0 +1,23 @@
...
+  }
+  emitAbort(): Promise<void> {
+    return Promise.resolve();
+  }
+  emitSuccess(): Promise<void> {
+    return Promise.resolve();
+  }
+  init(installationUuid: String, version: String, input: any): void {}
+
+  private static instance: NoTelemetry;
```
attilah, line 18: Same as for the other `ITelemetry` implementation.

packages/amplify-cli/src/__tests__/telemetry.test.ts
```diff
@@ -0,0 +1,100 @@
...
+import { getUrl } from '../domain/amplify-telemetry/getTelemetryUrl';
+import { Input } from '../domain/input';
+
+const baseOriginlUrl = 'https://cli.amplify';
+const pathToUrl = '/metrics';
+const originalUrl = `${baseOriginlUrl}${pathToUrl}`;
+
+describe('test telemetry', () => {
+  beforeAll(() => {
+    process.env = Object.assign(process.env, { AMPLIFY_CLI_BETA_USAGE_TRACKING_URL: originalUrl });
```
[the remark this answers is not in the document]
ammarkarachi: What do you mean by testing?

packages/amplify-cli/src/app-config/config.ts
ammarkarachi: ~/.amplify the directory is created during a plugin scan.

packages/amplify-cli/src/domain/amplify-telemetry/Telemetry.ts
ammarkarachi: We are not waiting on any response from the server. if the server does not return in that allotted time we don't wan't the cli to hang when it is exiting.

packages/amplify-cli/src/domain/amplify-telemetry/getTelemetryUrl.ts
ammarkarachi: No we are not caching it. It's kind of excessive to cache something that's been called twice in a lifetime when the software is running

packages/amplify-cli/src/domain/amplify-telemetry/getTelemetryUrl.ts
ammarkarachi: Yes the payload version is coupled together with the API version

packages/amplify-cli/src/domain/amplify-telemetry/VersionManager.ts
ammarkarachi: I want the API version to be mapped to payload version. This way it's explicit if there are any major changes we would have to make sure the API works with the new payload. No we are not caching it. It's kind of excessive to cache something that's been called twice in a lifetime

packages/amplify-cli/src/domain/amplify-telemetry/Telemetry.ts
ammarkarachi: This is how a singleton is created

packages/amplify-cli/src/domain/amplify-telemetry/NoTelemetry.ts
ammarkarachi: Same as telemetry

packages/amplify-cli/src/__tests__/telemetry.test.ts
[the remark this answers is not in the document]
ammarkarachi: We shouldn't be exposing a beta url to the public also.

PR #3669, opened 2020-03-13
fix(amplify-category-notifications): add apns sandbox channel

### LINKED ISSUE #3247: AWS Pinpoint Campaign testing push notifications with APNS Sandbox does not start working after `amplify add notifications`
**Describe the bug**
When I enable APNS push notifications with key certificate via `amplify add notifications`, it appears to be enabled, however when I try to create a campaign and send a test message to my endpointId, it fails with `Channel Disabled`. The endpoint is set up with a device token for APNS sandbox. When I check settings, push notification is enabled. I checked the amplify-cli code and it looks like it's hitting `updateApnsChannel`. When using AWS Console to add the key certificate, it then works (device receives the notification) when I send test messages or launch the campaign.

**Amplify CLI Version**
`4.12.0`

**To Reproduce**
1. `amplify add analytics` and `amplify push`
2. `amplify add notifications`
```
 amplify notifications add
? Choose the push notification channel to enable. APNS
? Choose authentication method used for APNs Key
? The bundle id used for APNs Tokens:  <Your App's BundleID like com.yourname.projectname>
? The team id used for APNs Tokens:  XXXXXXXXX
? The key id used for APNs Tokens:  ABCDEXXXXX
? The key file path (.p8):  AuthKey_ABCDEXXXXX.p8
✔ The APNS channel has been successfully enabled.
```
3. Make sure an app is registers for push notifications and then registers the token with APNS, and registers an endpoint with device token and APNS Sandbox with pinpoint.
4. Make sure that single endpoint is picked up when choosing a segment when creating a campaign. Try sending a notification using Test message to the endpointId (can be printed out from pinpoint sdk logs), and fails with some message (AWS Console sometimes isn't very verbose, may be easier to test using AWS CLI but I don't have that set up)
5. `amplify console analytics` and go to the Settings/Push Notifications/Edit, and under APNS re-add the key certificate
6. Try creating a campaign/test message and it works!

I suspect that going through the console, it may be making calls to both [update-apns-sandbox-channel](https://docs.aws.amazon.com/cli/latest/reference/pinpoint/update-apns-sandbox-channel.html) and [update-apns-channel](https://docs.aws.amazon.com/cli/latest/reference/pinpoint/update-apns-channel.html)

**Expected behavior**
`amplify add notifications` should allow us to specify apns sandbox or apns production since the key is used for both channels when Pinpoint sends authorization data to APNS. If I have to go to AWS Pinpoint Console to enable calls for APNS Sandbox, that kind of defeats the purpose of using amplify CLI for this step in the development. I think a solution could be by default; make a request to pinpoint to update both apns sandbox and production since that's what the key is used for. And ask `Choose APNS environment: [Production|Sandbox|Both]` and Both could be the default

**Additional context**
Some assumptions i've made
- AWS Console/Pinpoint console/Enable notifications/APNS/Add key flow adds it to both APNS sandbox and production. I've only tested that my app starts receiving notifications after using the console, and amplify-cli call doesn't enable it for me since i'm using APNS sandbox.
- Have not verified with AWS CLI to directly call update APNS sandbox and tested that notifications start working after that. Just know that using `amplify add notifications` results in not working

UnleashedMind:
APNS Sandbox channel will be enabled/disabled when regular APNS channel is enabled/disabled

fix #3247

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-category-notifications/__tests__/lib/channel-APNS.test.js
+++ b/packages/amplify-category-notifications/__tests__/lib/channel-APNS.test.js
@@ -10,18 +10,56 @@ const channelAPNS = require('../../lib/channel-APNS');
 describe('channel-APNS', () => {
   const mockServiceOutput = {};
   const mockChannelOutput = { Enabled: true };
+  mockServiceOutput[channelName] = mockChannelOutput;
+
   const mockPinpointResponseErr = {};
   const mockPinpointResponseData = {
     APNSChannelResponse: {},
   };
   const mockKeyConfig = {};
   const mockCertificateConfig = {};
+
   const mockPinpointClient = {
-    updateApnsChannel: jest.fn((_, callback) => {
-      callback(null, mockPinpointResponseData);
+    updateApnsChannel: jest.fn(() => {
+      return {
+        promise: () => {
+          return new Promise((resolve, reject) => {
+            resolve(mockPinpointResponseData);
+          });
+        },
+      };
+    }),
+    updateApnsSandboxChannel: jest.fn(() => {
+      return {
+        promise: () => {
+          return new Promise((resolve, reject) => {
+            resolve(mockPinpointResponseData);
+          });
+        },
+      };
     }),
   };
-  mockServiceOutput[channelName] = mockChannelOutput;
+
+  const mockPinpointClientReject = {
+    updateApnsChannel: jest.fn(() => {
+      return {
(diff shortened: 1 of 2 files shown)


lawmicha:
thanks for picking this up, i have verified this is working for me and have updated the docs
https://github.com/aws-amplify/docs/pull/1275

PR #3698, opened 2020-03-18
test: add test for custom stacks

jhockett:
Test custom stack flow with SES

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -769,6 +769,13 @@ jobs:
           when: always
       - store_artifacts:
           path: ../uitest_android_results
+  customStacks-amplify_e2e_tests:
+    working_directory: ~/repo
+    docker: *ref_0
+    resource_class: large
+    steps: *ref_1
+    environment:
+      TEST_SUITE: src/__tests__/customStacks.test.ts
   plugin-amplify_e2e_tests:
     working_directory: ~/repo
     docker: *ref_0
@@ -961,6 +968,7 @@ workflows:
             - graphql_e2e_tests
             - integration_test
             - amplify_console_integration_tests
+            - customStacks-amplify_e2e_tests
             - plugin-amplify_e2e_tests
             - datastore-modegen-amplify_e2e_tests
             - interactions-amplify_e2e_tests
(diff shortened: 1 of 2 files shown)


kaustavghosh06:
@yuth could you please glance at this PR as well?

jhockett:
> Why do we need a separate step in CircleCI for this?

I'm pretty sure the CircleCI changes were generated since this is a new file in the ```__tests__``` folder.

.circleci/config.yml
```diff
@@ -982,14 +990,18 @@ workflows:
...
-      - plugin-amplify_e2e_tests:
+      - customStacks-amplify_e2e_tests:
           filters: &ref_2
             branches:
               only:
                 - master
                 - split-test
           requires:
             - publish_to_local_registry
+      - plugin-amplify_e2e_tests:
```
kaustavghosh06, line 1001: Why do have this? Isn't this unrelated?

yuth:
> Why do we need a separate step in CircleCI for this?



> > Why do we need a separate step in CircleCI for this?
> 
> I'm pretty sure the CircleCI changes were generated since this is a new file in the `__tests__` folder.

Yes. This is done to split tests into separate job in CircleCI

.circleci/config.yml
yuth: This is diffing algorithm doing its thing

packages/amplify-e2e-tests/src/__tests__/customStacks.test.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    const meta = getProjectMeta(projRoot);
+
+    const { DeploymentBucketName: bucketName, Region: region } = meta.providers.awscloudformation;
+    expect(bucketName).toBeDefined();
+    expect(region).toBeDefined();
+    const bucketExists = await checkIfBucketExists(bucketName, region);
+    expect(bucketExists).toMatchObject({});
+
+    const { customResource } = meta.customCategory;
```
kaustavghosh06, line 110: Out here - I beleive either you should make an SDK call to SES to check if the AWS resources exists or you should have an "Output" in the Cloudformation file which gets populated in the amplify-meta file after deployment which you can check for.

packages/amplify-e2e-tests/src/__tests__/customStacks.test.ts
```diff
@@ -0,0 +1,116 @@
...
+  amplifyPushUpdate,
+  checkIfBucketExists,
+  createNewProjectDir,
+  deleteProject,
+  deleteProjectDir,
+  getProjectMeta,
+  initJSProjectWithProfile,
+} from 'amplify-e2e-core';
+import * as path from 'path';
+import { existsSync, mkdirSync, writeFileSync } from 'fs';
```
yuth, line 11: we usually use `fs-extra`

Issue #3705, opened 2020-03-19
Elasticsearch / searchable updates broken for "old" Amplify installations


houmark:
@SwaySway I believe PR #3391 which was discussed in #3359 broke all our Elasticsearch indexes.

Before Amplify v4.16.1 (or a few versions before) a record would land in ES something like this:

```
{
       .... // more
        "_index": "member",
        "_type": "doc",
        "_id": "8f5ea23e-40e3-4bad-7cd9-35e36b4cda26",
        "_score": 32.273067,
       .... // more
}
```

After upgrading to 4.16.1 it will land like this:

```
{
       .... // more
        "_index": "member",
        "_type": "doc",
        "_id": "id=8f5ea23e-40e3-4bad-7cd9-35e36b4cda26",  <-------- note id= prepended in the value
        "_score": 32.273067,
       .... // more
}
```

`_id` is now prepended with the key field where it used to be just the value.

The result is very damaging and completely broke our entire application as ES will create a new document for any update to an existing item in the index. The result of that is "duplicate" items all over our UI listing, one for the old outdated record and one for the new one.

**How can this be committed and enter Amplify without a HUGE warning AND a migration method documented? Do you think we are just playing around with non-production applications still?**

After updating to latest CLI last night, I'm sitting here pulling my hair out trying to figure out what is happening for many hours today and by using Kibana I was able to get a closer look at the index and realize this "simple" change which has huge effects on all existing users of a `@searchable`.

Do you have any suggestions for me @SwaySway? Delete the entire index and then code a migration job for ALL DynamoDB tables that are using `@searchable` to have every single item updated and re-triggered so they get re-created in Elasticsearch seems like a lot of work.

houmark:
@SwaySway I see your PR and not sure exactly what it does for me. What is your suggestion for me to do, as my PRODUCTION app is currently broken and is breaking more and more by the hour.

Is my best option to delete the index and run a job on all items in all tables which uses `@searchable` to have the Lambda re-add to the index?

SwaySway:
Hello @houmark I understand your frustration and we’re working on documenting a migration workflow to be made available in our docs for customers to migrate  existing data from DynamoDB into their Elasticsearch cluster with the change we introduced. At the same time, we’re working on a smoother upgrade experience in the CLI to inform our customers about this change and point them to the script to re-index if they hit a scenario such as this.

[...]

Issue #3602, opened 2020-03-06
4.14.1 elasticsearch transform bug


rarira:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[Amplify Console GitHub Issue Tracker](https://github.com/aws-amplify/amplify-console/issues)

**Describe the bug**
I have to upgrade from 4.13.1 to 4.14.1 because of this issue #3493 
and pagination token issue had been solved but there happens to be another big issue in elasticsearch transform..
I think #2860 happens again 
when there is existing string set data in elasticsearch index, the item can't be updated accordingly with dynamodb update and there was very strange outcome that elasticsearch stores duplicate item with different id instead of updating.

**Amplify CLI Version**
You can use `amplify -v` to check the amplify cli version on your system
4.14.1
when rolled back to 4.13.1 there is no problem

**Expected behavior**
stringset data can be dealt with properly
I hope there will be changelog update if there is a breaking change like this

SwaySway:
Hello @rarira 
This seems related to supporting `@key` in `@searchable` we'll investigate on a fix to avoid duplicate records. In the meantime one potential workaround you can do is use the lambda function from the old version below and in the latest release of the cli you can add that zip into a folder functions and it will be overridden.

Old Streaming Lambda Function: https://github.com/aws-amplify/amplify-cli/blob/16419f4d9e1733ed0ada064f9ced604083ee4703/packages/graphql-elasticsearch-transformer/streaming-lambda/python_streaming_function.py

brene:
Hi @rarira - I'm closing this per #3705

rarira:
@brene I just tested new 4.17.1 and it seems that it solved this issue but your backfill python script doesn't work as expected...

at first, I acknowledged that I'm not good at elasticsearch so it might be my fault.

I have a Conv model which has 'user1state' which is string and this string is 'ISO date string' or just plain string. 
```
type Conv
  @model
  @searchable {
  user1state: String
}
```

but your backfill script map this field automatically type "date" so other items with 'just plain string'  failed indexing.

```
"user1state": {
   "type": "date"
},

```

so I had used custom mapping before populating all of my indices as below 

```
PUT conv
{
   "mappiings" : {
...
...
    }
}

```

but in this case, backfill script doen't work at all..

SwaySway:
[...]

PR #3712, opened 2020-03-19
fix(graphql-elasticsearch-transformer): fix duplicate records in es lambda

SwaySway:
- fixed es streaming lambda to remove duplicates based on changes introduced supporting `@key`
- added a warning map based on transformer changes to notify customers of changes in the transformer
- BREAKING CHANGE: Customers using the CLI will be notified of a change in the lambda function which needs a migration flow to be introduced in docs

re #3602 #3705

*Issue #, if available:*

*Description of changes:*


rarira:
please be considerate about string set issue also..

packages/graphql-transformer-core/src/util/transfomConfigVersionMap.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Map of versions and there warning messages
+ */
+export const TRANSFORM_CONFIG_VERSION_MAP = {
+  3: directives => {
```
kaustavghosh06: what are the keys in the map out here? Would be helpful to describe it at the top of the file

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -161,12 +162,22 @@ async function transformerVersionCheck(context, resourceDir, cloudBackendDirecto
...
+    versionChangeMessage = TRANSFORM_CONFIG_VERSION_MAP[transformerVersion](usedDirectives);
+  }
 
   const resources = updatedResources.filter(resource => resource.service === 'AppSync');
 
-  if (showPrompt && usedDirectives.includes('auth') && resources.length > 0) {
+  if (versionChangeMessage && resources.length > 0) {
     if (context.exeInfo && context.exeInfo.inputParams && context.exeInfo.inputParams.yes) {
       context.print.warning(`\n${versionChangeMessage}\n`);
+      transformerVersion += 1;
```
kaustavghosh06: This logic is not scalable. If we have another bump in the transform conf version then this logic would need a big refactor. I would recommend putting all the warning messages in an array based on logic on transform conf version - after which you just have one question to ask in the flow - "Do you want to continue?"

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -161,12 +162,22 @@ async function transformerVersionCheck(context, resourceDir, cloudBackendDirecto
   // if we already asked the confirmation question before at a previous push
   // or during current operations we should not ask again.
   const showPrompt = !(cloudVersionExist || localVersionExist);
+  if (!showPrompt) {
+    transformerVersion = 3;
```
nikhname: Why do we go to version 3 here?

packages/graphql-transformer-core/src/util/transfomConfigVersionMap.ts
```diff
@@ -0,0 +1,15 @@
...
+    if (directives.includes('auth')) {
+      return 'The default behavior for @auth has changed in the latest version of Amplify\nRead here for details: https://aws-amplify.github.io/docs/cli-toolchain/graphql#authorizing-subscriptions';
+    }
+  },
+  5: directives => {
+    if (directives.includes('searchable')) {
+      return 'The behavior for @searchable has changed after version 4.14.1.\nRead here for details: https://aws-amplify.github.io/docs/cli-toolchain/graphql#searchable';
+    }
+  },
```
nikhname: Why isn't version 6 mapped to anything?

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
yuth: Feature flags are the way to solve this long term

yuth: We should not just increment this. We should know exactly what version of transformer we are at rather then incrementing it

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
yuth: This should be the version that we are currently at. What does TRANSFORM_CURRENT_VERSION do?

kaustavghosh06:
Does this also take care of the --yes flag during push? Please make sure to test the headless mode for push.

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -177,14 +188,15 @@ async function transformerVersionCheck(context, resourceDir, cloudBackendDirecto
       if (!response.transformerConfig) {
         process.exit(0);
       }
+      transformerVersion += 1;
```
yuth: This should be the current version of the transformer. If the customer goes and edits the version to 25 or what ever value we should not be making it 26. There is also the possiblilty of team workflow where one developer is in one version and the other is in a different version. When a older version is used, we should warn them as well as this could break things

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -177,14 +188,15 @@ async function transformerVersionCheck(context, resourceDir, cloudBackendDirecto
...
     }
   }
 
   // Only touch the file if it misses the Version property
   // Always set to the base version, to not to break existing projects when coming
   // from an older version of the CLI.
-  if (!localTransformerConfig.Version) {
-    localTransformerConfig.Version = TRANSFORM_BASE_VERSION;
+  if (!localTransformerConfig.Version || transformerVersion) {
+    localTransformerConfig.Version = transformerVersion || TRANSFORM_BASE_VERSION;
```
yuth: TRANSFORM_CURRENT_VERSION instead of base version?

packages/graphql-elasticsearch-transformer/streaming-lambda/python_streaming_function.py
```diff
@@ -212,7 +215,11 @@ def _lambda_handler(event, context):
             es_actions.append(json.dumps(action))
             # Payload line
             es_actions.append(doc_json)
-
+            # migration step remove old key if it exists
+            if ('id' in doc_fields) and (event_name == 'MODIFY') :
```
yuth: I am wondering if we should this route at all? This solves the problem for now but once every installation has this problem fixed, do we need this code? Would it become cruft?

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
SwaySway: The questionflow of adding the transformer version would always be `transformerVersion`. If we change the base to current we would not want to bump a customer to the current version should the transformer version not be available. Though double checking this I'm not sure if that the base version is needed at all.

packages/graphql-elasticsearch-transformer/streaming-lambda/python_streaming_function.py
SwaySway: Yes that is correct we can add an environment variable into the lambda to run this piece and have that disabled at a later time when it is not needed?

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
SwaySway: Let's say a customer is in version 25 and the current version is 27. If they acknowledge the prompt wouldn't they potentially miss a warning message for version 26?

packages/graphql-elasticsearch-transformer/streaming-lambda/python_streaming_function.py
kaustavghosh06: Yeah, I think this is okay to have here - On the top of my head, I don't see a cleaner and easier way to mitigate the re-indexing issue on the fly other than having it in the streaming function.

packages/graphql-transformer-core/src/util/transfomConfigVersionMap.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export const TRANSFORM_CONFIG_VERSION_MAP = {
+  3: directives => {
+    if (directives.includes('auth')) {
+      return 'The default behavior for @auth has changed in the latest version of Amplify\nRead here for details: https://aws-amplify.github.io/docs/cli-toolchain/graphql#authorizing-subscriptions';
+    }
+  },
+  5: directives => {
+    if (directives.includes('searchable')) {
+      return 'The behavior for @searchable has changed after version 4.14.1.\nRead here for details: https://aws-amplify.github.io/docs/cli-toolchain/graphql#searchable';
```
kaustavghosh06: Change this link to the correct placeholder which we'll be having in the docs.

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -43,6 +44,15 @@ function warnOnAuth(context, map) {
   }
 }
 
+class TransformerConfVersionNotSupportedError extends Error {
+  constructor() {
+    super(
+      "The transformer conf version entered is higher than the current supported version."
```
yuth: I think this message should be changed, Who is entering the transformer conf version? Probably the message should be
```
This project was pushed with newer version of Amplify CLI. Using {curreentVersion} would cause some features in the API to be removed 
```

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
yuth: Can we add unit tests testing 
1. No transform conf version
2. Transform conf version set to 3
3. Transform conf version set to current version
3. Transform conf set to 100

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -147,48 +147,73 @@ function getTransformerFactory(context, resourceDir, authConfig) {
...
   // this is where we check if there is a prev version of the transformer being used
   // by using the transformer.conf.json file
   const cloudTransformerConfig = await readTransformerConfiguration(cloudBackendDirectory);
   const cloudVersionExist = checkVersionExist(cloudTransformerConfig);
+  const cloudWarningExist = checkESWarningExists(cloudTransformerConfig);
 
   // check local resource if the question has been answered before
   const localTransformerConfig = await readTransformerConfiguration(resourceDir);
   const localVersionExist = checkVersionExist(localTransformerConfig);
+  const localWarningExist = checkESWarningExists(cloudTransformerConfig);
```
yuth: checkESWarningExists(~cloudTransformerConfig~localTransformerConfig);

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -147,48 +147,73 @@ function getTransformerFactory(context, resourceDir, authConfig) {
...
-        type: 'confirm',
-        message: `${versionChangeMessage}\nDo you wish to continue?`,
-        default: false,
-      });
-      if (!response.transformerConfig) {
-        process.exit(0);
-      }
+  if (resources.length > 0) {
+    if (showPrompt && usedDirectives.includes('auth')) {
+      warningMessage(context, versionChangeMessage);
```
yuth: what is the difference between warning and prompt?

packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -147,48 +147,74 @@ function getTransformerFactory(context, resourceDir, authConfig) {
...
 
   // check local resource if the question has been answered before
   const localTransformerConfig = await readTransformerConfiguration(resourceDir);
   const localVersionExist = checkVersionExist(localTransformerConfig);
+  const localWarningExist = checkESWarningExists(localTransformerConfig);
 
   // if we already asked the confirmation question before at a previous push
   // or during current operations we should not ask again.
   const showPrompt = !(cloudVersionExist || localVersionExist);
+  const showWarning = !(cloudWarningExist || localWarningExist);
```
yuth: these variables could be named better

simeon-smith:
This script causes an issue with the ID generation. See https://github.com/aws-amplify/amplify-cli/issues/3705#issuecomment-605118467.

juniarz:
@SwaySway An error occurs when any of the key values are non string.

I fixed the issue by adding a conversion:

`
def compute_doc_index(keys_raw, deserializer, formatIndex=False):
    index = []
    for key in sorted(keys_raw):
        if formatIndex:
            index.append('{}={}'.format(
                key, deserializer.deserialize(keys_raw[key])))
        else:
            index.append(deserializer.deserialize(keys_raw[key]))
    return '|'.join(index)
`

to

`
def compute_doc_index(keys_raw, deserializer, formatIndex=False):
    index = []
    for key in sorted(keys_raw):
        if formatIndex:
            index.append('{}={}'.format(
                key, str(deserializer.deserialize(keys_raw[key]))))
        else:
            index.append(str(deserializer.deserialize(keys_raw[key])))
    return '|'.join(index)
`

PR #3733, opened 2020-03-21
test(amplify-e2e-tests): removed amplify app test

ammarkarachi:
*Issue #, if available:*

*Description of changes:*


yarn.lock
```diff
@@ -2886,6 +2886,13 @@
   resolved "https://registry.yarnpkg.com/@tootallnate/once/-/once-1.0.0.tgz#9c13c2574c92d4503b005feca8f2e16cc1611506"
```
attilah: yarn.lock changes are unrelated, please remove from the PR.

packages/amplify-e2e-tests/src/__tests__/delete.test.ts
```diff
@@ -72,25 +72,25 @@ describe('amplify delete', () => {
...
-    await pullProject(projRoot2, { appId: AmplifyAppId, envName });
-    await initIosProjectWithProfile(projRoot2, {});
-    await deleteProject(projRoot);
-    expect(await appExists(AmplifyAppId, Region)).toBeTruthy();
-    // clean up
-    await deleteProject(projRoot2);
-    deleteProjectDir(projRoot2);
-    await deleteAmplifyApp(AmplifyAppId, Region);
-  });
+  // it('should not delete amplify app', async () => {
```
attilah: Add some comments on why this test is removed, for future reference and it would be better to use ```it.skip``` instead of commenting out.

ammarkarachi:
The tests were commented because if the number of apps on the Amplify Console exceed 25 the appId does not get assigned and it was breaking the test. With the functionality of the amplify app getting deleted after every amplify introduced by this change it would deleted the apps and avoid the creation of Amplify Apps.

Issue #3785, opened 2020-03-27
CLI stuck at Oauth  Which redirect signin URIs do you want to edit? 


ajonp:
**Describe the bug**
When trying to update auth to include a new lambda trigger I am unable to move past the question `Which redirect signin URIs do you want to edit?`

It lists all of the current redirects assigned to the project but I cannot hit enter.

**Amplify CLI Version**
4.17.2

**To Reproduce**
```
amplify auth update
Please note that certain attributes may not be overwritten if you choose to use defaults settings.

You have configured resources that might depend on this Cognito resource.  Updating this Cognito resource could have unintended side effects.

Using service: Cognito, provided by: awscloudformation
 What do you want to do? Walkthrough all the auth configurations
 Select the authentication/authorization services that you want to use: User Sign-Up, Sign-In, connected with AWS IAM controls (Enables per-user Storage features for images or oth
er content, Analytics, and more)
 Allow unauthenticated logins? (Provides scoped down permissions that you can control via AWS IAM) No
 Do you want to enable 3rd party authentication providers in your identity pool? No
 Do you want to add User Pool Groups? No
 Do you want to add an admin queries API? No
 Multifactor authentication (MFA) user login options: OFF
 Email based user registration/forgot password: Enabled (Requires per-user email entry at registration)
 Please specify an email verification subject: Your verification code
 Please specify an email verification message: Your verification code is {####}
 Do you want to override the default password policy for this User Pool? No
 Specify the app's refresh token expiration period (in days): 30
 Do you want to specify the user attributes this app can read and write? No
 Do you want to enable any of the following capabilities? Add User to Group
 Do you want to use an OAuth flow? Yes
 What domain name prefix you want us to create for you? compass5eea5dc0-5eea5dc0
 Which redirect signin URIs do you want to edit? (Press <space> to select, <a> to toggle all, <i> to invert selection)
❯◯ http://localhost:4200/
```

**Expected behavior**
I am able to select nothing and leave oauth untouched.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Mac
 - Node Version: v12.16.0

**Additional context**
I am currently bipassing this and choosing no oauth and then manually resetting everything back to how it was prior to the change.

kaustavghosh06:
@ajonp Can't you just hit enter and go to the next question?

ajonp:
@kaustavghosh06 no it will not let me move past using enter. No matter if I select nothing or everything.

PR #3791, opened 2020-03-28
fix(amplify-category-auth): fixed issue with updating urls in auth

ammarkarachi:
skipping validate for multilist and update in auth

#3785

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-category-auth/provider-utils/awscloudformation/question-factories/core-questions.js
+++ b/packages/amplify-category-auth/provider-utils/awscloudformation/question-factories/core-questions.js
@@ -42,6 +42,8 @@ function parseInputs(input, amplify, defaultValuesFilename, stringMapsFilename, 
   if (input.type && ['list', 'multiselect'].includes(input.type)) {
     if (context.updatingAuth && input.iterator) {
       question = iteratorQuestion(input, question, context);
+      // if selecting existing value to edit it's not require to validate inputs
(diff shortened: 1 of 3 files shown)


packages/amplify-category-auth/provider-utils/awscloudformation/question-factories/core-questions.js
```diff
@@ -42,6 +42,8 @@ function parseInputs(input, amplify, defaultValuesFilename, stringMapsFilename,
   if (input.type && ['list', 'multiselect'].includes(input.type)) {
     if (context.updatingAuth && input.iterator) {
       question = iteratorQuestion(input, question, context);
+      // if selecting existing value to edit it's not require to validate inputs
+      question.validate = () => true;
```
kaustavghosh06: Is this set only on update and not on an add?
We have some validations of the URI's on an add which we don't want to loose. Can you please confirm?

ammarkarachi: Yes if the update auth is present in the context, there is an update going on.
image

kaustavghosh06: Yes, but in update also, customers can modify their URL's - we should be validating those URL's as well.

PR #3857, opened 2020-04-03
refactor(cli): allow cli local installation and other plugin platform updates

UnleashedMind:
***Description of changes:***
1. Updated cli plugin platform to allow command execution with locally installed CLI
2. `amplify plugin scan`: error message will be printed out
3. `amplify plugin scan`: only scans unique directories so it will be more efficient
4. `amplify plugin add`:  will take relative paths input such as . and it will be resolved to full path, duplicates entry will be rejected. 
5. `amplify plugin configure` on pluginDirectories: will take relative paths input such as . and it will be resolved to full path, duplicates will not be allowed to be inserted into the list. 
6. `amplify plugin configure` on pluginPrefxies:  duplicates will not be allowed to be inserted into the list. 

**How to run locally installed Amplify CLI**

In the root directory of the project, run **`npx amplify <command args>`**

**Q&A**
**Q:** When CLI is installed locally inside a package, does it write to global plugin.json
**A:** yes, other than the amplify-dev setup for CLI development, the plugin platform information is stored in the `~/.amplify/plugins.json` file. 

**Q:**  If there are multiple projects each with its own installation of Amplify CLI how would it effect the plugin json and how we choose which plugin we want to use for a given task
**A:** When an user starts an Amplify CLI command, the CLI core retrieves the plugin platform information from the` ~/.amplify/plugins.json` file, and checks if the CLI core location stored in the file is the same as its own package location, if not, it will start an plugin scan and update the content of the file. So, there is no plugin package mismatch or cross reference, but there will be slow down in the command execution  due to the scan when the user switches from one local installation to another, or between local and global installations. 

**Q:**  If some one installs a plugin globally and uses it in project with local plugin installation, how would that work in team workflow
**A:** This is the case for third party plugin, If another team member runs the CLI from another computer without the plugin package installed, the CLI command will fail,  but the CLI will print out the error message saying it can not find the plugin, the user will then know to install that particular plugin package.


packages/amplify-e2e-tests/src/plugin/index.ts
```diff
@@ -1,34 +1,31 @@
...
 export * from './verifyPluginStructure';
+export * from './pluginConfigure';
 
 import { nspawn as spawn } from 'amplify-e2e-core';
 import { getCLIPath } from '../utils';
 
 export function help(cwd: string) {
   return new Promise((resolve, reject) => {
-    spawn(getCLIPath(), ['plugin', 'help'], { cwd, stripColors: true })
-      .wait(/.*/)
```
jhockett, line 10: Why remove the ```.wait(/.*/)```?

Shouldn't we be checking that the help command outputs something?

packages/amplify-e2e-tests/src/plugin/index.ts
```diff
@@ -1,34 +1,31 @@
...
+      if (!err) {
+        resolve();
+      } else {
+        reject(err);
+      }
+    });
   });
 }
 
 export function scan(cwd: string) {
```
jhockett, line 20: Why isn't this e2e test just checking the full output? I think we need better tests here. This is the exact current output for me:

```amplify plugin scan```
Scanning for plugins...
Plugin scan successful
Scanning for plugins...
Plugin scan successful

packages/amplify-e2e-tests/src/plugin/index.ts
UnleashedMind: We are not waiting for any  particular output to respond to. 
other example:
https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-e2e-tests/src/categories/codegen.ts#L6

packages/amplify-e2e-tests/src/plugin/index.ts
UnleashedMind: This method simulates the command line UX workflow for the command. The command itself does not have any interaction with the user after it starts. It either runs successfully or it throws error.

yuth:
Could you please make the following things clear with local installation
1. When CLI is installed locally inside a package, does it write to global plugin.json
2. If there are multiple projects each with its own installation of Amplify CLI how would it effect the plugin json and how we choose which plugin we want to use for a given task
3. If some one installs a plugin globally and uses it in project with local plugin installation, how would that work in team workflow

flywheelms:
Closing this PR and backlogging the requirement.
Will need updated use cases to drive a new design which addresses possible edge cases, testing and OPS support issues.

PR #3912, opened 2020-04-09
feat(amplify-appsync-simulator): add support for websocket subscriptions

### LINKED ISSUE #3008: Failed to construct 'WebSocket' when using amplify mock over http
Copied from an [upstream amplify-cli issue](https://github.com/aws-amplify/amplify-cli/issues/2935) as reported by @riacoding:

> Describe the bug
> Create a fresh react project and add amplify to it with an API. On running amplify mock api the graphql subscriptions are not working as expected with the AWSAppSyncRealTimeProvider having a connection failed exception after timeout.
> 
> Amplify CLI Version
> Amplify/Cli@4.5.0
> 
> To Reproduce
> Following demo from aws blog post by Nikhil Dabhade
> https://aws.amazon.com/blogs/mobile/amplify-framework-local-mocking/
> 
> npx create-react-app myapp
> amplify init
> amplify add api
> npm i aws-amplify
> replace App.js with code below
> 
> import React, { useEffect, useReducer } from "react";
> import Amplify from "@aws-amplify/core";
> import { API, graphqlOperation } from "aws-amplify";
> import { createTodo } from "./graphql/mutations";
> import { listTodos } from "./graphql/queries";
> import { onCreateTodo, onUpdateTodo } from "./graphql/subscriptions";
> 
> import config from "./aws-exports";
> Amplify.configure(config); // Configure Amplify
> 
> const initialState = { todos: [] };
> const reducer = (state, action) => {
>   switch (action.type) {
>     case "QUERY":
>       return { ...state, todos: action.todos };
>     case "SUBSCRIPTION":
>       return { ...state, todos: [...state.todos, action.todo] };
>     default:
>       return state;
>   }
> };
> 
> async function createNewTodo() {
>   const todo = { name: "Use AppSync", description: "Realtime and Offline" };
>   await API.graphql(graphqlOperation(createTodo, { input: todo }));
> }
> function App() {
>   const [state, dispatch] = useReducer(reducer, initialState);
> 
>   useEffect(() => {
>     getData();
>     const subscription = API.graphql(graphqlOperation(onCreateTodo)).subscribe({
>       next: eventData => {
>         const todo = eventData.value.data.onCreateTodo;
>         dispatch({ type: "SUBSCRIPTION", todo });
>       }
>     });
>     return () => {
>       subscription.unsubscribe();
>     };
>   }, []);
> 
>   async function getData() {
>     const todoData = await API.graphql(graphqlOperation(listTodos));
>     dispatch({ type: "QUERY", todos: todoData.data.listTodos.items });
>   }
> 
>   return (
>     <div>
>       <div className="App">
>         <button onClick={createNewTodo}>Add Todo</button>
>       </div>
>       <div>
>         {state.todos.map((todo, i) => (
>           <p key={todo.id}>
>             {todo.name} : {todo.description}
>           </p>
>         ))}
>       </div>
>     </div>
>   );
> }
> export default App;
> amplify mock api
> npm start
> 
> Expected behavior
> From the http://localhost:20002/ local endpoint using graphiql to create a mutation create a todo item. This would then be automatically updated on the http://localhost:3000/ endpoint via the graphql subscription.

The error appears to happen because `new Websocket` is called with a string that starts with `http://` instead of the required `ws://` or `wss://`.

`https://` is replaced with `wss://` [here](https://github.com/aws-amplify/amplify-js/blob/7bd8843ad3c71c34c4f40684cd3c986976870c21/packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider.ts#L550).

Suggest adding a line that replaces `http://` with `ws://` to enable supporting subscriptions with `amplify mock` which runs locally on `http://` by default.

yuth:
Added support for websocket subscriptions. Websocket server will be running on the same port and
accessible on /graphql endpoint. Extracted common code from operation server and moved it to helper
methods. Added tests for the helper methods

fix #3008

*Issue #, if available:*

*Description of changes:*


dwamianm:
Thank you for addressing this @yuth !

packages/amplify-appsync-simulator/src/server/index.ts
```diff
@@ -2,27 +2,57 @@ import { OperationServer } from './operations';
...
+    this.realTimeSubscriptionServer = new AppSyncSimulatorSubscriptionServer(simulatorContext, this.httpServer, '/graphql');
   }
-  async start() {
+
+  async start(): Promise<void> {
     await this.subscriptionServer.start();
     await this.operationServer.start();
+    await this.realTimeSubscriptionServer.start();
+
+    let port = this.config.port;
```
attilah, line 33: When a port in config comes in we should still run a portfinder on that one to see if it is free or not, in case if it is not free, we can print a warning for customers that we picked a different port from the configured one.

packages/amplify-appsync-simulator/src/server/index.ts
```diff
@@ -2,27 +2,57 @@ import { OperationServer } from './operations';
...
+const MAX_PORT = 9999;
 
 export class AppSyncSimulatorServer {
   private operationServer: OperationServer;
   private subscriptionServer: SubscriptionServer;
+  private httpServer: Server;
+  private realTimeSubscriptionServer: AppSyncSimulatorSubscriptionServer;
 
-  constructor(config: AppSyncSimulatorServerConfig, simulatorContext: AmplifyAppSyncSimulator) {
+  private _url: string;
```
attilah, line 20: nit: only this private var has _ prefix, lets use common form until we enforce lint at this level

packages/amplify-appsync-simulator/src/server/operations.ts
```diff
@@ -33,54 +30,29 @@ export class OperationServer {
...
-    this.app = express();
-    this.app.use(express.json({ limit: MAX_BODY_SIZE }));
-    this.app.use(cors());
-    this.app.post('/graphql', this.handleRequest.bind(this));
-    this.app.get('/api-config', this.handleAPIInfoRequest.bind(this));
-    this.app.use('/', express.static(STATIC_ROOT));
+    this._app = express();
+    this._app.use(express.json({ limit: MAX_BODY_SIZE }));
+    this._app.use(cors());
+    this._app.post('/graphql', this.handleRequest.bind(this));
```
attilah, line 36: we could make handlerequest and the other method arrow functions so no old school bind required.

packages/amplify-appsync-simulator/src/server/operations.ts
```diff
@@ -33,54 +30,29 @@ export class OperationServer {
...
-
-    if (!this.port) {
-      this.port = await portfinder.getPortPromise({
-        startPort: BASE_PORT,
-        stopPort: MAX_PORT,
-      });
-    }
-
-    this.server = this.app.listen(this.port);
+  async start() {}
```
attilah, line 42: why do we need these empty methods if this class is not an interface implementation of some generic server?

packages/amplify-appsync-simulator/src/server/subscription.ts
```diff
@@ -1,17 +1,16 @@
...
+import { DocumentNode, ExecutableDefinitionNode, ExecutionResult, FieldNode } from 'graphql';
+import { createServer as createHTTPServer, Server } from 'http';
 import { address as getLocalIpAddress } from 'ip';
-
+import { AddressInfo } from 'net';
+import portfinder from 'portfinder';
+import { inspect } from 'util';
 import { AmplifyAppSyncSimulator } from '..';
 import { AppSyncSimulatorServerConfig } from '../type-definition';
+import { Server as MQTTServer } from './subscription/mqtt-server';
```
attilah, line 12: Can't we change the export name to MQTTServer instead of renaming the import?

packages/amplify-appsync-simulator/src/server/subscription.ts
```diff
@@ -1,17 +1,16 @@
...
+import { WebsocketSubscriptionServer } from './subscription/websocket-server/server';
 
 const MINUTE = 1000 * 60;
 const CONNECTION_TIME_OUT = 2 * MINUTE; // 2 mins
```
attilah, line 16: nit: TIMEOUT

packages/amplify-appsync-simulator/src/server/subscription/websocket-server/utils/decode-header.ts
```diff
@@ -0,0 +1,10 @@
+import { parse } from 'url';
+export function decodeHeaderFromQueryParam(rawUrl: string, paramName: string = 'header'): Record<string, any> {
+  const url = parse(rawUrl);
+  const params = new URLSearchParams(url.query);
```
attilah, line 4: Don't we need to url decode querystring parameters in case they're coming in escaped form?

packages/amplify-appsync-simulator/src/server/websocket-subscription.ts
```diff
@@ -0,0 +1,64 @@
...
+import { AmplifyAppSyncSimulator } from '..';
+import { extractHeader, extractJwtToken, getAuthorizationMode } from '../utils/auth-helpers';
+import { AppSyncGraphQLExecutionContext } from '../utils/graphql-runner';
+import { runSubscription, SubscriptionResult } from '../utils/graphql-runner/subscriptions';
+import { ConnectionContext, WebsocketSubscriptionServer } from './subscription/websocket-server/server';
+
+export class AppSyncSimulatorSubscriptionServer {
+  private realtimeServer: WebsocketSubscriptionServer;
+  constructor(private simulatorContext: AmplifyAppSyncSimulator, private server: Server, private subscriptionPath: string = '/graphql') {
+    this.onSubscribe = this.onSubscribe.bind(this);
```
attilah, line 12: same as above, make them arrow function and no need for bind.

packages/amplify-appsync-simulator/src/__tests__/server/subscription/websocket-server/utils/decode-headers.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+describe('decodeHeaderFromQueryParam', () => {
+  it('should decode header query param from the the url', () => {
+    const header = {
+      authorization: 'some token',
+      someOtherValue: 'value',
+    };
+    const paramName = 'header';
+    const base64EncodedHeaderValue = Buffer.from(JSON.stringify(header)).toString('base64');
+    const queryParam = new URLSearchParams({
```
attilah, line 11: also test url encoded values

packages/amplify-appsync-simulator/src/server/operations.ts
```diff
@@ -33,16 +33,12 @@ export class OperationServer {
     this._app = express();
     this._app.use(express.json({ limit: MAX_BODY_SIZE }));
     this._app.use(cors());
-    this._app.post('/graphql', this.handleRequest.bind(this));
-    this._app.get('/api-config', this.handleAPIInfoRequest.bind(this));
+    this._app.post('/graphql', (request, response) => this.handleRequest(request, response));
```
attilah, line 36: this could be simplified, no need for extra arrow function:
```
 this._app.post('/graphql', this.handleRequest);
 ```

Issue #3860, opened 2019-10-19
@searchable directive Elasticsearch costs


AlessandroAnnini:
** Which Category is your question related to? **
Appsync + Elasticsearch
** What AWS Services are you utilizing? **
Cognito + Appsync
** Provide additional details e.g. code snippets **
I am using `@searchable` directive on one or two appsync graphql nodes and used it a few times for a search query but my amazon bill is very high, this is the Elasticsearch detail for September:
![image](https://user-images.githubusercontent.com/3852307/67141074-63673080-f260-11e9-9d7d-51a98327e55e.png)
Is it normal? I think that it is extremely expensive, how could I set the service so that I can contain the costs? This is holding me back from starting new projects!

undefobj:
As @raelmiu noted you need to remove `@searchable` from your schema to decommission the Elasticsearch instance. At this time AWS does not have a "serverless" solution for Full Text Search so keeping that GraphQL directive without Elasticsearch has no implementation. We can however open up options or give documentation for using another instance size if that would be helpful.

raelmiu:
@Undefobj that would indeed help, not only for developing at lower cost, but also for scaling the system via amplify Cli instead of console settings.

AlessandroAnnini:
Hi @raelmiu, I tried to remove `@searchable` from my schema but I had troubles with that. So you're telling that I can remove that keyword from a working schema? If so I will try again.

raelmiu:
@AlessandroAnnini Yes it did work for me. It get's removed after I `amplify push` to each environment. In my experience amplify is really fragile with backend updates though, so make sure you only make ONE change per push.

GeorgeBellTMH:
it would be nice if multiple environments could use the same es instance...even better is if multiple projects could...at the least we should be able to run dev/uat/prod off the same ES instance...

starpebble:
[...]

PR #4009, opened 2020-04-20
chore(graphql-elasticsearch-transformer): add default to lowest instance

akshbhu:
*Issue #3860

*Description of changes:*

* Changed the default instance to t2.micro

* Add the upgrade policy to upragde the version of instance inplace.


packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -443,6 +443,44 @@ async function transformGraphQLSchema(context, options) {
 
   if (directiveMap.directives.includes('searchable')) {
     searchableTransformerFlag = true;
+    if (context.input.command === 'add') {
```
SwaySway, line 446: We could also use`resourcesToBeCreated` here similar to apiProjectIsFromOldVersion.
https://github.com/aws-amplify/amplify-cli/blob/16a2b4f50a85372f733eef129e43b154a993b136/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js#L219

akshbhu:
Closing as we currently don't support t2.micro instances as they work on lower version 2.3

sakhmedbayev:
> Closing as we currently don't support t2.micro instances as they work on lower version 2.3

@akshbhu, what is the problem with the lower version 2.3?

It would be great if amplify work with t2.micro out of the box. I think a lot of people are disappointed with the high cost of the amplify's default elasticsearch instance, especially at the dev time.

Blkc:
This seems to be related to https://github.com/aws-amplify/amplify-cli/issues/3860#issuecomment-621373818

Issue #2488, opened 2019-10-03
Inform users of amplify-cli when there is an update available


daannijkamp:
**Is your feature request related to a problem? Please describe.**
When there is a new version of the Amplify CLI, I will not be notified.
Other Packages, for example [npm-check-updates](https://www.npmjs.com/package/npm-check-updates), give me a notification when a new version is available. For this they use [update-notifier](https://www.npmjs.com/package/update-notifier).

**Describe the solution you'd like**
I would like to see the update-notifier implemented.

**Describe alternatives you've considered**
Based on the weekly 4+ million downloads, [update-notifier](https://www.npmjs.com/package/update-notifier) seems to be a good standard.

**Additional context**
[Link](https://github.com/tjunnone/npm-check-updates/blob/master/bin/ncu#L14) to example code how [npm-check-updates](https://www.npmjs.com/package/npm-check-updates) uses [update-notifier](https://www.npmjs.com/package/update-notifier)

ericclemmons:
As an alternative, ZEIT has [update-check][2] which was a result of some issues [they've detailed in a twitter thread][1].

It is similar in size, has [~200k weekly downloads][4], and may be worth comparing to, or as a potential alternative _if_ we experience any [similar issues][1] with `update-notifier`.

[1]: https://twitter.com/notquiteleo/status/983193273224200192
[2]: https://github.com/zeit/update-check
[3]: https://github.com/zeit/now/blob/d8935cf121afe44e4e8c895ab327c9b74403e4e8/packages/now-cli/src/index.js#L120-L154
[4]: https://www.npmjs.com/package/update-check

ericclemmons:
Closed via #4046!

PR #4046, opened 2020-04-22
feat: notify if a CLI update is available

nikhname:
Will notify on every amplify call with a daily interval if an update to the CLI is available.

*Issue #, if available:*
re #2488 

*Description of changes:*
Notify if an update to the CLI is available with a daily interval.

### DIFF
--- a/packages/amplify-cli/src/index.ts
+++ b/packages/amplify-cli/src/index.ts
@@ -20,13 +20,6 @@ EventEmitter.defaultMaxListeners = 1000;
 // entry from commandline
 export async function run(): Promise<number> {
   try {
-    // Checks for available update on each use
-    const notifier = updateNotifier({
-      pkg,
-      updateCheckInterval: 0,
-    });
-    notifier.notify({ defer: false, isGlobal: true });
-
     let pluginPlatform = await getPluginPlatform();
     let input = getCommandLineInput(pluginPlatform);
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/index.ts
```diff
@@ -18,6 +20,13 @@ EventEmitter.defaultMaxListeners = 1000;
 // entry from commandline
 export async function run(): Promise<number> {
   try {
+    // Checks for available update on each use
+    const notifier = updateNotifier({
+      pkg,
+      updateCheckInterval: 0,
```
attilah, line 26: I think to no to annoy the users on each execution it could be a few days interval, looking that we're releasing ~weekly. 3 or 5 days?

packages/amplify-cli/src/index.ts
```diff
@@ -18,6 +20,13 @@ EventEmitter.defaultMaxListeners = 1000;
...
+    });
+    notifier.notify({ defer: false, isGlobal: true });
```
attilah, line 28: What do you think if the update notification would appear after the logo and such but after execution in general, otherwise it could be missed easily since we output a lot of information. What is the practice in other major tools that using this package?

nikhname: From looking at the behavior other CLI's (npm, angular, yeoman) and considering the amount of information we output, the notification should appear after execution. Making this change

PR #4079, opened 2020-04-26
Changes to use .NET global tools

andyhopp:
*Issue #, if available:*

*Description of changes:*
This PR changes the detection logic to only detect and not prompt to auto-install the Lambda tools and test tool for 3.1.
This PR also fixes an issue where .gitignore wasn't being copied from resources.


packages/amplify-dotnet-function-runtime-provider/src/utils/detect.ts
```diff
@@ -42,23 +42,26 @@ export const detectDotNetCore = async (): Promise<CheckDependenciesResult> => {
...
     };
   } else {
-    return {
+    const result = {
       hasRequiredDependencies: false,
-      errorMessage: `Expected ${executableName} minimum version ${currentSupportedVersion}, but found: ${installedSdks}`,
+      errorMessage: 'Unable to detect required dependencies:\n',
     };
+    if (!sdkInstalled) {
+      result.errorMessage += '- The .NET Core 3.1 SDK must be installed.\n';
```
attilah, line 56: Please add: 'It can be installed from https://dotnet.microsoft.com/download'

packages/amplify-dotnet-function-runtime-provider/src/utils/detect.ts
```diff
@@ -42,23 +42,26 @@ export const detectDotNetCore = async (): Promise<CheckDependenciesResult> => {
...
+    }
+    if (!toolInstalled) {
+      result.errorMessage += '- The Amazon.Lambda.Tools global tool must be installed.\n';
```
attilah, line 59: Please add: 'Install by running: dotnet tool install -g Amazon.Lambda.Tools'

packages/amplify-dotnet-function-runtime-provider/src/utils/detect.ts
```diff
@@ -42,23 +42,26 @@ export const detectDotNetCore = async (): Promise<CheckDependenciesResult> => {
...
+    }
+    if (!testToolInstalled) {
+      result.errorMessage += '- The Amazon.Lambda.TestTool-3.1 global tool must be installed.\n';
```
attilah, line 62: Please add: 'Install by running: dotnet tool install -g Amazon.Lambda.TestTool-3.1'

attilah: Also the installation of these tools must be added after dotnet install: https://github.com/aws-amplify/amplify-cli/blob/master/.circleci/config.base.yml#L185

andyhopp: Ah, OK! I don't have access to CCI, so I wasn't even aware of this file. Adding now...

PR #4092, opened 2020-04-27
test(amplify-e2e-tests): add tests for graphql schemas in doc

UnleashedMind:
Add tests for graphql schemas in the document Amplify CLI/API(GraphQL)/Directives

*Issue #, if available:*

*Description of changes:*


packages/amplify-e2e-tests/src/__tests__/api-directives-1.test.ts
```diff
@@ -0,0 +1,106 @@
...
+  });
+
+  afterEach(async () => {
+    await deleteProject(projectDir);
+    deleteProjectDir(projectDir);
+  });
+
+  //model
+  it('model usage1', async () => {
+    const testresult = await testSchema(projectDir, 'model', 'usage1');
```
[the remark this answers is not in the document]
UnleashedMind: updated

packages/amplify-e2e-tests/src/__tests__/api-directives-1.test.ts
```diff
@@ -0,0 +1,106 @@
+import { initJSProjectWithProfile, deleteProject, createNewProjectDir, deleteProjectDir } from 'amplify-e2e-core';
+import { testSchema } from '../api-directives';
+
+describe('api directives @model @key @connection @versioned', () => {
```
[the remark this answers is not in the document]
UnleashedMind: updated

packages/amplify-e2e-tests/src/api-directives/index.ts
```diff
@@ -0,0 +1,32 @@
...
+    const schemaDocDirPath = path.join(__dirname, directive, section);
+    if (!fs.existsSync(schemaDocDirPath)) {
+      throw new Error(`Missing testing schema documents directory ${directive}/${section}`);
+    }
+
+    //If certain schema needs special handling, the index file inside schemaDocDirPath
+    //can define its own runTest method, which will override the common implementation
+    let schemaDocTestingModule;
+    try {
+      schemaDocTestingModule = await import(schemaDocDirPath);
```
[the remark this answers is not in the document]
UnleashedMind: Some schema does not fit in the general test bench, so there need to be a runTest method for them individually. 
In the latest update of this PR,  such individual test methods are in the same test file where the schema being tested is located.

packages/amplify-e2e-tests/src/api-directives/common.ts
```diff
@@ -0,0 +1,95 @@
...
+    const docCompiledSchema = fs
+      .readFileSync(docCompiledSchemaFilePath)
+      .toString()
+      .trim();
+    const actualCompileSchema = fs
+      .readFileSync(actualCompiledSchemaFilePath)
+      .toString()
+      .trim();
+
+    if (docCompiledSchema !== actualCompileSchema) {
```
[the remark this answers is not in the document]
UnleashedMind: I removed this test in the latest update. 
The reason it was there was because there is one occurrence in the document, [@model/Generates](https://docs.amplify.aws/cli/graphql-transformer/directives#generates), in which the generated schema is explicitly laid out. It does not seem to be the theme of the document to always define the actual generated (gql-compiled) schema. 
Snapshot tests for the gql-compile generated schema can definitely be added for each of the schema if necessary.

UnleashedMind:
@yuth
In response to the following comments:
--------------
"The main goal with tests should be verification (which this change does) and readability of the test code so one knows what is going on (and make it easier to figure out why tests are failing) rather then reducing code duplication.

Could you make the schema, query and mutation inline and use the runQuery, runMutation in the test itself. Also use the proper asserting for tests instead of returning true and expecting tests to return true as it makes it harder to figure out why a test failed"
------------

The schema being tested, the mutations, queries and subscriptions to test the schema, and their expected responses are consolidated into one file for better readability. 
If the test for a schema does not fall into one of the common test patterns, the test method itself (runTest) is also in the same file as the schema.

The test compares the actual mutation, query and subscription responses with the expected results. As the responses can be objects of depth, the test goes to a max depth of 50 (configurable constant) trying to identify mismatches. I can't figure out a simple assertion for this. But if there is a mismatch, the cause of mismatch is clearly printed out, so there shouldn't be any problem identifying the failed test, and the part that failed it.

yuth:
> The schema being tested, the mutations, queries and subscriptions to test the schema, and their expected responses are consolidated into one file for better readability.
> If the test for a schema does not fall into one of the common test patterns, the test method itself (runTest) is also in the same file as the schema.

Consolidating them to one file is counter productive. The most important thing with tests is 
1. Testing of functionality
2. Readability as tests are living documentation of behaviors.

We have similar tests in our GraphQL E2E tests. In those tests, the schema and mutation is all contained in the same file and makes it easy to understand what is going on.

Schema
https://github.com/aws-amplify/amplify-cli/blob/8e0662eac8c88da9393f32c33457a597acf591ed/packages/graphql-transformers-e2e-tests/src/__tests__/DynamoDBModelTransformer.e2e.test.ts#L39-L77

Test
https://github.com/aws-amplify/amplify-cli/blob/8e0662eac8c88da9393f32c33457a597acf591ed/packages/graphql-transformers-e2e-tests/src/__tests__/DynamoDBModelTransformer.e2e.test.ts#L189-L224

> The test compares the actual mutation, query and subscription responses with the expected results. As the responses can be objects of depth, the test goes to a max depth of 50 (configurable constant) trying to identify mismatches. I can't figure out a simple assertion for that. But if there is a mismatch, the cause of mismatch is clearly printed out, so there shouldn't be any problem identifying the failed test, and the part that failed it.


Could you explain what you mean by max depth of 50? Jest has the assertion [`expect(result).toEqual({})`](https://jestjs.io/docs/en/expect#toequalvalue) which does compare objects. The other option is to use [`expect(result).toMatchInlineSnapshot()`](https://jestjs.io/docs/en/expect#tomatchinlinesnapshotpropertymatchers-inlinesnapshot) which keeps the exact result.

packages/amplify-e2e-core/src/categories/api.ts
```diff
@@ -298,3 +298,244 @@ export function addRestApi(cwd: string, settings: any) {
...
+      .wait('"amplify publish" will build all your local backend and frontend resources')
+      .run((err: Error) => {
+        if (!err) {
+          resolve();
+        } else {
+          reject(err);
+        }
+      });
+  });
+}
```
yuth, line 335: How is this different from the function below except for not passing an Schema file? Wouldn't it be possible to update this function to accept optional schema file and have both functionality?  https://github.com/aws-amplify/amplify-cli/blob/c522f295304410aeb1d6f60aaba9b466d3304ee1/packages/amplify-e2e-core/src/categories/api.ts#L47-L78

There is also another function which does the exact same thing
https://github.com/aws-amplify/amplify-cli/blob/c522f295304410aeb1d6f60aaba9b466d3304ee1/packages/amplify-e2e-core/src/categories/api.ts#L11-L46

packages/amplify-e2e-core/src/categories/api.ts
```diff
@@ -298,3 +298,244 @@ export function addRestApi(cwd: string, settings: any) {
...
```
yuth, line 541: Can we consolidate all of this into a single function which accepts an auth configuratation and then select the appropriate auth type without having 3 separate function which has so much code repeated

packages/amplify-e2e-core/src/categories/auth.ts
```diff
@@ -916,3 +916,81 @@ export function addAuthWithMaxOptions(cwd: string, settings: any) {
...
+      .wait('"amplify publish" will build all your local backend and frontend resources')
+      .run((err: Error) => {
+        if (!err) {
+          resolve();
+        } else {
+          reject(err);
+        }
+      });
+  });
+}
```
yuth, line 939: Duplicate of this https://github.com/aws-amplify/amplify-cli/blob/4322a326df8c5c0a89de5f5e8f46bcfd4e1ad770/packages/amplify-e2e-core/src/categories/auth.ts#L3-L21

packages/amplify-e2e-core/src/categories/auth.ts
```diff
@@ -916,3 +916,81 @@ export function addAuthWithMaxOptions(cwd: string, settings: any) {
...
+
+//update auth to add user group
+export function updateAuthAddFirstUserGroup(projectDir: string, groupName: string) {
```
yuth, line 942: Can this function accept an list of strings for group name
```suggestion
export function updateAuthWithUserpoolGroups(projectDir: string, groupNames: string[]) {
```

packages/amplify-e2e-core/src/init/initProjectHelper.ts
```diff
@@ -205,3 +205,46 @@ export function initNewEnvWithProfile(cwd: string, s: { envName: string }) {
...
+      .resumeRecording()
+      .wait('Try "amplify add api" to create a backend API and then "amplify publish" to deploy everything')
+      .run((err: Error) => {
+        if (!err) {
+          resolve();
+        } else {
+          reject(err);
+        }
+      });
+  });
```
yuth, line 249: Duplicate of https://github.com/aws-amplify/amplify-cli/blob/78642a2664e4449f7c44b99ecdf15e8e3ced901e/packages/amplify-e2e-core/src/init/initProjectHelper.ts#L115-L157. We just need to update this function to have default region as `us-east-2`

packages/amplify-e2e-tests/src/api-directives/common.ts
```diff
@@ -0,0 +1,366 @@
...
+      received,
+      expected,
+      depth: 0,
+    },
+  ];
+  return runCompare(queue);
+}
+
+const MAX_DEPTH = 50;
+function runCompare(queue: { received: any; expected: any; depth: number }[]): boolean {
```
yuth, line 324: Can we use [`.toMatchObject(object)`](https://jestjs.io/docs/en/expect#tomatchobjectobject) instead of rolling our own solution

packages/amplify-e2e-tests/src/api-directives/tests/predictions-usage.ts
```diff
@@ -0,0 +1,150 @@
...
+}
+
+/*
+This test will fail:
+There seems to be an AppSync bug, the error received:
+
+    ../error ApolloError: GraphQL error: Unable to parse the JSON document: 'Unexpected character ('m' (code 109)): was expecting comma to separate Object entries
+       at [Source: (String)"{
+        "version": "2018-05-29",
+        "method": "POST",
```
yuth, line 50: Why this comment? Are we doing anything about this in our tests?

packages/amplify-e2e-tests/src/api-directives/tests/predictions-usage.ts
```diff
@@ -0,0 +1,150 @@
...
+//schema
+export const schema = `
+type Query {
+  speakTranslatedImageText: String @predictions(actions: [identifyText, translateText, convertTextToSpeech])
+}
+`;
+
+//queries
+export const query = `
+#change: remove redaudant ($input: SpeakTranslatedImageTextInput!)
```
yuth, line 140: Did you do a PR in docs repo to fix this?

packages/amplify-e2e-tests/src/api-directives/tests/searchable-usage.ts
```diff
@@ -0,0 +1,32 @@
...
+    title
+    createdAt
+    updatedAt
+    upvotes
+  }
+}`;
+
+//queries
+export const query1 = `
+#error: add "s" for searchPosts
```
yuth, line 24: Did we fix this in docs repo?

packages/amplify-graphql-types-generator/test/__snapshots__/loading.ts.snap
```diff
@@ -384,6 +384,6 @@ Object {
 `;
 
 exports[`Validation should throw a helpful message when a file has invalid gql snippets 1`] = `
-"Could not parse graphql operations in test/fixtures/misc/invalid-gqlQueries.js
+"Could not parse graphql operations in packages/amplify-graphql-types-generator/test/fixtures/misc/invalid-gqlQueries.js
```
yuth, line 387: Please remove this as this is not relevant to this PR

packages/graphql-auth-transformer/src/__tests__/__snapshots__/OperationsArgument.test.ts.snap
```diff
@@ -5,15 +5,7 @@ exports[`Test "create", "update", "delete" auth operations 1`] = `"$util.toJson(
...
-"## [Start] Set default values. **
-$util.qr($context.args.input.put(\\"id\\", $util.defaultIfNull($ctx.args.input.id, $util.autoId())))
-#set( $createdAt = $util.time.nowISO8601() )
-## Automatically set the createdAt timestamp. **
-$util.qr($context.args.input.put(\\"createdAt\\", $util.defaultIfNull($ctx.args.input.createdAt, $createdAt)))
-## Automatically set the updatedAt timestamp. **
-$util.qr($context.args.input.put(\\"updatedAt\\", $util.defaultIfNull($ctx.args.input.updatedAt, $createdAt)))
-## [End] Set default values. **
-## [Start] Determine request authentication mode **
+"## [Start] Determine request authentication mode **
```
yuth, line 8: Please rebase this should not have changed here

packages/graphql-key-transformer/src/__tests__/__snapshots__/KeyTransformer.test.ts.snap
```diff
@@ -2,14 +2,7 @@
...
 Object {
-  "Mutation.createItem.req.vtl": "## [Start] Set default values. **
-#set( $createdAt = $util.time.nowISO8601() )
-## Automatically set the createdAt timestamp. **
-$util.qr($context.args.input.put(\\"createdAt\\", $util.defaultIfNull($ctx.args.input.createdAt, $createdAt)))
-## Automatically set the updatedAt timestamp. **
-$util.qr($context.args.input.put(\\"updatedAt\\", $util.defaultIfNull($ctx.args.input.updatedAt, $createdAt)))
-## [End] Set default values. **
-
+  "Mutation.createItem.req.vtl": "
```
yuth, line 5: this should not have changed here

packages/amplify-e2e-tests/src/api-directives/tests/auth-fieldLevelAuth5.ts
```diff
@@ -0,0 +1,66 @@
...
+      updatedAt: "2020-01-01T01:05:49.129Z"
+      content: "todo1 content"
+    }) {
+      id
+      owner
+      updatedAt
+      content
+    }
+}`;
+export const expected_result_mutation1 = {
```
yuth, line 32: Please dont use the variable naming magic. Like i suggested in an earlier comment, move this inside a test and then wrap it inside an object, so both mutation and results are wrapped in the same object

packages/amplify-e2e-tests/src/api-directives/tests/searchable-usage.ts
UnleashedMind: yes

packages/amplify-e2e-core/src/categories/function.ts
```diff
@@ -140,6 +140,77 @@ const coreFunction = (
...
+export function addSimpleFunction(projectDir: string, functionName: string) {
+  return new Promise((resolve, reject) => {
+    spawn(getCLIPath(), ['add', 'function'], { cwd: projectDir, stripColors: true })
+      .wait('Provide a friendly name for your resource to be used as a label for this category in the project:')
+      .sendLine(functionName)
+      .wait('Provide the AWS Lambda function name:')
+      .sendLine(functionName)
+      .wait('Choose the function runtime that you want to use:')
+      .sendCarriageReturn()
+      .wait('Choose the function template that you want to use:')
```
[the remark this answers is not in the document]
UnleashedMind: I update the code to use multiSelect and singleSelect. . 
But even with when multiSelect or singleSelect, when we re-order our prompts, the test will still fail.

packages/amplify-e2e-core/src/categories/api.ts
UnleashedMind: That particular function is removed from this PR. The add Api workflows with different auth settings in the PR are combined into a single function.

packages/amplify-e2e-core/src/categories/api.ts
UnleashedMind: done

packages/amplify-e2e-core/src/categories/storage.ts
```diff
@@ -353,3 +353,41 @@ export function addS3WithTrigger(cwd: string, settings: any) {
...
+      .sendCarriageReturn()
+      .wait('Please provide bucket name:')
+      .sendCarriageReturn()
+      .wait('Who should have access:')
+      .send(KEY_DOWN_ARROW) //'Auth and guest users'
+      .sendCarriageReturn()
+      .wait('What kind of access do you want for Authenticated users?')
+      .send(' ') //create/update
+      .send(KEY_DOWN_ARROW)
+      .send(' ') //red
```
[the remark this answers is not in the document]
UnleashedMind: This method is removed in the latest update of the PR.

packages/amplify-e2e-core/src/categories/function.ts
```diff
@@ -140,6 +140,77 @@ const coreFunction = (
...
+      .send(KEY_DOWN_ARROW) //Hellow world
```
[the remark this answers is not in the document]
UnleashedMind: done

packages/graphql-auth-transformer/src/__tests__/__snapshots__/OperationsArgument.test.ts.snap
UnleashedMind: done

packages/graphql-key-transformer/src/__tests__/__snapshots__/KeyTransformer.test.ts.snap
UnleashedMind: done

packages/amplify-e2e-tests/src/api-directives/common.ts
UnleashedMind: There are some technical difficulties. 
In the test, the actual received mutation, query and subscription results are not directly compared with the expected results specified in the test case file (e.g. using the *jest* *.toMatchObject(object)* method). 
- *id*, *createdAt* and *updatedAt* fields are generated (if not specified in the mutation), their values are not determined when the test starts to run. Those fields are generally given a value *“<check-defined>”* in the expected result specification, and the test only checks that they are defined in the actual received result. 
- If error is expected, the test only checks for the error type, but the actual error response contains a lot of fields.
- If actual response can contain the "__typename" field, even if the schema does not contain union of interface, and even if the mutation or query does not specify it, and the expected result does not include this extra field. 

To compare the expected and the actual result object, the test uses a BFS method to find mismatched field, if no mismatch is found after MAX_DEPTH (set at 50 in the test), the test passes.

PR #4145, opened 2020-05-01
fix(amplify-codegen): use ResDir directory to compute modelgen output

### LINKED ISSUE #3993: [codegen] Output path for generated Android models is hardcoded to `app`, which breaks if module name is anything else.
Running `amplify codegen models` for an Android project outputs generated files to 
the `app/src/main/java/` directory ([hardcoded here](https://github.com/aws-amplify/amplify-cli/blob/74281cc541f3226db3cc47ed37c1b70f7d7880a6/packages/amplify-codegen/src/commands/models.js#L108)).   If your module name is something other than the default, `app`, the source files are still put in an `app` folder, instead of your module's folder.

When running `amplify init`, the customer is already asked to specific the resource directory (usually `app/src/main/res`) where the GraphQL schema is located, and is used to store amplifyconfiguration.json.  From this path, you should be able to infer the codegen output directory as `../java`

Currently, I can workaround this issue by manually copying the generated files to the correct directory after they are generated.

yuth:
Updated model gen to use the Applications ResDirectory to compute the output path for generated
models

fix #3993

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -20,7 +20,7 @@ async function generateModels(context) {
 
   const allApiResources = await context.amplify.getResourceStatus('api');
   const apiResource = allApiResources.allResources.find(
-    resource => resource.service === 'AppSync' && resource.providerPlugin === 'awscloudformation'
+    resource => resource.service === 'AppSync' && resource.providerPlugin === 'awscloudformation',
   );
 
   if (!apiResource) {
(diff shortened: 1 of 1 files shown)


packages/amplify-codegen/src/commands/models.js
```diff
@@ -105,7 +105,9 @@ function getModelOutputPath(context) {
     case 'javascript':
       return 'src';
     case 'android':
-      return 'app/src/main/java/';
+      return projectConfig.android && projectConfig.android.config && projectConfig.android.config.ResDir
+        ? path.join(projectConfig.android.config.ResDir, '..', 'java')
+        : 'app/src/main/java/';
```
nikhname, line 110: Should we use path.join here instead of forward slashes for windows compatibility?

PR #4196, opened 2020-05-07
feat: support for overriding pipeline function templates in transformer

### LINKED ISSUE #4192: Customizing appsync functions created by @function directive
API, GraphQL

When using the @function directive, it creates the AppSync function with the data source as the lambda function, but the resolvers are static and cannot be overwritten (according to what I see), they are created inside `build/pipelineFunctions`.

I want to find a way to change that resolver, so for example I want to handle errors in a custom way, that allows me to send custom data when I throw errors. Is there a way to do that other than going to the AppSync console and manually changing the resolver?

**Amplify CLI Version** 
4.18.0

yuth:
Add support for overriding pipeline function tempalte in graphql transformer. Updated mock to
support this as well

fix #4192

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
+++ b/packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
@@ -439,13 +439,13 @@ async function transformGraphQLSchema(context, options) {
   }
 
   const buildConfig = {
+    ...options,
     buildParameters,
     projectDirectory: options.dryrun ? false : resourceDir,
     transformersFactory: transformerListFactory,
     transformersFactoryArgs: [searchableTransformerFlag, storageConfig],
     rootStackFileName: 'cloudformation-template.json',
     currentCloudBackendDirectory: previouslyDeployedBackendDir,
-    disableResolverOverrides: options.disableResolverOverrides,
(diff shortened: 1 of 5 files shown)


packages/amplify-provider-awscloudformation/lib/transform-graphql-schema.js
```diff
@@ -439,13 +439,13 @@ async function transformGraphQLSchema(context, options) {
...
 
   const buildConfig = {
+    ...options,
     buildParameters,
     projectDirectory: options.dryrun ? false : resourceDir,
     transformersFactory: transformerListFactory,
     transformersFactoryArgs: [searchableTransformerFlag, storageConfig],
     rootStackFileName: 'cloudformation-template.json',
     currentCloudBackendDirectory: previouslyDeployedBackendDir,
-    disableResolverOverrides: options.disableResolverOverrides,
```
SwaySway, line 448: Why are we removing disableResolverOverrides here?

yuth: Because we are spreading the value from options and this option will be set only when using mock

PR #4199, opened 2020-05-07
Added logging when Zipping Fails

aprowe:
Added reason for zipping fails
'Zipping artifacts failed.' is not enough to determine what when wrong. 
When I added this line, I got the error "Please ensure your build artifacts path exists. which immediately Pointed me towards the issue.

*Description of changes:*
Added `console.log` to zipping error


attilah:
@aprowe thanks for the PR, unfortunately `console.log` is not used in the CLI we have `context.print` methods to print errors. I am closing this PR as we cannot take as it is, but will look into how could we surface this error in a better way in the future.

PR #4258, opened 2020-05-13
feat(packages/amplify-codegen-appsync-model-plugin): add support for keyName in connection

yuth:
Update modelgen to support @connection directive with `keyName`. 

related change in Amplify JS: https://github.com/aws-amplify/amplify-js/pull/5778

### DIFF
--- a/packages/amplify-codegen-appsync-model-plugin/src/utils/process-connections.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/utils/process-connections.ts
@@ -51,16 +51,14 @@ export function getConnectedField(field: CodeGenField, model: CodeGenModel, conn
   if (connectionFields) {
     let keyDirective;
     if (keyName) {
-      throw new Error('DataStore does not support connection directive with keyName');
-      // When the library starts supporting remove the exception
-      // keyDirective = connectedModel.directives.find(dir => {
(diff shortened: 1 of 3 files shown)


packages/amplify-codegen-appsync-model-plugin/src/utils/process-connections.ts
```diff
@@ -51,16 +51,14 @@ export function getConnectedField(field: CodeGenField, model: CodeGenModel, conn
...
-      // keyDirective = connectedModel.directives.find(dir => {
-      //   return dir.name === 'key' && dir.arguments.name === keyName;
-      // });
-      // if (!keyDirective) {
-      //   throw new Error(
-      //     `Error processing @connection directive on ${model.name}.${field.name}, @key directive with name ${keyName} was not found in connected model ${connectedModel.name}`,
-      //   );
-      // }
+      keyDirective = connectedModel.directives.find(dir => {
+        return dir.name === 'key' && dir.arguments.name === keyName;
```
attilah, line 55: `dir.arguments` defined always or pre-checked when execution gets here?

yuth: GraphQL JS ensures that

PR #4268, opened 2020-05-14
feat(amplify-app): group amplify files in xcode integration

drochetti:
**Notes:**

- change `amplifyxc.config` to `amplifytools.xcconfig` to match standard
xcode config files extension
- create an Xcode group called `AmplifyConfig` and add all
config-related files
  - `amplifytools.xcconfig`
  - `amplifyconfiguration.json`
  - `awsconfiguration.json`
  - `schema.graphql`
- create an Xcode group called `AmplifyModels` and add the generated
models to it and also as project source files
- files are added as references and not copied, so updates to the
original files are automatically reflected

### DIFF
--- a/packages/amplify-app/src/index.js
+++ b/packages/amplify-app/src/index.js
@@ -6,7 +6,7 @@ const emoji = require('node-emoji');
 const { spawnSync, spawn } = require('child_process');
 const frameworkConfigMapping = require('./framework-config-mapping');
 const args = require('yargs').argv;
-const { addDataStoreFiles } = require('./xcodeHelpers');
+const { addAmplifyFiles } = require('./xcodeHelpers');
 const ini = require('ini');
 const semver = require('semver');
 const stripAnsi = require('strip-ansi');
@@ -400,9 +400,8 @@ async function createIosHelperFiles() {
     fs.writeFileSync(amplifyConfigFile, configJsonStr);
   }
 
-  // add models and schema
   if (fs.existsSync(amplifyDir)) {
-    await addDataStoreFiles();
+    await addAmplifyFiles();
   }
 }
 
--- a/packages/amplify-app/src/index.js
+++ b/packages/amplify-app/src/index.js
@@ -6,7 +6,7 @@ const emoji = require('node-emoji');
 const { spawnSync, spawn } = require('child_process');
 const frameworkConfigMapping = require('./framework-config-mapping');
 const args = require('yargs').argv;
-const { addFileToXcodeProj } = require('./xcodeHelpers');
+const { addDataStoreFiles } = require('./xcodeHelpers');
 const ini = require('ini');
 const semver = require('semver');
 const stripAnsi = require('strip-ansi');
@@ -375,7 +375,7 @@ async function createAndroidHelperFiles() {
 }
 
 async function createIosHelperFiles() {
-  const configFile = './amplifyxc.config';
+  const configFile = './amplifytools.xcconfig';
   const awsConfigFile = './awsconfiguration.json';
   const amplifyConfigFile = './amplifyconfiguration.json';
   const amplifyDir = './amplify';
@@ -391,20 +391,18 @@ async function createIosHelperFiles() {
     configxc.envName = 'amplify';
     fs.writeFileSync(configFile, ini.stringify(configxc));
   }
-  await addFileToXcodeProj(configFile);
 
   if (!fs.existsSync(awsConfigFile)) {
     fs.writeFileSync(awsConfigFile, configJsonStr);
   }
-  await addFileToXcodeProj(awsConfigFile, true);
 
   if (!fs.existsSync(amplifyConfigFile)) {
     fs.writeFileSync(amplifyConfigFile, configJsonStr);
   }
-  await addFileToXcodeProj(amplifyConfigFile, true);
 
+  // add models and schema
   if (fs.existsSync(amplifyDir)) {
-    await addFileToXcodeProj(amplifyDir);
+    await addDataStoreFiles();
   }
 }
 
@@ -462,10 +460,10 @@ async function showIOSHelpText() {
   console.log();
   console.log(chalk.green('Some next steps:'));
   console.log(
-    'Setting "modelgen" to true in amplifyxc.config will allow you to generate models/entities for your GraphQL models in your next xcode build',
+    'Setting "modelgen" to true in amplifytools.xcconfig will allow you to generate models/entities for your GraphQL models in your next xcode build',
   );
   console.log(
-    'Setting "push" to true in the amplifyxc.config will build all your local backend resources and provision them in the cloud in your next xcode build',
+    'Setting "push" to true in the amplifytools.xcconfig will build all your local backend resources and provision them in the cloud in your next xcode build',
   );
   console.log('');
 }
--- a/packages/amplify-app/src/xcodeHelpers.js
+++ b/packages/amplify-app/src/xcodeHelpers.js
@@ -1,7 +1,7 @@
-const xcode = require('xcode');
+const fs = require('fs-extra');
 const glob = require('glob');
 const path = require('path');
-const fs = require('fs-extra');
+const xcode = require('xcode');
 
 /**
  * @typedef {Object} PBXGroup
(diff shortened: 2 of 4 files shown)


drochetti:
Screen Shot 2020-05-13 at 7 13 00 PM

These changes make sure that the Xcode project will have the generated files added and grouped correctly.

There's a PR on the iOS repo to leverage this changes: https://github.com/aws-amplify/amplify-ios/pull/445

undefobj:
@nikhname should we move the config file for push under this new AmplifyConfig directory? That way it keeps project root clean.

drochetti:
> @nikhname should we move the config file for push under this new AmplifyConfig directory? That way it keeps project root clean.

no need, the `AmplifyConfig` and `AmplifyModels` are not actual folders. They are just logical Xcode views to group things.

packages/amplify-app/src/xcodeHelpers.js
```diff
@@ -15,42 +40,143 @@ async function getXcodeProjectDir() {
...
+ * @returns {PBXGroupRef|undefined}
+ */
+function getGroupByName(project, name) {
+  /** @type PBXGroupRef */
+  let rootGroup = null;
+
+  const groups = Object.entries(project.hash.project.objects.PBXGroup);
+  for (let i = 0; i < groups.length; i++) {
+    const [key, value] = groups[i];
+    // console.log(value);
```
wooj2, line 56: remove console.log?

packages/amplify-app/src/xcodeHelpers.js
```diff
@@ -15,42 +40,143 @@ async function getXcodeProjectDir() {
...
+ */
+function groupHasFile(group, name) {
+  return group.pbxGroup.children.filter(child => child.comment === name).length > 0;
+}
+
+/**
+ * @public
+ * @return {Promise<void>}
+ */
+async function addDataStoreFiles() {
```
wooj2, line 113: We seem to be adding more than just dataStoreFiles in this function (but the function name is addDataStoreFiles).... maybe we should rename this function to something like addAmplifyFiles() which calls smaller functions like addGeneratedModels, addConfigGroup, etc..

packages/amplify-app/src/index.js
```diff
@@ -391,20 +391,18 @@ async function createIosHelperFiles() {
...
     fs.writeFileSync(awsConfigFile, configJsonStr);
   }
-  await addFileToXcodeProj(awsConfigFile, true);
 
   if (!fs.existsSync(amplifyConfigFile)) {
     fs.writeFileSync(amplifyConfigFile, configJsonStr);
   }
-  await addFileToXcodeProj(amplifyConfigFile, true);
 
+  // add models and schema
```
wooj2, line 403: Technically we are adding more than just models and schema by calling this function (we are adding the a*config.json files as well)... we should probably update this comment to be more accurate or just remove the comment.

packages/amplify-app/src/xcodeHelpers.js
drochetti: amplify-app is not resilient enough to work with things outside of `Amplify.DataStore`. I can rename the function, but I'm reluctant to break it down into smaller functions until we actually need because of Promises/Error Handlings/etc.

My goal in mentioning DataStore here is find-ability, `amplify-app` is primarily used for DataStore but there's no mention of it.

packages/amplify-app/src/xcodeHelpers.js
```diff
@@ -15,42 +40,143 @@ async function getXcodeProjectDir() {
...
+  const projectDir = getXcodeProjectDir();
+  const rootDir = path.resolve(projectDir, '..', '..');
+  const project = xcode.project(projectDir);
+  return new Promise((resolve, reject) => {
+    const schemaFilePattern = path.join(rootDir, 'amplify', 'backend', 'api', '*', 'schema.graphql');
+    const [schemaFile] = glob.sync(schemaFilePattern);
+    if (!schemaFile) {
+      reject(new Error('schema.graphql file not found'));
+      return;
+    }
```
yuth, line 132: Should we fail if we dont find the schema file?

packages/amplify-app/src/xcodeHelpers.js
```diff
@@ -15,42 +40,143 @@ async function getXcodeProjectDir() {
...
+/**
+ * @public
+ * @return {Promise<void>}
+ */
+async function addDataStoreFiles() {
+  const projectDir = getXcodeProjectDir();
+  const rootDir = path.resolve(projectDir, '..', '..');
+  const project = xcode.project(projectDir);
+  return new Promise((resolve, reject) => {
+    const schemaFilePattern = path.join(rootDir, 'amplify', 'backend', 'api', '*', 'schema.graphql');
```
yuth, line 127: Amplify CLI also support multiple schema file as long as they are inside `schema` directory. So we should update this glob pattern to reflect that

packages/amplify-app/src/xcodeHelpers.js
```diff
@@ -15,42 +40,143 @@ async function getXcodeProjectDir() {
...
+        const amplifyConfigGroup = getOrCreateGroup(project, 'AmplifyConfig');
+
+        // add the amplifytools config file
+        project.addFile('./amplifytools.xcconfig', amplifyConfigGroup.uuid);
+
+        // adding resources (i.e. files that are added to the app bundle)
+        project.addPbxGroup([], 'Resources', 'Resources', '"<group>"');
+        project.addResourceFile('amplifyconfiguration.json', null, amplifyConfigGroup.uuid);
+        project.addResourceFile('awsconfiguration.json', null, amplifyConfigGroup.uuid);
+        project.removePbxGroup('Resources');
```
yuth, line 173: Why are we removing this?

packages/amplify-app/src/xcodeHelpers.js
drochetti: updated to `addAmplifyFiles()` but kept the implementation, let me know what you think

packages/amplify-app/src/xcodeHelpers.js
drochetti: I assume so because without a schema the project is in an invalid state.

packages/amplify-app/src/xcodeHelpers.js
drochetti: Can you clarify that? What would be the correct pattern?

packages/amplify-app/src/xcodeHelpers.js
drochetti: This is the same logic we had before, just moved from the `index.js` here.

nikhname: The Resources group is added so that we can add the resource files, the group is unneeded outside of this operation so its removed. It's a limitation of the cordova xcode library that the group is required in the first place

packages/amplify-app/src/xcodeHelpers.js
```diff
@@ -15,42 +40,152 @@ async function getXcodeProjectDir() {
...
+
+  const groups = Object.entries(project.hash.project.objects.PBXGroup);
+  for (let i = 0; i < groups.length; i++) {
+    const [key, value] = groups[i];
+    if (typeof value !== 'string') {
+      // only the root pbx group can have no name, path or description
+      const isRoot = name === undefined && value.name === undefined && value.path === undefined;
+      const matchesName = typeof name === 'string' && (value.name === name || value.path === name);
+
+      if (isRoot || matchesName) {
```
nikhname, line 61: Couldn't this lead to this function returning the root group when you're looking for a group that matches a certain name?

packages/amplify-app/src/xcodeHelpers.js
nikhname: We may be extending this plugin to allow customers to scaffold full amplify projects in the future (not just Datastore), so we may want to allow the schema to be removed

packages/amplify-app/src/xcodeHelpers.js
drochetti: No because `isRoot` checks if `name` is `unfefined` and `matchesName` checks if it's an actual `string`

packages/amplify-app/src/xcodeHelpers.js
yuth: `amplify/backend/api/*/{schema/*.graphql,schema.graphql}` should do it

packages/amplify-app/src/xcodeHelpers.js
yuth: `amplify/backend/api/*/{schema/*.graphql,schema.graphql}` should do it

PR #4285, opened 2020-05-15
fix(amplify-appsync-simulator): replace "extend" library with lodash

FilipPyrek:
There was "extend" dependency imported in
amplify-appsync-simulator/src/server/subscription/mqtt-server/server.ts but it was missing in
package.json

*Description of changes:*
In https://github.com/bboure/serverless-appsync-simulator/issues/22 I stumbled upon error `Cannot find module 'extend'`. I've traced this error to [amplify-appsync-simulator/src/server/subscription/mqtt-server/server.ts#L5](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-appsync-simulator/src/server/subscription/mqtt-server/server.ts#L5)
I found out that the dependency `extend` is not in `package.json`. 

So accordingly to @jhockett suggestion, I've replaced the `extend` library with "extend" function from `lodash`.

### DIFF
--- a/packages/amplify-appsync-simulator/src/server/subscription/mqtt-server/server.ts
+++ b/packages/amplify-appsync-simulator/src/server/subscription/mqtt-server/server.ts
@@ -2,7 +2,7 @@ import Connection from 'mqtt-connection';
 import ws from 'ws';
 import steed from 'steed';
 import pino from 'pino';
-import extend from 'extend';
+import { defaultsDeep } from 'lodash';
 import nanoid from 'nanoid';
 import { TrieListener } from './trie-listener';
 import { Client } from './client';
@@ -97,7 +97,7 @@ export class MQTTServer extends EventEmitter {
 
   constructor(options: MQTTServerOptions = {}, private callback = (err: any, data?: any) => {}) {
     super();
-    this.options = extend(true, {}, DEFAULT_OPTIONS, options);
+    this.options = defaultsDeep(options, DEFAULT_OPTIONS);
 
     this._dedupId = 0;
     this.clients = {};


jhockett:
Thanks so much for your contribution @FilipPyrek and for finding this bug. We try to limit our dependencies as much as possible though. Since we already import `lodash` in the `amplify-appsync-simulator` package, I propose we use `_.extend`

FilipPyrek:
> Thanks so much for your contribution @FilipPyrek and for finding this bug. We try to limit our dependencies as much as possible though. Since we already import `lodash` in the `amplify-appsync-simulator` package, I propose we use `_.extend` in place of the imported `extend`.
> 
> Closed the PR by accident, sorry about that!

Thanks @jhockett. Good catch! I definitely agree with the dependencies limitation. If you want I can refactor the PR so it uses `_.extend`.

FilipPyrek:
I've refactored the PR according to your suggestion @jhockett 👍

packages/amplify-appsync-simulator/src/server/subscription/mqtt-server/server.ts
```diff
@@ -97,7 +97,7 @@ export class MQTTServer extends EventEmitter {
 
   constructor(options: MQTTServerOptions = {}, private callback = (err: any, data?: any) => {}) {
     super();
-    this.options = extend(true, {}, DEFAULT_OPTIONS, options);
+    this.options = extend({}, DEFAULT_OPTIONS, options);
```
yuth, line 100: May be [`_.defaultsDeep`](https://lodash.com/docs/4.17.15#defaultsDeep) a better choice

FilipPyrek: Good choice @yuth 👍 Changed according to your suggestion.

PR #4310, opened 2020-05-19
fix(amplify-provider-awscloudformation): fix bug for no credential file

### LINKED ISSUE #4284: Support for Session-Based Credentials
**Describe the bug**
I prefer _not_ to store long-lived credentials in my `~/.aws/credentials`, to increase my security posture.

The AWS CLI supports the use of short-lived, session-scoped credentials. This topic is [discussed here](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html#using-temp-creds-sdk-cli).

I am trying to provide credentials to `amplify` CLI in the same way.

When I do `amplify init .`, the command will _fail_, if I am using environment-based session credentials:
```
Using default provider  awscloudformation
AWS access credentials can not be found.
? Setup new user No

For more information on AWS Profiles, see:
https://docs.aws.amazon.com/cli/latest/userguide/cli-multiple-profiles.html

? accessKeyId:  ASIAXWWQ3X**********
? secretAccessKey:  /fNQ4zvx+C6dh2yFAud/********************
? region:  us-east-1
init failed
Error: Profile configuration is missing for: default
    at Object.getProfiledAwsConfig (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/lib/system-config-manager.js:82:11)
    at Object.getAwsConfig (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/lib/configuration-manager.js:638:45)
    at Object.run (/usr/local/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/lib/initializer.js:23:50)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

**Amplify CLI Version**
4.13.4

**To Reproduce**
1. `amplify init .`
2. Blah blah...
3. Setup new user? No.
4. Copy paste access key / secret key. Choose `us-east-1` as region.
5. Observe it blow up

**Expected behavior**
When I say "No" to setup new user, the CLI should inspect my environment and see that I have session credentials available. It shouldn't prompt me to paste stuff.

**Screenshots**

 - Mac OS X @ 10.14.6 (18G3020)
 - Node @ v12.12.0

**Additional context**
Information about my environment:
```
cat ~/.aws/credentials | wc
       0       0       0
```
```
# Redacting the values, but they're there.
declare | awk -F\= '/^AWS_/ { print $1 }'
AWS_ACCESS_KEY_ID
AWS_SECRET_ACCESS_KEY
AWS_SESSION_TOKEN
```

UnleashedMind:
fix a bug in init workflow when there is no credentials file and the customer select not to set up
new user and directly type in credentials

fix #4284

*Issue #, if available:*

*Description of changes:*


kaustavghosh06:
Can we please add E2E tests for the broken flow?

UnleashedMind:
E2E test added

packages/amplify-e2e-tests/src/init-special-cases/index.ts
```diff
@@ -0,0 +1,85 @@
...
+    throw e;
+  } finally {
+    if (fs.existsSync(configFilePathHide)) {
+      fs.renameSync(configFilePathHide, configFilePath);
+    }
+    if (fs.existsSync(credentialsFilePathHide)) {
+      fs.renameSync(credentialsFilePathHide, credentialsFilePath);
+    }
+  }
+}
```
yuth, line 42: the better approach here is to use mock/spy instead of renaming the files, which makes one less thing to ensure.

See if you can use 
```js
const existSyncOrig = fs.existsSync;
jest.spyOn(fs, 'existsSync').mockImplementation((path) => {
   return [credentialsFilePath, configFilePath].includes(path) ? false :
 existSyncOrig(path)});

```

UnleashedMind: I tried different combinations, but using jest.spyOn does not work, seems like jest mock can not penetrate into the spawned process. 
I updated the test to use another mechanism to avoid having to rename the credential files.

yuth: I would rather do the first approach then, instead of using a custom script to invoke the CLI

UnleashedMind: Agree, got back to the original test setup.

PR #4317, opened 2020-05-19
remove util.toJson because it includes the double quotes as string

ykbryan:
*Issue #, if available:* https://github.com/aws-amplify/amplify-cli/issues/4346

*Description of changes:*
Fixing the bug to set variable as value instead of json and excludes double quotes in the s3 object key name. Refer to the screenshot below.

Screenshot 2020-05-19 22 50 06

Screenshot 2020-05-19 22 56 51


ykbryan:
Is there any question on this? 
Is this replicable? I am able to replicate this on Vue and React.

SwaySway:
Hello @ykbryan 
While removing the `$util.toJson` would solve the double quotes issue. We do want to sanitize the input coming in from the argument. We can see if we can rearrange it so we can still sanitize the input and avoid double quotes.

Sanitize Input Docs: https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html#sanitizing-inputs

ykbryan:
> Hello @ykbryan
> While removing the `$util.toJson` would solve the double quotes issue. We do want to sanitize the input coming in from the argument. We can see if we can rearrange it so we can still sanitize the input and avoid double quotes.
> 
> Sanitize Input Docs: https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html#sanitizing-inputs

I see. I followed and copied the code from documentation (https://docs.amplify.aws/cli/graphql-transformer/directives#usage-5). Can you see if you can replicate this issue and maybe provide a fix?

SwaySway:
Following up on this closing this in favor of creating a new PR to include sanitization and double quote removal.

PR #4365, opened 2020-05-24
layer update and push  flow

akshbhu:
*Description of changes:*

Lambda Layer Update  flow:
 
-Command line prompts to take user input on what to update

-Update the template file based on user input

-added basic tests for the update flow


packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
```diff
@@ -3,7 +3,7 @@ import { LayerParameters } from './utils/layerParams';
 import { supportedServices } from '../supported-services';
 import { ServiceName, provider } from './utils/constants';
 import { category as categoryName } from '../../constants';
-import { copyFunctionResources, createLayerCfnFile, createLayerFolders, createParametersFile } from './utils/storeResources';
+import { copyFunctionResources, createLayerCfnFile, createLayerFolders, createParametersFile,createLayerParametersFile,updateLayerCfnFile } from './utils/storeResources';
```
jhockett, line 6: nit: spacing

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
```diff
@@ -139,6 +141,67 @@ async function addLayerResource(
...
+  }
+  switch (service) {
+    case ServiceName.LambdaFunction:
+      return updateFunctionResource(context, category, service, parameters, resourceToUpdate);
+    case ServiceName.LambdaLayer:
+      return updateLayerResource(context, category, service, serviceConfig, parameters);
+    default:
+      throw BAD_SERVICE_ERR;
+  }
+}
```
edwardfoyle, line 158: Instead of having this switch statement here and duplicated in the addResource method above, we should consider having a different controller here: https://github.com/aws-amplify/amplify-cli/blob/lambdaLayers/packages/amplify-category-function/src/provider-utils/supported-services.ts#L20,L29 so that once we have determined the service config no more disambiguation is necessary

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -51,31 +144,34 @@ function layerNameQuestion(context: any) {
...
     {
-      type: 'list',
+      type: 'checkbox',
       name: 'layerPermissions',
       message: 'Who should have permission to use this layer?',
       choices: [
         {
           name: 'Only the current AWS account',
           value: Permissions.private,
+          checked: _.includes(params.layerPermissions, Permissions.private) ? true : false,
```
edwardfoyle, line 157: ternary is redundant. `_.includes()` returns true / false

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -51,31 +144,34 @@ function layerNameQuestion(context: any) {
   ];
 }
 
-function layerPermissionsQuestion() {
+function layerPermissionsQuestion(params: Partial<LayerParameters>) {
```
edwardfoyle, line 147: if this function only needs the layerPermissions, that should be the only input

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
```diff
@@ -186,7 +185,14 @@ async function getSelectionsFromContributors<T>(
...
         choices: selections,
-        default: selectionOptions.listOptionsField === 'runtimes' ? 'nodejs' : undefined,
+        default:
+          selectionOptions.service === ServiceName.LambdaFunction
+            ? selectionOptions.listOptionsField === 'runtimes'
+              ? 'nodejs'
+              : undefined
+            : selectionOptions.runtimeState === undefined
+            ? undefined
+            : prevruntime,
```
edwardfoyle, line 195: This should be its own function and use if statements rather than ternary

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
```diff
@@ -162,6 +149,18 @@ async function getSelectionsFromContributors<T>(
...
+    prevruntime = selections
+      .filter(runtime => {
+        for (let val of selectionOptions.runtimeState) {
+          if (val === runtime.name) return runtime;
+        }
+      })
+    prevruntime.forEach(selection =>{
+      _.assign(selection,{checked : true})
+    });
+  }
```
edwardfoyle, line 163: this should be it's own function and be called from within the else block below. It looks like you're trying to set {checked: true} on the selections that appear in runtimeState in which case I think all of this simplifies to
```js
selections
  .filter(selection => selectionOptions.runtimeState.includes(selection.name))
  .forEach(selection => _.assign(selection, {checked: true})
```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
```diff
@@ -195,6 +201,7 @@ async function getSelectionsFromContributors<T>(
   if (!Array.isArray(selection)) {
     selection = [selection];
   }
+  selection = _.uniq(selection); // remove duplicates
```
edwardfoyle, line 204: how could there be duplicates?

packages/amplify-category-function/src/commands/function/update.ts
```diff
@@ -19,7 +22,15 @@ module.exports = {
         }
         return providerController.updateResource(context, categoryName, result.service);
       })
-      .then(() => context.print.success('Successfully updated resource'))
+      .then(result => {
+        const { print } = context;
+        const { name, service } = result;
+        if (service === ServiceName.LambdaFunction) {
```
jhockett, line 28: This if/else statement is currently doing the same thing for both services

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
```diff
@@ -238,3 +246,7 @@ interface ListOption {
   name: string;
   value: string;
 }
+
+function isLayerParameter(params: Partial<LayerParameters> | Partial<FunctionParameters> ): params is Partial<LayerParameters> {
+  return (((params as Partial<LayerParameters>) !== undefined) && ((params as Partial<LayerParameters>).runtimes !== undefined));
```
edwardfoyle, line 251: nit don't need the type assertion to check for undefined. Also, I would say this is a typeguard for `Pick<LayerParameters, 'runtimes'>` rather than `Partial<LayerParameters>`

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
```diff
@@ -139,6 +141,67 @@ async function addLayerResource(
...
+): Promise<void> {
+  if (!serviceConfig) {
+    throw `amplify-category-function is not configured to provide service type ${service}`;
+  }
+
+  if (!parameters) {
+    parameters = {};
+    parameters.providerContext = {
+      provider: provider,
+      service: ServiceName.LambdaLayer, // TODO: switch to using service
```
jhockett, line 175: I'd change this to `service: service` and remove the TODO comment

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,28 +1,121 @@
...
+      }
+    }
+    // if (islayerVersionChanged) {
+    //   _.assign(templateParameters, await inquirer.prompt(layerVersionQuestion(context)));
+    // }
+  }
+  return templateParameters;
+}
+
+// function layerVersionQuestion(context: any) {
```
edwardfoyle, line 94: remove unused code here and above

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/constants.ts
```diff
@@ -1,5 +1,10 @@
-export const chooseServiceMessage = 'Select which capability you want to add:';
+import { category } from '../../../constants';
+
+export const categoryName = category;
+
+export const chooseServiceMessage = ['Select which capability you want to add:', 'Select which capability you want to update:'];
```
edwardfoyle, line 5: seems like the update message should be a different constant

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,28 +1,121 @@
...
+
+  if (resources.length === 0) {
+    context.print.error('No Lambda Layer resource to update. Please use "amplify add function" command to create a new Function');
+    process.exit(0);
+    return;
+  }
+  const resourceQuestion = [
+    {
+      name: 'resourceName',
+      message: 'Please select the Lambda Layer you would want to update',
```
jhockett, line 46: remove 'would' from this sentence.
`message: 'Please select the Lambda Layer you want to update',`

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -102,6 +102,25 @@ export function createLayerCfnFile(context, parameters, layerDirPath) {
...
+    path.join(layerDirPath, parameters.layerName + '-awscloudformation-template.json'),
+    generateLayerCfnObj(parameters),
+    true,
+  );
+  context.amplify.updateamplifyMetaAfterResourceUpdate(categoryName, parameters.layerName,'runtimes', parameters.runtimes);
+
+}
+
+export function createLayerParametersFile(context,parameters,layerDirPath){
+  let parametersFileName = 'layer-parameters.json';
```
edwardfoyle, line 116: this should use the const you defined

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -102,6 +102,25 @@ export function createLayerCfnFile(context, parameters, layerDirPath) {
...
+  context.amplify.writeObjectAsJson(
+    path.join(layerDirPath, parameters.layerName + '-awscloudformation-template.json'),
+    generateLayerCfnObj(parameters),
+    true,
+  );
+  context.amplify.updateamplifyMetaAfterResourceUpdate(categoryName, parameters.layerName,'runtimes', parameters.runtimes);
+
+}
+
+export function createLayerParametersFile(context,parameters,layerDirPath){
```
edwardfoyle, line 115: nit run the linter

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,28 +1,121 @@
...
+      type: 'list',
+      choices: resources,
+    },
+  ];
+  const resourceAnswer = await inquirer.prompt(resourceQuestion);
+  templateParameters.layerName = String(resourceAnswer.resourceName);
```
jhockett, line 52: What's the purpose of this type conversion? When is `resourceAnswer.resourceName` not a string?

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -102,6 +102,25 @@ export function createLayerCfnFile(context, parameters, layerDirPath) {
...
+  let parametersFileName = 'layer-parameters.json';
+  fs.ensureDirSync(layerDirPath);
+  const parametersFilePath = path.join(layerDirPath, parametersFileName);
+  const currentParameters = fs.existsSync(parametersFilePath) ? context.amplify.readJsonFile(parametersFilePath) : {};
+  const jsonString = JSON.stringify({ ...currentParameters, ...parameters }, null, 4);
+  fs.writeFileSync(parametersFilePath, jsonString, 'utf8');
```
edwardfoyle, line 121: use `context.amplify.writeObjectAsJson()`

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -102,6 +102,25 @@ export function createLayerCfnFile(context, parameters, layerDirPath) {
...
+  context.amplify.updateamplifyMetaAfterResourceUpdate(categoryName, parameters.layerName,'runtimes', parameters.runtimes);
+
+}
+
+export function createLayerParametersFile(context,parameters,layerDirPath){
+  let parametersFileName = 'layer-parameters.json';
+  fs.ensureDirSync(layerDirPath);
+  const parametersFilePath = path.join(layerDirPath, parametersFileName);
+  const currentParameters = fs.existsSync(parametersFilePath) ? context.amplify.readJsonFile(parametersFilePath) : {};
+  const jsonString = JSON.stringify({ ...currentParameters, ...parameters }, null, 4);
```
edwardfoyle, line 120: does this spread operation work if someone removes all of the layerPermissions?

packages/amplify-cli/src/extensions/amplify-helpers/service-select-prompt.js
```diff
@@ -47,10 +47,17 @@ function serviceQuestionWalkthrough(context, supportedServices, category, custom
     });
   }
 
+  let customFlowQuestion;
+  if(context.input.command === 'add'){
+    customFlowQuestion = customQuestion[0];
+  }else{
+    customFlowQuestion = customQuestion[1];
+  }
+
```
edwardfoyle, line 56: get rid of this in favor of separate constants as mentioned above

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,28 +1,121 @@
...
+
+  // get layer-patameters
```
jhockett, line 54: ultra nit: typo in comment

packages/amplify-e2e-tests/src/__tests__/layer.test.ts
```diff
@@ -20,4 +20,28 @@ describe('amplify add lambda layer', () => {
     await initJSProjectWithProfile(projRoot, {});
     await addLayer(projRoot);
   });
+
+  it.only('init a project and add/update simple layer', async () => {
```
edwardfoyle, line 24: probably remove .only() unless you intend for the other tests to be ignored for now

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -76,20 +29,81 @@ export default function generateLayerCfnObj(parameters) {
...
-          Ref: 'LambdaLayerPermission',
-        },
-      },
     },
   };
+  const POLICY_RETAIN = DeletionPolicy.Retain;
+  const layer = new Lambda.LayerVersion({
+    CompatibleRuntimes: parameters.runtimes.map(runtime => runtime.cloudTemplateValue),
+    Content: {
+      S3Bucket: '<S3Bucket>',
```
jhockett, line 38: This was changed in my PR to use refs (see above diff)

```js
    Content: {
      S3Bucket: Fn.Ref('deploymentBucketName'),
      S3Key: Fn.Ref('s3Key'),
    },
```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -76,20 +29,81 @@ export default function generateLayerCfnObj(parameters) {
...
+  cfnObj.Resources["LambdaLayer"] = layer;
 
-    cfnObj.Resources = { ...cfnObj.Resources, ...layerVersionPermission };
-  }
+  parameters.layerPermissions.forEach(permissions => {
+   if (permissions === Permissions.awsAccounts) {
+      assignLayerPermissionsAccounts(cfnObj,parameters,permissions);
+    }else{
+      assignLayerPermissions(cfnObj,parameters,permissions)
+    }
```
jhockett, line 55: nit: please fix spacing, indentation, and missing semi-colon

packages/amplify-e2e-core/src/categories/lambda-layer.ts
```diff
@@ -32,7 +32,67 @@ export function addLayer(cwd: string, settings?: any) {
...
+          '/*',
+      );
+      chain.wait(layerRuntimeDirRegex);
+    }
+
+    chain
+      .wait('Include any files you want to share across runtimes in this folder:')
+      .wait('"amplify function update <function-name>" - configure a function with this Lambda layer')
+      .wait('"amplify push" builds all of your local backend resources and provisions them in the cloud')
+      .sendEof()
```
jhockett, line 61: Ideally we would refactor this duplicated test code into its own function

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
akshbhu: fixed

packages/amplify-category-function/src/commands/function/update.ts
akshbhu: fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
akshbhu: done

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
akshbhu: fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
akshbhu: Ya , I didnt checked that earlier , thanks
fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
akshbhu: fixed

packages/amplify-category-function/src/commands/function/update.ts
```diff
@@ -19,7 +22,15 @@ module.exports = {
...
+          print.success(`Successfully added resource ${name} locally.`);
+        } else if (service === ServiceName.LambdaLayer) {
+          print.success(`Successfully added resource ${name} locally.`);
+        }
```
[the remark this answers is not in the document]
akshbhu: fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
akshbhu: fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
akshbhu: This assigns permissions in the layer parameters in this function. Thats why Params is passed

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
akshbhu: done

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
akshbhu: Thanks for this Edward. Done!!

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
akshbhu: This is because earlier I was doing some other implementation.
Thanks for catching this . Fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
akshbhu: If I am using params.runtime , its giving error as runtimes is not part of functionParameters.
Thats  why I put a typeguard to differentiate between layer and function parameters.

Is there any other way to do this?

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
akshbhu: done!!!

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
akshbhu: done!!!

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
akshbhu: done

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
akshbhu: Ohh Nice catch , fixed

packages/amplify-cli/src/extensions/amplify-helpers/service-select-prompt.js
akshbhu: fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
akshbhu: done

packages/amplify-e2e-core/src/categories/lambda-layer.ts
akshbhu: done

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
akshbhu: Do you mean I should changed it as LambdaController and Layer controller for the add/update function respectively?

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
akshbhu: This assigns permissions in the layer parameters in this function. Thats why Params is passed

edwardfoyle: I'm saying it looks like this function only uses params.layerPermissions so why not just pass that in directly?

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
edwardfoyle: You need the typeguard for the second check, just not the first one

packages/amplify-cli/src/extensions/amplify-helpers/service-select-prompt.js
```diff
@@ -47,10 +47,12 @@ function serviceQuestionWalkthrough(context, supportedServices, category, custom
     });
   }
 
+  let customFlowQuestion = customQuestion;
```
edwardfoyle, line 50: no need to assign this variable

packages/amplify-e2e-core/src/categories/lambda-layer.ts
```diff
@@ -154,3 +110,30 @@ const getRuntimeDisplayName = (runtime: LayerRuntimes) => {
...
+  chain.wait(layerRuntimeDirRegex);
+}
+
+chain
+  .wait('Include any files you want to share across runtimes in this folder:')
+  .wait('"amplify function update <function-name>" - configure a function with this Lambda layer')
+  .wait('"amplify push" builds all of your local backend resources and provisions them in the cloud')
+  .sendEof()
+  return chain;
+}
```
jhockett, line 139: nit: indentation

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
edwardfoyle: Yeah, like instead of using the same controller and having switch / if statement within each function to switch behavior, we should just have two separate controllers referenced in the service config

packages/amplify-cli/src/extensions/amplify-helpers/service-select-prompt.js
akshbhu: fixed

packages/amplify-category-function/src/commands/function/update.ts
```diff
@@ -1,16 +1,19 @@
-import { category as categoryName } from '../../constants';
 import { supportedServices } from '../../provider-utils/supported-services';
+import { chooseServiceMessageUpdate } from '../../provider-utils/awscloudformation/utils/constants';
+import { category as categoryName } from '../../constants';
 
 const subcommand = 'update';
 
+let options;
```
jhockett, line 7: `options` looks unused in this file

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -51,31 +141,34 @@ function layerNameQuestion(context: any) {
...
           value: Permissions.awsOrg,
+          checked: _.includes(params, Permissions.awsOrg),
         },
         {
           name: 'Public (everyone on AWS can use this layer)',
           value: Permissions.public,
+          checked: _.includes(params, Permissions.public),
         },
       ],
-      default: Permissions.private,
```
jhockett, line 78: I think it makes more sense if we keep the default as private and remove private as an option.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
```diff
@@ -238,3 +228,27 @@ interface ListOption {
...
+}
+
+function defaultSelection(selectionOptions : PluginSelectionOptions<FunctionRuntimeCondition>,selections){
+  if(selectionOptions.service === ServiceName.LambdaFunction){
+    if(selectionOptions.listOptionsField === 'runtimes'){
+      return 'node.js'
```
jhockett, line 239: Previously, `'nodejs'` was being passed. Did you mean to add the `.`?

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
```diff
@@ -186,7 +173,8 @@ async function getSelectionsFromContributors<T>(
         name: 'selection',
         message: selectionOptions.selectionPrompt,
         choices: selections,
-        default: selectionOptions.listOptionsField === 'runtimes' ? 'nodejs' : undefined,
+        default:
+          defaultSelection(selectionOptions,selections),
```
jhockett, line 177: nit: should be 1 line

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -474,6 +484,46 @@ function updateIdPRolesInNestedStack(context, nestedStack, authResourceName) {
...
+  const distDir = path.join(resourcePath, 'dist');
+  if (!fs.existsSync(distDir)) {
+    fs.mkdirSync(distDir);
+  }
+  const destination = path.join(distDir,zipFilename );
+  // verify the folder structure
+
+  // create write stream
+  let zip = ziparchiver.create('zip');
+  let output = fs.createWriteStream(destination);
```
jhockett, line 504: It doesn't look like these variables get modified, `const` might be more appropriate

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -474,6 +484,46 @@ function updateIdPRolesInNestedStack(context, nestedStack, authResourceName) {
...
+  const layerVersionArn = layerObj.parameters.layerVersionsArray[0];
+  const zipFilename = 'latest-build.zip';
+
+  // package the function
+  const distDir = path.join(resourcePath, 'dist');
+  if (!fs.existsSync(distDir)) {
+    fs.mkdirSync(distDir);
+  }
+  const destination = path.join(distDir,zipFilename );
+  // verify the folder structure
```
jhockett, line 500: nit: remove comment since we decided to not verify structure in case user wants to configure their own custom paths in the `/opt` folder

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
```diff
@@ -2,15 +2,17 @@ import { FunctionRuntime, ProviderContext } from 'amplify-function-plugin-interf
...
   providerContext: ProviderContext;
   authorizedAccountIds?: string;
   authorizedOrgId?: string;
+  layerVersion: string;
+  layerVersionsArray?: number[];
 };
 
 export enum Permissions {
-  private = 'private',
+  private = 'private',// remove
```
jhockett, line 15: nit: fix this comment

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -99,9 +99,27 @@ export function createLayerCfnFile(context, parameters, layerDirPath) {
     providerPlugin: parameters.providerContext.provider,
     service: parameters.providerContext.service,
     runtimes: parameters.runtimes,
+    build: true,
```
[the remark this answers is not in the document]
akshbhu: Its for amplify push command , So that we can use some part of functions flow to package layer

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
akshbhu: done

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -474,6 +484,46 @@ function updateIdPRolesInNestedStack(context, nestedStack, authResourceName) {
 
   Object.assign(nestedStack.Resources, idpUpdateRoleCfn);
 }
+ function packageLayer(context,resource){
+  const resourcePath = path.join(context.amplify.pathManager.getBackendDirPath(), resource.category, resource.resourceName);
+  //const resourcePathSrc = path.join(resourcePath,'src');
```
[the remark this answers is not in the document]
akshbhu: done

packages/amplify-provider-awscloudformation/lib/push-resources.js
akshbhu: done

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/functionPluginLoader.ts
akshbhu: fixed

packages/amplify-provider-awscloudformation/lib/push-resources.js
akshbhu: Nice catch!! , fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,122 +1,134 @@
...
+    return 0;
+  }
+  const parametersFilePath = path.join(resourceDirPath, layerParametersFileName);
+  let prevParameters =  context.amplify.readJsonFile(parametersFilePath);
+  const prevlayerData = layerMetadataFactory(prevParameters.parameters);
+  let latestVersionPushed = prevlayerData.listVersions().reduce((a,b)=>{
+    return Math.max(a,b);
+  });
+  return latestVersionPushed;
+}
```
jhockett, line 134: nit: no newline at end of file

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,122 +1,134 @@
...
+    templateParameters.runtimes = runtimeReturn.map(val => val.runtime);
+  } else {
+    islayerVersionChanged = false;
+  }
+
+  // get the latest version from #currentcloudbackend
+  let latestVersionPushed =  getLastestVersionPushed(context,templateParameters.layerName);
+  let latestVersion = layerData.listVersions().reduce((a,b)=>{
+    return Math.max(a,b);
+  });
```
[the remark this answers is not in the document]
akshbhu: done!! thanks

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
```diff
@@ -160,9 +166,71 @@ export async function updateResource(
...
+
+  print.info('');
+  print.info('Include any files you want to share across runtimes in this folder:');
+  print.info(`amplify/backend/function/${parameters.layerName}/opt/data`);
+  print.info('"amplify function update <function-name>" - configure a function with this Lambda layer');
+  print.info('"amplify push" builds all of your local backend resources and provisions them in the cloud');
+}
+
+export async function updateFunctionResource(context, category, service, parameters, resourceToUpdate) {
+  let answers;
```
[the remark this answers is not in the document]
akshbhu: fixed

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -48,7 +48,7 @@ export async function createWalkthrough(
 /**
  * FIXME this function needs to be refactored so it doesn't have side-effects of writing to CFN files
  */
-export async function updateWalkthrough(context, lambdaToUpdate?: string) {
+export async function updateWalkthrough(context, parameters?:Partial<FunctionParameters>, lambdaToUpdate?: string) {
```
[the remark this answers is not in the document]
akshbhu: this is added to maintain same function definition in function and layers walkthrough

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
```diff
@@ -181,7 +249,7 @@ export async function updateResource(
     }
     saveMutableState(context, parameters);
   } else {
-    parameters = await serviceConfig.walkthroughs.updateWalkthrough(context, resourceToUpdate);
+    parameters = await serviceConfig.walkthroughs.updateWalkthrough(context,parameters,resourceToUpdate);
```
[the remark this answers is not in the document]
akshbhu: done

packages/amplify-e2e-core/src/categories/lambda-layer.ts
```diff
@@ -132,4 +154,31 @@ function getLayerRuntimeInfo(runtime: LayerRuntimes) {
     default:
       throw new Error(`Invalid runtime value: ${runtime}`);
   }
+};
+
+function printFlow(chain : ExecutionContext , settings : any ,layerDirRegex , runtimeDisplayNames){
```
edwardfoyle, line 159: you don't need to wait on every line. It would be sufficient to just wait on the last 2 lines which would eliminate a lot of that regex stuff

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerHelpers.ts
```diff
@@ -0,0 +1,156 @@
...
+    let obj : LayerPermission
+      if(permission === Permissions.public){
+        obj = {
+          type : Permissions.public,
+        }
+      }
+      else if(permission === Permissions.awsOrg){
+        obj = {
+          type : Permissions.awsOrg,
+          orgs : parameters.authorizedOrgId.split(',')
```
edwardfoyle, line 138: There shouldn't be tight coupling between the name of the prompts and the object properties here. Also, we should never be dealing with parsing input strings except for immediately after receiving customer input. You can use inquirers [filter feature](https://www.npmjs.com/package/inquirer#question) to convert the input to an array, set or whatever

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerHelpers.ts
```diff
@@ -0,0 +1,156 @@
...
+            }
+            set.add(orgId);
+          })
+          return true;
+        },
+        default : defaultorgs.toString()
+    }];
+}
+
+export function createVersionsMap(parameters : Partial<LayerParameters>,version : string){
```
edwardfoyle, line 125: this method only needs parameters.layerPermissions

Also, why are we representing the permissions in 2 different ways? Seems like we should refactor so a conversion like this isn't necessary

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
```diff
@@ -1,12 +1,15 @@
 import { FunctionRuntime, ProviderContext } from 'amplify-function-plugin-interface';
+import _ from 'lodash';
+import path from 'path';
 
 export type LayerParameters = {
   layerName: string;
-  layerPermissions: Permissions;
+  layerPermissions: Permissions[];
```
edwardfoyle, line 7: why do we have both layerPermissions and layerVersionMap? Doesn't the layerVersionMap contain the permissions for each version (including the current one)?

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -475,6 +485,54 @@ function updateIdPRolesInNestedStack(context, nestedStack, authResourceName) {
...
+      reject(new Error('Failed to zip code.'));
+    });
+
+    zip.pipe(output);
+    let folders = glob.sync(resourcePath + '/*');
+    folders.forEach(folder =>{
+      if(fs.lstatSync(folder).isDirectory() && path.basename(folder) !== 'dist'){
+        zip.directory(folder,path.basename(folder));
+      }
+    })
```
edwardfoyle, line 521: would write as
```js
glob.sync(resourcePath + '/*')
  .filter(folder => fs.lstatSync(folder).isDirectory())
  .filter(folder => path.basename(folder) !== 'dist')
  .forEach(folder => zip.directory(folder, path.basename(folder)))
```

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -475,6 +485,54 @@ function updateIdPRolesInNestedStack(context, nestedStack, authResourceName) {
   Object.assign(nestedStack.Resources, idpUpdateRoleCfn);
 }
 
+ function packageLayer(context,resource){
```
edwardfoyle, line 488: this function should really be in the function category plugin. Ideally it wouldn't depend directly on a category, but it already depends on auth and predictions, so may as well add functions

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -76,20 +64,100 @@ export default function generateLayerCfnObj(parameters) {
...
+  return result;
+}
+
+function joinWithEnv(separator: string, stringToJoin: string) {
+  return Fn.If('HasEnvironmentParameter', Fn.Join(separator, [stringToJoin, Fn.Ref('env')]), stringToJoin);
+}
+
+function createLayerversionArn(layerData, layerName, version, flag) {
+  //arn:aws:lambda:us-west-2:136981144547:layer:layers089e3f8b-dev:1
+  if (flag) {
```
jhockett, line 153: `flag` should have a more descriptive name

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
```diff
@@ -15,3 +18,165 @@ export enum Permissions {
...
+          accounts = permission.accounts;
+        }
+      });
+    }
+    return accounts;
+  }
+
+  listOrgAccess(): string[] {
+    let orgs: string[] = [];
+    if (!(this.permissions === undefined || this.permissions.length == 0)) {
```
jhockett, line 133: `!(this.permissions === undefined || this.permissions.length == 0)` uses == instead of ===.

You can simplify this to:
`this.permissions && this.permissions.length > 0`

packages/amplify-category-function/src/provider-utils/supportedServicesType.ts
```diff
@@ -16,7 +16,7 @@ export interface ServiceConfig<T> {
 
 export interface WalkthroughProvider<T> {
   createWalkthrough: (context: any, params: Partial<T>) => Promise<Partial<T>>;
-  updateWalkthrough: (context: any, resourceToUpdate?: string) => Promise<Partial<T>>;
+  updateWalkthrough: (context: any, params?: Partial<T>,resourceToUpdate?: string) => Promise<Partial<T>>;
```
jhockett, line 19: nit: spacing

packages/amplify-e2e-core/src/categories/lambda-layer.ts
jhockett: It's there for validation. If someone makes a text change by accident, the tests will catch it.

packages/amplify-e2e-core/src/categories/lambda-layer.ts
```diff
@@ -128,4 +150,31 @@ function getLayerRuntimeInfo(runtime: LayerRuntimes) {
...
```
jhockett, line 155: nit: spacing in function definition

packages/amplify-e2e-core/src/utils/sdk-calls.ts
```diff
@@ -74,6 +74,16 @@ export const getFunction = async (functionName: string, region: string) => {
   return await service.getFunction({ FunctionName: functionName }).promise();
 };
 
+export const getLayerVersion = async (functionArn: string, region: string) => {
+  const service = new Lambda({ region });
+  return await service.getLayerVersionByArn({ Arn: functionArn }).promise();
+};
+
+export const listVersions = async(layerName: string , region: string) =>{
```
jhockett, line 82: nit: spacing

packages/amplify-graphql-types-generator/test/__snapshots__/loading.ts.snap
```diff
@@ -384,6 +384,6 @@ Object {
 `;
 
 exports[`Validation should throw a helpful message when a file has invalid gql snippets 1`] = `
-"Could not parse graphql operations in test/fixtures/misc/invalid-gqlQueries.js
+"Could not parse graphql operations in packages/amplify-graphql-types-generator/test/fixtures/misc/invalid-gqlQueries.js
```
jhockett, line 387: why is this here?

packages/amplify-e2e-core/src/categories/lambda-layer.ts
edwardfoyle: If you want to check all the lines except that regex one that would be fine. But that regex is a lot of complexity for marginal test benefit in my opinion. It could also easily be a small unit test around the printout of the method rather than in the e2e tests

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -220,54 +222,62 @@ function packageResources(context, resources) {
...
-          cfnMeta.Resources.LambdaFunction.Properties.Code = {
-            S3Bucket: s3Bucket,
-            S3Key: s3Key,
-          };
-        }
+    return resource.service === 'LambdaFunction'
+      ? buildResource(context, resource)
+      : packageLayer(context, resource)
+          .then(result => {
+            // Upload zip file to S3
```
jhockett, line 229: Need parenthesis around the ternary operation, otherwise the `.then()` functions will only be called on `packageLayer()`

```ts
 return (resource.service === 'LambdaFunction'
      ? buildResource(context, resource)
      : packageLayer(context, resource))
          .then(result => {
```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
```diff
@@ -15,3 +18,165 @@ export enum Permissions {
...
+          accounts: permission.accounts,
+        };
+        this.permissions.push(accountsPermissions);
+      }
+    });
+  }
+
+  listAccoutAccess(): string[] {
+    let accounts: string[] = [];
+    if (!(this.permissions === undefined || this.permissions.length == 0)) {
```
edwardfoyle, line 121: this check isn't necessary because the constructor initializes the permissions array. Also not necessary in any of the other methods below

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
edwardfoyle: as mentioned above, you don't need this check at all

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
```diff
@@ -15,3 +18,165 @@ export enum Permissions {
...
+  isPrivate(): boolean {
+    let retVal: boolean = false;
+    if (!(this.permissions === undefined || this.permissions.length == 0)) {
+      this.permissions.forEach(permission => {
+        if (permission.type === Permissions.private) {
+          retVal = true;
+        }
+      });
+    }
+    return retVal;
```
edwardfoyle, line 152: simplifies to
```js
return !!this.permissions.map(perm => perm.type).find(type => type === Permissions.private)
```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
```diff
@@ -15,3 +18,165 @@ export enum Permissions {
...
+  }
+
+  isPublic(): boolean {
```
edwardfoyle, line 155: simplifies to the same thing above, but replace private with public

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
```diff
@@ -15,3 +18,165 @@ export enum Permissions {
...
+        }
+      });
+    }
+    return retVal;
+  }
+}
+
+export const layerMetadataFactory: LayerMetadataFactory = (context: any, layerName: string, isPush: boolean = false): LayerMetadata => {
+  let projectBackendDirPath;
+  if (!isPush) {
```
edwardfoyle, line 170: better to use positive conditional whenever possible. So revers this to be `if (isPush)`

Better yet, why not have this method take in the backendDirPath directly instead of the isPush flag. That way no if statement is required

packages/amplify-graphql-types-generator/test/__snapshots__/loading.ts.snap
edwardfoyle: Yeah this needs to be reverted. It will cause graphql tests to fail

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -220,54 +222,62 @@ function packageResources(context, resources) {
...
-          };
-        } else {
-          cfnMeta.Resources.LambdaFunction.Properties.Code = {
-            S3Bucket: s3Bucket,
-            S3Key: s3Key,
-          };
-        }
+    return resource.service === 'LambdaFunction'
+      ? buildResource(context, resource)
+      : packageLayer(context, resource)
```
edwardfoyle, line 227: need parens around this so that the then() clauses get attached to the buildResource or packageLayer, not just packageLayer

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -220,54 +222,62 @@ function packageResources(context, resources) {
...
+                };
+              }
+            } else {
+              cfnMeta.Resources.LambdaLayer.Properties.Content = {
+                S3Bucket: s3Bucket,
+                S3Key: s3Key,
+              };
+            }
+            const jsonString = JSON.stringify(cfnMeta, null, '\t');
+            fs.writeFileSync(cfnFilePath, jsonString, 'utf8');
```
edwardfoyle, line 279: use context.amplify.writeObjectAsJson()

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -475,6 +485,52 @@ function updateIdPRolesInNestedStack(context, nestedStack, authResourceName) {
   Object.assign(nestedStack.Resources, idpUpdateRoleCfn);
 }
 
+function packageLayer(context, resource) {
```
edwardfoyle, line 488: I would still recommend moving this to the function category plugin and importing it here. Logic specific to category functionality doesn't belong in the provider plugin

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,122 +1,131 @@
...
-  return [
+  if (resources.length === 0) {
+    context.print.error('No Lambda Layer resource to update. Please use "amplify add function" command to create a new Function');
+    process.exit(0);
+    return;
+  }
+  const resourceQuestion = [
     {
+      name: 'resourceName',
+      message: 'Please select the Lambda Layer you want to update',
```
edwardfoyle, line 58: nit "Select the Lambda Layer to update:"

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,122 +1,131 @@
...
-          name: 'Public (everyone on AWS can use this layer)',
-          value: Permissions.public,
-        },
-      ],
-      default: Permissions.private,
+      choices: resources,
     },
   ];
-}
+  const resourceAnswer = await inquirer.prompt(resourceQuestion);
```
edwardfoyle, line 63: Not a huge issue, but if there's only one layer in the project, you don't need to prompt for this

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,122 +1,131 @@
...
+    let map = createVersionsMap(layerInputParameters, String(versionAnswer.layerVersion));
+    templateParameters.layerVersionsMap[Object.keys(map)[0]] = map[Object.keys(map)[0]];
+  }
+  _.assign(templateParameters, { layerVersion: String(latestVersion) });
+
+  if (latestVersion === latestVersionPushed) {
+    _.assign(templateParameters, { build: false });
+  } else {
+    _.assign(templateParameters, { build: true });
+  }
```
edwardfoyle, line 129: simplifies to
```js
_.assign(templateParameters, {build: latestVersion === latestVersionPushed})
```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -76,20 +64,100 @@ export default function generateLayerCfnObj(parameters) {
...
+
+function assignLayerPermissions(layerData: LayerMetadata, version: string, layerName: string, flag: boolean) {
+  const layerVersionPermissionBase = {
+    Action: 'lambda:GetLayerVersion',
+    LayerVersionArn: createLayerversionArn(layerData, layerName, version, flag),
+  };
+
+  const result = [];
+  if (layerData.getVersion(Number(version)).isPublic()) {
+    result.push({
```
edwardfoyle, line 103: I would early return in the isPublic and isPrivate condition because they are mutually exclusive

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -76,20 +64,100 @@ export default function generateLayerCfnObj(parameters) {
...
 }
+
+function assignLayerPermissions(layerData: LayerMetadata, version: string, layerName: string, flag: boolean) {
+  const layerVersionPermissionBase = {
+    Action: 'lambda:GetLayerVersion',
+    LayerVersionArn: createLayerversionArn(layerData, layerName, version, flag),
+  };
+
+  const result = [];
+  if (layerData.getVersion(Number(version)).isPublic()) {
```
edwardfoyle, line 102: assign `layerData.getVersion(Number(version))` to a const because it's used in multiple places

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -76,20 +64,100 @@ export default function generateLayerCfnObj(parameters) {
...
+    }),
+  );
+  return result;
+}
+
+function joinWithEnv(separator: string, stringToJoin: string) {
+  return Fn.If('HasEnvironmentParameter', Fn.Join(separator, [stringToJoin, Fn.Ref('env')]), stringToJoin);
+}
+
+function createLayerversionArn(layerData, layerName, version, flag) {
```
edwardfoyle, line 151: type the inputs

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -76,20 +64,100 @@ export default function generateLayerCfnObj(parameters) {
...
+  //arn:aws:lambda:us-west-2:136981144547:layer:layers089e3f8b-dev:1
+  if (flag) {
+    // if runtime/Content updated
+    if (layerData.getLatestVersion() === Number(version)) {
+      return Fn.Ref('LambdaLayer');
+    }
+  }
+  return Fn.Sub('arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:layer:${layerName}:${layerVersion}', {
```
edwardfoyle, line 159: why can't this be used regardless of the layer version?

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -122,7 +124,33 @@ export function saveCFNParameters(
...
+  context.amplify.updateamplifyMetaAfterResourceUpdate(categoryName, parameters.layerName, 'runtimes', parameters.runtimes);
+  context.amplify.updateamplifyMetaAfterResourceUpdate(categoryName, parameters.layerName, 'versionsMap', parameters.layerVersionsMap);
+  context.amplify.updateamplifyMetaAfterResourceUpdate(categoryName, parameters.layerName, 'build', parameters.build);
+}
+
+export function createLayerParametersFile(context, parameters, layerDirPath) {
+  fs.ensureDirSync(layerDirPath);
+  const parametersFilePath = path.join(layerDirPath, layerParametersFileName);
+  const jsonString = JSON.stringify({ parameters }, null, 4);
+  fs.writeFileSync(parametersFilePath, jsonString, 'utf-8');
```
edwardfoyle, line 150: use context.amplify.writeObjectAsJson()

packages/amplify-category-function/src/commands/function/update.ts
```diff
@@ -19,9 +20,12 @@ module.exports = {
         }
         return providerController.updateResource(context, categoryName, result.service);
       })
-      .then(() => context.print.success('Successfully updated resource'))
+      .then(() => {
+        context.print.info('');
+      })
```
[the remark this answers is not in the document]
akshbhu: message is their in update layer resource

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -139,15 +139,15 @@ export function updateLayerCfnFile(context, parameters, layerDirPath) {
...
   context.amplify.updateamplifyMetaAfterResourceUpdate(categoryName, parameters.layerName, 'build', parameters.build);
 }
 
 export function createLayerParametersFile(context, parameters, layerDirPath) {
   fs.ensureDirSync(layerDirPath);
   const parametersFilePath = path.join(layerDirPath, layerParametersFileName);
-  const jsonString = JSON.stringify({ parameters }, null, 4);
-  fs.writeFileSync(parametersFilePath, jsonString, 'utf-8');
+  //const jsonString = JSON.stringify({ parameters }, null, 4);
+  context.amplify.writeObjectAsJson(parametersFilePath, { parameters }, true);
```
edwardfoyle, line 150: remove commented out code. Also, I don't think we need to be wrapping all the parameters in a top level "parameters" field (ie, remove the braces around parameters)

packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -18,6 +18,7 @@ const archiver = require('../src/utils/archiver');
 const amplifyServiceManager = require('./amplify-service-manager');
 const ziparchiver = require('archiver');
 const glob = require('glob');
+const { packageLayer } = require('../../amplify-category-function/lib/provider-utils/awscloudformation/utils/functionPluginLoader');
```
edwardfoyle, line 21: see other comment about the packageLayer function. TLDR, we shouldn't be importing based on a path. just import from the `amplify-category-function` module

brene:
gif
cc @akshbhu @edwardfoyle @jhockett

PR #4382, opened 2020-05-27
feat(graphql-key-transformer): auto population of id and timestamp

### LINKED ISSUE #1657: using createdAt as a sortKey in @key directive results mandatory id and createdAt fields in CreateItemInput and CreateItem mutation in GraphQL Transform

### LINKED ISSUE #2671: @model and @key used together prevent fields from generating
**Describe the bug**
Using `@model` and `@key` prevent the `id`, `createdAt` and `updatedAt` fields from being generated if any of those fields are used as fields in `@key`.

The error received in AppSync when creating the record: "Variable 'createpostinput' has coerced Null value for NonNull type 'ID!'"

I saw the VTL mutation (Mutation.createPost.req.vtl) generated applies output of `graphql-key-transformer` package before the `graphql-dynamodb-transformer` package outcome. This means the `@key` annotation uses the `$ctx.args.input.id` supplied in the request without the "cleanup" made in the other package.

When the `graphql-dynamodb-transformer` runs it overrides `$ctx.args.input.id` with `$util.defaultIfNullOrBlank($ctx.args.input.id, $util.autoId())` when the field is null. This means the VTL template interprets two different values for the same field. See screen shots for more context.

**Amplify CLI Version**
3.17.0

**To Reproduce**
add model to schema.graphql

```
type Post @model @key(fields: ["id", "createdAt"]) {
  id: ID!
  createdAt: AWSDateTime!
}
```

push changes to the cloud
```
# amplify push
```

create a post without supplying the id or createdAt field
```
mutation createPost($createpostinput: CreatePostInput!){
  createPost(input: {}){
    id
    createdAt
  }
}
```

**Expected behavior**
A new row should be created in the Post DynamoDB table with the id value built from $util.autoId() and the createdAt time should be built by $util.time.nowISO8601() in the VTL template

**Screenshots**
image
image

 - OS: Windows
 - Node Version: v8.11.4

**Additional context**
As a work around, the user must supply values for ID, createdAt, and/or updatedAt fields.

yuth:
Added support for auto population of id, createdAt and updatedAt fields even when they are part of
the primary key

fix #2671
fix #1657 
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
+++ b/packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
@@ -313,7 +313,7 @@ export class DynamoDBModelTransformer extends Transformer {
       });
       const hositedInitalization = this.resources.initalizeDefaultInputForCreateMutation(createInput, timestampFields);
       const resourceId = ResolverResourceIDs.DynamoDBCreateResolverResourceID(typeName);
-      this.addInitalizationMetaData(ctx, resourceId, hositedInitalization);
+      this.addInitalizationMetadata(ctx, resourceId, hositedInitalization);
       ctx.setResource(resourceId, createResolver);
       ctx.mapResourceToStack(typeName, resourceId);
       const args = [makeInputValueDefinition('input', makeNonNullType(makeNamedType(createInput.name.value)))];
@@ -708,7 +708,7 @@ export class DynamoDBModelTransformer extends Transformer {
     return true;
   }
 
-  private addInitalizationMetaData(ctx: TransformerContext, resourceId: string, initCode: string): void {
+  private addInitalizationMetadata(ctx: TransformerContext, resourceId: string, initCode: string): void {
     const ddbMetadata = ctx.metadata.has(METADATA_KEY) ? ctx.metadata.get(METADATA_KEY) : {};
     ddbMetadata.hoistedRequestMappingContent = {
       ...ddbMetadata?.hoistedRequestMappingContent,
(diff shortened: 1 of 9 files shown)


packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -692,4 +707,13 @@ export class DynamoDBModelTransformer extends Transformer {
     }
     return true;
   }
+
+  private addInitalizationMetaData(ctx: TransformerContext, resourceId: string, initCode: string): void {
```
attilah, line 711: 'Metadata' typo

packages/graphql-dynamodb-transformer/src/DynamoDBModelTransformer.ts
```diff
@@ -31,6 +31,8 @@ import {
 import { ModelDirectiveArgs, getCreatedAtFieldName, getUpdatedAtFieldName } from './ModelDirectiveArgs';
 import { ResourceFactory } from './resources';
 
+const METADATA_KEY = 'DynamoDBTransformerMetadata';
```
attilah, line 34: I like the use of metadata here, I was thinking of something different, but have 1 key per transformer in there and utilize key-value pairs for all values, even if just a boolean, but not depending only on the key's presence. It is up to you if you want to change it now, we can do that later as well.

yuth: For now its internal to dynamodb transformer. If we are exposing this, we should also design for the ordering of these, as some of these when they get rendered in mapping template

PR #4394, opened 2020-05-28
fix(amplify-appsync-simulator): fixes util.isNullOrEmpty

### LINKED ISSUE #4393: amplify-appsync-simulator util.isNullOrEmpty returns true for a boolean true value
When invoking the method with a boolean value of true, true is returned a result. Indicating that arg was equal to null.

eg. $util.isNullOrEmpty(true) === true

rickfoxcroft:
fixes util.isNullOrEmpty returning incorrect values when invoked with a boolean or string value

fix  #4393


rickfoxcroft:
@yuth @attilah Are we any closer to assessing whether this can be merged. There are two linked issues depending on this?

yuth:
I tested the behavior of using non string type in AppSync service and it looks like the service renders the literal as is if the parameter passed to $util.isNullOrEmpty is of any other types. 

To have the same behavior there is some additional change that would be needed in Velocity template engine

rickfoxcroft:
> I tested the behavior of using non string type in AppSync service and it looks like the service renders the literal as is if the parameter passed to $util.isNullOrEmpty is of any other types.
> 
> To have the same behavior there is some additional change that would be needed in Velocity template engine

Could you please provide more detail on the test you performed and the additional change that is required?

Also, I have updated the tests to add more test cases.

yuth:
> Could you please provide more detail on the test you performed and the additional change that is required?

So I used the AppSync consoles, Velocity template test feature. My test context contains these values
```JSON
{
    "arguments": {
        "id": "BEoDcq9lqwVO0ShcHNR77Yk8pqjnBLu0",
        "isFalse": false,
        "isTrue": true,
        "isEmptyString": "",
        "isNonEmptyString": "foo bar",
        "isNull": null
    },
    "source": {
        "echo": "Hello, world!"
    },
    "result": {
        "id": "OrdbGHOruchIqNirE9qQ1Rd9W3mD6hiQ",
        "name": "Hello, world!",
        "description": "Hello, world!"
    }
}
```
and request template contains
```
#set($foo = $util.isNullOrEmpty(true))
{
  "version": "2017-02-28",
  "operation": "GetItem",
  "key": #if( $modelObjectKey ) $util.toJson($modelObjectKey) #else {
  "id": $util.dynamodb.toDynamoDBJson($ctx.args.id),
  "isTrueNotNull": $util.isNullOrEmpty($ctx.args.isTrue),
  "isFalseNotNull": $util.isNullOrEmpty($ctx.args.isFalse),
  "isNonEmptyString": $util.isNullOrEmpty($ctx.args.isNonEmptyString)
  "isEmptyString": $util.isNullOrEmpty($ctx.args.isEmptyString)
  "isStringNotNull": $util.isNullOrEmpty(12),
  "isNull": $util.isNullOrEmpty($ctx.args.isNull)
  #if($ctx.args.isTrue) 
   isTrue = true
  #end
  
  #if($ctx.args.isFalse) 
     isFalse = false
  #end
} #end
}
${foo}
```

and this is what gets rendered by VTL on AppSync service

Screen Shot 2020-06-04 at 5 46 08 PM

yuth:
Closing the PR as we can't implement the exact behavior of appsync with this implementation of mock.

PR #4421, opened 2020-05-31
feat(amplify-appsync-simulator): implement BatchInvoke

### LINKED ISSUE #4404: Lambda BatchInvoke on local environment
When calling BatchInvoke on a Lambda function in app sync the payload and response changes so that now app sync expects an array of values for both. This is fine, except when writing code and testing locally BatchInvoke does not work. Calling the Lambda function locally with Amplify will always treat the function call as a regular "Invoke" and so its difficult to test. It's annoying to have two different responses, one for local and one for the real appsync environment.

Am I missing something here? Whats the best way to BatchInvoke a Lambda function and have it consistent across local and once deployed?

bboure:
*Issue #, if available:*
Fixes #4404

*Description of changes:*
Implements, `BatchInvoke` for lambda resolvers.
Payloads are batched by `parentType` and `fieldName`

### DIFF
--- a/packages/amplify-appsync-simulator/package.json
+++ b/packages/amplify-appsync-simulator/package.json
@@ -29,6 +29,7 @@
     "aws-sdk": "^2.608.0",
     "chalk": "^3.0.0",
     "cors": "^2.8.5",
+    "dataloader": "^2.0.0",
     "event-to-promise": "^0.8.0",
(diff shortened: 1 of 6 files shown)


bboure:
@yuth  would be great to have feedback on this. 
Cheers

PR #4458, opened 2020-06-04
feat(graphql-key-transformer): add query automatically for named keys

yuth:
Automatically add query field for named queries


packages/graphql-key-transformer/src/__tests__/KeyTransformer.test.ts
```diff
@@ -112,3 +114,88 @@ test('KeyTransformer should fail if a non-existing type field is defined as key
 
   expect(() => transformer.transform(invalidSchema)).toThrowError(InvalidDirectiveError);
 });
+
+test('KeyTransformer should generate queryFeild automatically', () => {
```
attilah, line 118: nit: queryFeild typo

yuth:
> @key is a other side of a @connection? I think in that case we should:
> a. not generate a query field automatically, only if the generateQuery is true
> b. not generate a query as it is prohibited, so if the generateQuery is true we should error out

I had this implemented this in [earlier commit](https://github.com/aws-amplify/amplify-cli/pull/4458/commits/4657d1b5d183713e830eb6e32ab18c4cb7a92238#diff-ca2c9971cf82b7ea4207497c2bc7e611R913-R938) but it was decided in API review to  keep the directive behavior consistent.

PR #4465, opened 2020-06-04
fix(amplify-provider-awscloudformation): Stack delete condition

arseneoaa:
Currently when updating/deleting the stack both authParamsJson and unauthParamsJson keep their previous condition statement.
When the new stack is created the previous values for 'Condition': {'StringEquals': {'cognito-identity.amazonaws.com:aud': idpId},' for example do not get updated.

The fix is just about deleting the condition on the object inside the Statement array instead of on the Statement array itself.

*Issue #, if available:*

*Description of changes:*


UnleashedMind:
The fix looks good.
Could you add e2e test for the PR? Sample e2e tests can be found in the amplify-e2e-tests package. 
I'm happy offer assistance if needed. Just send an email to amplify-cli@amazon.com with subject Re: PR#4465.

UnleashedMind:
I added the e2e tests

packages/amplify-e2e-tests/src/aws-matchers/iamMatcher.ts
```diff
@@ -27,3 +27,49 @@ export const toBeIAMRoleWithArn = async (roleName: string, arn?: string) => {
...
+    pass = false;
+    message = 'IAM GetRole threw Error: ' + e.message;
+  }
+
+  const result = {
+    message: () => message,
+    pass,
+  };
+
+  return result;
```
jhockett, line 74: Small nit: I would return the object directly

```ts
  return {
    message: () => message,
    pass,
  };
```

packages/amplify-e2e-tests/src/aws-matchers/iamMatcher.ts
```diff
@@ -27,3 +27,49 @@ export const toBeIAMRoleWithArn = async (roleName: string, arn?: string) => {
   };
   return result;
 };
+
+export const toHaveAssumeRolePolicyConditionMatchingIdpId = async (roleName: string, idpId: string) => {
```
jhockett, line 31: Nit: I think something like `toHaveValidAssumeRolePolicyCondition` would be easier to understand at a glance.

packages/amplify-e2e-tests/src/aws-matchers/iamMatcher.ts
```diff
@@ -27,3 +27,49 @@ export const toBeIAMRoleWithArn = async (roleName: string, arn?: string) => {
...
+      } else {
+        return false;
+      }
+    });
+
+    if (pass) {
+      message = 'Found Matching Condition';
+    } else {
+      message = 'Matching Condition does not exist';
+    }
```
jhockett, line 63: Simplifies to 
```ts
message = pass ?  'Found Matching Condition' : 'Matching Condition does not exist';
```

arseneoaa:
Thanks a lot @jhockett .
Sorry I did not have the time to look into how to do the e2e tests.

Issue #3096, opened 2019-12-29
nested-cloudformation-stack.yml uses backup (tilde) files for custom CFN


berenddeboer:
**Describe the bug**

The generated `amplify/backend/awscloudformation/nested-cloudformation-stack.yml` uses the backup file for it's custom stack:

```
		"AuroraCluster": {
			"Type": "AWS::CloudFormation::Stack",
			"Properties": {
				"TemplateURL": "https://s3.amazonaws.com/amplify-amplify-js-app-ireland-145750-deployment/amplify-cfn-templates/Aurora/cloudformation-template.yml~",
				"Parameters": {
					"rdsDatabaseName": "TESTDB",
					"env": "ireland"
				}
			}
		}
``` 

I'm editing with Emacs, so it tends to create a backup file with the same name but ending in `~` (tilde).

**Amplify CLI Version**
4.9.

**To Reproduce**
Copy `amplify/backend/<custom-category>/<custom-resource>/cloudformation-template.yml` to `cloudformation-template.yml~` and execute `amplify push` to update `nested-cloudformation-stack.yml`.

**Expected behavior**
It should never use backup files. It should look for exact endings of '.yml' or '.json', not any filename having this as part of the filename.

**Desktop (please complete the following information):**
 - OS: Ubuntu 19.10.
 - Node Version: 10.18.0

berenddeboer:
This happens at more places. Just saw this during an `amplify push`:

```
Invalid extension .json~ for stack CustomResources.json~
An error occured during the push operation: Invalid extension .json~ for stack CustomResources.json~
```

attilah:
@berenddeboer while we're working on fixing the issue as a workaround you could reconfigure emacs to always create the backups in 1 directory instead where the file exists or disable backups.

PR #4478, opened 2020-06-05
fix: #3096 - glob *template*.+(yaml|yml|json) files only as cfn template

attilah:
*Issue #, if available:*

fix: #3096 - glob *template*.+(yaml|yml|json) files only as cfn template

*Description of changes:*

Strict and proper globbing or file patterns when looking for cloudformation templates in category directories.

For non-model types then the 1st level type had nested non-model types, @auth was not propagated properly on nested types.

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/push-resources.js
+++ b/packages/amplify-provider-awscloudformation/lib/push-resources.js
@@ -21,6 +21,8 @@ const amplifyServiceManager = require('./amplify-service-manager');
 const spinner = ora('Updating resources in the cloud. This may take a few minutes...');
 const nestedStackFileName = 'nested-cloudformation-stack.yml';
 const optionalBuildDirectoryName = 'build';
+const cfnTemplateGlobPattern = '*template*.+(yaml|yml|json)';
+const parametersJson = 'parameters.json';
 
 async function run(context, resourceDefinition) {
   try {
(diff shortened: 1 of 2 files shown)


packages/amplify-provider-awscloudformation/lib/push-resources.js
```diff
@@ -199,9 +200,9 @@ function validateCfnTemplates(context, resourcesToBeUpdated) {
     const { category, resourceName } = resourcesToBeUpdated[i];
     const backEndDir = context.amplify.pathManager.getBackendDirPath();
     const resourceDir = path.normalize(path.join(backEndDir, category, resourceName));
-    const files = fs.readdirSync(resourceDir);
-    // Fetch all the Cloudformation templates for the resource (can be json or yml)
-    const cfnFiles = files.filter(file => file.indexOf('template') !== -1 && file.indexOf('.') !== 0);
+    const cfnFiles = glob.sync('*template*.+(yaml|yml|json)', {
```
yuth, line 203: Do we want to support files which have something else after template like `myApi.template.first.json`
Can this be pulled to a const.
```suggestion
   const CFN_GLOB_PATTERN = '*template*.+(yaml|yml|json)';
    const cfnFiles = glob.sync(CFN_GLOB_PATTERN , {
```

attilah: Currently we don't have any requirement on this, this changed followed the same pattern what we had before, which allows this, later we can revisit it and even just enable anything with those extensions except parameters.json (already included in the update) if we want to remote the 'template' restriction later.
Good catch, moved to constant for all codepaths

Issue #4460, opened 2020-06-04
API push fails if key expire set 1


asideHazle:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[Amplify Console GitHub Issue Tracker](https://github.com/aws-amplify/amplify-console/issues)

**Describe the bug**
API push fails if key expire set 1

**Amplify CLI Version**
4.21.1

**To Reproduce**
```bash
amplify add api
? Please select from one of the below mentioned services: GraphQL
? Provide API name: amplifytest
? Choose the default authorization type for the API API key
? Enter a description for the API key:
? After how many days from now the API key should expire (1-365): 1
? Do you want to configure advanced settings for the GraphQL API No, I am done.
? Do you have an annotated GraphQL schema? No
? Do you want a guided schema creation? Yes
? What best describes your project: One-to-many relationship (e.g., “Blogs” with
 “Posts” and “Comments”)
? Do you want to edit the schema now? Yes
```

**Expected behavior**

```bash
Following resources failed

Resource Name: 7eeywqbd6na5voygrlzgvyxciuGraphQLSchema (AWS::AppSync::GraphQLSchema)
Event Type: create
Reason: Resource creation cancelled


Resource Name: GraphQLAPIKey (AWS::AppSync::ApiKey)
Event Type: create
Reason: API key must be valid for a minimum of 1 days. (Service: AWSAppSync; Status Code: 400; Error Code: ApiKeyValidityOutOfBoundsException; Request ID: c490e268-4743-426f-8122-cf98c3f1fba3)


✖ An error occurred when pushing the resources to the cloud

Resource is not in the state stackUpdateComplete
An error occured during the push operation: Resource is not in the state stackUpdateComplete
```

**Screenshots**
![스크린샷 2020-06-04 오후 2 36 38](https://user-images.githubusercontent.com/35303552/83718871-d64add80-a670-11ea-9fb7-5fb6c9cc9df3.png)

**Desktop (please complete the following information):**
 - OS:  Mac
 - Node Version : 12.16.2

**Additional context**
When try expire value set 2, Push is success

SwaySway:
Closing this as the fix has been merged and is in the latest release.

PR #4493, opened 2020-06-08
fix(graphql-auth-transformer): add a time delay when creating apiKey

SwaySway:
added a time delay to the expiration date

re #4460

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-auth-transformer/src/resources.ts
+++ b/packages/graphql-auth-transformer/src/resources.ts
@@ -106,7 +106,9 @@ export class ResourceFactory {
       expirationDays = apiKeyConfig.apiKeyExpirationDays;
     }
     // add delay expiration time is valid upon resource creation
-    const expirationDateInSeconds = 60 /* s */ * 60 /* m */ * 24 /* h */ * expirationDays /* d */ + 60 * 2;
+    let expirationDateInSeconds = 60 /* s */ * 60 /* m */ * 24 /* h */ * expirationDays /* d */;
+    // Add a 2 minute time delay if set to 1 day: https://github.com/aws-amplify/amplify-cli/issues/4460
+    if (expirationDays == 1) expirationDateInSeconds + 60 * 2;
     const nowEpochTime = Math.floor(Date.now() / 1000);
     return new AppSync.ApiKey({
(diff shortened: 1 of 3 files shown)


packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -105,7 +105,8 @@ export class ResourceFactory {
     if (apiKeyConfig && apiKeyConfig.apiKeyExpirationDays) {
       expirationDays = apiKeyConfig.apiKeyExpirationDays;
     }
-    const expirationDateInSeconds = 60 /* s */ * 60 /* m */ * 24 /* h */ * expirationDays; /* d */
+    // add delay expiration time is valid upon resource creation
+    const expirationDateInSeconds = 60 /* s */ * 60 /* m */ * 24 /* h */ * expirationDays /* d */ + 60 * 2;
```
attilah, line 109: I think that we only need to adjust the value if ```expirationDays===1```, otherwise we can run into an issue with 365 days and perhaps the 2 minutes adjustment could be higher like 30-60 minutes to factor in the time that large stack deployments could take.

Also please update the comment to describe why the adjustment is needed.

nth: E2E test for both cases to see that we're not regressing because of this adjustment.

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -105,7 +105,10 @@ export class ResourceFactory {
     if (apiKeyConfig && apiKeyConfig.apiKeyExpirationDays) {
       expirationDays = apiKeyConfig.apiKeyExpirationDays;
     }
-    const expirationDateInSeconds = 60 /* s */ * 60 /* m */ * 24 /* h */ * expirationDays; /* d */
+    // add delay expiration time is valid upon resource creation
+    let expirationDateInSeconds = 60 /* s */ * 60 /* m */ * 24 /* h */ * expirationDays /* d */;
+    // Add a 2 minute time delay if set to 1 day: https://github.com/aws-amplify/amplify-cli/issues/4460
+    if (expirationDays == 1) expirationDateInSeconds += 60 * 2;
```
jhockett, line 111: nit: ===

PR #4533, opened 2020-06-11
perf: optimize appsync file upload and bucket exist check

attilah:
*Description of changes:*

perf: optimize appsync file upload to use globbing to collect the files and upload all the files in parallel
perf: cache environment and bucket name for S3 operations
perf: optimize bucket exist check to use head operation instead of list and filtering
chore: remove duplicated hashing function

Test schema with 20 simple @model types ~300 files.
 
before: 2 minutes 19 seconds 
after: 2.9 seconds

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.js
@@ -19,18 +19,32 @@ class S3 {
   }
 
   uploadFile(s3Params) {
-    const projectDetails = this.context.amplify.getProjectDetails();
-    const { envName } = this.context.amplify.getEnvInfo();
-    const projectBucket = projectDetails.amplifyMeta.providers
-      ? projectDetails.amplifyMeta.providers[providerName].DeploymentBucketName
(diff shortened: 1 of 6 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.js
```diff
@@ -19,18 +19,32 @@ class S3 {
...
   uploadFile(s3Params) {
-    const projectDetails = this.context.amplify.getProjectDetails();
-    const { envName } = this.context.amplify.getEnvInfo();
-    const projectBucket = projectDetails.amplifyMeta.providers
-      ? projectDetails.amplifyMeta.providers[providerName].DeploymentBucketName
-      : projectDetails.teamProviderInfo[envName][providerName].DeploymentBucketName;
-    s3Params.Bucket = projectBucket;
+    // envName and bucket does not change during execution, cache them into a class level
+    // field.
+    if (this.uploadState === undefined) {
```
yuth, line 24: Can we move this to a function that has memoization instead of doing it here?

PR #4539, opened 2020-06-12
fix: e2e tests for layers working

akshbhu:
*Description of changes:*

* Fixes e2e tests for layers

* Change logic to add CFN parameters of layers to functions


packages/amplify-e2e-tests/src/__tests__/layer.test.ts
```diff
@@ -62,22 +63,53 @@ describe('amplify add lambda layer', () => {
...
       layerName: `testlayer${shortId}`,
-      versionChanged: true,
     };
     await initJSProjectWithProfile(projRoot, {});
-    await addLayer(projRoot, settings);
-    await updateLayer(projRoot, settings);
+    await addLayer(projRoot, settingsAdd);
+    const settingsUpdate = {
+      runtimes: ['java'],
+      layerName: `testlayer${shortId}`,
```
jhockett, line 74: Let's add a const above so that the different settings objects are in sync

```js
   const [shortId] = uuid().split('-');
   const layerName = `testlayer${shortId}`;
   const settingsAdd = {
     layerName,
     runtimes: ['nodejs'],
     versionChanged: true,
   };
```

jhockett: Same thing in the next test.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -172,14 +172,15 @@ export async function updateWalkthrough(context, lambdaToUpdate?: string) {
...
-      };
+  functionParameters.lambdaLayers.forEach(val => {
+    if ((val as ProjectLayer).resourceName) {
+      const param: string = `function${(val as ProjectLayer).resourceName}Arn`;
+      if (cfnContent.Parameters[`${param}`] === undefined) {
+        cfnContent.Parameters[`${param}`] = {
+          Type: 'String',
+          Default: `${param}`,
+        };
+      }
```
jhockett, line 183: I would write something like this. I think using a resourceName const makes this a bit more readable.

```ts
const resourceName = _.get((val as ProjectLayer), ['resourceName'], null);
if (resourceName) {
  const param: string = `function${resourceName}Arn`;
  if (cfnContent.Parameters[`${param}`] === undefined) {
    cfnContent.Parameters[`${param}`] = { Type: 'String', Default: `${param}` };
  }
}
```

packages/amplify-e2e-tests/src/__tests__/layer.test.ts
akshbhu: done thanks!!

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
akshbhu: Nice suggestion!!

Issue #4158, opened 2020-05-04
Publish to s3 runs sequentially...


GeorgeBellTMH:
When uploading to s3 during a publish Amplify uploads only one file at a time since it is using await sequential. Is there a better way of doing this that uploads multiple files at once without waiting for promises to complete?
```js
const spinner = new Ora('Uploading files...');
  try {
    spinner.start();
    await sequential(uploadFileTasks);
    spinner.succeed('Uploaded files successfully.');
  } catch (e) {
    spinner.fail('Error has occured during file upload.');
    throw e;
  }
}
```

yuth:
Thanks for bringing this up to our attention. Yes I think it would be better to have a thread pool for uploading so we can have a few concurrent uploads at once instead of doing it sequentially.

GeorgeBellTMH:
I am curious if we could just use Promise.all().... in local testing this seems to have a significant improvement in speed on large projects, projects with many files and even smaller projects.

It seems like Promise.all() is effectively a thread pool since the runtime (node) only will allow X amounts of uploads at a time, and all others are queued. Since publish is the only thing running in the node instance at this point in the code it doesn't seem like it would impact anything else.

The only risk is that if a file fails, it would be a random file, instead of the sequentially next file in the directory...but since publish isn't atomic to begin with, any failure during publish is already a problem.

PR #4546, opened 2020-06-12
perf: fulfill promises to upload files to S3 concurrently

nchaloult:
*Issue #, if available:*
#4158

*Description of changes:*
When uploading multiple files to S3 after running `$ amplify publish`, those uploads are handled one after the other when they could be handled concurrently. Rather than fulfilling promises to upload each file sequentially, this PR proposes that we let the node runtime handle fulfilling those promises concurrently with `Promise.all()`.


nchaloult:
I accidentally opened this PR prematurely. Will close for now.

PR #4551, opened 2020-06-12
fix: add hashing function to track layer version changes

jhockett:


packages/amplify-category-function/src/provider-utils/awscloudformation/utils/packageLayer.ts
```diff
@@ -1,19 +1,39 @@
...
+import _ from 'lodash';
+import { createLayerParametersFile } from './storeResources';
+import { hashElement } from 'folder-hash';
 
-export function packageLayer(context, resource) {
+export async function packageLayer(context, resource) {
   const resourcePath = path.join(context.amplify.pathManager.getBackendDirPath(), resource.category, resource.resourceName);
+  const layerHash = await hashLayerDir(resourcePath);
+  const layerParameters = context.amplify.readJsonFile(path.join(resourcePath, 'layer-parameters.json'));
+  const versions = Object.keys(layerParameters.layerVersionMap).sort();
```
akshbhu, line 13: you can use LayerStateMetadata interface to get latestVersion
```js
const layerData = layerMetadataFactory(backendDirPath, layerName);
const latestVersion = layerData.getlatestVersion();
```

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
```diff
@@ -210,13 +210,13 @@ export async function updateLayerResource(
       projectName: context.amplify.getProjectDetails().projectConfig.projectName,
     };
   }
-  await serviceConfig.walkthroughs.updateWalkthrough(context, undefined, parameters);
+  parameters = await serviceConfig.walkthroughs.updateWalkthrough(context, undefined, parameters);
```
akshbhu, line 213: I think we dont have to return parameters from this function as we are already passing parameters as a reference in this function

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -52,39 +41,32 @@ export function generatePermissionCfnObj(context: any, parameters: LayerParamete
...
   const POLICY_RETAIN = DeletionPolicy.Retain;
+  const latestVersion = layerData.getLatestVersion();
   const layer = new Lambda.LayerVersion({
     CompatibleRuntimes: parameters.runtimes.map(runtime => runtime.cloudTemplateValue),
     Content: {
       S3Bucket: Fn.Ref('deploymentBucketName'),
       S3Key: Fn.Ref('s3Key'),
     },
-    Description: 'Lambda Layer',
+    Description: `Lambda Layer ${latestVersion}`,
```
edwardfoyle, line 62: nit `Lambda Layer version ${latestVersion}`

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerParams.ts
```diff
@@ -155,7 +165,7 @@ export const getLayerMetadataFactory = (projectBackendDirPath: string): LayerMet
       return undefined;
     }
     const parametersFilePath = path.join(resourceDirPath, layerParametersFileName);
-    const obj = JSON.parse(fs.readFileSync(parametersFilePath, 'utf8')).parameters;
+    const obj = JSON.parse(fs.readFileSync(parametersFilePath, 'utf8'));
```
edwardfoyle, line 168: would be nice to use context.amplify.readJsonFile()

We could pass in context instead of projectBackendDirPath and compute projectBackendDirPath in this function.

If that's going to be difficult to change, then just leave it

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerHelpers.ts
```diff
@@ -6,14 +6,15 @@ export interface LayerInputParams {
   layerPermissions?: Permission[];
   authorizedAccountIds?: string;
   authorizedOrgId?: string;
+  hash?: string;
```
edwardfoyle, line 9: Does this belong in this interface? Seems like the other fields in this interface are parameters to the layer CFN template, but this is for our own bookkeeping

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerHelpers.ts
```diff
@@ -6,14 +6,15 @@ export interface LayerInputParams {
...
 }
 
 export function layerVersionQuestion(choices) {
   return [
     {
       type: 'list',
       name: 'layerVersion',
-      message: 'Select the version number to update for given Lambda Layer: ',
+      message: 'Select the version number to update for given Lambda Layer:',
```
edwardfoyle, line 17: nit "Select the layer version to update:"

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerHelpers.ts
```diff
@@ -129,6 +130,7 @@ export function layerOrgAccessQuestion(defaultorgs?: string[]) {
 
 export function createVersionsMap(parameters: LayerInputParams, version: string) {
```
edwardfoyle, line 131: this function should have a return type besides object

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
edwardfoyle: I think this is better because it makes it obvious that parameters is being modified

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/packageLayer.ts
```diff
@@ -1,19 +1,44 @@
...
+  const layerData = getLayerMetadataFactory(backendPath)(resource.resourceName);
+  let latestVersion: number = layerData.getLatestVersion();
+  const curLayerHash = await hashLayerDir(resourcePath);
+  const previousHash = layerData.getHash(latestVersion);
+  const layerParameters = context.amplify.readJsonFile(path.join(resourcePath, 'layer-parameters.json'));
+
+  if (previousHash && previousHash !== curLayerHash.hash) {
+    ++latestVersion; // Content changes detected, bumping version
+    layerParameters.layerVersionMap[`${latestVersion}`] = {
+      permissions: [{ type: 'private' }],
```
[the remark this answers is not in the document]
jhockett: This will be changed in a future PR. This is a placeholder to make sure `amplify push` works as intended.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/packageLayer.ts
```diff
@@ -31,13 +56,29 @@ export function packageLayer(context, resource) {
...
       .filter(folder => fs.lstatSync(folder).isDirectory())
       .forEach(folder => zip.directory(folder, path.basename(folder)));
     zip.finalize();
   });
 }
 
+export async function hashLayerDir(layerPath: string): Promise<any> {
+  const hashOptions = {
+    folders: { exclude: ['.*', 'dist'] },
+    files: { exclude: ['README.txt', 'layer-parameters.json', 'parameters.json', '*-awscloudformation-template.json'] },
```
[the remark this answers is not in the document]
akshbhu: If the customer just changed the permissions in CFN, wont that be a problem because we will bump a version locally as hash will change.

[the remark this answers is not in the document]
edwardfoyle: Hmm yeah that's true but won't we run into problems if say, they manually update the compatible runtimes of the layer? Maybe we're okay with not covering that edge case. I guess the alternative is to include a hash of just the lambdalayer part of the CFN rather than the entire CFN

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/packageLayer.ts
```diff
@@ -54,9 +53,10 @@ export async function packageLayer(context, resource) {
       reject(new Error('Failed to zip code.'));
     });
 
+    const libPath = path.join(resourcePath, 'lib', '*');
```
edwardfoyle, line 56: I think you want '**' (to include contents of subdirectories as well)
And you may also want ** for the optPath as well

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/packageLayer.ts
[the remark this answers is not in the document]
edwardfoyle: I see we're still excluding the CFN template (which I think is okay) but we need to document that you should not make manual changes to the layer CFN, or if you do you may also need to update the version metadata that we store

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/packageLayer.ts
jhockett: The single '*' is intentional. If you use '**' the zipped folder would look like:
```
lib/nodejs/node_modules/
nodejs/node_modules/
node_modules/
opt/
```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/packageLayer.ts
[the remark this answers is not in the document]
jhockett: What's the best way to document this? Possible options that I can think of:
In a comment inside the CFN template
A message from the CLI on Layer creation
In the docs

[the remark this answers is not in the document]
akshbhu: I think it's better if we can put these in the docs not to make manual changes to CFN file for now.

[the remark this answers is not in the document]
edwardfoyle: I would also put a message in the CLI on layer creation and maybe even a reminder in the update message. We can't put comments in the CFN template because strict JSON can't have comments

PR #4565, opened 2020-06-15
e2e tests for NodeJS and Python

akshbhu:
*Description of changes:*

* Added Test for NodeJS and Python Runtime for layers

* Skipping Java Tests 

* Added a fix for building function if layer is also present

### DIFF
--- a/packages/amplify-e2e-tests/layerdata/python/index.py
+++ b/packages/amplify-e2e-tests/layerdata/python/index.py
@@ -1,7 +1,6 @@
 from testfunc import testString
 def handler(event, context):
-  print('received event:')
-  print(event)
+  print('received event:', event)
   return {
     'message': testString
   }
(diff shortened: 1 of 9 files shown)


packages/amplify-e2e-tests/layerdata/python/index.py
```diff
@@ -0,0 +1,7 @@
+from testfunc import testString
+def handler(event, context):
+  print('received event:')
+  print(event)
```
jhockett, line 4: This can be one line in python, similar to console.log
```py
print('received event:', event)
```

packages/amplify-e2e-tests/layerdata/python/testfunc.py
```diff
@@ -0,0 +1 @@
+testString = "Hello from Lambda!"
```
jhockett, line 1: Nit: no newline at end of file

PR #4570, opened 2020-06-16
fix(graphql-dynamodb-transformer): support model without id

### LINKED ISSUE #4502: amplify cli v4.21.1 introduces graphql mutation dynamo error
**Describe the bug**
`amplify push` with v4.21.1 results in an api stack that where sending `createThing` graphql mutations results in consistant failure with the following error:

```json
{
  "data": {
    "createFoo": null
  },
  "errors": [
    {
      "path": [
        "createFoo"
      ],
      "data": null,
      "errorType": "DynamoDB:AmazonDynamoDBException",
      "errorInfo": null,
      "locations": [
        {
          "line": 2,
          "column": 3,
          "sourceName": null
        }
      ],
      "message": "One or more parameter values were invalid: Type mismatch for key id expected: S actual: NULL (Service: AmazonDynamoDBv2; Status Code: 400; Error Code: ValidationException; Request ID: REDACTED)"
    }
  ]
}
```

The same action with v4.21.0 (and earlier) results in no error.

**Amplify CLI Version**
@aws-amplify/cli@4.21.1 (@aws-amplify/cli@4.21.0 is fine)

**To Reproduce**
`amplify push` and then send a `createFoo` mutation to the api.

Schema:

```graphql
type Foo
  @model
  @key(
    name: "byBar"
    fields: ["barID"]
    queryField: "fooByBar"
  ) {
  barID: ID!
  Bar: Bar! @connection(fields: ["barID"])
  something: [Something]
    @connection(keyName: "byBar", fields: ["barID"])
}
```

Autogenerated mutation that creates error when used:

```graphql

mutation CreateFoo(
  $input: CreateFooInput!
  $condition: ModelFooConditionInput
) {
  createFoo(input: $input, condition: $condition) {
    barID
    bar {
      (stuff)
    }
  }
}
```
With the input type being:

```graphql
input CreateFooInput {
  barID: ID!
}
```

**Expected behavior**
I expect the mutation to create a new Foo entry in dynamo and no error to be thrown.

**Screenshots**

 - OS: Mac
 - Node Version: v12.16.2

yuth:
Updatd graphql-dynamodb-transformer to automatically add id to the model when
id is missing in the type. Id will not be added if `@key` directive changes
the primary index

fix #4502 
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/graphql-key-transformer/src/KeyTransformer.ts
+++ b/packages/graphql-key-transformer/src/KeyTransformer.ts
@@ -116,6 +116,10 @@ export class KeyTransformer extends Transformer {
   private updateSchema = (definition: ObjectTypeDefinitionNode, directive: DirectiveNode, ctx: TransformerContext) => {
     this.updateQueryFields(definition, directive, ctx);
     this.updateInputObjects(definition, directive, ctx);
+    const isPrimaryKey = this.isPrimaryKey(directive);
+    if (isPrimaryKey) {
+      this.removeAutocreatedPrimaryKey(definition, directive, ctx);
+    }
   };
 
   /**
@@ -193,6 +197,19 @@ export class KeyTransformer extends Transformer {
         ctx.mapResourceToStack(definition.name.value, queryResolverId);
         ctx.setResource(queryResolverId, queryResolver);
       }
+    }
+  };
+
+  private removeAutocreatedPrimaryKey = (definition: ObjectTypeDefinitionNode, directive: DirectiveNode, ctx: TransformerContext): void => {
+    const schemaHasIdField = definition.fields.find(f => f.name.value === 'id');
+    if (!schemaHasIdField) {
+      const obj = ctx.getObject(definition.name.value);
+      const fields = obj.fields.filter(f => f.name.value !== 'id');
+      const newObj: ObjectTypeDefinitionNode = {
+        ...obj,
+        fields,
+      };
+      ctx.updateObject(newObj);
     }
   };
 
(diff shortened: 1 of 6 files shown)


packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -115,6 +115,10 @@ export class KeyTransformer extends Transformer {
   private updateSchema = (definition: ObjectTypeDefinitionNode, directive: DirectiveNode, ctx: TransformerContext) => {
     this.updateQueryFields(definition, directive, ctx);
     this.updateInputObjects(definition, directive, ctx);
+    const isPrimaryKey = this.isPrimaryKey(directive);
+    if (isPrimaryKey) {
+      this.removeAutocreatedPrimaryKey(definition, directive, ctx);
```
attilah, line 120: nit: AutoCreated

yuth:
> A migration test which ensures that the GraphQL output of the same schema where `@key` transformer added the `id` field will produce the very same output from the new CLI with this change where this behavior is done with `@model` transformer.

I am not sure how we can do migration test for this. This changes the behavior of `@model` directive by adding `id` field and exposing that in the type if the `model`  does not change the `hashKey` with `@key` directive.

This used to work in CLI version 4.20.0 but stopped working in later versions. But prior to 4.20.0, the id was not exposed in the type and an model without the `id` field could not be used with `@connection`. This fixes that issue

packages/graphql-key-transformer/src/KeyTransformer.ts
yuth: Fixed

PR #4575, opened 2020-06-16
perf: fulfill promises to upload files to S3 concurrently

nchaloult:
*Issue #, if available:*
#4158 

*Description of changes:*
When uploading multiple files to S3 after running `$ amplify publish`, those uploads are handled one after the other when they could be handled concurrently. Rather than fulfilling promises to upload each file sequentially, this PR proposes that we let the node runtime handle fulfilling those promises concurrently with `Promise.all()`.

### DIFF
--- a/packages/amplify-category-auth/provider-utils/awscloudformation/utils/trigger-file-uploader.js
+++ b/packages/amplify-category-auth/provider-utils/awscloudformation/utils/trigger-file-uploader.js
@@ -2,7 +2,6 @@ const { readdirSync, existsSync } = require('fs');
 const { createReadStream } = require('fs-extra');
 const Ora = require('ora');
 const mime = require('mime-types');
-const sequential = require('promise-sequential');
 const { getAuthResourceName } = require('../../../utils/getAuthResourceName');
 
 const providerName = 'awscloudformation';
@@ -30,15 +29,12 @@ async function uploadFiles(context) {
       return null;
     }
     const fileList = readdirSync(assetPath);
-    const uploadFileTasks = [];
-    fileList.forEach(file => {
-      uploadFileTasks.push(async () => await uploadFile(s3Client, bucketName, `${assetPath}/${file}`, file));
-    });
+    const uploadFileTasks = fileList.map((file) => uploadFile(s3Client, bucketName, `${assetPath}/${file}`, file));
 
     const spinner = new Ora('Uploading files...');
     try {
       spinner.start();
-      await sequential(uploadFileTasks);
+      await Promise.all(uploadFileTasks);
       spinner.succeed('Uploaded files successfully.');
     } catch (e) {
       spinner.fail('Error has occured during file upload.');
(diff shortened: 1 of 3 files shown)


nchaloult:
It looks like the original member of the open source community who created #4158 opened a PR of their own to resolve this issue: https://github.com/aws-amplify/amplify-cli/pull/4508. After taking a look at it, I'm not sure if their work is correct. I'd like to hear what everyone else thinks.

`uploadFileTasks` is an array of anonymous functions that return promises, because `sequential()` expects an array of functions that return promises as its argument. `Promise.all()`, however, expects an iterable of promises as its argument, so the logic that builds the items in the `uploadFileTasks` array needs to be changed accordingly if `sequential()` is to be swapped out for `Promise.all()`.

I confirmed through local testing with `amplify-dev` that if `sequential()` call is replaced with `Promise.all()`, but no changes are made to the way that the `uploadFileTasks` array is constructed, no files are ever uploaded to an S3 bucket.

GeorgeBellTMH:
I am fine with closing my PR...it wasn't working, and I have had other stuff on my plate...

packages/amplify-category-auth/provider-utils/awscloudformation/utils/trigger-file-uploader.js
```diff
@@ -32,13 +31,13 @@ async function uploadFiles(context) {
     const fileList = readdirSync(assetPath);
     const uploadFileTasks = [];
     fileList.forEach(file => {
-      uploadFileTasks.push(async () => await uploadFile(s3Client, bucketName, `${assetPath}/${file}`, file));
+      uploadFileTasks.push(uploadFile(s3Client, bucketName, `${assetPath}/${file}`, file));
```
ammarkarachi, line 34: Why don't we simplify this to

```js
const uploadFileTasks = fileList.map(file => uploadFile(s3Client, bucketName, `${assetPath}/${file}`, file))
```

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
```diff
@@ -28,13 +27,13 @@ async function run(context, distributionDirPath) {
   }
 
   fileList.forEach(filePath => {
-    uploadFileTasks.push(() => uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, cloudFrontS3CanonicalUserId));
+    uploadFileTasks.push(uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, cloudFrontS3CanonicalUserId));
```
ammarkarachi, line 30: Same as before 
```js
const uploadFileTasks = fileList.map(file => uploadFile(s3Client, bucketName, `${assetPath}/${file}`, file))
```

packages/amplify-category-auth/provider-utils/awscloudformation/utils/trigger-file-uploader.js
nchaloult: Good idea! Done 👍

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
nchaloult: Good idea! Done 👍

Issue #4115, opened 2020-04-29
Don't allow amplify delete command with trailing text


toddsampson:
>**Note**: If your feature-request is regarding the AWS Amplify Console service, please log it in the 
[official AWS Amplify Console forum](https://forums.aws.amazon.com/forum.jspa?forumID=314&start=0)

**Is your feature request related to a problem? Please describe.**
I just killed by amplify stack by accident. I typed `amplify delete api` instead of `amplify remove api` by mistake. I hit yes instinctively. (I've been using Amplify for some time now but haven't removed many features.) By the time I noticed the description for delete and hit ctrl+c the damage was done.

**Describe the solution you'd like**
This would be easy to fix if the CLI didn't allow any text following the `amplify delete` command. It is clearly ignoring what was typed, but it didn't error out.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

PR #4576, opened 2020-06-16
feat: add check for extra command line args provided with amplify delete

nchaloult:
*Issue #, if available:*
#4115 

*Description of changes:*
Since `$ amplify delete` is a sensitive and dangerous command, there should be checks in place to ensure that users truly mean to execute it. `$ amplify delete` may be confused with `$ amplify remove <category>`. If `$ amplify delete` is executed with any additional command line arguments, then an error message should be displayed indicating that the `delete` command does not accept any arguments, and a suggestion should be printed that asks the user if they meant to use the `remove` command instead of `delete`.

### DIFF
--- a/packages/amplify-cli/src/__tests__/commands/delete.test.ts
+++ b/packages/amplify-cli/src/__tests__/commands/delete.test.ts
@@ -1,0 +1,96 @@
+import { run as runDeleteCmd } from '../../commands/delete';
+
+// amplify delete run method calls process.exit() in certain situations. To avoid this affecting tests when they run, spy on process.exit()
+// calls and intercept them, performing a NOP, instead.
+//
+// https://stackoverflow.com/questions/46148169/stubbing-process-exit-with-jest
+const mockExit = jest.spyOn(process, 'exit').mockImplementation(() => undefined as never);
+
+describe('amplify delete: ', () => {
+  it('delete run method should exist', () => {
+    expect(runDeleteCmd).toBeDefined();
+  });
+
+  const mockContextNoCLArgs = {
+    amplify: {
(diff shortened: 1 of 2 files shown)


ammarkarachi:
@nchaloult We have a command to support our headless users `amplify delete --force` I believe this would break them

nchaloult:
> @nchaloult We have a command to support our headless users `amplify delete --force` I believe this would break them

I tested this locally with `$ amplify-dev delete --force` and `$ amplify-dev delete <any-other-argument>`. The `delete --force` command seems to behave as expected, even after the changes in this PR. I believe that's because of the first `if` statement in the `getConfirmation()` func in `packages/amplify-cli/src/extensions/amplify-helpers/delete-project.js`

```
if (context.input.options && context.input.options.force) {
  return {
    proceed: true,
    deleteS3: true,
    deleteAmpilfyApp: true,
  };
}
```

nchaloult:
Are there any other situations where `$ amplify delete` may expect additional command line arguments?

ammarkarachi:
Can you write a unit test that tests this change?

nchaloult:
> Can you write a unit test that tests this change?

I wrote a first draft of a set of tests.

I'm concerned about the fact that the `delete.run()` method calls `process.exit(1)` if additional command line arguments are present. I did some reading about how Jest handles situations where functions that tests call try to interrupt or kill a node process, but I'm still not sure if this will be a problem when these new tests run on this project's CI pipeline. What do you think?

nchaloult:
Addressed the concerns with `process.exit()`, and moved the new test file to a new directory that the CI pipeline recognizes and runs.

Issue #4579, opened 2020-06-16
`amplify delete` — change prompt default value from "yes" to "no"


nchaloult:
**Is your feature request related to a problem? Please describe.**
After running `$ amplify delete`, an "are you sure you want to continue" prompt appears. The default value for that prompt is "yes."

**Describe the solution you'd like**
Since this command is dangerous and sensitive, I feel like users should have to explicitly indicate that they'd like to run the `delete` command by answering "yes" to this prompt.

**Describe alternatives you've considered**
N/A

**Additional context**
Semi-related to #4115 

I'm proposing that this prompt

    $ amplify delete
    $ ? Are you sure you want to continue? ... (Y/n)

look like this instead:

    $ amplify delete
    $ ? Are you sure you want to continue? ... (y/N)

PR #4580, opened 2020-06-16
feat: change `amplify delete` prompt default value from yes to no

nchaloult:
*Issue #, if available:*
#4579 

*Description of changes:*
After running `$ amplify delete`, an "are you sure you want to continue" prompt appears. This PR proposes to change that prompt's default response value from "yes" to "no."

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/delete-project.js
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/delete-project.js
@@ -61,6 +61,7 @@ async function getConfirmation(context, env) {
       `Are you sure you want to continue? This CANNOT be undone. (This would delete ${environmentText} of the project from the cloud${
         env ? '' : ' and wipe out all the local files created by Amplify CLI'
       })`,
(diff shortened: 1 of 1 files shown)


ammarkarachi:
@nchaloult Good job following through. This might not be apparent but there are tests under /packages/amplify-e2e-tests/src/__tests__/. That test this functionality. Can you run those tests and see if they fail?

I suspect there are instances when we don't really send a 'yes' to the command line but rather just a carriage return

nchaloult:
> Can you run those tests and see if they fail?

The tests in `packages/amplify-e2e-tests/src/__tests__/delete.test.ts` passed on my machine 👍 

It looks like a 'y' is sent to stdin by `deleteProject()` in `packages/amplify-e2e-core/src/init/deleteProject.ts `:
```
...
.wait('Are you sure you want to continue?')
.sendLine('y')
.sendCarriageReturn()
...
```

ammarkarachi:
LGTM!
nice

PR #4602, opened 2020-06-17
fix: change scope of hashed files for AppSync

attilah:
*Issue #, if available:*

feat(graphql-key-transformer): add query automatically for named keys (#4458)
fix: change scope of hashed files for AppSync …
chore: fix expect string for new plugin e2e test

*Description of changes:*

Add back the following feature: Automatically add query field for named queries and expose a new argument `generateQuery` to control generation of query.

Change scope of hashed files for AppSync to guarantee change detection by CloudFormation by including build artifacts of schema compilation

Migration test is covered in a separate PR


packages/amplify-provider-awscloudformation/lib/upload-appsync-files.js
```diff
@@ -218,11 +220,30 @@ async function uploadAppSyncFiles(context, resourcesToUpdate, allResources, opti
  * should return the same hash.
  */
 async function hashDirectory(directory) {
```
SwaySway, line 222: This would create a new S3DeploymentRootKey. Wouldn't we also want to update [resource-status](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-cli/src/extensions/amplify-helpers/resource-status.js) as well to indicate an update is needed?

attilah: Not at part of this change, we've a separate issue around last push timestamp and hashes.

PR #4610, opened 2020-06-18
chore: remove unnecessary stack trace log when adding services

sebastiancrossa:
*Issue #, if available:*
aws-amplify#4283

*Description of changes:*
The unnecessary stack traces that were displayed through the console when a user tried to run amplify add <service> without initializing amplify first was because of a lacking try/catch block which displays only the error message, so I added that onto the parts where the addition of services was being called.

### DIFF
--- a/packages/amplify-category-predictions/commands/predictions/add.js
+++ b/packages/amplify-category-predictions/commands/predictions/add.js
@@ -27,12 +27,12 @@ module.exports = {
         print.success('Some next steps:');
         print.info('"amplify push" builds all of your local backend resources and provisions them in the cloud');
         print.info(
-          '"amplify publish" builds all of your local backend and front-end resources (if you added hosting category) and provisions them in the cloud'
+          '"amplify publish" builds all of your local backend and front-end resources (if you added hosting category) and provisions them in the cloud',
         );
         print.info('');
       })
       .catch(err => {
-        context.print.info(err.stack);
+        context.print.error(err.message);
         context.print.error('An error occurred when adding the predictions resource');
       }),
 };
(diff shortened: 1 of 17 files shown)


packages/amplify-category-predictions/commands/predictions/update.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
         context.print.success(`Successfully updated resource ${resourceName} locally`);
       })
       .catch(err => {
-        context.print.info(err.stack);
+        context.print.error(err.message);
```
ammarkarachi, line 23: This is going to swallow errors on update there is a possibility of users losing information of their errors on update of a valid project

packages/amplify-category-predictions/commands/predictions/add.js
```diff
@@ -27,12 +27,12 @@ module.exports = {
...
         print.info('"amplify push" builds all of your local backend resources and provisions them in the cloud');
         print.info(
-          '"amplify publish" builds all of your local backend and front-end resources (if you added hosting category) and provisions them in the cloud'
+          '"amplify publish" builds all of your local backend and front-end resources (if you added hosting category) and provisions them in the cloud',
         );
         print.info('');
       })
       .catch(err => {
-        context.print.info(err.stack);
+        context.print.error(err.message);
```
ammarkarachi, line 35: Same here, we still need the stack trace in case of other errors.

ammarkarachi:
@sebastiancrossa 
It's great that you have written tests for the amplify execute command but the way to go about it would be to have the smallest impact to change the current functionality. 

If the project is not an amplify project any other command that does not initialize, configure or pull an existing project should result in an error/information without a stack trace. Refer to [docs](https://docs.amplify.aws/cli) for more information.
`amplify pull`
`amplify configure`
`amplify init`

Issue #4621, opened 2020-06-19
Error on push with Python Lambda function


dave-moser:
**Describe the bug**
After adding a Python Lambda function to my Amplify project and attempting to push, I receive a UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)

This causes the push to exit with failure.

**Amplify CLI Version**
4.21.4

**To Reproduce**
- amplify add function
- select Python and generate the function
- amplify push

**Expected behavior**
Push should complete without error.

**Screenshots**
![image](https://user-images.githubusercontent.com/54151448/85177901-f7472d00-b231-11ea-93fa-259b74b0ecc3.png)

**Desktop (please complete the following information):**
 - OS: Mac
 - Node Version. v12.14.1

**Additional context**
If I manually set an environment variable: PIPENV_VERBOSITY=-1 I am able to run the push successfully.

SwaySway:
Closing this as the fix has been merged and will be included in the next release after v4.24,

PR #4673, opened 2020-06-25
fix(amplify-python-function-runtime-provider): change exec to execa

SwaySway:
changed execa, removed outputOnStderr as it's not needed for py3, remove console.log from build
process since the err is thrown

ref #4621

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-python-function-runtime-provider/src/util/buildUtils.ts
+++ b/packages/amplify-python-function-runtime-provider/src/util/buildUtils.ts
@@ -5,7 +5,7 @@ import glob from 'glob';
 
 export async function pythonBuild(params: BuildRequest): Promise<BuildResult> {
   if (!params.lastBuildTimestamp || isBuildStale(params.srcRoot, params.lastBuildTimestamp)) {
-    const pipenvLogs = await execAsStringPromise('pipenv install', { cwd: params.srcRoot }, undefined, true);
+    const pipenvLogs = await execAsStringPromise('pipenv install', { cwd: params.srcRoot });
     console.log(pipenvLogs);
     return Promise.resolve({ rebuilt: true });
   }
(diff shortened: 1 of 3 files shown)


packages/amplify-python-function-runtime-provider/src/util/buildUtils.ts
```diff
@@ -5,7 +5,7 @@ import glob from 'glob';
 
 export async function pythonBuild(params: BuildRequest): Promise<BuildResult> {
   if (!params.lastBuildTimestamp || isBuildStale(params.srcRoot, params.lastBuildTimestamp)) {
-    const pipenvLogs = await execAsStringPromise('pipenv install', { cwd: params.srcRoot }, undefined, true);
+    const pipenvLogs = await execAsStringPromise('pipenv install', { cwd: params.srcRoot }, undefined);
```
edwardfoyle, line 8: nit don't need to specify undefined for the last param

packages/amplify-python-function-runtime-provider/src/util/pyUtils.ts
```diff
@@ -31,19 +31,11 @@ export function majMinPyVersion(pyVersion: string): string {
...
-  errorMessage?: string,
-  outputOnStderr: boolean = false,
-): Promise<string> {
-  const { stdout, stderr } = await promisify(exec)(command, opts);
-  if (outputOnStderr) {
-    return stderr.toString('utf8').trim();
+export async function execAsStringPromise(command: string, opts?: ExecOptions, errorMessage?: string): Promise<string> {
+  try {
+    const { stdout } = await execa.command(command, opts);
+    return stdout;
```
edwardfoyle, line 37: nit could write as `return (await execa.command(command, opts)).stdout`

chenei:
@SwaySway can you please fix the comments? :)
Really waiting for this PR.

kevinpallado:
Still having the same problem

image

OS: windows 10
amplify version : 4.25.0
python version : 3.8.5

yes1vinod:
Even I am having the same problem

hogarthww-labs:
We're also having this problem

Issue #4638, opened 2020-06-22
DataStore owner + IAM auth rule combination not working


mdoesburg:
**Which Category is your question related to?**
- DataStore
- `@auth` directive

**Amplify CLI Version** 
4.21.3

**Provide additional details e.g. code snippets**
Hey @undefobj, you stated [here](https://github.com/aws-amplify/amplify-cli/issues/1678#issuecomment-505039534) that using a User Pool user inside a Lambda to call the AppSync API is bad practice, and though I agree, I am unable to make the IAM auth rule work in conjunction with the owner rule on the DataStore.

This is my GraphQL schema:
```graphql
type SharedContent
  @model
  @auth(rules: [
    { allow: owner },
    { allow: private, provider: iam }
  ])
{
  id: ID!
  content: String
  owner: String
}
```
It generates the following subscriptions:
```graphql
onCreateSharedContent(owner: String!): SharedContent @aws_subscribe(mutations: ["createSharedContent"]) @aws_iam @aws_cognito_user_pools
onUpdateSharedContent(owner: String!): SharedContent @aws_subscribe(mutations: ["updateSharedContent"]) @aws_iam @aws_cognito_user_pools
onDeleteSharedContent(owner: String!): SharedContent @aws_subscribe(mutations: ["deleteSharedContent"]) @aws_iam @aws_cognito_user_pools
```
And results in the following warnings/errors:
```
[WARN] 26:54.58 DataStore - Connection failed: {"errors":[{"message":"Validation error of type MissingFieldArgument: Missing field argument owner @ 'onDeleteSharedContent'"}]}
[WARN] 26:54.97 DataStore - Connection failed: {"errors":[{"message":"Validation error of type MissingFieldArgument: Missing field argument owner @ 'onCreateSharedContent'"}]}
[WARN] 26:54.156 DataStore - Connection failed: {"errors":[{"message":"Validation error of type MissingFieldArgument: Missing field argument owner @ 'onUpdateSharedContent'"}]}
```

I've posted about it [here](https://github.com/aws-amplify/amplify-js/issues/6108#issuecomment-647708908) as well. The only way I can currently keep all subscriptions working, and have the capability to run mutations from my Lambda is by using [Schema 5](https://github.com/aws-amplify/amplify-js/issues/6108#schema-5) from my examples, and logging in as a User Pool user that is part of the Admin group inside the Lambda. The Lambda is created using `amplify function add` and has proper access.

SwaySway:
[...]

PR #4719, opened 2020-06-30
fix(graphql-auth-transformer): allow optional owner arg on multi auth

SwaySway:
allow owner argument to optional when allow private, public, and/or static group access on
subscriptions

re #4638

*Issue #, if available:*

*Description of changes:*


packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1860,14 +1860,17 @@ operations will be generated by the CLI.`,
         // check if owner is enabled in auth
         const ownerRules = rules.filter(rule => rule.allow === OWNER_AUTH_STRATEGY);
         const needsDefaultOwnerField = ownerRules.find(rule => !rule.ownerField);
-        const hasStaticGroupAuth = rules.find(rule => rule.allow === GROUPS_AUTH_STRATEGY && !rule.groupsField);
+        // if there are any public, private, and/or static group rules then the owner argument is optional
```
jpignata, line 1863: I don't really follow this comment with the code below. Can you expound a bit upon the problem you're solving, and the desired future state, in the pull request description?

SwaySway: The owner argument becomes non-nullable if it's only auth rule on subscriptions, otherwise if there are other auth rules (such as public, private, static group rules) the owner argument becomes optional. This way if an argument such as `onCreatePost` is called with IAM auth the optional argument owner is not required because it was authorized via IAM.

packages/graphql-transformers-e2e-tests/src/__tests__/SubscriptionsWithAuthTest.e2e.test.ts
```diff
@@ -209,7 +209,8 @@ beforeAll(async () => {
 
     type Post @model
         @auth(rules: [
-            {allow: owner, ownerField: "postOwner"}
+            { allow: owner, ownerField: "postOwner" }
+            { allow: private, operations: [read], provider: iam }
```
yuth, line 213: I think we should create a separate type for this instead of changing the cognito based auth type to include IAM.

SwaySway: Adding this change tests that we can subscribe using IAM auth on a mutation that used UserPools auth. This also ensures we can call the subscription without an argument.

yuth: But do we have any tests where owner based auth makes subscription parameter non-billable?

SwaySway: good point added a unit test to confirm when only an owner rule is used the owner arg should be non nullable

PR #4734, opened 2020-07-02
test: add migration test for Lambda layers

jhockett:
Migration test for adding layer to pre-update function
Refactor metadata validation for Lambda layers
Update config.yml to latest e2e changes

### DIFF
--- a/packages/amplify-e2e-core/src/categories/lambda-layer.ts
+++ b/packages/amplify-e2e-core/src/categories/lambda-layer.ts
@@ -36,13 +36,13 @@ export function validatePushedVersion(projRoot: string, layerName: string, versi
   return false;
 }
 
-export function addLayer(cwd: string, settings?: any) {
+export function addLayer(cwd: string, settings?: any, testingWithLatestCodebase: boolean = false) {
   const defaultSettings = {
     permissions: [],
   };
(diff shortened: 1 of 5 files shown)


packages/amplify-e2e-core/src/categories/lambda-layer.ts
```diff
@@ -2,28 +2,25 @@ import fs from 'fs-extra';
...
-  if (layerExists) {
-    let validDir = fs.pathExistsSync(path.join(layerDir, 'opt'));
-    if (runtimes && runtimes.length) {
-      for (let runtime of runtimes) {
-        if (!fs.pathExistsSync(path.join(layerDir, getLayerRuntimeInfo(runtime).path))) {
-          return false;
-        }
+  let validDir = fs.pathExistsSync(path.join(layerDir, 'opt'));
+  if (runtimes && runtimes.length) {
+    for (let runtime of runtimes) {
```
nikhname, line 17: nit - could this be done with a foreach loop?

jhockett: Yes, but since I'm returning from the for loop, that's gonna require some slightly different logic. Unless there's a compelling reason I'm missing, I'm going to leave as is.

PR #3387, opened 2020-02-10
fix(amplify-appsync-simulator): forward stash to responseMappingTemplate

marcoreni:
*Description of changes:*
According to docs, stash should live across functions inside a pipeline resolver. Currently, the stash returned from the requestMappingTemplate is not forwarded to the responseMappingTemplate, therefore parameters set during request are not passed through. 

Ref: https://docs.aws.amazon.com/appsync/latest/devguide/pipeline-resolvers.html#ctx-stash

This PR fixes this by forwarding the requestTemplateResult stash to the responseMappingTemplate.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4738, opened 2020-07-03
fix(amplify-appsync-simulator): forward stash to responseMappingTemplate

perbeijer:
stash from requestMappingTemplate should be available in responseMappingTemplate

*Description of changes:*
https://github.com/aws-amplify/amplify-cli/pull/3387 fixed this for pipeline resolvers, but the same applies to unit resolvers as well on real AWS. See stash documentation at https://docs.aws.amazon.com/appsync/latest/devguide/resolver-context-reference.html 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1116,9 +1116,9 @@ workflows:
           filters:
             branches:
               only:
-                # - master
+                - master
                 - graphqlschemae2e
-                # - beta
+                - beta
           requires:
             - build
             - mock_e2e_tests
@@ -1161,7 +1161,7 @@ workflows:
             - test
             - mock_e2e_tests
             - graphql_e2e_tests
-            # - integration_test
+            - integration_test
             - amplify_console_integration_tests
             - amplify_migration_tests_latest
             - amplify_migration_tests_v4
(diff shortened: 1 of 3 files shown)


yuth:
Stash is available to resolvers only when they are pipeline resolvers. The changes in this PR adds that to Unit resolvers which is not supported in the service. Closing the PR

perbeijer:
> Thank you for the PR. Unfortunately stash is available to to are pipeline resolvers. The changes in this PR adds that to Unit resolvers which is not supported in the service. Closing the PR

Are you really sure about this? I have used the stash successfully on AWS to without pipeline. A quote from the docs page:

> The same stash instance lives through a single resolver execution. This means that you can use the stash to pass arbitrary data across request and response mapping templates, and across functions in a pipeline resolver.

Only the last part mentions pipeline. The first part applies to request and response in a single execution.

perbeijer:
I have created a small reproducible test for stash here https://github.com/perbeijer/stashtest

Deployed with "serverless deploy" this shows stash working when deployed on aws. Running locally with appsync simulator (using "serverless offline start") it doesn't work without this PR.

I have a deployed Appsync of this test running here https://fge5gp4r3bhkfgmv6p7gw4qhbu.appsync-api.us-east-1.amazonaws.com/graphql with apikey da2-z47mm2ytivaspbw7yplruopncm call it with graphql query { test }

PR #4754, opened 2020-07-07
test(amplify-console-integration-tests): fix afterEach

SwaySway:
*Issue #, if available:*

*Description of changes:*
- allow delete project to pass a project config when calling cloudformation

### DIFF
--- a/packages/amplify-console-integration-tests/src/pullAndInit/amplifyConsoleOperations.ts
+++ b/packages/amplify-console-integration-tests/src/pullAndInit/amplifyConsoleOperations.ts
@@ -26,7 +26,11 @@ export async function deleteAllAmplifyProjects(amplifyClient?: any) {
 
 export async function deleteAmplifyStack(stackName: string, cfnClient?: any) {
   if (!cfnClient) cfnClient = getConfiguredCFNClient();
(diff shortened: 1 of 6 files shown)


packages/amplify-console-integration-tests/src/pullAndInit/amplifyConsoleOperations.ts
```diff
@@ -26,7 +26,11 @@ export async function deleteAllAmplifyProjects(amplifyClient?: any) {
 
 export async function deleteAmplifyStack(stackName: string, cfnClient?: any) {
   if (!cfnClient) cfnClient = getConfiguredCFNClient();
-  await cfnClient.deleteStack({ StackName: stackName }).promise();
+  try {
+    await cfnClient.deleteStack({ StackName: stackName }).promise();
+  } catch (err) {
+    // do nothing
```
jhockett, line 32: nit: Elaborate the comment to explain why nothing happens here

SwaySway: This is mainly added since the stack may not exist and it will through an error. Though for this kind of err we can ignore as it's not pertinent to the test.

PR #4797, opened 2020-07-10
ci:split test to run in multiple regions concurrently

UnleashedMind:
Split the e2e tests into multiple regions so we can concurrently run more tests to reduce the overall execution time of the e2e tests in the release pipeline.

*Issue #, if available:*

*Description of changes:*


packages/amplify-e2e-core/src/utils/pinpoint.ts
```diff
@@ -99,8 +116,14 @@ export function initProject(cwd: string) {
       .wait('secretAccessKey')
       .sendLine(settings.secretAccessKey)
       .resumeRecording()
-      .wait('region')
-      .sendLine(settings.region)
+      .wait('region');
```
yuth, line 119: We have an util to do single and mutliselect. Please use that instead of a custom implmentation
https://github.com/aws-amplify/amplify-cli/blob/82d10933754230c311bf5f24dc1c59dfa393ce63/packages/amplify-e2e-core/src/utils/selectors.ts#L9-L22

You can see how it is used here
https://github.com/aws-amplify/amplify-cli/blob/82d10933754230c311bf5f24dc1c59dfa393ce63/packages/amplify-e2e-core/src/configure/index.ts#L39-L48

UnleashedMind: We can not use the multiselect util, those are for type 'checkbox' that uses space to select, the region selection is of type 'list' in inquirer.

yuth: There is a `singleSelect` function, which is used for region selection when setting up the profile, which is also a region picker.

UnleashedMind: updated

Issue #4795, opened 2020-07-10
Amplify pull stuck on fetching updates (latest CLI version)


mdoesburg:
**Describe the bug**
When running `amplify pull` on the latest CLI version, the console stays stuck on:

*⠏ Fetching updates to backend environment: dev from the cloud.*

Rolling back to 4.23.0 fixes the issue.

**Amplify CLI Version**
4.23.1

jhockett:
Hi @mdoesburg what categories are you using? Can you share your `amplify status`?

mdoesburg:
@jhockett 
```bash
| Category      | Resource name               | Operation | Provider plugin   |
| ------------- | --------------------------- | --------- | ----------------- |
| Api           | xxxxxxApi                   | No Change | awscloudformation |
| Api           | xxxxxxDataSyncApi           | No Change | awscloudformation |
| Auth          | xxxxxxCognito               | No Change | awscloudformation |
| Auth          | userPoolGroups              | No Change | awscloudformation |
| Function      | xxxxxxUserMigration         | No Change | awscloudformation |
| Function      | xxxxxxDynamoTrigger         | No Change | awscloudformation |
| Function      | xxxxxxTestLambda            | No Change | awscloudformation |
| Function      | xxxxxxPreTokenGeneration    | No Change | awscloudformation |
| Function      | xxxxxxDataSync              | No Change | awscloudformation |
| Analytics     | xxxxxxPinpoint              | No Change | awscloudformation |
| Notifications | xxxxxxPinpoint              | No Change |                   |
```

TrungRueta:
version 4.24.0.

I got same issue today. I updated from 4.23 yesterday and today when do `amplify pull` or even `amplify pull --restore` it always freeze at `Fetching updates to backend environment:...` . i try to wait near 30 mins and it not finish. Check files on project look like download done, but terminal command freeze forever.

aavargasp:
```
amplify --version
4.24.1
```

I started having the same problem after updating to version 4.24 and today I updated to 4.24.1 hoping that the issue was resolved but wasn't the case.

Here my project resources:

Current Environment: dev

```
| Category | Resource name              | Operation | Provider plugin   |
| -------- | -------------------------- | --------- | ----------------- |
| Auth     | xxxxxxxxxxxxxxxxxxxxxxxxxx | No Change | awscloudformation |
| Function | xxxxxxxxxxxxxxxxxxxxxxxxxx | No Change | awscloudformation |
[...]

PR #4836, opened 2020-07-15
fix(cli): moving the spinner above category initialization tasks

SwaySway:
stopping the spinner before running category init tasks since the spinner can swallow outputs (such
as questions via inquirer)
related issue: https://github.com/sindresorhus/ora/issues/90

re #4795

change type to intersection type in e2e

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-cli/src/initialize-env.js
+++ b/packages/amplify-cli/src/initialize-env.js
@@ -9,7 +9,7 @@ const { getProviderPlugins } = require('./extensions/amplify-helpers/get-provide
 
 async function initializeEnv(context, currentAmplifyMeta) {
   const currentEnv = context.exeInfo.localEnvInfo.envName;
-  const isPulling = context.input.command === 'pull' || (context.input.command === 'env' && context.input.subCommands[0] === 'pull');
+  let isPulling = context.input.command === 'pull' || (context.input.command === 'env' && context.input.subCommands[0] === 'pull');
 
   try {
     const { projectPath } = context.exeInfo.localEnvInfo;
@@ -43,7 +43,7 @@ async function initializeEnv(context, currentAmplifyMeta) {
             categoryInitializationTasks.push(() => initEnv(context));
           }
         } catch (e) {
-          context.print.warning(`Could not run initEnv for ${category}`);
+          context.print.warning(`Could not load initEnv for ${category}`);
         }
       });
     });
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/initialize-env.js
```diff
@@ -63,15 +63,16 @@ async function initializeEnv(context, currentAmplifyMeta) {
     );
     await sequential(initializationTasks);
 
+    spinner.succeed(
+      isPulling ? `Successfully pulled backend environment ${currentEnv} from the cloud.` : 'Initialized provider successfully.',
+    );
+    // if pull is successful further errors would result from initializing the env
+    isPulling = false;
```
jhockett, line 70: Since isPulling will always be false, we can remove these lines and the subsequent checks below

```js
    ...
    // Generate AWS exports/configurtion file
    await context.amplify.onCategoryOutputsChange(context, currentAmplifyMeta);
    context.print.success('Initialized your environment successfully.');
  } catch (e) {
    spinner.fail('There was an error initializing your environment.');
    throw e;
  }
```

SwaySway: It can still fail while running `await sequential(initializationTasks);` hence the pull message. Though I figure since the code here does the same thing on a pull or init we can change the error message to be the same as you suggested.

jhockett: Ah that's an oversight by me. I guess it's just the context.print.success() call that can be simplified. Edited the above comment.

SwaySway: Changed the fail message to be the same. We can remove the `isPulling` as well if we think the message would be more concise to just say it's initializing the env.

packages/amplify-cli/src/initialize-env.js
```diff
@@ -43,7 +43,7 @@ async function initializeEnv(context, currentAmplifyMeta) {
             categoryInitializationTasks.push(() => initEnv(context));
           }
         } catch (e) {
-          context.print.warning(`Could not run initEnv for ${category}`);
+          context.print.warning(`Could not load initEnv for ${category}`);
```
jhockett, line 46: Nit: ```context.print.warning(`Could not load environment for ${category}`);```

PR #4866, opened 2020-07-17
Update Resolver Ver in @model

SwaySway:
*Issue #, if available:*

*Description of changes:*
- update resolver config to 2018 for better error handling on failed operations
- return errorType for ver 2018 resolvers
  - previously for 2018 the error type was null
- additive update on e2e and mock e2e tests

### DIFF
--- a/packages/amplify-appsync-simulator/src/velocity/index.ts
+++ b/packages/amplify-appsync-simulator/src/velocity/index.ts
@@ -112,7 +112,7 @@ export class VelocityTemplate {
       stash: convertToJavaTypes(stash || {}),
       source: convertToJavaTypes(source),
       result: convertToJavaTypes(result),
-      error,
+      error: error ? { ...error, type: error.extensions.errorType } : error,
     };
 
     if (typeof prevResult !== 'undefined') {
--- a/packages/amplify-appsync-simulator/src/velocity/index.ts
+++ b/packages/amplify-appsync-simulator/src/velocity/index.ts
@@ -112,6 +112,8 @@ export class VelocityTemplate {
       stash: convertToJavaTypes(stash || {}),
       source: convertToJavaTypes(source),
       result: convertToJavaTypes(result),
+      // surfacing the errorType to ensure the type is included in $ctx.error
+      // Mapping Template Errors: https://docs.aws.amazon.com/appsync/latest/devguide/troubleshooting-and-common-mistakes.html#mapping-template-errors
       error: error ? { ...error, type: error.extensions.errorType } : error,
     };
 
--- a/packages/graphql-auth-transformer/src/resources.ts
+++ b/packages/graphql-auth-transformer/src/resources.ts
@@ -25,6 +25,7 @@ import {
   ifElse,
   newline,
   methodCall,
+  RESOLVER_VERSION_ID,
 } from 'graphql-mapping-template';
 import { ResourceConstants, NONE_VALUE } from 'graphql-transformer-common';
 import GraphQLApi, {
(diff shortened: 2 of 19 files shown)


packages/amplify-appsync-simulator/src/velocity/index.ts
```diff
@@ -112,7 +112,7 @@ export class VelocityTemplate {
       stash: convertToJavaTypes(stash || {}),
       source: convertToJavaTypes(source),
       result: convertToJavaTypes(result),
-      error,
+      error: error ? { ...error, type: error.extensions.errorType } : error,
```
undefobj, line 115: I suggest you add a comment block with the reasons for doing this change and linking to the AppSync docs, that way future engineers can understand the context

packages/graphql-auth-transformer/src/resources.ts
```diff
@@ -900,7 +901,7 @@ identityClaim: "${rule.identityField || rule.identityClaim || DEFAULT_IDENTITY_F
       TypeName: subscriptionTypeName,
       RequestMappingTemplate: print(
         raw(`{
-    "version": "2018-05-29",
+    "version": "${RESOLVER_VERSION_ID}",
```
undefobj, line 904: One thought - Can we tell in general what versions customers are using in templates? For example if they manually create a resolver with the older version perhaps we could flag a warning similar to what we do today for types that are not annotated with `@auth`?

Also have we given any thought to implications for existing projects, or do we expect this to be transparent?

packages/graphql-transformers-e2e-tests/src/__tests__/ModelAuthTransformer.e2e.test.ts
```diff
@@ -528,6 +528,7 @@ describe(`ModelAuthTests`, async () => {
     );
     expect(updateResponse.data.updatePost).toEqual(null);
     expect(updateResponse.errors.length).toEqual(1);
+    expect((updateResponse.errors[0] as any).data).toBeNull();
```
undefobj, line 531: As any >.<

packages/graphql-mapping-template/src/dynamodb.ts
```diff
@@ -288,10 +284,12 @@ export class DynamoDBMappingTemplate {
...
 
-  public static dynamoDBResponse(
-    expression: Expression = ref('util.error($ctx.error.message, $ctx.error.type, $ctx.result)')
-  ): CompoundExpressionNode {
-    return compoundExpression([ifElse(ref('ctx.error'), expression, ref('util.toJson($ctx.result)'))]);
+  public static dynamoDBResponse(isSyncEnabled: boolean, returnExpression?: Expression): CompoundExpressionNode {
+    const errorExpresion = isSyncEnabled
+      ? ref('util.error($ctx.error.message, $ctx.error.type, $ctx.result)')
+      : ref('util.error($ctx.error.message, $ctx.error.type)');
+    const resultExpression = returnExpression ? returnExpression : ref('util.toJson($ctx.result)');
```
undefobj, line 291: Small nit: For readability it's best to keep formatting strategy with line breaks consistent in a codebase, so if one ternary expression puts the cases on separate lines, the subsequent should do the same.

packages/graphql-auth-transformer/src/resources.ts
SwaySway: iirc there isn't a mechanism in place which can run a LGTM for VTL. If anything we could the resolver changelog into the custom resolver docs.

packages/graphql-transformers-e2e-tests/src/__tests__/ModelAuthTransformer.e2e.test.ts
SwaySway: Good point here there is a few instances of this in some of the tests, though this change might be best in it's own PR.

Issue #2702, opened 2019-11-07
API (GraphQL with DynamoDB): Creating GSI reusing table's Partition Key


lazy-var:
Not completely sure if this is a bug or feature request, but I need assistance on it.

`graphql-transformer`, `code-gen?`, `bug`, `help-wanted`, `feature-request`, `enhancement`.

**Describe the bug**
I need to create a table GSI (Global Secondary Index), setting as Partition Key the same field that is used as Partition Key on the main table (Primary Index).

**Amplify CLI Version**
3.16.0

**To Reproduce**
i.e.:
I have already created a model (running stack) with:
`@key(fields: ["partitionKey", "sortKey"])`
And afterwards I want to add:
`@key(name: "GlobalSecondaryIndex", fields: ["partitionKey", "GSIsortKey"])`
When running `amplify api push` I get:

> Attempting to add a local secondary index to the tableName table in the stackName stack. Local secondary indexes must be created when the table is created.
> There was an error pushing the API resource
> Attempting to add a local secondary index to the tableName table in the stackName stack. Local secondary indexes must be created when the table is created.
> Cause: Adding a @key directive where the first field in 'fields' is the same as the first field in the 'fields' of the primary @key.
> How to fix: Change the first field in 'fields' such that a global secondary index is created or delete and recreate the model.

**Expected behavior**
Instead of a LSI (Local Secondary Index) I expect to create a GSI. This design works if it's done manually on DynamoDB website.

An expected behavior for `@key (name: "gsi-name", fields [<partition-key>, <sort-key>]` could be:
- on table creation -> create LSI
- on already deployed table -> create GSI

**Additional context**
Related issues: #1725 #1917

attilah:
@Soek22 could you tell me what benefit are you looking for, what do you want to achieve when you create LSI/GSI for the same fields that is already key fields on your type?

Soek22:
Hello @attilah , not exactly same fields:
For primary index: `@key(fields: ["partitionKey", "sortKey"])`
For GSI (secondary index): `@key(name: "GlobalSecondaryIndex", fields: ["partitionKey", "GSIsortKey"])`

I need them to share the same field as partition key(`partitionKey`), but sort them with different field (`sortKey` and `GSIsortKey`). 

This allows me to retrieve same data with two different data access patterns, plus sort data by different fields.

> ie.: partitionKey=user_id; sortKey=item_id; GSIsortKey=item_name
[...]

PR #4667, opened 2020-06-24
feat(graphql-key-transformer): add a lock flag to @key

SwaySway:
- added a lock flag to @key to toggle the behavior to indicate whether a secondary index is local or
global
-  by adding this flag it also changes the default behavior to create a GSI

re #2702

*Issue #, if available:*

*Description of changes:*

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #4979, opened 2020-07-30
feat(graphql-key-transformer): add type arg to specify @key type

SwaySway:
*Issue #, if available:*
re #2702
*Description of changes:*
 - added a type argument to `@key` to toggle the behavior to indicate whether a secondary index is local or
global
- Added sanity check when removing an LSI from an already existing table
- `@key` Validations
  - check that an LSI can only be added when when a primary composite key is defined
  - check that an LSI has the same field length as the primary composite key
- Removed unused imports from amplify e2e
- Updated `@connection` unit tests since the schema provided would fail if actually deployed.

### DIFF
--- a/packages/amplify-e2e-tests/schemas/migrations_key/cant_remove_lsi.graphql
+++ b/packages/amplify-e2e-tests/schemas/migrations_key/cant_remove_lsi.graphql
@@ -1,0 +1,11 @@
+type Todo
+  @model
+  @key(fields: ["id", "version"])
+  @key(name: "SomeLSI", fields: ["id", "name"], type: GSI)
+  @key(name: "SomeGSI", fields: ["name", "id"]) {
+  id: ID!
+  id2: ID!
+  name: String!
+  name2: String!
+  version: String!
+}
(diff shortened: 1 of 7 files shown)


SwaySway:
Addressing some comments in the previous review #4667 

> * What is the default value of the `lock` argument, that preserves the current behavior?

The current behavior remains as is where the transform where infer that an LSI is being created when using the hash field of the primary key

> * The use of the `lock` argument is only valid for non-named `@key`s so we should have a validation in place to make 
sure Customers cannot use it in an invalid combination, also have tests for that.

We do have migration tests for this test cases covered currently are 
- Changing GSI to LSI is not possible as the table already exists
- Renaming LSI is not possible once the table has been created

The PR does add tests for one more use case here which is LSI -> GSI
- The sanity check throws an error when an LSI has been removed from an already created table.

> * I would like to see E2E tests that is covering the removal of locked and non-locked keys after they were pushed.

- Added an E2E migration test to track LSI removal

SwaySway:
Closing in favor #5115

PR #5033, opened 2020-08-06
feat(frontend-plugin): initial feature commit

haverchuck:
*Description of changes:*
Adds a new frontend plugin for Flutter projects.  Currently codegen is not supported and our initial preview is for auth, storage and analytics only, and so we display a warning message at project init.  We have a flutter project being created and tested in the auth e2e tests; most other frontend plugins do not contain tests but we will add these prior to preview release.  


packages/amplify-app/src/index.js
```diff
@@ -168,6 +168,7 @@ async function guessPlatform() {
     javascript: 'amplify-frontend-javascript',
     android: 'amplify-frontend-android',
     ios: 'amplify-frontend-ios',
+    flutter: 'amplify-frontend-flutter',
```
attilah, line 171: amplify-app needs a dependency on this new package

packages/amplify-cli/package.json
```diff
@@ -52,6 +52,7 @@
     "amplify-dotnet-function-runtime-provider": "1.2.0",
     "amplify-dotnet-function-template-provider": "1.3.0",
     "amplify-frontend-android": "2.13.4",
+    "amplify-frontend-flutter": "0.0.1",
```
attilah, line 55: If we start with 0.0.1 the published version will be 0.1.1, not? I think it is ok to start with 1.0.0 here so 1.1.0 will be the published version if you don't know a shortcut to make lerna to make it 1.0.0

packages/amplify-cli/src/domain/amplify-toolkit.ts
```diff
@@ -35,6 +35,7 @@ export class AmplifyToolkit {
   private _pushResources: any;
   private _storeCurrentCloudBackend: any;
   private _readJsonFile: any;
+  private _readJsonFromDart: any;
```
attilah, line 38: This is flutter specific I'd not pollute the context and the toolkit if this function is not used from anywhere else, move it to frontend package

packages/amplify-e2e-core/src/init/initProjectHelper.ts
```diff
@@ -100,6 +100,36 @@ export function initAndroidProjectWithProfile(cwd: string, settings: Object) {
...
+  return new Promise((resolve, reject) => {
+    spawn(getCLIPath(), ['init'], { cwd, stripColors: true })
+      .wait('Enter a name for the project')
+      .sendLine(s.name)
+      .wait('Enter a name for the environment')
+      .sendLine(s.envName)
+      .wait('Choose your default editor:')
+      .sendLine(s.editor)
+      .wait("Choose the type of app that you're building")
+      .send('j')
```
attilah, line 113: We have list selector functions for e2e to make it better, use those.

packages/amplify-cli/src/extensions/amplify-helpers/write-json-to-dart.js
```diff
@@ -0,0 +1,15 @@
+import fs from 'fs-extra';
+import path from 'path';
+
+function writeJsonToDart(dest, obj) {
+  const destPath = path.parse(dest).dir;
+  if (!fs.existsSync(destPath)) {
+    fs.mkdirSync(destPath, { recursive: true });
+  }
+  const dartContent = `const amplifyconfig = ''' ${obj}''';`;
```
SwaySway, line 9: Any particular reason we are creating a dart file with the exports existing a string instead of writing them to a json file?

packages/amplify-codegen/src/commands/add.js
```diff
@@ -20,7 +20,7 @@ const fs = require('fs-extra');
 const askForFrontend = require('../walkthrough/questions/selectFrontend');
 const askForFramework = require('../walkthrough/questions/selectFramework');
 
-const frontends = ['android', 'ios', 'javascript'];
+const frontends = ['android', 'ios', 'javascript', 'flutter'];
```
SwaySway, line 23: Considering Flutter is more an SDK, might it be better to move flutter under android as a framework?

packages/amplify-cli/src/extensions/amplify-helpers/write-json-to-dart.js
haverchuck: Yeah - for easier importation into Flutter project.

packages/amplify-codegen/src/commands/add.js
haverchuck: Flutter is for multiplatform development, so flutter projects will sit on top of all frontend types.

packages/amplify-app/src/index.js
haverchuck: @attilah Do we have a way to easily disable particular commands?  It would be nice to just shut off api, app, codegen, etc until those features are supported.

packages/amplify-cli/src/domain/amplify-toolkit.ts
haverchuck: Done!

packages/amplify-cli/package.json
haverchuck: Doesn't 1.0 suggest that the feature is stable?  This is feature is explicitly in preview.

packages/amplify-app/src/index.js
haverchuck: We are not supporting amplify-app yet.  This is just a developer preview.

PR #5062, opened 2020-08-11
fix: added sharedId in manual auth flow

akshbhu:
*Description of changes:*

When adding auth manually configure everything, two IAM roles are created that have 'undefined' in their name.This PR adds SharedId to the props , so that IAM roles generated has proper Id instead of undefined


akshbhu:
Closing as this issue is already resolved

Issue #4992, opened 2020-08-01
Cannot mock lambda function locally


prophet1996:
**Describe the bug**
Unable to mock feebeeclientpaymentfunction. Mocking a function with layers is not supported. To test in the cloud: run "amplify push" to deploy your function to the cloud and then run "amplify console function" to test your function in the Lambda console.

**Amplify CLI Version**
4.25.0

**To Reproduce**
Create a lambda function and try to run mock

**Expected behavior**
should run the mock function

**Screenshots**
![Screenshot from 2020-08-01 14-42-46](https://user-images.githubusercontent.com/16836510/89098468-88025280-d405-11ea-8141-9efc9425cd69.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu Xenial
 - Node Version. v13.11.0

akshbhu:
Hi @prophet1996 

Currently we doesn't support mock functionality of functions attached with layers , you can only mock functions that aren't attached to layers.
If you just want to mock a function , you can remove the layer attached to function and mock it but careful when you are pushing as you have to again attach the layer as per your use-case.

If you are facing other problems , feel free to comment and ill reopen this issue again

prophet1996:
hi @akshbhu  i don't have layers in this,

akshbhu:
Hi @prophet1996 

Can you tell me steps you did before reaching this error?

Also can you share amplifymeta.json file , if the file contains sensitive information , you can mail us at amplify-cli@amaon.com with issueNo. as subject.

pkudinov:
I have the same problem on 4.26.0.
Here are my steps:
```amplify add function
? Select which capability you want to add: Lambda function (serverless function)
? Provide a friendly name for your resource to be used as a label for this category in the project: rendermarkdown
? Provide the AWS Lambda function name: rendermarkdown
? Choose the runtime that you want to use: NodeJS
? Choose the function template that you want to use: Hello World
? Do you want to access other resources in this project from your Lambda function? Yes
? Select the category hosting, function, storage
Hosting category has a resource called amplifyhosting
Policies cannot be added for hosting/amplifyhosting
? Function has 3 resources in this project. Select the one you would like your Lambda to access management17df9d05CreateAuthChallenge
? Select the operations you want to permit formanagement17df9d05CreateAuthChallenge read
? Storage has 3 resources in this project. Select the one you would like your Lambda to access Application:@model(appsync)
? Select the operations you want to permit for Application:@model(appsync) read

[...]

PR #5064, opened 2020-08-11
fix: change logic to separate functions from layers in mock

akshbhu:
*Issue #4992 *

*Description of changes:*

* Changed the logic to mock functions with layers when dependency on a function

* Added tests for the same

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1116,9 +1116,9 @@ workflows:
           filters:
             branches:
               only:
-                # - master
+                - master
                 - graphqlschemae2e
-                # - beta
+                - beta
           requires:
             - build
             - mock_e2e_tests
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -1116,9 +1116,9 @@ workflows:
           filters:
             branches:
               only:
-                # - master
+                - master
                 - graphqlschemae2e
-                # - beta
+                - beta
```
jhockett, line 1121: Is this related to the mock changes?

.circleci/config.yml
```diff
@@ -1161,7 +1161,7 @@ workflows:
             - test
             - mock_e2e_tests
             - graphql_e2e_tests
-            # - integration_test
+            - integration_test
```
jhockett, line 1164: Is this related to the mock changes?

.circleci/config.yml
akshbhu: No , this is because of issues in circleCI yesterday. I am now changing it back to original

.circleci/config.yml
akshbhu: same as above

prophet1996:
Hi, @akshbhu just wondering when will this be available on taking npm update

PR #5083, opened 2020-08-13
feat(amplify-codegen): add inquiries for user input to configure modelgen output

### LINKED ISSUE #3585: Can you change the location of where models are generated when using amplify codegen models?
Datastore, Amplify codegen models

Right now when I run amplify codegen models, the models are currently created in the src/models directory. Is there any way to change this?

I imagine that it would act similarly to how the normal codegen where the file location can be changed in the .graphqlconfig.yml

### LINKED ISSUE #4548: Specify codegen target directory
A customer is experiencing that fixed codegen target deviates from the actual `src` directory of the project. 

Enable specifying desired target `src` directory when running `amplify codegen models`

N/A

**Additional context**
[Customer issue on Android platform](https://github.com/aws-amplify/amplify-android/issues/572)

AaronZyLee:
*Issue #, if available:*
fix #3585 fix #4548
*Description of changes:*

- Running `amplify codegen models` will first read from `amplify/.config/project-config.json` to get the output path. If it does not exist, the default output path will be used.

- Added an inquiry option when you run `amplify codegen models --configPath`, which will provide a customized output path for users.

- The path is stored in `amplify/.config/project-config.json` with the entry `modelgen.outputPath` after the first time modelgen runs. You can change the value either directly in this file or add `--configPath` flag in cmd. The first approach is recommended for iOS/Android developer using amplify tool plugins.

## Testing Done

- [x] passed API test in android studio gradle plugin

- [x] passed API test in xcode build process

- [x] added and passed unit tests for modelgen

## Examples

- `amplify codegen models`
Screen Shot 2020-08-24 at 5 43 31 PM

- `amplify codegen models --configPath`
Screen Shot 2020-08-24 at 5 44 51 PM

### DIFF
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -104,8 +104,8 @@ function loadSchema(apiResourcePath) {
 async function getModelOutputPath(context) {
   const projectConfig = context.amplify.getProjectConfig();
 
-  if(projectConfig.modelgen && projectConfig.modelgen.outputPath && !context.parameters.options.configpath) {
-    context.print.success('\nThe output path for modelgen is already setup. You can edit the path in .config/project-config.json\n');
+  if(projectConfig.modelgen && projectConfig.modelgen.outputPath && !context.parameters.options.configPath) {
+    context.print.success('\nThe output path for modelgen is already configured. You can edit the path in .config/project-config.json\n');
     return projectConfig.modelgen.outputPath;
   }
 
@@ -125,7 +125,7 @@ async function getModelOutputPath(context) {
   }
 
   let outputPath = defaultPath;
-  if(context.parameters.options.configpath) {
+  if(context.parameters.options.configPath) {
     const answer = await askForModelOutputPath(defaultPath);
     outputPath = answer.outputPath;
   }
@@ -147,7 +147,7 @@ async function askForModelOutputPath(defaultPath) {
   const answer = await inquirer.prompt({
     type: 'input',
     name: 'outputPath',
-    message: 'Please type your relative output path for model',
+    message: 'Enter the path to the directory where models will be output. The path should be relative to the project root',
     default: defaultPath
   });
   return answer;
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -103,11 +103,11 @@ function loadSchema(apiResourcePath) {
 
 async function getModelOutputPath(context) {
   const projectConfig = context.amplify.getProjectConfig();
-  if(!projectConfig.frontend){
+  if(!projectConfig.frontend) {
     throw new Error('Frontend type is not configured.');
   }
 
-  if(projectConfig.modelgen && projectConfig.modelgen.outputPath){
+  if(projectConfig.modelgen && projectConfig.modelgen.outputPath) {
     context.print.success('\nThe output path for modelgen is already setup. You can edit the path in .config/project-config.json\n');
     return projectConfig.modelgen.outputPath;
   }
@@ -128,7 +128,7 @@ async function getModelOutputPath(context) {
   }
 
   let outputPath = defaultPath;
-  if(!context.parameters.options.yes){
+  if(!context.parameters.options.yes) {
     const answer = await askForModelOutputPath(defaultPath);
     outputPath = answer.outputPath;
   }
@@ -136,7 +136,7 @@ async function getModelOutputPath(context) {
   return outputPath;
 }
 
-async function askForModelOutputPath(defaultPath){
+async function askForModelOutputPath(defaultPath) {
   const answer = await inquirer.prompt({
     type: 'input',
     name: 'outputPath',
--- a/packages/amplify-codegen/src/commands/models.js
+++ b/packages/amplify-codegen/src/commands/models.js
@@ -112,8 +112,9 @@ async function getModelOutputPath(context) {
   let defaultPath = '.';
   switch (projectConfig.frontend) {
     case 'javascript':
-      defaultPath = 'src/models';
-      break;
+      return projectConfig.javascript && projectConfig.javascript.config && projectConfig.javascript.config.SourceDir
+        ? path.normalize(projectConfig.javascript.config.SourceDir)
+        : 'src';
     case 'android':
       defaultPath = projectConfig.android && projectConfig.android.config && projectConfig.android.config.ResDir
         ? path.normalize(path.join(projectConfig.android.config.ResDir, '..', 'java'))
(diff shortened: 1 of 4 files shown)


packages/amplify-codegen/src/commands/models.js
```diff
@@ -99,19 +101,48 @@ function loadSchema(apiResourcePath) {
...
     case 'ios':
-      return 'amplify/generated/models';
-    default:
-      return '.';
+      defaultPath =  'amplify/generated/models';
+      break;
   }
+
+  let outputPath = defaultPath;
+  if(!context.parameters.options.yes){
```
jamesonwilliams, line 131: Keep in mind that the current version of the Gradle plugin _won't_ be provided `--yes`. So, there are four scenarios you need to account for:
1. Old CLI version, old plugin (current behavior, it kinda works.)
2. Old CLI version, new plugin (the user doesn't update CLI, but does have your plugin updates.)
3. New CLI version, old plugin (user updates the CLI, but uses the current plugin version.)
4. New CLI version, new plugin (this one is easy, this is what you _want_ to happen. I assume this will work.)

What happens in cases (2), (3)? If it breaks, how will you convey a message to the user, to tell them to update?

packages/amplify-codegen/src/commands/models.js
```diff
@@ -99,19 +101,48 @@ function loadSchema(apiResourcePath) {
...
 }
 
-function getModelOutputPath(context) {
+async function getModelOutputPath(context) {
   const projectConfig = context.amplify.getProjectConfig();
+  if(!projectConfig.frontend){
+    throw new Error('Frontend type is not configured.');
+  }
+
+  if(projectConfig.modelgen && projectConfig.modelgen.outputPath){
```
jamesonwilliams, line 110: In all places in this PR, formatting should include a space between `(` and `{` chars. e.g.,

```javascript
if (something.booleanValue) {
    // note the whitespace padding around if expression
}
```

packages/amplify-codegen/src/commands/models.js
AaronZyLee: Good points to consider. We have no problems on (1) and (4). I just figure out how to test it locally and try scenario (2) and (3). I get everything right in (2). (adding extra prameters will not impact the old modelgen).  However, I get it stuck in (3) since the command line executed by old plugin does not include `--yes` and will stop on the inquiry and no models are generated. One way to fix this problem is that we add a warning print in the new modelgen code to remind users of getting plugin updated.

AaronZyLee: I just have a good idea about this issue. Just switch the two operations and there is no need to change code in SDK libraries. More specifically, the question will only pop up when `--configpath` flag is passed(I change the flag name since the function of flag is changed). And when no flags exist in cmd, it simply generate models with default path just the same as the old one and no questions will pop up, while the only difference is that the output path is stored in `amplify/.config/project-meta.json`.  Since no code changes to both android and ios plugin and both old and new version of CLI generate the same output on running `amplify codegen models`, all the issues are perfectly resolved

AaronZyLee: I run the API test in android studio with amplify tools plugin and find this method much better than the previous one, which is really beyond my expectation. You can have customized output path this time even by using gradle plugins. Simply changing the `modelgen.outputPath` field in `project-meta.json` and users are all set.

packages/amplify-codegen/src/commands/models.js
```diff
@@ -99,19 +101,55 @@ function loadSchema(apiResourcePath) {
   throw new Error('Could not load the schema');
 }
 
-function getModelOutputPath(context) {
+async function getModelOutputPath(context) {
   const projectConfig = context.amplify.getProjectConfig();
+
+  if(projectConfig.modelgen && projectConfig.modelgen.outputPath && !context.parameters.options.configpath) {
```
jamesonwilliams, line 107:
```suggestion
  if (projectConfig.modelgen && projectConfig.modelgen.outputPath && !context.parameters.options.configPath) {
```

packages/amplify-codegen/src/commands/models.js
```diff
@@ -99,19 +101,55 @@ function loadSchema(apiResourcePath) {
...
+    context.print.success('\nThe output path for modelgen is already setup. You can edit the path in .config/project-config.json\n');
```
jamesonwilliams, line 108:
```suggestion
    context.print.success('\nThe output path for modelgen is already configured. You can edit the path in .config/project-config.json\n');
```

packages/amplify-codegen/src/commands/models.js
```diff
@@ -99,19 +101,55 @@ function loadSchema(apiResourcePath) {
...
     case 'ios':
-      return 'amplify/generated/models';
-    default:
-      return '.';
+      defaultPath =  'amplify/generated/models';
+      break;
+  }
+
+  let outputPath = defaultPath;
+  if(context.parameters.options.configpath) {
```
jamesonwilliams, line 128:
```suggestion
  if (context.parameters.options.configPath) {
```

packages/amplify-codegen/src/commands/models.js
```diff
@@ -99,19 +101,55 @@ function loadSchema(apiResourcePath) {
...
+  context.amplify.updateProjectConfig(projectRoot, 'modelgen', {outputPath: outputPath});
+
+  return outputPath;
+}
+
+async function askForModelOutputPath(defaultPath) {
+  const answer = await inquirer.prompt({
+    type: 'input',
+    name: 'outputPath',
+    message: 'Please type your relative output path for model',
```
jamesonwilliams, line 150: The contents of this message is probably one of the most important things in this PR. Does it make sense to include an example?
```suggestion
    message: 'Enter the path to the directory where generated models will be output. The path should be relative to the project root (e.g. ./src/main/java/com/yourcompany/generated.)',
```

However, my message above is getting long.

packages/amplify-codegen/tests/commands/models.test.js
```diff
@@ -0,0 +1,122 @@
...
+        info: jest.fn(),
+    }
+}
+
+const MOCK_PROJECT_ROOT = 'MOCK_PROJECT_ROOT';
+const MOCK_SCHEMA = 'SCHEMA.GRAPHQL';
+const MOCK_RESOURCE_NAME = 'MOCK_RESOURCE_NAME';
+const MOCK_BACKEND_PATH = 'MOCK_BACKEND_PATH';
+const MOCK_RESOURCES = {
+    allResources:[
```
jamesonwilliams, line 34:
```suggestion
    allResources: [
```

packages/amplify-codegen/src/commands/models.js
AaronZyLee: I use the message without content in parentheses. User can see the default path as an example when the question pops up.

AaronZyLee: I include the examples in the PR description FYI.

AaronZyLee:
Having config regarding codegen in different places is not a good idea. Close the PR temporarily and come up with a better solution.

Issue #4641, opened 2020-06-23
Sanity check in GraphQL transformer doesn't detect multiple @key changes causing CloudFormation deployment failures


kaustavghosh06:
**Describe the bug**
One of the expectation of the sanity check in the GraphQL transformer is to detect changes for multiple @key updates/additions/deletions on a single table and fail pre-deployment to CloudFormation. For the following updates to the annotated schema, the sanity check doesn't catch the multiple additions of @key and continues with the CloudFormation deployment and the CloudFormation deployment after 10 minutes errors out with the following error:
```
Resource Name: Todo-je6oynaxxrfmhig5aogdhzhtqi (AWS::DynamoDB::Table)
Event Type: update
Reason: Cannot perform more than one GSI creation or deletion in a single update
URL: https://console.aws.amazon.com/cloudformation/home?region=us-east-1#/stacks/arn%3Aaws%3Acloudformation%3Aus-east-1%3A827149277658%3Astack%2Famplify-keytest-dev-144503-apikeytest-15HAL01D7NL6R-Todo-1JHK03F669U3U%2F01ba18c0-b4f2-11ea-985c-12da3ecd6186/events

```

My initial schema
```graphql
type Todo @model 
@key(name: "byName", fields: ["name", "description"], queryField: "byNameDescription") {
  id: ID!
  name: String!
  description: String!
}
```
I run `amplify push` thereafter and then change my schema to the following after completion of the deployment
```graphql
type Todo @model 
@key(name: "byName", fields: ["name", "description"], queryField: "byNameDescription")
@key(name: "byTitle", fields: ["title", "description"], queryField: "byTitleDescription")
@key(name: "byPurpose", fields: ["purpose", "description"], queryField: "byPurposeDescription")
@key(name: "byReason", fields: ["reason", "description"], queryField: "byReasonDescription") {
  id: ID!
  name: String!
  description: String!
  title: String!
  purpose: String!
  reason: String!
}
```

**Expected behavior**

Running  `amplify push`  after these changes should stop me from deploying to CloudFormation using the Sanity Check mechanism implemented as a part of the GraphQL transformer but it doesn't.

**Amplify CLI Version**
4.21.3

akshbhu:
Hi @kaustavghosh06

I am able to reproduce this bug.Thanks for reporting this

PR #5141, opened 2020-08-21
fix: Validation check when mutating more than 1 GSI in update flow

akshbhu:
*Issue #4641 

*Description of changes:*

- Added a sanity check to throw error when mutating more than 1 GSI

- added test for the same


packages/amplify-e2e-tests/schemas/migrations_key/cant_add_more_gsi.graphql
```diff
@@ -0,0 +1,13 @@
+type Todo
+  @model
+  @key(fields: ["id", "version"])
+  @key(name: "SomeLSI", fields: ["id", "name"])
```
yuth, line 4: LSI would fail the test as well. Can you remove the LSI from here as the intial schema does not have it

packages/graphql-transformer-core/src/util/sanity-check.ts
```diff
@@ -212,6 +212,50 @@ export function cantAddAndRemoveGSIAtSameTime(diff: Diff, currentBuild: Diffable
...
+        diff.path.length >= 6 &&
+        diff.path[5] === 'GlobalSecondaryIndexes'
+      ) {
+        if (diff.item.kind === 'N') {
+          gsiCountAdded += 1;
+        }
+        if (diff.item.kind === 'D') {
+          gsiCountRemoved += 1;
+        }
+        if (gsiCountAdded > 1 || gsiCountRemoved > 1) {
```
yuth, line 249: I think there is also a  case where one can't add and delete an GSI in the same update. This does not seem to handle that case

akshbhu: I have added the case for update and delete GSI at same time

Issue #4946, opened 2020-07-27
Amplify Push Timeout Error BUG/ISSUE


KayIlory:
This issue is evident with there are about 30 graphql schema. From Amplify CLI version 4.22 going forward, this issue has been consistent and causing issues. It hangs for a while on the cursor in command line and return a Amplify Push Timeout Error. This issue does not exist in 4.21.3. 

[schema.zip](https://github.com/aws-amplify/amplify-cli/files/4983327/schema.zip)

![image](https://user-images.githubusercontent.com/51476926/88570230-a0ecbb80-d033-11ea-9870-5dca50b264c5.png)

KayIlory:
Attached is a list of 25 graphql schemas that might help replicate the issue using CLI version 4.22 and over.

ammarkarachi:
Hi @KayIlory, The issue exists when a timeout is received when attempting to write to S3 we are working on a fix to retry after a timeout has been received

KayIlory:
@ammarkarachi thanks for acknowledging the bug, any update on when we will get a fix for this.

KayIlory:
> @ammarkarachi thanks for acknowledging the bug, any update on when we will get a fix for this.

and in what version the fix will be present so that I can test as well.

ammarkarachi:
@KayIlory This should be fixed now in the latest version can you try it out?

Issue #4113, opened 2020-04-29
amplify push timeout: An error occured during the push operation: Your socket connection to the server was not read from or written to within the timeout period


andreialecu:
**Describe the bug**

For the past few days, as the schema started growing it has been almost impossible to reliably push any changes via `amplify push --yes`

The "Updating resources in the cloud. This may take a few minutes..." message only appears for one in several push attempts. 

Some more logs/status updates relating to what it does at every step between compiling the schema and updating the resources in the cloud would be helpful.

I'm not sure if it's some sort of intermittent outage related to S3 or other resources amplify push might be using. This is **not** an internet problem on my end.

Running `amplify api push` results in a stack trace that mentions `S3` as the reson for the timeout.

I have attached a screenshot below.

Error is: 
`Your socket connection to the server was not read from or written to within the timeout period. Idle connections will be closed.
An error occured during the push operation: Your socket connection to the server was not read from or written to within the timeout period. Idle connections will be closed.`

**Amplify CLI Version**

4.18.0

**Expected behavior**

Pushing should not timeout.

**Screenshots**

![image](https://user-images.githubusercontent.com/697707/80586275-61ef9f80-8a1d-11ea-9c44-3f106ac5cd65.png)

**Desktop (please complete the following information):**

node -v: 13.7.0
Win10 with WSL2 (Ubuntu)

**Additional context**

This started happening yesterday, presumably related to schema changes but I'm not sure. The processing step before it starts "Updating resources in the cloud. This may take a few minutes..." has been taking longer and longer. 

Has been reported several times but the issues are closed:

https://github.com/aws-amplify/amplify-cli/issues/462
https://github.com/aws-amplify/amplify-cli/issues/501
https://github.com/aws-amplify/amplify-cli/issues/2121

andreialecu:
Here's a screenshot of the error during `amplify function push --yes`:

![image](https://user-images.githubusercontent.com/697707/80587859-fce97900-8a1f-11ea-8147-cdff577ec9a1.png)

andreialecu:
Here's another screenshot of having to run the command 3 times:

![image](https://user-images.githubusercontent.com/697707/80588167-7f723880-8a20-11ea-960e-0056c6dfdf9a.png)

ammarkarachi:
@andreialecu This usually happens when Lambda sizes are too large, what's the size of your Lambda package that you are trying to deploy?

andreialecu:
[...]

PR #5158, opened 2020-08-23
fix(amplify-provider-awscloudformation): large file upload

ammarkarachi:
*Issue #, if available:*
#4113 #4946 

*Description of changes:*
The s3 uploadfile switches to `upload` when a file exceeding 5 MB is uploaded. This fixes the issue with error `Your socket connection to the server was not read from or written to within the timeout period`

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.js
@@ -2,6 +2,10 @@ const aws = require('./aws.js');
 const _ = require('lodash');
 const providerName = require('../../lib/constants').ProviderName;
 const configurationManager = require('../../lib/configuration-manager');
+const fs = require('fs-extra');
+const ora = require('ora');
+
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.js
```diff
@@ -2,6 +2,10 @@ const aws = require('./aws.js');
 const _ = require('lodash');
 const providerName = require('../../lib/constants').ProviderName;
 const configurationManager = require('../../lib/configuration-manager');
+const fs = require('fs-extra');
+const ora = require('ora');
+
+const minChunkSize = 5 * 1024 * 1024; // 5 MB https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/S3/ManagedUpload.html#minPartSize-property
```
yuth, line 8: Can we do some profiling of multipart upload for different sizes content and before selecting 5MB as the threshold for multipart upload

ammarkarachi: https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
I chose this because that's minimum size and don't really see that threshold being hit of. I can go only higher from 5 MB  up to 5 GB. We can still profile for higher sizes

PR #5164, opened 2020-08-24
feat(amplify-codegen): add schema compile process in codegen commands

### LINKED ISSUE #4882: [request?] `amplify codegen` does not reflect updated schema from local copy

**Problem 1: codegen doesnt codegen from local copy**

when i update my local schema in `/amplify/backend/api/MYAPI/schema.graphql` and run `amplify codegen`, the generated code in `src/graphql` doesn't update at all.

we are not sure if this is a bug or working as intended. but it's clear that in order to update the generated code i have to `amplify push`.

**Problem 2: CLI gives the impression that codegen is done before push**

I have to `amplify push` before i can `codegen` the new model:

image

this causes a LONG wait time and completely breaks my flow, for information i already have on my machine. 

**I ran `amplify push` ~~40 minutes~~ 1 hour ago and am still waiting for my updated graphql schema so i can continue working. this is not a sustainable workflow when i'm just changing one little field on my schema.**

**Repro steps**

1. `amplify add api`
2. `amplify codegen`
3. change one field in the graphql schema
4. `amplify codegen` again
5. observe that nothing has changed in `src/graphql`

i understand you do this in order to always codegen the right thing. but perhaps offer to codegen based off my local model first, if i accept the risks of mismatch between the model i have on my machine and the model i have in the cloud.

do nothing

AaronZyLee:
*Issue #, if available:*
fix #4882 
*Description of changes:*
Added graphql schema compiling in command `amplify codegen`, `amplify codegen statements` and `amplify codegen types`. 

### DIFF
--- a/packages/amplify-codegen/src/commands/generateStatementsAndType.js
+++ b/packages/amplify-codegen/src/commands/generateStatementsAndType.js
@@ -47,9 +47,11 @@ async function generateStatementsAndTypes(context, forceDownloadSchema, maxDepth
     ({ projectPath } = context.amplify.getEnvInfo());
   }
 
-  await context.amplify.executeProviderUtils(context, 'awscloudformation', 'compileSchema', {
-    forceCompile: true,
-  });
+  if (!project.schema.endsWith('.json') && fs.existsSync(path.join(projectPath, project.schema))){
+    await context.amplify.executeProviderUtils(context, 'awscloudformation', 'compileSchema', {
+      forceCompile: true,
+    });
+  }
 
   let downloadPromises;
   if (!withoutInit) {
--- a/packages/amplify-codegen/src/commands/generateStatementsAndType.js
+++ b/packages/amplify-codegen/src/commands/generateStatementsAndType.js
@@ -7,6 +7,7 @@ const constants = require('../constants');
 const { ensureIntrospectionSchema, getAppSyncAPIDetails } = require('../utils');
 const path = require('path');
 const fs = require('fs-extra');
+const ensureSchemaCompiled = require('../utils/ensureSchemaCompiled');
 
 async function generateStatementsAndTypes(context, forceDownloadSchema, maxDepth) {
   let withoutInit = false;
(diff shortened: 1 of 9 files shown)


packages/amplify-codegen/src/commands/generateStatementsAndType.js
```diff
@@ -47,6 +47,10 @@ async function generateStatementsAndTypes(context, forceDownloadSchema, maxDepth
     ({ projectPath } = context.amplify.getEnvInfo());
   }
 
+  await context.amplify.executeProviderUtils(context, 'awscloudformation', 'compileSchema', {
+    forceCompile: true,
+  });
+
```
yuth, line 53: We might need to add a check here to ensure that there is an valid GraphQL API in the project before executing this. The CLI allows customers to add external api using `amplify add codegen --apiId <your-api-id>`. If the API is an external API, then we download the introspection schema and generate types and statements.

packages/amplify-codegen/src/commands/generateStatementsAndType.js
```diff
@@ -47,9 +47,11 @@ async function generateStatementsAndTypes(context, forceDownloadSchema, maxDepth
     ({ projectPath } = context.amplify.getEnvInfo());
   }
 
-  await context.amplify.executeProviderUtils(context, 'awscloudformation', 'compileSchema', {
-    forceCompile: true,
-  });
+  if (!project.schema.endsWith('.json') && fs.existsSync(path.join(projectPath, project.schema))){
```
yuth, line 50: I would prefer the check rely on existence of a API resource along with looking at the file extension. There is a possibility that the API is removed, but the `.graphql.yml` file is not updated. You could follow something similar to what we do in mock. Also move the check to an helper method instead of having the same check in 3 different places

https://github.com/aws-amplify/amplify-cli/blob/eee972f44363a6fb518bcda45ffbed940144853d/packages/amplify-util-mock/src/api/api.ts#L237-L253

AaronZyLee: Fair enough. I will add this part to the check

yuth (approved):

PR #5178, opened 2020-08-26
feat: resource Tagging Support

ammarkarachi:
Enable to customers to add tags using `/amplify/backend/tags.json` file

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-cli-core/src/tags/Tags.ts
+++ b/packages/amplify-cli-core/src/tags/Tags.ts
@@ -20,11 +20,11 @@ export function ReadValidateTags(tagsFilePath: string): Tag[] {
 }
 
 export function validate(tags: Tag[]): void {
-  const set = new Set(['Key', 'Value']);
+  const allowedKeySet = new Set(['Key', 'Value']);
 
   //check if Tags have the right format
   _.each(tags, tags => {
-    if (_.some(Object.keys(tags), r => !set.has(r))) throw new Error('Tag thould be of type Key: string, Value: string');
+    if (_.some(Object.keys(tags), r => !allowedKeySet.has(r))) throw new Error('Tag thould be of type Key: string, Value: string');
   });
 
   //check if Tag Key is repeated
--- a/packages/amplify-cli-core/src/tags/Tags.ts
+++ b/packages/amplify-cli-core/src/tags/Tags.ts
@@ -1,0 +1,35 @@
+import { JSONUtilities } from '../jsonUtilities';
+import _ from 'lodash';
+
+export interface Tag {
+  Key: string;
+  Value: String;
+}
+
+export function ReadValidateTags(tagsFilePath: string): Tag[] {
+  const tags = JSONUtilities.readJson<Tag[]>(tagsFilePath, {
+    throwIfNotExist: false,
+    preserveComments: false,
+  });
+
+  if (!tags) return [];
+
+  validate(tags);
+
+  return tags;
+}
+
+export function validate(tags: Tag[]): void {
+  const set = new Set(['Key', 'Value']);
+
+  //check if Tags have the right format
+  _.each(tags, tags => {
+    if (_.some(Object.keys(tags), r => !set.has(r))) throw new Error('Tag thould be of type Key: string, Value: string');
+  });
+
+  //check if Tag Key is repeated
+  if (_.uniq(tags.map(r => r.Key)).length !== tags.length) throw new Error("'Key' should be unique");
+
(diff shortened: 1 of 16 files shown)


packages/amplify-cli-core/src/tags/Tags.ts
```diff
@@ -0,0 +1,35 @@
...
+
+  if (!tags) return [];
+
+  validate(tags);
+
+  return tags;
+}
+
+export function validate(tags: Tag[]): void {
+  const set = new Set(['Key', 'Value']);
```
jhockett, line 23: nit: use a more descriptive variable name

packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
```diff
@@ -321,7 +337,11 @@ export async function showResourceTable(category, resourceName, filteredResource
...
 
   table(tableOptions, { format: 'markdown' });
+  if (tagsUpdated) {
+    print.info('Resource Tags Update Detected');
+  }
+  const changedResourceCount =
+    resourcesToBeCreated.length + resourcesToBeUpdated.length + resourcesToBeDeleted.length + tagsUpdated ? 1 : 0;
 
-  const changedResourceCount = resourcesToBeCreated.length + resourcesToBeUpdated.length + resourcesToBeDeleted.length;
   return changedResourceCount;
```
jhockett, line 346: Why are we only returning 1 or 0?

If we are only returning 1 or 0, should the variable name still be `changedResourceCount`?

packages/amplify-cli/src/init-steps/s9-onSuccess.ts
```diff
@@ -74,7 +75,24 @@ export function generateLocalEnvInfoFile(context: $TSContext) {
   stateManager.setLocalEnvInfo(projectPath, context.exeInfo.localEnvInfo);
 }
 
-function generateAmplifyMetaFile(context: $TSContext) {
+function generateLocalTagsFile(context) {
```
jhockett, line 78:
```suggestion
function generateLocalTagsFile(context: $TSContext) {
```

packages/amplify-e2e-core/src/utils/projectMeta.ts
```diff
@@ -26,6 +26,11 @@ function getProjectMeta(projectRoot: string) {
   return JSON.parse(fs.readFileSync(metaFilePath, 'utf8'));
 }
 
+function getProjectTags(projectRoot: string) {
+  const metaFilePath = path.join(projectRoot, 'amplify', '#current-cloud-backend', 'tags.json');
+  return JSON.parse(fs.readFileSync(metaFilePath, 'utf8'));
```
jhockett, line 31:
```suggestion
  return JSONUtilities.readJson(metaFilePath);
```

packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
ammarkarachi: Tags is a single aspect either it's modified or not, Tags also doesn't fit in to the whole category thing

packages/amplify-e2e-core/src/utils/projectMeta.ts
ammarkarachi: It's the `amplify-e2e-core` package

packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
jhockett: I was concerned because the method is called elsewhere, but it doesn't look like anything uses the returned count besides evaluating it to be truthy or falsey.

I suggest returning a boolean in this case:

```ts
return resourcesToBeCreated.length + resourcesToBeUpdated.length + resourcesToBeDeleted.length + tagsUpdated > 0;
```

ammarkarachi: https://repl.it/repls/LimitedWanRecords#index.js
I don't think we should mix types ((number + number + boolean))

jhockett: Oh sorry, I thought tagsUpdated was the number of tags updated.

In that case, maybe this would work:
```ts
return resourcesToBeCreated.length + resourcesToBeUpdated.length + resourcesToBeDeleted.length > 0 || tagsUpdated;
```

PR #5196, opened 2020-08-28
fix(amplify-provider-awscloudformation): include region configure url

### LINKED ISSUE #4735: [Possible Bug] amplify configure url string has `undefined` for region
**Describe the bug**

run `amplify configure` and look closely at the URL

image

**Amplify CLI Version**
4.21.3 

**Expected behavior**

probably it is supposed to say the region. ALTERNATIVELY if the region doesnt matter we should probably not ask for it.

yuth:
Updated the user addition URL to include selected region instead of undefined. Updated code to use async/await
instead of promises

fix #4735

*Issue #, if available:*

*Description of changes:*


yuth:
> is it easy to write a test for this?

Not in the current state, we will have to refactor to make it easier to test

PR #5222, opened 2020-09-01
Configureissue

UnleashedMind:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-provider-awscloudformation/lib/system-config-manager.js
+++ b/packages/amplify-provider-awscloudformation/lib/system-config-manager.js
@@ -19,18 +19,12 @@ function setProfile(awsConfig, profileName) {
   if (fs.existsSync(credentialsFilePath)) {
     makeFileOwnerReadWrite(credentialsFilePath);
     credentials = ini.parse(fs.readFileSync(credentialsFilePath, 'utf-8'));
-  } else {
-    const fd = fs.openSync(credentialsFilePath, 'w', 0o600);
-    fs.closeSync(fd);
   }
 
   if (fs.existsSync(configFilePath)) {
     makeFileOwnerReadWrite(configFilePath);
     config = ini.parse(fs.readFileSync(configFilePath, 'utf-8'));
-  } else {
-    const fd = fs.openSync(configFilePath, 'w', 0o600);
-    fs.closeSync(fd);
-  }
+  } 
 
   let isCredSet = false;
   Object.keys(credentials).forEach(key => {
@@ -62,9 +56,9 @@ function setProfile(awsConfig, profileName) {
       region: awsConfig.region,
     };
   }
-
-  fs.writeFileSync(credentialsFilePath, ini.stringify(credentials));
-  fs.writeFileSync(configFilePath, ini.stringify(config));
+  
+  fs.writeFileSync(credentialsFilePath, ini.stringify(credentials), {mode: 0o600});
+  fs.writeFileSync(configFilePath, ini.stringify(config), {mode: 0o600});
 }
 
 async function getProfiledAwsConfig(context, profileName, isRoleSourceProfile) {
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -17,10 +17,17 @@ function setProfile(awsConfig, profileName) {
   let credentials = {};
   let config = {};
   if (fs.existsSync(credentialsFilePath)) {
```
flywheelms, line 19: This logic seems inverted.

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -55,7 +62,6 @@ function setProfile(awsConfig, profileName) {
   }
 
   fs.writeFileSync(credentialsFilePath, ini.stringify(credentials));
```
yuth, line 64: Would it be possible to change the file permission when you're writing it rather that doing it on top

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
```diff
@@ -17,10 +17,17 @@ function setProfile(awsConfig, profileName) {
   let credentials = {};
   let config = {};
   if (fs.existsSync(credentialsFilePath)) {
+    makeFileOwnerReadWrite(credentialsFilePath);
     credentials = ini.parse(fs.readFileSync(credentialsFilePath, 'utf-8'));
+  } else {
+    fs.openSync(credentialsFilePath, 'w', 0o600);
```
yuth, line 23: Shouldn't you close the file?

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
UnleashedMind: theoretically yes, but fs has a bug and I checked it's not working properly

yuth: In what case it does not work?

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
UnleashedMind: no, because we need to write to the file in the method `setProfile`

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
UnleashedMind: like in our case, it does not set the mode correctly

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
UnleashedMind: yes, nice catch

yuth: I don't see the opened file descriptor used in `writeSync`, so this file descriptor needs to be closed

yuth:
```suggestion
     credentialFd =fs.openSync(credentialsFilePath, 'w', 0o600);
```

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
yuth, line 66:
```suggestion
  fs.writeFileSync(credentialFd, ini.stringify(credentials));
  fs.closeSync(credentialFd)
```

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
UnleashedMind: nice catch. 
fixed

packages/amplify-provider-awscloudformation/lib/system-config-manager.js
UnleashedMind: fs.writeFileSync first opens the file for write, write to the file, and then closes the file, so you don't need to call closeSync after fs.writeFileSync.

yuth: We should see if the user has permission to write these files and show proper error message as this could be one of the edge cases where user does not have permission to write the file

UnleashedMind: This works: 

fs.writeFileSync(credentialsFilePath, ini.stringify(credentials), {mode: 0o600});

Updated the PR to use the above code

PR #5246, opened 2020-09-03
fix: update default file permissions for credentials/config

jhockett:

### DIFF
--- a/packages/amplify-cli-core/src/state-manager/pathManager.ts
+++ b/packages/amplify-cli-core/src/state-manager/pathManager.ts
@@ -3,6 +3,11 @@ import * as fs from 'fs-extra';
 import { homedir } from 'os';
 
 const PathConstants = {
+  // in home directory
+  DotAwsDir: '.aws',
+  Credentials: 'credentials',
+  Config: 'config',
+
   // in project root
   AmplifyDirName: 'amplify',
   DotAmplifyDirName: '.amplify',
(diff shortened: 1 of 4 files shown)


packages/amplify-cli-core/src/state-manager/pathManager.ts
```diff
@@ -89,6 +94,12 @@ export class PathManager {
       PathConstants.BackendConfigFileName,
     ]);
 
+  getDotAWSDirPath = (): string => path.normalize(path.join(homedir(), PathConstants.DotAwsDir));
+
+  getCredentialsFilePath = (): string => path.normalize(path.join(this.getDotAWSDirPath(), PathConstants.Credentials));
```
attilah, line 99: can we call these `getAWS*` methods as `getConfigFilePath` is way to generic even for the CLI.

packages/amplify-cli/src/index.ts
```diff
@@ -111,6 +115,13 @@ export async function run() {
   }
 }
 
+function verifyFilePermissions(filePath) {
```
attilah, line 118: `ensureFilePermissions` probably a better name since we also changing it, not just verifying.

attilah (changes requested):
Added comments about naming the methods, for `pathManager` I'd like to see it, the other is for easier readibility.

attilah (approved):

UnleashedMind (approved):

Issue #5212, opened 2020-08-31
Autogenerated parameters does not follow allowed format.


emiltorlen:
**Describe the bug**

When using underscore in the field name this generates resource names in the stack which is not supported when pushing the code. The error message then becomes 
`Resource name <model><field>Resolver is non alphanumeric.`

**Amplify CLI Version**
`amplify -v
4.27.3
`

**To Reproduce**

Consider the following graphql schema

```graphql
type Contact
@model
@auth(
  rules: [{ allow: owner, ownerField: "owner", operations: [read, update] }]
) {
  id: ID!
  firstname: String
  lastname: String
  email: AWSEmail #maybe change name? bug with creating empty
  phone: AWSPhone
  country_dial_code: String
  address_street: String
  address_postal_code: String
  @auth(
    rules: [
      { allow: owner, operations: [read] }
    ]
  )
  user_id: String
  @auth(
    rules: [
      { allow: owner, operations: [read] }
    ]
  )
  createdAt: AWSDateTime
  @auth(
    rules: [
      { allow: owner, operations: [read] }
    ]
  )
  updatedAt: AWSDateTime
  @auth(
    rules: [
      { allow: owner, operations: [read] }
    ]
  )
}
```

When running `amplify push` this generates resource names under  

`amplify/backend/api/<project>/build/stacks/Contact.json`

```json
{
    "AWSTemplateFormatVersion": "2010-09-09",
    "Description": "An auto-generated nested stack.",
    "Metadata": {},
    "Parameters": {
            ...
      },
    "Resources": {
        ...,
        "Contactuser_idResolver": {
            "Type": "AWS::AppSync::Resolver",
            "Properties": {
                "ApiId": {
                    "Ref": "GetAttGraphQLAPIApiId"
                },
                "DataSourceName": "NONE",
                "FieldName": "user_id",
                "TypeName": "Contact",
                "RequestMappingTemplateS3Location": {
                    "Fn::Sub": [
                        "s3://${S3DeploymentBucket}/${S3DeploymentRootKey}/resolvers/${ResolverFileName}",
                        {
                            "S3DeploymentBucket": {
                                "Ref": "S3DeploymentBucket"
                            },
                            "S3DeploymentRootKey": {
                                "Ref": "S3DeploymentRootKey"
                            },
                            "ResolverFileName": {
                                "Fn::Join": [
                                    ".",
                                    [
                                        "Contact",
                                        "user_id",
                                        "req",
                                        "vtl"
[...]

PR #5263, opened 2020-09-05
fix(graphql-transformer-common): support underscore names

SwaySway:
support underscore names by making logical ids unique to support graphql spec
updated e2e test with underscore named field

re #5212

*Issue #, if available:*

*Description of changes:*


attilah (changes requested):
Please make the test related changes what we discussed offline, that is the only thing left I think.

attilah (approved):

PR #5297, opened 2020-09-10
fix(amplify-go-function-runtime-provider): support json response from go functions

Jhorlin:
*Issue #5293

Added JSON.parse in a try catch block for go invoked functions.


edwardfoyle (approved):
LGTM! Thanks for the contribution @Jhorlin

PR #5301, opened 2020-09-10
chore(amplify-category-function): add missing cloudform-types dependency

jhockett:


packages/amplify-category-function/package.json
```diff
@@ -24,6 +24,7 @@
     "amplify-category-storage": "2.7.6",
     "amplify-cli-core": "1.3.0",
     "amplify-function-plugin-interface": "1.4.1",
+    "cloudform-types": "6.0.0",
```
attilah, line 27: Please use the same version what we are already using to make sure we're not breaking anything.

attilah (changes requested):

attilah (approved):

PR #5315, opened 2020-09-11
feat: added commented out cors headers

ammarkarachi:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
@@ -141,6 +141,9 @@ function copyTemplateFiles(context: any, parameters: FunctionParameters | Functi
     });
     templateParams = _.assign(templateParams, triggerEnvs);
   }
+  templateParams = _.assign(templateParams, {
+    uncommentCorsHeader: process.env.AMPLIFY_CLI_LAMBDA_CORS_HEADER === 'true',
+  });
 
   context.amplify.copyBatch(context, copyJobs, templateParams, false);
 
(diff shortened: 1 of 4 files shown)


edwardfoyle (approved):

nikhname (approved):

packages/amplify-e2e-core/src/categories/lambda-function.ts
```diff
@@ -206,6 +208,19 @@ export const addLambdaTrigger = (chain: ExecutionContext, cwd: string, settings:
...
+  const uncommentedCode = code
+    .replace('//  headers: {', '    headers: {')
+    .replace('//      "Access-Control-Allow-Origin": "*"', '        "Access-Control-Allow-Origin": "*"')
+    .replace('//  },', '    },');
+  fs.writeFileSync(codeHandlerPath, uncommentedCode, {
+    flag: 'w',
+    encoding: 'utf8',
+  });
+};
+
```
yuth, line 223: Why don't we add an helper function that determines if we want to allow CORs or not based on a global/env variable. And in the code turn set that global variable instead of using string replace which can be error prone

ammarkarachi: Done

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -141,6 +141,9 @@ function copyTemplateFiles(context: any, parameters: FunctionParameters | Functi
     });
     templateParams = _.assign(templateParams, triggerEnvs);
   }
+  templateParams = _.assign(templateParams, {
+    uncommentCorsHeader: process.env.AMPLIFY_CLI_LAMBDA_CORS_HEADER === 'true',
```
yuth, line 145: Can we rename this to `enableCors` instead of `unCommentCorsHeader`

yuth (approved):

PR #5321, opened 2020-09-14
feat(amplify-cli-core): visually hide "immutable files" for VSCode users

r0zar:
I'm proposing a developer experience improvement to Amplify, specifically targeting new users.

Consider the following scenario:

A new user decide is trying out Amplify for the first time.

Forgetting to use the CLI, they attempt to modify the metadata files directly.

They didn't read the docs, and certainly the not the small section at the end explain the Reference > Files and Folders section.

So the only outcome is they get stuck. This causes frustration and a bad new user experience.

Good thing it's totally avoidable. This change visually hides those metadata files so new users aren't tempted to fiddle with them.

With basic workspace configuration settings, we can provide a better user experience for new users by directing their energies and focus towards things in their control, like the templates and source code. 

Of course, these defaults are easily opted out of by updating the settings.json file.

*Issue #, if available:*

*Description of changes:*
If a user selects VSCode during `amplify init`, create or safely update their .vscode/settings.json file to exclude immutable files.
Based on [Files and Folders where manual edits are not allowed](https://docs.amplify.aws/cli/reference/files)

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/editor-selection.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/editor-selection.ts
@@ -81,13 +81,13 @@ function hideNoManualEdit(editor) {
       const workspaceSettingsPath = '.vscode/settings.json';
       const exclusionRules = {
         'files.exclude': {
-          'amplify/.config': true,
-          'amplify/**/*-parameters.json': true,
-          'amplify/**/amplify.state': true,
-          'amplify/**/transform.conf.json': true,
-          'amplify/#current-cloud-backend': true,
-          'amplify/backend/amplify-meta.json': true,
-          'amplify/backend/awscloudformation': true,
+          '**/amplify/.config': true,
+          '**/amplify/**/*-parameters.json': true,
+          '**/amplify/**/amplify.state': true,
+          '**/amplify/**/transform.conf.json': true,
+          '**/amplify/#current-cloud-backend': true,
+          '**/amplify/backend/amplify-meta.json': true,
+          '**/amplify/backend/awscloudformation': true,
         },
       };
       try {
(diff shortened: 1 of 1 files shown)


r0zar:
Feedback and open dialog encouraged! I discussed this feature with a few other devs, but would love to get a wider range of opinions on it. Thanks and looking forward to hearing your thoughts.

renebrandel:
Pasting in my message from Discord: 
I'm actually curious what the other devs in this channel think about it?
I see arguments for both sides: 1. "Hide all the files that I'm not supposed to modify anyway!" or 2. "I want to see everything that's going on in my project. This isn't just a big dependency dump folder like node_modules. I don't want to accidentally modify a folder thinking that there's nothing inside"

cc @sw-yx @dabit3 @aspittel any opinions on this from your experience?

r0zar:
To provide an example:

With excludes:
Screenshot from 2020-10-15 02-37-12

Without excludes:
Screenshot from 2020-10-15 02-37-44

aspittel:
I'm actually pretty for it. I've taught create-react-app a ton of times and it *always* terrifies new developers to see all the files (even though I tell them right away they don't need to touch most of them). I'm pro as few visible files as possible, it feels simpler.

dabit3:
For me I'd say that if I'm likely to never touch it ever, then I don't want to see it (i.e. maybe #currentcloudbackend & any others that are similar)

renebrandel (approved):

r0zar:
:tada:

r0zar:
Just retested this in in a few configurations locally and everything looks good.

As a note for monorepo projects with multiple amplify apps in different packages. This will create and apply .vscode settings files for each package, but they will only take effect if you open vscode with that package as the workspace route. I think that's the best user experience anyhow.

Issue #5139, opened 2020-03-04
Models not saving due to incorrect data model, again.


jesper-bylund:
**Describe the bug**
This morning all my saves had stopped working. Again.

**To Reproduce**
1. Add any model, with an undefined "createdAt".
2. DataStore sync comes back with an error. This error also seems to be impossible to solve.

**Expected behavior**
Sync should work. Or the error should allow me to debug.
It is also surprising that my app changes behavior even though _I'm not making any changes_. How is that happening? No packages were updated?

**Screenshots**
Original error:
<img width="670" alt="Screenshot 2020-03-04 at 17 00 40" src="https://user-images.githubusercontent.com/186083/75898528-760e8a80-5e3a-11ea-8580-560fe5b16e02.png">

Empty string added:
<img width="672" alt="Screenshot 2020-03-04 at 17 00 09" src="https://user-images.githubusercontent.com/186083/75898554-81fa4c80-5e3a-11ea-90a1-7347c2ea6e03.png">

Date string added:
<img width="674" alt="Screenshot 2020-03-04 at 17 00 16" src="https://user-images.githubusercontent.com/186083/75898572-8aeb1e00-5e3a-11ea-852a-542522a37b63.png">

Date ISO string added:
<img width="658" alt="Screenshot 2020-03-04 at 17 03 39" src="https://user-images.githubusercontent.com/186083/75898607-94748600-5e3a-11ea-96a9-c40e980c38ae.png">

Seems like AppSync has changed how it handles models, but updating my packages does nothing?

PR #5327, opened 2020-09-14
fix(amplify-codegen-appsync-model-plugin): DataStore Array Support

seang96:
Issue #, if available: #5139

*Description of changes:*
I added a check to getTypeInfo to check if the current node is NonNullType and its' child node is ListType to set isListNullable, new attribute in TypeInfo, to to true.
The metadata-visitor then sets isArrayRequired to true whe isListNullable is not allowed.
To add support in DataStore, this PR should do: https://github.com/aws-amplify/amplify-js/pull/6784

This PR should support the table below:

declaration | null | [] | [null] | [{ foo: 'BAR' }]
-- | -- | -- | -- | --
[Type!]! | no | yes | no | yes
[Type]! | no | yes | yes | yes
[Type!] | yes | yes | no | yes
[Type] | yes | yes | yes | yes

### DIFF
--- a/packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-json-metadata-visitor.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-json-metadata-visitor.ts
@@ -205,7 +205,7 @@ export class AppSyncJSONVisitor<
         };
 
         if (field.isListNullable !== undefined) {
-          fieldMeta.isArrayNullable = !field.isListNullable;
+          fieldMeta.isArrayNullable = field.isListNullable;
         }
 
         const association: AssociationType | void = this.getFieldAssociation(field);
--- a/packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-json-metadata-visitor.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-json-metadata-visitor.ts
@@ -201,9 +201,13 @@ export class AppSyncJSONVisitor<
           isArray: field.isList,
           type: this.getType(field.type),
           isRequired: !field.isNullable,
-          isArrayRequired: field.isListNullable !== undefined ? !field.isListNullable : undefined,
           attributes: [],
         };
+
+        if (field.isListNullable !== undefined) {
+          fieldMeta.isArrayRequired = !field.isListNullable;
+        }
+
         const association: AssociationType | void = this.getFieldAssociation(field);
         if (association) {
           fieldMeta.association = association;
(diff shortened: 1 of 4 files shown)


packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-json-metadata-visitor.ts
```diff
@@ -202,6 +203,11 @@ export class AppSyncJSONVisitor<
           isRequired: !field.isNullable,
           attributes: [],
         };
+
+        if (field.isListNullable !== undefined) {
+          fieldMeta.isArrayNullable = !field.isListNullable;
```
yuth, line 208: With the rename, this should be changed to 
```suggestion
          fieldMeta.isArrayNullable = field.isListNullable;
```

seang96: @yuth, Ah I did not think about this. I added a commit to fix this.

yuth (approved):

seang96:
@yuth, I discovered an issue due to the rename to isArrayNullable having the opposite meaning means in getTypeInfo it should only return if isListNullable is true instead of if it is false. I am submitting a PR that fixes this.

danrivett:
@yuth just an FYI that this looks like the cause that broke our front-end when we upgraded from `@aws-amplify/cli` package version from `4.29.3` to `4.29.4` since types that are defined as:

```
foo: [Bar!]
```

(e.g. the array is optional) stopped being generated as:

```typescript
readonly foo?: Bar[];
```

but instead got generated as:

```typescript
readonly foo: Bar[];
```

I haven't yet read this ticket in detail to know if that was an intentional change or not, but if it was, I would suggest it wasn't a patch-level change since it caused our RN app using those types to start generating TypeScript errors.

I think overall the change is likely a good one as we were having to work around DataStore not being able to handle empty arrays and so instead we were deliberately setting the values to `null` as that was working for us. So the change looks like a good change, it was more the surprise of it being released as a patch release and not seeing anything in the [changelog](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-cli/CHANGELOG.md) under `4.29.4` mentioning this change.

seang96:
@danrivett I looked into this, #5450 is in the latest version. The schema is generating correctly with isArrayNullable true and isRequired as true, but the types are not correct. I'll look into this and submit a PR for @aws-amplify/datastore. Sorry for the bug!

danrivett:
> @danrivett I looked into this, #5450 is in the latest version. The schema is generating correctly with isArrayNullable true and isRequired as true, but the types are not correct. I'll look into this and submit a PR for types. Sorry for the bug!

Thanks for the update @seang96, no problem.

Thank you for looking into this. For now we've reverted to `4.29.3` so unblocked.

PR #5335, opened 2020-09-15
fix: add owner field if owner auth is included in auth rules

zeynepsu:
*Issue #, if available:*
https://github.com/aws-amplify/amplify-js/issues/5687

*Description of changes:*
Add owner field to generated schema if it does not exist in the original schema but owner auth is included in auth rules.

### DIFF
--- a/packages/amplify-codegen-appsync-model-plugin/src/utils/process-auth.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/utils/process-auth.ts
@@ -1,4 +1,4 @@
-import { CodeGenDirectives, CodeGenDirective } from '../visitors/appsync-visitor';
+import { CodeGenDirectives, CodeGenDirective, CodeGenModel, CodeGenField } from '../visitors/appsync-visitor';
 export enum AuthProvider {
   apiKey = 'apiKey',
   iam = 'iam',
@@ -92,4 +92,31 @@ export function processAuthDirective(directives: CodeGenDirectives): AuthDirecti
       },
     };
   });
+}
+
+export function getOwnerAuthRules(modelObj: CodeGenModel): any {
+  let rules = Array<any>();
+  for (let directive of modelObj.directives) {
+    if ("auth" === directive.name) {
+      for (let rule of directive.arguments.rules) {
+        if (rule.allow === AuthStrategy.owner) {
+          rules.push(rule);
+        }
+      }
+    }
+  }
+  return rules;
+}
+
+export function getOwnerFieldName(rule: any): string | undefined {
+  if (rule.ownerField === undefined) {
+    return "owner";
+  } else {
+    return rule.ownerField
+  }
+}
+
+export function hasOwnerField(fields: CodeGenField[], ownerField: string): boolean {
+  let ownerFields = fields.filter(field => field.name === ownerField);
+  return ownerFields && ownerFields.length !== 0
 }
--- a/packages/amplify-codegen-appsync-model-plugin/src/utils/process-auth.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/utils/process-auth.ts
@@ -94,8 +94,8 @@ export function processAuthDirective(directives: CodeGenDirectives): AuthDirecti
   });
 }
 
-export function getOwnerAuthRules(modelObj: CodeGenModel): any {
-  let rules = Array<any>();
+export function getOwnerAuthRules(modelObj: CodeGenModel): AuthRule[] {
+  let rules = Array<AuthRule>();
   for (let directive of modelObj.directives) {
     if ("auth" === directive.name) {
       for (let rule of directive.arguments.rules) {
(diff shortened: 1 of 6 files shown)


packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-typescript-visitor.ts
```diff
@@ -83,12 +96,22 @@ export class AppSyncModelTypeScriptVisitor<
       .withName(modelName)
       .export(true);
 
+    let ownerFieldExists = false;
     modelObj.fields.forEach((field: CodeGenField) => {
+      if ("owner" === this.getFieldName(field) && "String" === this.getNativeType(field)) {
```
yuth, line 101: The field name that stores owner value can be changed with `ownerField` in auth directive. Model gen should support overriding this
```gql
type Post
  @model
  @auth(
    rules: [
      { allow: owner, ownerField: "ownedBy", operations: [create, update, delete, read] },
    ])
{
  id: ID!
  title: String!
  # ownedBy: String would be automatically added
}
```
You can find the definition of `auth` directive here
https://docs.amplify.aws/cli/graphql-transformer/directives#definition-2

yuth (changes requested):

yuth (commented):
Realized that this would not work without corresponding change in the graphql transformer to expose the owner fields. Without that, the owner information will be available only in the DynamoDB and only accessible to resolvers, but will not be exposed in type.

packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-typescript-visitor.ts
```diff
@@ -70,6 +70,30 @@ export class AppSyncModelTypeScriptVisitor<
...
+    return ownerField
+  }
+
+  protected getOwnerAuthRule(modelObj: CodeGenModel): any {
+    for (let directive of modelObj.directives) {
+      if ("auth" === directive.name) {
+        for (let rule of directive.arguments.rules) {
+          if ("owner" === rule.allow) {
+            return rule;
+          }
```
yuth, line 91: `@auth` directive supports multiple owner rules, for instance this is an valid auth rule
```gql
type Todo
@model
@auth(
  rules: [
    { allow: owner, ownerField: "ownedBy" }
    { allow: owner, ownerField: "owner" }
  ]
) {
  id: ID!
  name: String!
  description: String
}
```

iartemiev (changes requested):
This also requires changes to `amplify-codegen/src/commands/models.js` in order to add the owner field into schema.js that gets generated when calling `amplify codegen models`

packages/amplify-codegen-appsync-model-plugin/src/utils/process-auth.ts
```diff
@@ -93,3 +93,30 @@ export function processAuthDirective(directives: CodeGenDirectives): AuthDirecti
...
+        if (rule.allow === AuthStrategy.owner) {
+          rules.push(rule);
+        }
+      }
+    }
+  }
+  return rules;
+}
+
+export function getOwnerFieldName(rule: any): string | undefined {
```
yuth, line 111: Can you define type for `rule` instead of `any`

packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-typescript-visitor.ts
```diff
@@ -89,6 +90,17 @@ export class AppSyncModelTypeScriptVisitor<
...
+    if (ownerAuthRules !== undefined) {
+      for (let rule of ownerAuthRules) {
+        let ownerFieldName = getOwnerFieldName(rule);
+        if (ownerFieldName !== undefined && !hasOwnerField(modelObj.fields, ownerFieldName)) {
+        modelDeclarations.addProperty(ownerFieldName, "String", undefined, 'DEFAULT', {
+          readonly: true,
+          optional: true,
+        });
+      }
+    }
```
yuth, line 103:
```suggestion
    let ownerAuthRules = getOwnerAuthRules(modelObj);
    if (ownerAuthRules !== undefined) {
      for (let rule of ownerAuthRules) {
        let ownerFieldName = getOwnerFieldName(rule);
        if (ownerFieldName !== undefined && !hasOwnerField(modelObj.fields, ownerFieldName)) {
          modelDeclarations.addProperty(ownerFieldName, "String", undefined, 'DEFAULT', {
            readonly: true,
            optional: true,
          });
         }
      }
    }
```

yuth (changes requested):

iartemiev:
Working as expected with a simple owner auth schema:
```graphql
type Post @model @auth(rules: [{ allow: owner }]) {
  id: ID!
  title: String!
}
```
However, subscriptions are failing with multiple owner auth rules, e.g.
```graphql
type Post
@model
@auth(
  rules: [
    { allow: owner, ownerField: "owner" }
    { allow: owner, ownerField: "editors" }
  ]
) {
  id: ID!
  title: String!
  owner: String
  editors: [String]
}

```
They're failing with the current CLI release as well though (not just this PR). @yuth when you have a moment, let's discuss what we're expecting the generated subscriptions to look like for the 2nd schema. Maybe the bug for this is on the library side.

iartemiev (approved):
From the library side: I'm good with us merging this. 
The issue with multiple ownerFields I mentioned in my previous comment is outside the scope of this PR, so we can address it later

yuth (approved):

PR #5342, opened 2020-09-16
fix: fix appsync permission assignment from functions

### LINKED ISSUE #4306: Grant lambdas access to other project resources for GraphQL API is either broken or very misleading
**Describe the bug**
The workflow you go through to "update permissions granted to the Lambda function to perform on other resources in your project" does not work - at least not in the way that this piece of documentation implies that it does: https://docs.amplify.aws/cli/function. If you don't choose the `create` operation, you won't actually be able to access an AppSync api from the lambda.

**Amplify CLI Version**
4.20.0

**To Reproduce**
image

Go through the `amplify update function` flow as shown in the above image, being sure to choose an operation other than `create` - `read` for example.

Also, follow all the other steps outlined in the amplify docs here: https://docs.amplify.aws/cli/function#signing-a-request-from-lambda - except instead of a create mutation, feel free to switch to one that is implied to be available based on the operation you chose (a list query for example).

Then, run the lambda and note that you get back an unauthorized error.

**Expected behavior**
Lambda's response contains response of the AppSync query.

 - OS: Ubuntu
 - Node Version. 13.14.0

**Additional context**
When going through the permission grant flow, the outcome is the generation of a CloudFormation template in the lambda's directory. This contains a subtree `AmplifyResourcesPolicy.Properties.PolicyDocument.Statement.Action`, and this is where the problem is. The actions granted by choosing the `read` in the permission grant flow are `appsync:Get*` and `appsync:List*`; `update` gets you `appsync:Update*` and `delete` gets you `appsync:Delete*`. `create` on the other hand gives you `appsync:Create*`, `appsync:StartSchemaCreation`, and critically `appsync:GraphQL`.

An examination of the available Actions under the AppSync category in the IAM Policy Simulator reveals the following: 

image

It looks like all the Create*, Get*, Update*, and Delete* Actions are meta-actions, which would allow your lambda to manipulate AppSync itself - not _use_ it.

I'm not sure if this is technically a bug, because you may well want a lambda to be able to perform meta-actions on your AppSync instance. But that's definitely not the use case advertised in the amplify docs on functions that I've linked twice in here. Those docs technically work because they recommend choosing the `create` option in the workflow, but the way the article is written, that looks like an arbitrary choice that shouldn't affect the outcome, and/or it is implied that because the author was using a mutation that did creation, _that's_ why `create` was chosen.

**Workaround**
Just choose the `create` option if you want a lambda to be able to access your AppSync API (and optionally go into the cloudformation template and delete the extraneous `Create*` and `StartSchemaCreation` Actions).

**Suggested Fix**
Update the CLI to include some more language about how the `create`, `read`, `update`, and `delete` choices relate to meta actions on the AppSync api itself; add a new category named `access` or `call GraphQL endpoint` or something; move the Action `appsync:GraphQL` from `create` to this new category; update the docs at https://docs.amplify.aws/cli/function

r0zar:
fix #4306

*Issue #, if available:*
4306

*Description of changes:*
The current implementation of appsync permissions is using an incorrect understanding of [IAM policies for AppSync](https://docs.aws.amazon.com/appsync/latest/devguide/security.html). This PR updates the IAM Policy generation process for lambdas from `amplify function create/update` walkthroughs.

This PR is ready for merging. :rocket: 

### DIFF
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
@@ -13,10 +13,7 @@ import {
 import { FunctionParameters, FunctionDependency } from 'amplify-function-plugin-interface/src';
 import { appsyncTableSuffix } from '../utils/constants';
 import { getAppSyncResourceName } from '../utils/appSyncHelper';
-import { stateManager } from 'amplify-cli-core';
-import { FeatureFlags } from 'amplify-cli-core';
-
-const generateGraphQLPermissions = () => FeatureFlags.getBoolean('generateGraphQLPermissions');
+import { stateManager, FeatureFlags } from 'amplify-cli-core';
 
 /**
  * This whole file desperately needs to be refactored
@@ -29,6 +26,9 @@ export const askExecRolePermissionsQuestions = async (
   category?: string,
   serviceName?: string,
 ): Promise<ExecRolePermissionsResponse> => {
+  // feature flag for graphQL permission bugfix as part of PR-5342
+  const generateGraphQLPermissions = FeatureFlags.getBoolean('appSync.generateGraphQLPermissions');
+
   const amplifyMeta = stateManager.getMeta();
 
   const categories = Object.keys(amplifyMeta).filter(category => category !== 'providers');
@@ -82,7 +82,7 @@ export const askExecRolePermissionsQuestions = async (
     }
 
     if (_.isEmpty(resourcesList)) {
-      context.print.warning(`No resources found for ${category}`);
+      context.print.warning(`No resources found for ${selectedCategory}`);
       continue;
     }
 
@@ -90,7 +90,7 @@ export const askExecRolePermissionsQuestions = async (
       let selectedResources = [];
       if (resourcesList.length > 1) {
         // There's a few resources in this category. Let's pick some.
-        const resourceQuestion = selectResourcesInCategory(resourcesList, currentPermissionMap, category);
+        const resourceQuestion = selectResourcesInCategory(resourcesList, currentPermissionMap, selectedCategory);
         const resourceAnswer = await inquirer.prompt([resourceQuestion]);
         selectedResources = _.concat(resourceAnswer.resources);
       } else {
@@ -99,14 +99,13 @@ export const askExecRolePermissionsQuestions = async (
       }
 
       for (let resourceName of selectedResources) {
-        // PR-5342
         // If the resource is AppSync, use GraphQL operations for permission policies.
         // Otherwise, default to CRUD permissions.
-        const serviceType = _.get(amplifyMeta, [category, resourceName, 'service']);
+        const serviceType = _.get(amplifyMeta, [selectedCategory, resourceName, 'service']);
         let options;
         switch (serviceType) {
           case 'AppSync':
-            options = generateGraphQLPermissions() ? graphqlOperations : crudOptions;
+            options = generateGraphQLPermissions ? graphqlOperations : crudOptions;
             break;
           default:
             options = crudOptions;
@@ -114,14 +113,15 @@ export const askExecRolePermissionsQuestions = async (
         }
 
         // In case of some resources they are not in the meta file so check for resource existence as well
-        const isMobileHubImportedResource = _.get(amplifyMeta, [category, resourceName, 'mobileHubMigrated'], false);
+        const isMobileHubImportedResource = _.get(amplifyMeta, [selectedCategory, resourceName, 'mobileHubMigrated'], false);
         if (isMobileHubImportedResource) {
-          context.print.warning(`Policies cannot be added for ${category}/${resourceName}, since it is a MobileHub imported resource.`);
+          context.print.warning(
+            `Policies cannot be added for ${selectedCategory}/${resourceName}, since it is a MobileHub imported resource.`,
+          );
           continue;
         } else {
-          const resourceType = _.get(amplifyMeta, [category, resourceName, 'service']);
-          const currentPermissions = fetchPermissionsForResourceInCategory(currentPermissionMap, category, resourceName);
-          const permissionQuestion = selectPermissions(options, currentPermissions, resourceType);
+          const currentPermissions = fetchPermissionsForResourceInCategory(currentPermissionMap, selectedCategory, resourceName);
+          const permissionQuestion = selectPermissions(options, currentPermissions, resourceName);
           const permissionAnswer = await inquirer.prompt([permissionQuestion]);
           const resourcePolicy: any = permissionAnswer.options;
 
@@ -280,10 +280,10 @@ const selectCategories = (choices: DistinctChoice<any>[], currentPermissionMap: 
   default: fetchPermissionCategories(currentPermissionMap),
 });
 
-const selectPermissions = (choices: DistinctChoice<any>[], currentPermissions: any, resourceType: any) => ({
+const selectPermissions = (choices: DistinctChoice<any>[], currentPermissions: any, resourceName: any) => ({
   type: 'checkbox',
   name: 'options',
-  message: `Select the operations you want to permit for ${resourceType}`,
+  message: `Select the operations you want to permit on ${resourceName}`,
   choices,
   validate: answers => (_.isEmpty(answers) ? 'You must select at least one operation' : true),
   default: currentPermissions,
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
@@ -1,9 +1,9 @@
 import { constructCFModelTableNameComponent, constructCFModelTableArnComponent } from '../utils/cloudformationHelpers';
-import inquirer from 'inquirer';
+import inquirer, { CheckboxQuestion, DistinctChoice } from 'inquirer';
 import path from 'path';
 import * as TransformPackage from 'graphql-transformer-core';
 import _ from 'lodash';
-import { topLevelCommentPrefix, topLevelCommentSuffix, envVarPrintoutPrefix, CRUDOperation } from '../../../constants';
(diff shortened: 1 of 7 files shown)


jamesonwilliams:
PHAL @aws-amplify/amplify-cli (@kaustavghosh06 @renebrandel @UnleashedMind @yuth @attilah)

renebrandel:
hi @r0zar ! I'll try to get the team to review it on Friday. Thanks for starting this!

r0zar:
@renebrandel :wave:  Completed my final updates. This PR is ready for merging now AFAIK

renebrandel:
@attilah will test this PR and see if we need to introduce a feature flag. At this point it looks good.

attilah:
@r0zar Thanks for the PR, I went through it and I see the reason for the changes, but because of existing deployments we have to preserve the existing functionality so it needs to be put behind a feature flag that can be enabled by default for new projects and not for existing projects.

I have one concern about the generated IAM policy and that is that I can't see that the policy generation code includes the schema types. If I remember correctly you have to have access to the types as well, since if I have a policy for `UpdatePost` that does not implicitly grant me access to the `Post` type. This brings to the other thing I am missing is an E2E test which creates a function and an AppSync API and invokes a GraphQL operation.

Since this change has to go behind a feature flag we don't have to deal with the legacy permissions especially we don't have to map it.

If you need further help on the above suggested changes, ping me and I we can setup a meeting to get it through the finish line.

r0zar:
Thanks for the review. 

Yeah let's find some time next week so I get aligned on this and put together a todo list.

attilah:
@r0zar sounds good, please test out the `types/*` policy I miss before that if you can.

attilah:
Synced up offline with @r0zar will add the E2E tests and the Feature flag to preserve the existing functionality and will update the PR.

attilah:
@r0zar how you see your time working on this, do you have an ETA?

r0zar:
I finally had some time open up this week and was going to make a start on it. Hopefully I can knock it out in a few sessions.

attilah:
@r0zar let me know if I can be any help.

r0zar:
Ran a rebase to keep my changes.

r0zar:
Ok, fixed the terrible rebase. :sweat_smile:

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/constants.ts
```diff
@@ -9,9 +9,9 @@ export const layerParametersFileName = 'layer-parameters.json';
 export const parametersFileName = 'parameters.json';
 export const provider = 'awscloudformation';
 export const appsyncTableSuffix = '@model(appsync)';
-export const resourceAccessSetting = 'Resource access permissions';
-export const cronJobSetting = 'Scheduled recurring invocation';
-export const lambdaLayerSetting = 'Lambda layers configuration';
+export const resourceAccessSetting = '🔒 Resource access permissions';
+export const cronJobSetting = '⏰ Schedule recurring invocation';
+export const lambdaLayerSetting = '📦 Lambda layers configuration';
```
jhockett, line 14: Make sure associated Lambda e2e test helper functions are updated to wait on the updated text.

r0zar: Gotcha! thanks.

r0zar:
:tada:

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
```diff
@@ -1,9 +1,9 @@
 import { constructCFModelTableNameComponent, constructCFModelTableArnComponent } from '../utils/cloudformationHelpers';
-import inquirer from 'inquirer';
+import inquirer, { CheckboxQuestion, DistinctChoice } from 'inquirer';
 import path from 'path';
 import * as TransformPackage from 'graphql-transformer-core';
-import _ from 'lodash';
-import { topLevelCommentPrefix, topLevelCommentSuffix, envVarPrintoutPrefix, CRUDOperation } from '../../../constants';
+import _, { get, capitalize, concat, isEmpty } from 'lodash';
```
attilah, line 5: Please use `import _ from 'lodash';` and don't deconstruct its functions, it's more readable that way.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -60,7 +60,7 @@ export async function createWalkthrough(
   // list out the advanced settings before asking whether to configure them
   context.print.info('');
   context.print.success('Available advanced settings:');
-  advancedSettingsList.forEach(setting => context.print.info('- '.concat(setting)));
+  advancedSettingsList.forEach(setting => context.print.info(setting));
```
attilah, line 63: Why removed the '-' prefix? We rarely change prompts and DX without API reviews. Sync up on this with @renebrandel .

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/constants.ts
attilah: @renebrandel please ack the icon changes.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/permissionMapUtils.ts
```diff
@@ -1,5 +1,8 @@
+import { FeatureFlags } from 'amplify-cli-core';
 import _ from 'lodash';
 
+const featureFlagPR5342 = () => FeatureFlags.getBoolean('useAppSyncPermissionsInLambdaPolicies')
```
attilah, line 4: Just name it normally as a variable, no need for GH reference, also FeatureFlags has a <group>.<feature>, I'd suggest to use `appSync`.`generateGraphQLPermissionsForCRUD` since this feature flag is not for `graphqlTransformer`. We can debate the name but this is what it does IMHO, suggest something similar if you have a better in mind.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/permissionMapUtils.ts
```diff
@@ -8,6 +11,36 @@ export const fetchPermissionResourcesForCategory = (permissionMap, category: str
...
+export const fetchPermissionsForResourceInCategory = (
+  permissionMap: object = {},
+  category: string,
+  resourceName: string,
+  serviceType: string,
+) => {
+  const permissions = _.get(permissionMap, [category, resourceName], []);
+  // If AppSync and legacy CRUD permissions BUT they opted into the feature flag, update policy.
+  if (serviceType == 'AppSync' && hasCRUDPermissions(permissions) && featureFlagPR5342()) {
+    return updatePermissionsForBackwardsCompatibility(permissionMap, category, resourceName);
```
attilah, line 23: Why is the FF branch of the if doing anything that is backward compatible? It should not, right?

packages/amplify-cli-core/src/feature-flags/featureFlags.ts
```diff
@@ -504,6 +504,12 @@ export class FeatureFlags {
         defaultValueForExistingProjects: false,
         defaultValueForNewProjects: false,
       },
+      {
+        name: 'useAppSyncPermissionsInLambdaPolicies',
```
attilah, line 508: As suggested above it should go into a new `appSync` group.

attilah (changes requested):
Added a few comments on the recent changes, but overall almost there!

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/permissionMapUtils.ts
r0zar: how do you feel about the category being `iamPolicies`? This distinction would be helpful to make clear it's related to how the policies are generated and managed.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
r0zar: The `-` served as the bullet-point in a pseudo unordered list. I replaced that with the icons in the PR. It results in a more visually comprehensible list of choices that can be understood better internationally.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/permissionMapUtils.ts
r0zar: I wrote it so that enabling a feature flag with incorrect permissions would autocorrect their policies. I now see that the way feature flags are being used in amplify is that by turning them on early, the user may deals with the consequences. I see this results in code with less 'artifacts' so I'll update to fit this design.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
r0zar: :heavy_check_mark:

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/permissionMapUtils.ts
r0zar: I moved the FF check to line 102 of `execPermissionsWalkthrough.ts`. Still exists as an artifact in the code, but there is no more autocorrecting of permissions. Is this the intention?

r0zar:
I squashed and force pushed the commit with only the relevant changes to keep the scope a little tighter.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
```diff
@@ -25,32 +26,31 @@ export const askExecRolePermissionsQuestions = async (
   category?: string,
```
r0zar, line 26: This value seems to be coming in as undefined for most scenarios. Is it still used?

attilah (approved):
LGTM, thanks for hanging in there ;-)

PR #5359, opened 2020-09-18
feat(amplify-codegen-appsync-model-plugin): add appsync dart visitor

AaronZyLee:
*Issue #, if available:*

*Description of changes:*

- [x] Added dart declaration block and visitor to generate model code for the use of Amplify Datastore. Directives implemented:

    - [x] `@model`   

    - [x] `@connection`

    - [x] `@auth`

    - [x] `@key`
- [x] Added unit tests for dart model generation


packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
```diff
@@ -0,0 +1,2131 @@
...
+exports[`AppSync Dart Visitor Model Directive should generate a class for a Simple Model 1`] = `
+"import 'package:flutter/foundation.dart';
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+
+
+/** This is an auto generated class representing the SimpleModel type in your schema. */
+@immutalbe
+class SimpleModel extends Model {
+  static const classType = SimpleModelType();
+  final String id;
```
phani-srikar, line 27: We might want a line spacing between `static const` and `final` variables?

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
```diff
@@ -0,0 +1,2131 @@
...
+  static final QueryField OWNER = QueryField(fieldName: \\"owner\\");
+  static var schema = Model.defineSchema(define: (ModelSchemaDefinition modelSchemaDefinition) {
+    modelSchemaDefinition.name = \\"customClaim\\";
+    modelSchemaDefinition.pluralName = \\"customClaims\\";
+    
+    modelSchemaDefinition.authRules = [
+      AuthRule(
+        authStrategy: AuthStrategy.OWNER,
+        ownerField: \\"owner\\",
+        identityClain: \\"user_id\\",
```
phani-srikar, line 309: this is probably `identityClaim`

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
```diff
@@ -0,0 +1,2131 @@
...
+  String getId() {
+    return id;
+  }
+  
+  const Todo._internal({@required this.id, this.tasks});
+  
+  factory Todo({@required String id, List<Task> tasks}) {
+    return Todo._internal(
+      id: id == null ? UUID.getUUID() : id,
+      tasks: tasks != null ? List.unmodifiable(tasks) : tasks);
```
phani-srikar, line 1447: This should be `List.unmodifiable(tasks) : null`?

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
```diff
@@ -0,0 +1,2131 @@
...
+  Post fromJson(Map<String, dynamic> jsonData) {
+    return Post.fromJson(jsonData);
+  }
+}"
+`;
+
+exports[`AppSync Dart Visitor Model with Connection Directive should generate classes for models with one to many connection 1`] = `
+"import 'package:flutter/foundation.dart';
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+import 'package:collection/collection.dart';
```
phani-srikar, line 1425: Is it possible to sort all the imports before rendering?

phani-srikar (changes requested):
Good work. Few corrections.

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
```diff
@@ -0,0 +1,2131 @@
...
+      bar == other.bar;
+  }
+  
+  @override
+  int get hashCode => toString().hashCode;
+  
+  @override
+  String toString() {
+    var buffer = new StringBuffer();
+    buffer.write(\\"SimpleModel {\\");
```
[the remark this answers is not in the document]
AaronZyLee: These are generated by jest. They are not actually added in the real output.

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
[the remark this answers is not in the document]
AaronZyLee: I tried to add the space at first but it will make the code generator more complicated. So I just skip the part.

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
AaronZyLee: The first two are base imports for all generated files. The collection is only added when there is list type variable. This part is hard code.

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
```diff
@@ -0,0 +1,2131 @@
...
+"import 'package:flutter/foundation.dart';
+import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
+
+
+/** This is an auto generated class representing the authorBook type in your schema. */
+@immutalbe
+class authorBook extends Model {
+  static const classType = authorBookType();
+  final String id;
+  final String author_id;
```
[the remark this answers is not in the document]
AaronZyLee: These fields will be removed if they are as the `fieldName` in connection type 'BELONGS_TO'

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
AaronZyLee: I got this from the reference model. It will always be `null` in this case though.

phani-srikar: oh it's the same thing +1

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
phani-srikar: We can maintain a set of packages to be imported (including base and additional imports) and sort them before actually rendering them right?

AaronZyLee: Add sort in the new commit

packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/__snapshots__/appsync-dart-visitor.test.ts.snap
```diff
@@ -0,0 +1,2131 @@
...
+  
+  Map<String, dynamic> toJson() =>
+    {'id': id, 'author_id': author_id, 'book_id': book_id, 'author': author, 'book': book};
+
+  static final QueryField ID = QueryField(fieldName: \\"id\\");
+  static final QueryField AUTHOR_ID = QueryField(fieldName: \\"author_id\\");
+  static final QueryField BOOK_ID = QueryField(fieldName: \\"book_id\\");
+  static final QueryField AUTHOR = QueryField(fieldName: \\"author\\");
+  static final QueryField BOOK = QueryField(fieldName: \\"book\\");
+  static var schema = Model.defineSchema(define: (ModelSchemaDefinition modelSchemaDefinition) {
```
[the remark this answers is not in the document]
AaronZyLee: The same as the first one. We can add a new line but it will make the generator more complicated.

packages/amplify-codegen-appsync-model-plugin/src/configs/dart-config.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const typeToEnumMap: { [name: string] : string} = {
+  String: 'string',
+  Int: 'int',
+  Float: 'double',
+  Boolean: 'bool',
+  AWSDate: 'date',
+  AWSDateTime: 'dateTime',
+  AWSTime: 'time',
+  AWSTimestamp: 'timestamp',
```
attilah, line 16: A few AWS* types missing here, is that intentional? Also the mapped values are different then in scalar map, why?

packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -0,0 +1,549 @@
...
+  }
+
+  protected generateModelClass(model: CodeGenModel): string {
+    //class wrapper
+    const classDeclarationBlock = new DartDeclarationBlock()
+      .asKind('class')
+      .withName(this.getModelName(model))
+      .extends(['Model'])
+      .withComment(`This is an auto generated class representing the ${model.name} type in your schema.`)
+      .annotate(['immutalbe']);
```
attilah, line 119: immutable typo

packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-dart-visitor.ts
```diff
@@ -0,0 +1,549 @@
...
+  }
+
+      /**
+   * Get the list of fields that can be are writeable. These fields should exclude the following
+   * fields that are connected and are either HAS_ONE or HAS_MANY
+   * @param model
+   */
+  protected getNonConnectedField(model: CodeGenModel): CodeGenField[] {
+    return model.fields.filter(f => {
+      if (!f.connectionInfo) return true;
```
attilah, line 543: use blocks everywhere for readability.

attilah (changes requested):
Overall looks good, a few questions and comments and some formatting issues.

What happens if the GraphQL schema contains a field or type name that is a reserved keyword in Dart? I don't see any validation around these things.

Test only contains happy path tests, there are no non-happy paths to test for codegen?

As for formatting it would be a nice addition, that if the [`dartfmt`](https://dart.dev/tools/dartfmt) tool present, then invoke it on the generated file, so it would look conform to default dart styles.

I know that running the

packages/amplify-codegen-appsync-model-plugin/src/plugin.ts
```diff
@@ -37,6 +38,13 @@ export const plugin: PluginFunction<RawAppSyncModelConfig> = (
     case 'javascript':
       visitor = new AppSyncModelJavascriptVisitor(schema, config, {});
       break;
+    case 'dart':
```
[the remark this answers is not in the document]
AaronZyLee: The target is read from the `frontend` field of `project-config.json`, which is created by flutter frontend plugins. The code for this part is in `amplify-codegen/src/commands/models.js`. That's why I add the flutter as an additional case of dart.

packages/amplify-codegen-appsync-model-plugin/src/plugin.ts
```diff
@@ -37,6 +38,13 @@ export const plugin: PluginFunction<RawAppSyncModelConfig> = (
...
+    case 'flutter':
+      visitor = new AppSyncModelDartVisitor(schema, config, {
```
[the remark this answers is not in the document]
AaronZyLee: The code here is similar to the case for android and ios. I tried to include them in the 2nd argument but it did not actually read the property that passed in it.

packages/amplify-codegen-appsync-model-plugin/src/configs/dart-config.ts
AaronZyLee: This is not the mapping for graphql type to dart native type but to a enum type designed by flutter team. This is not finalized since flutter team is still discussing with ios and android team. I will update this after they have a final result.

packages/amplify-codegen-appsync-model-plugin/src/plugin.ts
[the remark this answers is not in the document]
AaronZyLee: I reorganize the entry point in `models.js` by adding a platform to language mapping while removing platform cases and remaining only language ones in modelgen plugin in the latest commit.

packages/amplify-codegen-appsync-model-plugin/src/plugin.ts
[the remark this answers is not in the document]
AaronZyLee: I just check the original code. The constructor use its super class method. `selectedType` and `generate` are two additional configs and cannot be passed except they are in the 3rd argument.

packages/amplify-codegen-appsync-model-plugin/src/configs/dart-config.ts
AaronZyLee: Update: add the missing AWS types

AaronZyLee:
Updates since last review:

- Add dart reserved keyword validation and tests

- Add dart formatting for output code

- Change the codegen target to language-only and add platform to language mapping in amplify-codegen

Also some changes for the modelgen output:

- add license to top of file

- include const in classType variable

- add getInstanceType method

- in fromJson method change how nested model fromJson is called

- queryField goes “modelName.id” instead of just id

attilah (approved):
LGTM, thanks for the changes, good job!

fjnoyp:
- ModelFieldDefinition is mispelled as “ModelFieldDeinition”

- ModelProvider import package has a space in the declaration: import ‘package: amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart’;

The next 2 comments reference the below image. 
- in fromJson, when receiving a nested map, the syntax must be as shown in red (the green is what is code gened right now). Without this change if the map received is null the program will crash
- in toJson, nested models must have a “?” as they can be null.  Again, the syntax should be what is red
Below is a picture for reference:
image

- In the fromJson method, objects must be converted into their data type.  So for DateTime, Date, Time, Timestamp, Collection<Model>, Model, ENUM must be converted into their respective type.  We can provide this full mapping for you

fjnoyp:
If I declare a enum in my .graphQL : 

```
enum TestEnum { 
  yes,
  no,
  maybe
}
```

And then run amplify-dev codegen models,

The enum doesn't appear to be generated.  Instead a blank file is created: 
```

/*
* Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
*
* Licensed under the Apache License, Version 2.0 (the "License").
* You may not use this file except in compliance with the License.
* A copy of the License is located at
*
*  http://aws.amazon.com/apache2.0
*
* or in the "license" file accompanying this file. This file is distributed
* on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
* express or implied. See the License for the specific language governing
* permissions and limitations under the License.
*/

import 'package:amplify_datastore_plugin_interface/amplify_datastore_plugin_interface.dart';
import 'package:flutter/foundation.dart';

```

fjnoyp:
The following is a mapping of .graphQL types to Dart types for the Codegen model: 

  ID: 'String',
  String: 'String',
  Int: 'Int',
  Float: 'Double',
  Boolean: 'Bool',
  **AWSDate: 'DateTime'
  AWSTime: 'DateTime',
  AWSDateTime: 'DateTime',
  AWSTimestamp: 'String',**
  AWSEmail: 'String',
  AWSJSON: 'String',
  AWSURL: 'String',
  AWSPhone: 'String',
  AWSIPAddress: 'String',

fjnoyp:
AWSDate: 'DateTime'
AWSTime: 'DateTime',
AWSDateTime: 'DateTime'

DateTime objects should be converted to a string via this method: 

        'created': created.toIso8601String()

instead of this method: 

        'created': formatDateToISO8601(created),

fjnoyp:
Could you update the toJson methods to remove null values - the red is the current generation, the green is the desired generation:

fjnoyp:
Can you make timestamp fields be of type timestamp?

    modelSchemaDefinition.addField(ModelFieldDefinition.field(
        key: AllTypeModel.TIMESTAMPTYPE,
        isRequired: true,
        ofType: ModelFieldType(ModelFieldTypeEnum.**timestamp**)));

fjnoyp:
Hi @lizeyu there is an issue with the fromJson method.
For fields of the following type:
AWSDate: ‘DateTime’
AWSTime: ‘DateTime’,
AWSDateTime: ‘DateTime’,
They must be parsed.  The red is what should be there - the green is what is currently generated.

fjnoyp:
Hi @AaronZyLee minor issue, the fields of AWSDate and AWSDateTime should be different.  One should be of ModelFieldTypeEnum.date while the other should be ModelFieldTypeEnum.dateTime

    modelSchemaDefinition.addField(ModelFieldDefinition.field(
        key: AllTypeModel.DATETYPE,
        isRequired: true,
        **ofType: ModelFieldType(ModelFieldTypeEnum.date)));**

    modelSchemaDefinition.addField(ModelFieldDefinition.field(
        key: AllTypeModel.DATETIMETYPE,
        isRequired: true,
        **ofType: ModelFieldType(ModelFieldTypeEnum.dateTime)));**

Issue #4761, opened 2020-07-07
amplify mock function myPreSignup errors with: Cannot read property 'filter' of undefined


kjetilge:
**Describe the bug**
When running ```amplify mock function myPreSignup``` on a fresh untouched PreSignup function I get the error:
```Cannot read property 'filter' of undefined```

I suspect this error came with the latest CLI version.

**Amplify CLI Version**
4.23.0

**To Reproduce**
Create an AUTH with PreSignup function then run:
```amplify mock function yourPreSignup```

**Expected behavior**
It should not result in error.

**Desktop (please complete the following information):**
 - OS: Mac
 - Node Version 10.17.0

**Comment by jhockett** [comment]
Hi @kjetilge, I was able to reproduce this. Thanks for the bug report! I'll take this to the team to begin investigating the root cause.

**Comment by kjetilge** [comment]
Great!😃

**Comment by jhockett** [comment]
A fix has been merged and will go out in the next release

**Comment by berenddeboer** [comment]
What's going on guys? All last couple of releases are broken.

**Comment by byennen** [comment]
same issue after upgrading to 4.23.0; rolling back!

**Comment by arvenom** [comment]
Any updates on this?

**Comment by jhockett** [comment]
> Any updates on this?

The bug fix went out several versions ago. Are you still experiencing the issue @arvenom ?

**Comment by kjetilge** [comment]
This is still a problem in version 4.29.3. Hope this can be fixed soon.

**Comment by kjetilge** [comment]
@jhockett - could you have another look at this ?

**Comment by jhockett** [comment]
Hi @kjetilge, I'm working on another project at the moment, but I will make sure the team is aware. Did this bug just re-appear in 4.29.3?

**Comment by kjetilge** [comment]
> Hi @kjetilge, I'm working on another project at the moment, but I will make sure the team is aware. Did this bug just re-appear in 4.29.3?

I have not tried it in a while so I don't know.
Also if I try to run this code: 
```
// index.js
console.log('Hello')
callback(null, event);
```
I get the same error text (in red) and no "Hello"

However, if I try to update the function and give it permission to access api read, then it behaves like this:
```
amplify mock function myPreSignup
? Provide the path to the event JSON object relative to /Users/kjetilge/Lokale filer/GitHub/my/amplify/backend/function/myPreSignup src/event.json
Could not find ref for "modules". Using unsubstituted value.
Could not find ref for "DOMAINWHITELIST". Using unsubstituted value.
[...]

PR #5446, opened 2020-09-29
fix: check on  dependsOn Array

akshbhu:
*Issue #4761

*Description of changes:*

* Added a Check on DependsOn Array

### DIFF
--- a/packages/amplify-category-function/src/index.ts
+++ b/packages/amplify-category-function/src/index.ts
@@ -176,13 +176,13 @@ export function isMockable(context: any, resourceName: string): IsMockableRespon
     };
   }
   const { service, dependsOn } = resourceValue;
-  let dependsOnLayers;
-  if (Array.isArray(dependsOn) && dependsOn.length !== 0) {
-    dependsOnLayers = dependsOn
-      .filter(dependency => dependency.category === 'function')
-      .map(val => _.get(context.amplify.getProjectMeta(), [val.category, val.resourceName]))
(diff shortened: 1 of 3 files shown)


packages/amplify-category-function/src/index.ts
```diff
@@ -176,10 +176,13 @@ export function isMockable(context: any, resourceName: string): IsMockableRespon
...
-    .filter(dependency => dependency.category === 'function')
-    .map(val => _.get(context.amplify.getProjectMeta(), [val.category, val.resourceName]))
-    .filter(val => val.service === ServiceName.LambdaLayer);
+  let dependsOnLayers;
+  if (Array.isArray(dependsOn) && dependsOn.length !== 0) {
+    dependsOnLayers = dependsOn
+      .filter(dependency => dependency.category === 'function')
+      .map(val => _.get(context.amplify.getProjectMeta(), [val.category, val.resourceName]))
+      .filter(val => val.service === ServiceName.LambdaLayer);
+  }
```
jhockett, line 185: LGTM! Personally I'd use a ternary operator instead of the `if` like this:

```ts
  const dependsOnLayers = Array.isArray(dependsOn) ? dependsOn
    .filter(dependency => dependency.category === 'function')
    .map(val => _.get(context.amplify.getProjectMeta(), [val.category, val.resourceName]))
    .filter(val => val.service === ServiceName.LambdaLayer) : [];
```

That's just preference though. What do you think?

jhockett (approved):
LGTM!

PR #5450, opened 2020-09-30
fix(amplify-codegen-appsync-model-plugin): fixed nullable lists

seang96:
*Issue #, if available:*

*Description of changes:*
isListNullable needs to return true when list is nullable due to the name change from PR #5327 

### DIFF
--- a/packages/amplify-codegen-appsync-model-plugin/src/utils/get-type-info.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/utils/get-type-info.ts
@@ -12,6 +12,7 @@ export function getTypeInfo(typeNode: TypeNode, schema: GraphQLSchema): TypeInfo
     return {
       ...getTypeInfo(typeNode.type.type, schema),
       isList: true,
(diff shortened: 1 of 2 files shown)


packages/amplify-codegen-appsync-model-plugin/src/utils/get-type-info.ts
```diff
@@ -12,7 +12,6 @@ export function getTypeInfo(typeNode: TypeNode, schema: GraphQLSchema): TypeInfo
     return {
       ...getTypeInfo(typeNode.type.type, schema),
       isList: true,
-      isListNullable: false,
```
yuth, line 15: Can we be explicit in here by setting this to false

seang96: Yeah, I'll update that... I didn't do this explicitly in the past since it increased the number of test cases that needed modification.

yuth (approved):

PR #5484, opened 2020-10-04
Fix 5436

### LINKED ISSUE #5436: Using custom `ownerField:` with `@auth` should automatically generate schema with custom field

**Describe the bug**
When using `ownerField` in the `@auth`directive, the custom field should be present in the graphql schema

**Amplify CLI Version**
4.29.2

**To Reproduce**
Create a new iOS amplify project with the following graph ql schema
```gql
type Post @model @auth(rules: [{ allow: owner, ownerField: "dominus" }]) {
  id: ID!
  title: String!
}
```

Kick off model generation:
`amplify api gql-compile`

```gql
$ cat amplify/backend/api/scenarioatest4/build/schema.graphql|more
type Post {
  id: ID!
  title: String!
  _version: Int!
  _deleted: Boolean
  _lastChangedAt: AWSTimestamp!
  createdAt: AWSDateTime!
  updatedAt: AWSDateTime!
}
```

**Expected behavior**
Expected output:
```gql
$ cat amplify/backend/api/scenarioatest4/build/schema.graphql|more
type Post {
  id: ID!
  title: String!
  dominus: String
  _version: Int!
  _deleted: Boolean
  _lastChangedAt: AWSTimestamp!
  createdAt: AWSDateTime!
  updatedAt: AWSDateTime!
}
```

Notice that `dominus: String` was automatically added because we are using `ownerField`

**Screenshots**
N/A
 - OS: Mac os
 - Node Version: v14.2.0

**Additional context**

yuth:
Update the transformer to automatically add owner fields to type
when owner auth is used and the owner fields are not explictly
declared in the type. This change is behind a feature flag
graphQLTransformer.addMissingOwnerFields and gets enabled automatically
only on new projects.

when the feature flag is enabled

```gql
type Record
  @model
  @auth(
    rules: [
      { allow: owner, operations: [update, delete, create] }
      {
        allow: owner
        ownerField: "custom_owner"
        identityClaim: "sub"
        operations: [update, delete, create]
      }
      { allow: owner, ownerField: "editor" }
      { allow: groups, groupsField: "groups" }
    ]
  ) {
  id: ID!
  name: String
  editor: [String]
}
```
produces
```gql
type Record {
  id: ID!
  name: String
  editor: [String]
  createdAt: AWSDateTime!
  updatedAt: AWSDateTime!
  owner: String
  custom_owner: String
}
```

fix #5436


packages/graphql-transformer-core/src/util/transformConfig.ts
```diff
@@ -99,6 +99,12 @@ export interface TransformConfig {
    * Such as sync configuration for appsync local support
    */
   ResolverConfig?: ResolverConfig;
+
+  /**
+   * List of custom transformer plugins
+   */
+  transformers?: string[];
+  warningESMessage?: boolean;
```
SwaySway, line 107: Aren't we using Elasticsearch Warning for this? If we are changing it we might want to use a feature flag?

SwaySway, line 107: Is there a need for these values in the transform config? If we moving away from using `ElasticsearchWarning` should we create a feature flag for that?

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1922,7 +1929,34 @@ operations will be generated by the CLI.`,
...
+   * @param object ObjectDefinitionNode
+   * @param rules Authorization rules
+   */
+  private addOwnerFieldsToObject(ctx: TransformerContext, typeName: string, rules: AuthRule[]): void {
+    if (ctx.featureFlags.getBoolean('addMissingOwnerFields', true)) {
+      const object = ctx.getObject(typeName);
+      const ownerRules = rules.filter(rule => rule.allow === 'owner');
+      const existingFields = Object.keys(getFieldArguments(object));
+      const ownerFileds = Array.from(new Set(ownerRules.map(rule => rule.ownerField || DEFAULT_OWNER_FIELD)).values());
+      const fieldsToadd = ownerFileds.filter(field => !existingFields.includes(field));
```
SwaySway, line 1944: nit suggestion
```suggestion
      const ownerFields = Array.from(new Set(ownerRules.map(rule => rule.ownerField || DEFAULT_OWNER_FIELD)).values());
      const fieldsToAdd = ownerFields.filter(field => !existingFields.includes(field));
```

packages/graphql-transformer-core/src/util/transformConfig.ts
yuth: These are the values that are getting passed here. Typescript complained about these missing fields in config when being used [here](https://github.com/yuth/amplify-cli/blob/5f674c843e45c6a230bee8bf26cef8bc2f40c8cd/packages/amplify-provider-awscloudformation/src/transform-graphql-schema.ts#L73-L76) and [here](https://github.com/yuth/amplify-cli/blob/5f674c843e45c6a230bee8bf26cef8bc2f40c8cd/packages/amplify-provider-awscloudformation/src/transform-graphql-schema.ts#L192-L195)

SwaySway (approved):

ammarkarachi (approved):

PR #5493, opened 2020-10-05
fix(amplify-codegen-appsync-model-plugin): types

seang96:
Fix types regarding nullable lists

*Issue #, if available:* Mentioned in PR #5327

*Description of changes:*
Fixes an issue with typing with newly added nullable array support for JavaScript and Typescript. Thank you @danrivett for discovering it.


packages/amplify-codegen-appsync-model-plugin/src/__tests__/visitors/appsync-javascript-visitor.test.ts
```diff
@@ -100,7 +100,7 @@ export enum SimpleEnum {
 
 export declare class SimpleNonModelType {
   readonly id: string;
-  readonly names?: string[];
+  readonly names?: (string | null)[];
```
danrivett, line 103: I don't know if there's a unit test to make sure that: 

```graphql
foo: [Bar!]
```

Is generated correctly?

As in that case we would want the following generated since the elements themselves should be non-null, just the array nullable:

```ts
readyonly foo?: Bar[];
```

seang96: @danrivett, I added a new Bar type to the tests and included it in the SimpleModel to cover this case.

packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-typescript-visitor.ts
```diff
@@ -216,7 +216,11 @@ export class AppSyncModelTypeScriptVisitor<
   }
 
   protected getListType(typeStr: string, field: CodeGenField): string {
-    return `${typeStr}[]`;
+    let type: string = typeStr;
+    if (field.isList && field.isNullable) {
```
danrivett, line 220: I'm not familiar with the codegen code so treat accordingly, but would:

```suggestion
    if (field.isNullable) {
```

Be sufficient here?

Do we need to check the `isList` type since this function is always returning an array type regardless (line 223) and so it seems this if-block only cares if the list element type is nullable or not.

seang96: Yeah I imagine just isNullable is sufficient, I was originally placing this code elsewhere which actually needed it until I found getListType. I am also learning the library haha

danrivett (approved):
LGTM (as a layperson here at least).

danrivett:
@yuth If you can review when you're available, since you reviewed the previous related PRs, it would be much appreciated. Thanks.

packages/amplify-codegen-appsync-model-plugin/src/visitors/appsync-typescript-visitor.ts
```diff
@@ -87,15 +87,15 @@ export class AppSyncModelTypeScriptVisitor<
...
+        optional: field.isList ? field.isListNullable : field.isNullable,
       });
     });
     let ownerAuthRules = getOwnerAuthRules(modelObj);
     if (ownerAuthRules !== undefined) {
       for (let rule of ownerAuthRules) {
         let ownerFieldName = getOwnerFieldName(rule);
         if (ownerFieldName !== undefined && !hasOwnerField(modelObj.fields, ownerFieldName)) {
-          modelDeclarations.addProperty(ownerFieldName, "String", undefined, 'DEFAULT', {
+          modelDeclarations.addProperty(ownerFieldName, 'String', undefined, 'DEFAULT', {
```
AaronZyLee, line 98: I think this is where the merging conflict is. I removed this part in my previous merged PR.

seang96: @AaronZyLee , I did not notice the merge conflict... My editor must've refactored to single quotes. I am pushing the merge that resolves the conflict now.

danrivett:
@yuth pinging again as this fixes a regression introduced in version 4.29.4 and until it's merged I cannot upgrade from 4.29.3.

yuth (approved):
Thank you for the PR

PR #5498, opened 2020-10-06
fix(amplify-nodejs-function-runtime-provider): invoke waits close 

### LINKED ISSUE #4439: amplify-nodejs-function-runtime-provider can't handle big responses

chucre:
invoke waits close event before rejects

execa module waits untill all output is drained, but it can sends the 'close' event after the the
promise is rejected. Or the promise is rejected after close event but before before functions
finishs. The change will upto 2 seconds to event close is been sent and waits until the close call
back is completed

fix #4439

*Issue #, if available:*

*Description of changes:*


edwardfoyle (approved):
@chucre thanks for the fix and the comprehensive tests! LGTM!

Issue #4532, opened 2020-06-11
✖ Zipping artifacts failed.


mluis:
**Describe the bug**
amplify publish builds correctly but fails on zip artifacts.

**Amplify CLI Version**
4.21.3

**To Reproduce**
$ amplify publish

**Expected behavior**
amplify should be enable to zip and perform publish.
the message should be accurate on how and where it failed.

**Desktop (please complete the following information):**
 - OS: Mac 10.13.6
 - Node Version: v13.2.0

**Additional context**
Please ask.

SwaySway:
Hello @mluis
What kind of error message is shown?
To better reproduce this are you able to share your project to amplify-cli@amazon.com?

mluis:
The only message shown was "Zipping artifacts failed." and the app quit. Perhaps I should had written a better issue title.

I had to setup everything again. On my end the project is now running although I couldn't understand what was wrong before.

jonnyparko:
check that it is building in the directory you expect

mluis:
@jonnyparko I think the cli should state exactly what's the problem.

jonnyparko:
@mluis Ah gotcha I see that now. My bad!

patelatharva:
In case of React app, your_app/amplify/.config/project-config.json should be something like this:
```
{
    "projectName": "blah",
    "version": "3.0",
    "frontend": "javascript",
    "javascript": {
        "framework": "react",
        "config": {
            "SourceDir": "src",
            "DistributionDir": "dist",
            "BuildCommand": "npm run-script build",
            "StartCommand": "npm run-script start"
        }
    },
    "providers": [
        "awscloudformation"
    ]
}
```

Rolando-Barbella:
Any update on this one? there is not even an error on the console, what is _Zipping artifacts failed_ supposed to mean?

Rolando-Barbella:
If you are using gatsby, you need to make sure that `"DistributionDir": "public",`

Freakrishnal:
I'm getting ZIPPING ARTIFACTS FAILED when i tried **amplify publish** from amplify CLI. The problem is no error is written. I didn't do any fancy thing. This is my first time and i simply followed tutorial steps in amplify.

Rolando-Barbella:
@Freakrishnal if show your __app/amplify/.config/project-config.json_ file, or even your project, that could help

Freakrishnal:
[...]

PR #5544, opened 2020-10-09
(fix) suggest fix for zipping failure message

sw-yx:
closes https://github.com/aws-amplify/amplify-cli/issues/4532

*Description of changes:* this builds in a fix to the failure message so that users know how to resolve their own issues.

### DIFF
--- a/packages/amplify-console-hosting/hosting/manual/publish.js
+++ b/packages/amplify-console-hosting/hosting/manual/publish.js
@@ -9,7 +9,7 @@ const ora = require('ora');
 
 const ZIPPING_MESSAGE = 'Zipping artifacts.. ';
 const ZIPPING_SUCCESS_MESSAGE = 'Zipping artifacts completed.';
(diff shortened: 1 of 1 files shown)


renebrandel:
@UnleashedMind - LMK what you think? Are there other common error cases that we should address here? I'm okay with the string.

packages/amplify-console-hosting/hosting/manual/publish.js
```diff
@@ -9,7 +9,7 @@ const ora = require('ora');
 
 const ZIPPING_MESSAGE = 'Zipping artifacts.. ';
 const ZIPPING_SUCCESS_MESSAGE = 'Zipping artifacts completed.';
-const ZIPPING_FAILURE_MESSAGE = 'Zipping artifacts failed. This is often because your DistributionDir is configured to a nonexistent folder, e.g. "build" instead of "dist". Check /amplify/.config/project-config.json and fix accordingly.';
+const ZIPPING_FAILURE_MESSAGE = 'Zipping artifacts failed. This is often due to an invalid distribution directory path. Run "amplify configure project" to check if your Distribution Directory is pointing to a valid path.';
```
sw-yx: i will not make a fuss about this, but personally i prefer the one I wrote compared to yours. less repetitive, more informative, gives example, and resolution method is faster.

PR #5564, opened 2020-10-13
fix(amplify-provider-awscloudformation): apigw resource download

### LINKED ISSUE #5557: CLI is stuck and not working later when I add a REST api with a lambda function.
**Describe the bug**
My project uses AppSync and other services (pinpoint, aurora, cognito....)
Everything works fine.
I added a REST api with a lambda function (I wanted to generate a small website to validate some tokens sent to emails).
"amplify api add"
I add the new (the second one) api. With a lambda function. No API restrictions.
When I execute "amplify push", it seems the process works, but at some point, it is stuck forever:
Screenshot 2020-10-10 at 23 37 22
After 10 or 20 min, if I press COMMAND+C, I see this error:
Screenshot 2020-10-10 at 23 37 33
When I check Cloud Formation Stacks using AWS console, the new api and lambda are there, in fact, they are working fine.
However, when I edit my lambda function and press "amplify push" or "amplify status", then this is what I see:
Screenshot 2020-10-11 at 11 23 37
I checked that folder and I only see my appsync api.
Perhaps the issue is that the CLI does not support an AppSync and REST api at the same time, I don't know.

So, at this point I am completely stuck. Amplify does not work anymore.
Please, what do I do? I cannot continue working! And I cannot restore the previous state.

EDIT: I tried calling "amplify pull" in another computer but I receive this error (I don´t know if it is related).
Screenshot 2020-10-11 at 15 18 29

EDIT2:
**I tried to reproduce it in 2 empty projects I just created and I see the bug in both of them** (so, the issue is not in my project).
I recorded a video so you can follow the same steps.
Here the link:
https://www.dropbox.com/s/aojjorfz9rg1rfe/bug.mov?dl=0
(if you need a higher resolution video, let me know)
Basically, this is what you have in the video:
I setup an amplify project from scratch.
I add an AppSync api, with cognito (email).
I add pinpoint.
I add notifications (APNS with key)
calling “amplify push” works.
I add REST api with one lambda function.
When I call “amplify push”, it appears it works, but after a while, it is completely stuck.
You can see the same step/spinner for 10 minutes in the video.
Then I press COMMAND+C, I modify a lambda function and I try to upload everything again.
You can see the error saying a folder is empty and it is completely impossible to do anything.

Can you check this? My project is completely stuck due to this bug.
If you have any problem reproducing it, let me know.
Thanks a lot.

**Amplify CLI Version**
4.29.6

**To Reproduce**
(check the video which is in the EDIT2 section)
In my project happens all the time if I take the same steps:
Add rest api
With a lambda function
no api restrictions
(if I remove the api and the the lambda and try again, then it happens all the time)
I don't know if this happens in other projects.
BTW, my project uses appsync, pinpoint, aurora, lambda, cognito.... I don't know if the CLI is prepared to have 2 apis in the same project (appsync and rest). In any case, all the services are working.

**Expected behavior**
No stuck

**Screenshots**

 - OS: Mac Catalina 10.15.7 (19H2)
 - Node Version. v13.2.0

yuth:
Extract-zip package was updated to v2 which removed support for
callback based zip extraction. The API GW model download code was
not updated to adhere to new signature. Changed the method to use
the new extract zip signature.
To detect any issues in future updated the the module and dependecies
to Typescript to catch these erros at compile time.

fix #5557

*Issue #, if available:*

*Description of changes:*


packages/amplify-provider-awscloudformation/src/download-api-models.ts
```diff
@@ -29,41 +30,27 @@ function downloadAPIModels(context, allResources) {
...
-        });
-      });
-  });
+async function extractAPIModel(context: $TSContext, resource, framework) {
+  const apigw = await APIGateway.getInstance(context);
+  const apigwParams = getAPIGWRequestParams(context, resource, framework);
+
+  const apiName = resource.output.ApiName;
+
+  const data = await apigw.apigw.getSdk(apigwParams).promise();
```
SwaySway, line 39: Shouldn't it be the following?
```suggestion
  const data = await apigw.getSdk(apigwParams).promise();
```

yuth: APIGW class is just an wrapper which has instance of actual class. The only reason I changed the wrapper class was not create an async constructor of the class

SwaySway (approved):

PR #5586, opened 2020-10-14
Feat: Selective Sync query for DataStore

akshbhu:
*Description of changes:*

* Added Sync Resolver for which supports `@model` and `@key` directives and passing query parameters at runtime to VTL template

* Added tests 


packages/graphql-key-transformer/package.json
```diff
@@ -27,7 +27,8 @@
     "graphql-dynamodb-transformer": "6.19.14",
     "graphql-mapping-template": "4.15.2",
     "graphql-transformer-common": "4.17.10",
-    "graphql-transformer-core": "6.21.5"
+    "graphql-transformer-core": "6.21.5",
+    "ts-md5": "^1.2.7"
```
attilah, line 31: please use "md5" that we are already using in transformer.

packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -63,6 +66,7 @@ interface KeyArguments {
 }
 
 export class KeyTransformer extends Transformer {
+  SyncConfig?: SyncConfig;
```
attilah, line 69: if it is a private prop camelCase it

packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -132,6 +135,8 @@ export class KeyTransformer extends Transformer {
     const createResolver = ctx.getResource(ResolverResourceIDs.DynamoDBCreateResolverResourceID(definition.name.value));
     const updateResolver = ctx.getResource(ResolverResourceIDs.DynamoDBUpdateResolverResourceID(definition.name.value));
     const deleteResolver = ctx.getResource(ResolverResourceIDs.DynamoDBDeleteResolverResourceID(definition.name.value));
+    // update sync resolver if @key directive is present
+    const syncResolver = ctx.getResource(ResolverResourceIDs.SyncResolverResourceID(definition.name.value));
```
attilah, line 139: This could be moved down before first usage.

packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -963,3 +978,245 @@ function addCompositeSortKey(
...
+    set(ref('PkMap'), obj({})),
+    set(ref('PkTypeMap'), obj({})),
+    set(ref('SkTypeMap'), obj({})),
+    set(ref('PkType'), obj({})),
+    set(ref('SkType'), obj({})),
+  );
+  return block(`Set map initialization for @key`, expressions);
+}
+
+function setSyncKeyexpressionForHashKey(queryExprReference: string) {
```
attilah, line 1065: casing: setSyncKeyExpressionForHashKey

packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -963,3 +978,245 @@ function addCompositeSortKey(
...
+        set(ref(`${queryExprReference}.expression`), str(`#pk = :pk`)),
+        set(ref(`${queryExprReference}.expressionNames`), obj({ [`#pk`]: str('$pk') })),
+        set(ref(`${queryExprReference}.expressionValues`), obj({ [`:pk`]: obj({ $PkType: str(`$ctx.args.get($pk)`) }) })),
+      ]),
+    ),
+  );
+  return block(`Set Primary Key initialization @key`, expressions);
+}
+
+function setSyncKeyexpressionForRangeKey(queryExprReference: string) {
```
attilah, line 1081: casing here too.

packages/amplify-e2e-tests/src/schema-api-directives/tests/key-howTo4.ts
```diff
@@ -0,0 +1,257 @@
...
+          content: 'order1',
+          username: 'user1',
+          createdAt: '2015-01-01T01:05:49.129Z',
+          editor: 'akz',
+          data1: 'example1',
+          data2: 'example2',
+        },
+        {
+          content: 'order3',
+          username: 'akz',
```
attilah, line 236: user2 would do :-D no personal info in code

packages/graphql-transformer-core/package.json
```diff
@@ -28,7 +28,8 @@
     "fs-extra": "^8.1.0",
     "glob": "^7.1.6",
     "graphql": "^14.5.8",
-    "graphql-transformer-common": "4.17.10"
+    "graphql-transformer-common": "4.17.10",
+    "ts-md5": "^1.2.7"
```
attilah, line 32: same as at the top

packages/graphql-transformer-core/src/GraphQLTransform.ts
```diff
@@ -331,6 +332,20 @@ export class GraphQLTransform {
     context: TransformerContext,
   ) {
     let index = 0;
+
+    // hashing key directive to get the order in @key transformer
+    let ind = 0;
```
attilah, line 337: index, ind, can we come up with a better naming and less abbreviations? like "dir" down below, that increases the code readibility a lot.

attilah (changes requested):

packages/graphql-key-transformer/package.json
akshbhu: done!!

packages/graphql-key-transformer/src/KeyTransformer.ts
akshbhu: done!!

packages/amplify-e2e-tests/src/schema-api-directives/tests/key-howTo4.ts
akshbhu: Ahh , I was going to change that :D

packages/graphql-transformer-core/src/GraphQLTransform.ts
akshbhu: done

attilah (approved):

packages/graphql-transformer-core/src/GraphQLTransform.ts
```diff
@@ -331,6 +332,20 @@ export class GraphQLTransform {
     context: TransformerContext,
   ) {
     let index = 0;
+
+    // hashing key directive to get the order in @key transformer
+    let keyDirIndex = 0;
+    let syncKeyMap: Map<string | Int32Array, number> = new Map();
+    if (transformer.name === 'KeyTransformer') {
```
SwaySway, line 339: I'm not sure we should add the a direct dependency here to KeyTransformer in the the core package. Perhaps we could move the logic back into the key transformer and use the after function to create the final resolver for each model?

packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -963,3 +977,245 @@ function addCompositeSortKey(
...
+    ),
+  ]);
+}
+
+function makeSyncQueryResolver() {
+  const requestVariable = 'QueryRequest';
+  const expressions: Expression[] = [];
+  expressions.push(
+    ifElse(
+      raw('!$scan'),
```
SwaySway, line 1150: In the else statement it's still creating a sync operation payload.

packages/graphql-key-transformer/src/sync_vtl_remove/Query.syncComments.req.vtl
```diff
@@ -0,0 +1,164 @@
...
+                    && $ctx.args.sortDirection == "DESC" )
+    #set( $QueryRequest.scanIndexForward = false )
+  #else
+    #set( $QueryRequest.scanIndexForward = true )
+  #end
+#if( $context.args.nextToken ) #set( $QueryRequest.nextToken = $context.args.nextToken ) #end
+  #if( $filterMap != {} )
+    #set( $QueryRequest.filter = $util.parseJson($util.transform.toDynamoDBFilterExpression($filterMap)) )
+  #end
+  #if( $index != "dbTable" )
```
SwaySway, line 146: Is this check used to see if there is at least one defined index one which to query on?

packages/graphql-key-transformer/src/sync_vtl_remove/Query.syncComments.req.vtl
```diff
@@ -0,0 +1,164 @@
...
+## [Start] Set Primary Key initialization @key **
+#set( $modelQueryExpression = {} )
+#if( !$util.isNull($pk) )
+  #set( $modelQueryExpression.expression = "#pk = :pk" )
+  #set( $modelQueryExpression.expressionNames = {
+  "#pk": "$pk"
+} )
+  #set( $modelQueryExpression.expressionValues = {
+  ":pk": {
+      "$PkType": "$ctx.args.get($pk)"
```
SwaySway, line 84: Do we only need the PkType to see the expression? If so could we use the `$util.dynamodb.toDynamoDBJson()` utility here?
https://docs.aws.amazon.com/appsync/latest/devguide/resolver-util-reference.html#dynamodb-helpers-in-util-dynamodb

SwaySway (changes requested):
Left a few comments. I think we should reconsider how we are appending the sync key logic into the resolver.

packages/graphql-key-transformer/src/sync_vtl_remove/Query.syncComments.req.vtl
akshbhu: yes and if there is no index defined , it will remove the index from VTL and directly query on table

packages/graphql-key-transformer/src/KeyTransformer.ts
akshbhu: yeah , if no (pk,sk) matches in VTL , the operation will be a scan operation rather than query.

packages/graphql-transformer-core/src/GraphQLTransform.ts
akshbhu: done

packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -167,6 +196,11 @@ export class KeyTransformer extends Transformer {
           deleteResolver.Properties.RequestMappingTemplate,
         ]);
       }
+      if (syncResolver) {
```
SwaySway, line 199: nit: duplicate code here maybe we can store the result of `this.isPrimaryKeyDirective(directive)` in a variable and pass that into `constructSyncResolver`?

packages/graphql-key-transformer/src/sync_vtl_remove/Query.syncComments.req.vtl
SwaySway: nit: could we check here if the value of index is nullorempty as opposed to a default value? Do we still need to add DbTable Key/Value pairs in the PkMap?

packages/graphql-key-transformer/src/KeyTransformer.ts
SwaySway: Can we bubble up the logic on where to add query related key/value pairs above the payload? (such as queryExp, index, and scanIndexForward)

packages/graphql-key-transformer/src/KeyTransformer.ts
```diff
@@ -74,6 +76,31 @@ export class KeyTransformer extends Transformer {
...
+    if (resolverMap.size) {
+      resolverMap.forEach((syncVTLContent, resource) => {
+        if (syncVTLContent) {
+          resource.Properties.RequestMappingTemplate = '';
+          resource.Properties.RequestMappingTemplate = joinSnippets([
+            print(generateSyncResolverInit()),
+            syncVTLContent,
+            print(setSyncQueryFilterSnippet()),
+            print(setSyncKeyExpressionForHashKey(ResourceConstants.SNIPPETS.ModelQueryExpression)),
+            print(setSyncKeyExpressionForRangeKey(ResourceConstants.SNIPPETS.ModelQueryExpression)),
```
SwaySway, line 97: Do we need to add the Hash/Range Key Expressions on every Resolver Map?

SwaySway (changes requested):
Can you also include a snapshot test to check to test the generated resolver?

packages/graphql-key-transformer/src/KeyTransformer.ts
akshbhu: yeah , we need to add hash/range key expressions when @key directive is present

packages/graphql-key-transformer/src/sync_vtl_remove/Query.syncComments.req.vtl
akshbhu: I have to add some default value to make sure selected (pk,sk) tuple belongs to some table or GSI or they are just filter parameters. If the default value isn't present , Its difficult to tell whether its a query operation on table or just a scan on table

packages/graphql-key-transformer/src/KeyTransformer.ts
akshbhu: Its not duplicate. I am passing a istable boolean with value true/false depends on its for ddb table or GSI

SwaySway (approved):
Discussed Changes offline. LGTM

PR #5587, opened 2020-10-15
fix(graphql-transformer-core): "innerDiffs is not iterable" error on push

dsypniewski:
Issue: #5689

This change fixes a problem with "innerDiffs is not iterable" when there are changes to a schema but after diffing changed structures nothing is actually detected.
This can occur when for example when changing order of `@key` directives, all of the keys are the same and there is no change the only thing changing is their order in the schema.


edwardfoyle (approved):
LGTM! Thanks for the fix @dsypniewski

PR #5629, opened 2020-10-19
fix: ci undefined flag

ammarkarachi:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
+++ b/packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
@@ -3,19 +3,19 @@ import { Input } from '../input';
 import { getLatestPayloadVersion } from './VersionManager';
 import ci from 'ci-info';
 export class UsageDataPayload {
-  sessionUuid: String;
-  installationUuid: String;
-  amplifyCliVersion: String;
+  sessionUuid: string;
+  installationUuid: string;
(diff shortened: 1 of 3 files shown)


packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
```diff
@@ -14,7 +14,7 @@ export class UsageDataPayload {
   osRelease: String;
   nodeVersion: String;
   state: String;
-  isCi: Boolean | undefined;
+  isCi: Boolean;
```
yuth, line 17: Why are you using object type instead of primitive type?
```suggestion
  isCi: boolean;
```

ammarkarachi: I can move it to primitive type wouldn't affect the other end I believe

yuth: It should not. But was there any reason for  not choosing primitive type?

yuth (approved):

packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
ammarkarachi: Not really any reason but I looked back and using primitive is better practice

PR #5636, opened 2020-10-20
More descriptive error on env cmd

r0zar:
In a multi-backend workflow, if a teammate removes your backend/env from team-provider-info, amplify will fail on most env commands with a non-helpful error. This fixes that.

New CLI response:
image

*Issue #, if available:*

*Description of changes:*
On the failing code, this applies a more descriptive error message.

### DIFF
--- a/packages/amplify-cli/src/commands/pull.ts
+++ b/packages/amplify-cli/src/commands/pull.ts
@@ -3,6 +3,7 @@ import { attachBackend } from '../attach-backend';
 import { constructInputParams } from '../amplify-service-helper';
 import { run as envCheckout } from './env/checkout';
 import { stateManager } from 'amplify-cli-core';
+import { get } from 'lodash';
 
 export const run = async context => {
   const inputParams = constructInputParams(context);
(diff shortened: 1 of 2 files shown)


r0zar:
WIP: Currently works for `amplify env pull` need to enable for `amplify pull` as well.

sw-yx (approved):

renebrandel:
Really nice! I wonder if explicitly mentioning `team-provider-info.json` is revealing too much of an implementation detail. What do you think of `${env} environment not found. Try running "amplify env add" to add a new environment."

This way there is a recovery step for the customer.

r0zar:
I agree. I was also thinking about adding a recovery step as well, ill add that.

r0zar:
@renebrandel I was hoping there would be a way to update the `team-provider-info.json` file based on a backend I have defined. but in this scenario, `amplify pull` and `amplify env pull` both throw this error. How would I go about "re-attaching" my backend if it's been removed from the `team-provider-info.json` file?

renebrandel (approved):

r0zar:
@renebrandel I know you recommended to not expose implementation details. In testing resolutions for this issue, the only solution  I found (to re-sync the local backend with the cloud backend) was to manually update `team-provider-info.json` with the missing config/definition. Let me know if you know of a better way to give the user the information they need to fix the issue without being overly specific! thanks! :+1:

PR #5334, opened 2020-09-15
fix: changed default config to user defined config

akshbhu:
*Issue #5226 

*Description of changes:*

* Checks for the project config for sourceDir 

* Added tests for the same

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5654, opened 2020-10-21
fix(amplify-codegen): sourceDir path issues in graphql schema location

### LINKED ISSUE #5483: Amplify push fails with error creating directory
**Describe the bug**
After adding api via amplify add api, then trying to push changes to cloud using amplify push, I receive following error at the time of adding resources to local folder

✖ An error occurred when pushing the resources to the cloud

EROFS: read-only file system, mkdir '/graphql'
An error occurred during the push operation: EROFS: read-only file system, mkdir '/graphql'

I am on Mac. My previous CLI version was worked perfectly fine. Latest version giving this problem.
I am doing this under fresh React Native CLI project. 

**Amplify CLI Version**
4.29.4

**To Reproduce**
Steps to reproduce the behavior or terminal output if applicable

**Expected behavior**

**Screenshots**

 - OS: Mac
 - Node Version. v12.16.2

**Additional context**

wei:
*Issue #, if available:*
Resolves #5483

*Description of changes:*
#5334 c63014558547f75a0afa7e506116d8bb124c44d2 changed `amplify-codegen` from hard coding `src` to using `SourceDir` in the config file. 
This resulted in the default sourceDir value `/` (system root) for reactnative being used. This PR corrects the default dir values to `./`.

The commit/bug in question affects releases starting from CLI version `4.29.4`.

**UPDATE:** the root cause has been identified to be in `doPushCallback.js` where the GraphQL schema location did not use `path.join` with project root. The PR has been updated to make the correction.

*Other:*
Docs PR: aws-amplify/docs#2581

### DIFF
--- a/packages/amplify-app/src/framework-config-mapping.js
+++ b/packages/amplify-app/src/framework-config-mapping.js
@@ -8,8 +8,8 @@ const reactConfig = {
 };
 
 const reactNativeConfig = {
-  SourceDir: '/',
-  DistributionDir: '/',
+  SourceDir: './',
+  DistributionDir: './',
   BuildCommand: `${npm} run-script build`,
   StartCommand: `${npm} run-script start`,
 };
@@ -36,7 +36,7 @@ const vueConfig = {
 };
 
 const emberConfig = {
-  SourceDir: '/',
+  SourceDir: './',
   DistributionDir: 'dist',
   BuildCommand: `${npm} run-script build -- -e production`,
   StartCommand: `${npm} run-script start`,
--- a/packages/amplify-codegen/src/callbacks/postPushCallback.js
+++ b/packages/amplify-codegen/src/callbacks/postPushCallback.js
@@ -1,3 +1,6 @@
+const path = require('path');
+const { pathManager } = require('amplify-cli-core');
+
 const loadConfig = require('../codegen-config');
 const generateStatements = require('../commands/statements');
 const generateTypes = require('../commands/types');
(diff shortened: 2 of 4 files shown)


UnleashedMind:
I don't think this is the proper fix for the issue.
The SourceDir and DistributionDir are meant to be used with projectPath like this https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-frontend-javascript/lib/frontend-config-creator.js#L49

wei:
@UnleashedMind In that case, 

https://github.com/aws-amplify/amplify-cli/blob/b624e0fff58659d0aeb13bc3e79b437071295aa3/packages/amplify-codegen/src/commands/models.js#L148

needs to be changed to 

```javascript
 const modelFolder = path.join(projectPath, getModelOutputPath(context), 'models'); 
```

Would this be an acceptable fix instead of changing the default values? At this moment, users can use `./` to circumvent the issue which has been proven and added to docs.

wei:
@UnleashedMind In that case, 

https://github.com/aws-amplify/amplify-cli/blob/c63014558547f75a0afa7e506116d8bb124c44d2/packages/amplify-codegen/src/utils/getSchemaDownloadLocation.js#L6-L22

^ This function will return `/graphql/models.json`

and 

https://github.com/aws-amplify/amplify-cli/blob/b624e0fff58659d0aeb13bc3e79b437071295aa3/packages/amplify-codegen/src/callbacks/postPushCallback.js#L8-L21

Is not joining `projectPath` with it resulting in the graphql config's schema to point to `/graphql/models.json` (system root). 

Would adding `path.join(projectPath, schemaLocation)` in the second snippet above be an acceptable fix? As of now, users can use `./` to circumvent the issue which has been proven and added to docs. 

Perhaps we can still also change the default `/` to `./` because `path.join(projectPath, '/')` yields the same result as `path.join(projectPath, './')` but the latter prevents any confusion with system root. 

Please advise which fix you'd prefer. Thank you!

UnleashedMind:
IMHO, we should use path.join to ensure that the path is correctly set. 
Using './' directly without path join with the project root does not guarantee we will also get to the actual source or build directory.

wei:
@UnleashedMind I agree we need to update the snippet I included to use path.join

I was just asking if you are fine with us also change / to ./ for clarity.

wei:
@UnleashedMind This PR has been updated to add path.join to the graphql schema location which was previous missing. I've also updated unit test to match the correct behavior. Please review.

mauerbac:
@kaustavghosh06 - can you review this please?

yuth:
closing this in favor of https://github.com/aws-amplify/amplify-cli/pull/6512

PR #5705, opened 2020-10-26
fix(amplify-appsync-simulator): appsync scalars

### LINKED ISSUE #5207: Value returned by amplify-appsync-simulator for AWSJSON fields doesn't match AppSync
Currently 'amplify-appsync-simulator' uses 'graphql-type-json' to represent AWSJSON type. 'graphql-type-json' sets `identity` as a function to parse and serialize value.

When you query field of AWSJSON type from AppSync, response contains string value with serialized JSON.
'amplify-appsync-simulator' returns parsed value (object in my use case) in same situation.

**Amplify CLI Version**
I'm using "amplify-appsync-simulator@1.23.2" through "serverless-appsync-simulator" and don't have "amplify-cli" installed.

**To Reproduce**
Test AppSync locally with schema, that contains fields of AWSJSON type.

**Expected behavior**
Strings with serialized JSON are returned to client for AWSJSON fields.

 - OS: Mac
 - Node Version. v12.18.0

bboure:
*Issue #, if available:*
Fixes #5207

*Description of changes:*

Several issues here.

1) AWS scalars were not used at all
AWS scalars were defined, but never injected in the schema. This is now fixed.
**Please let me know if this was intentional** (for some reason)

2) `AWSPhone` was needs to be instantiated in order to be injected into the schema.
Instead, the class itself was passed. 
`AWSPhone` also seems to accept [some options](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-appsync-simulator/src/schema/appsync-scalars/index.ts#L13) for the country. Since this was not used at all until now, I just left the defaults.

3) `AWSJSON`
Finally, this fixes #5207 
AWSJSON are stringified in AppSync but were returned as objects in the simulator. 

According to [the doc](https://docs.aws.amazon.com/appsync/latest/devguide/scalars.html) and the reverse engineering I have been doing, it should act as follow:

**As an Input**
JSON strings are parsed as values before getting to mapping template. This means:
`"{\"Foo\":\"Bar\"}"` becomes `{Foo: "Bar"}` (object, or `Map`)
`"true"` (the string) becomes `true` (the boolean)
`"1"` (the string) becomes `1` (the integer)
`"\"Hello\""` (double quotes are important) becomes the `Hello` string
etc.

Any array of the above should also work. Example: `"[\"Hello\", \"World\"]"`

`"Hello"` is an invalid json representation and fails.
Any non-string value should fail.

**As an output**
Basically, any output will be encoded as a JSON
`{Foo: "Bar"}` becomes `{"Foo": "Bar"}`, and so on... (Basically, the reverse of the above)

### DIFF
--- a/packages/amplify-appsync-simulator/package.json
+++ b/packages/amplify-appsync-simulator/package.json
@@ -76,7 +76,7 @@
       "!**/*.d.ts"
     ],
     "testURL": "http://localhost/",
-    "testRegex": "(src/__tests__/.*.test.*)$",
+    "testRegex": "(src/__tests__/.*.test.ts)$",
     "moduleFileExtensions": [
       "ts",
       "tsx",
--- a/packages/amplify-appsync-simulator/src/__tests__/scalars/AWSJSON.test.ts
+++ b/packages/amplify-appsync-simulator/src/__tests__/scalars/AWSJSON.test.ts
@@ -1,0 +1,62 @@
+import { scalars } from '../../schema/appsync-scalars';
+
+describe('AWSJSON parse', () => {
+  it('Should reject an invalid json encoded object', () => {
+    function parse() {
+      scalars.AWSJSON.parseValue('{Foo: "Not Valid"}');
+    }
+    expect(parse).toThrowErrorMatchingSnapshot();
+  });
+  it('Should reject a non json string', () => {
+    function parse() {
+      scalars.AWSJSON.parseValue('hello world');
+    }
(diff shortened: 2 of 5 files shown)


edwardfoyle (approved):

SwaySway (approved):
LGTM - thanks for the contribution!

ElvenMonky:
@bboure Thanks a lot for figuring this out!

PR #5783, opened 2020-11-05
fix: do not attempt to modify a file when reading it

ankon:
*Issue #, if available:*

*Description of changes:*

The configuration files can exist on a read-only mountpoint, and attempting to change the permissions
will fail then _even if the permissions are fine as they are_.

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/system-config-manager.js
+++ b/packages/amplify-provider-awscloudformation/src/system-config-manager.js
@@ -245,7 +245,6 @@ function getCacheFilePath(context) {
 function getProfileConfig(profileName) {
   let profileConfig;
   if (fs.existsSync(configFilePath)) {
-    makeFileOwnerReadWrite(configFilePath);
     const config = ini.parse(fs.readFileSync(configFilePath, 'utf-8'));
     Object.keys(config).forEach(key => {
       const keyName = key.replace('profile', '').trim();
@@ -260,7 +259,6 @@ function getProfileConfig(profileName) {
 function getProfileCredentials(profileName) {
   let profileCredentials;
   if (fs.existsSync(credentialsFilePath)) {
-    makeFileOwnerReadWrite(credentialsFilePath);
     const credentials = ini.parse(fs.readFileSync(credentialsFilePath, 'utf-8'));
 
     Object.keys(credentials).forEach(key => {
@@ -299,7 +297,6 @@ function getProfileRegion(profileName) {
 function getNamedProfiles() {
   let namedProfiles;
   if (fs.existsSync(configFilePath)) {
-    makeFileOwnerReadWrite(configFilePath);
     const config = ini.parse(fs.readFileSync(configFilePath, 'utf-8'));
     namedProfiles = {};
     Object.keys(config).forEach(key => {


ankon:
FWIW: This was introduced with https://github.com/aws-amplify/amplify-cli/pull/5222, but I don't seem to able to find an actual problem description of why this change was needed.

ammarkarachi:
@ankon I agree with your sentiment about not changing the mode if it is already read-only but removing `makeFileOwnerReadWrite` is not a solution I see that would work for all our customers. If you would like to submit a solution where it ignores setting the mode to `rw` if the mode is already at `r` we can work on merging that fix.

ankon:
Thanks for your reply!

>  but removing makeFileOwnerReadWrite is not a solution I see that would work for all our customers. 

I'm sure you know a lot more about the different cases for your customers, could you maybe explain the underlying concern/problem that the users hit? I understand that when you're _modifying_ the file you do want to ensure the mode is proper, and I could also follow with a warning to be issued when reading and the mode is improper -- but changing the mode while _reading_ sounds wrong.

Note that this PR does leave the call in place for the _modifying_ case, it only removes the additional places where it is called from a _reading_ path.

UnleashedMind:
@ankon 
Thanks for your contribution. 
I see the point you brought up, but I share the opinion of @ammarkarachi.

Below are the requirements for the permissions for the two files:
1. The permissions to access those two files should only be granted to owner
2. Read only mode for the `<home>/.aws/config` and `<home>/.aws/credentials` files does not work well with the Amplify CLI, the CLI needs the permission to update the profiles in those files. 

The makeFileOwnerReadWrite calls when reading the files is the Amplify CLI's attempt to ensure point 1 whenever those files are accessed.

ankon:
> Read only mode for the <home>/.aws/config and <home>/.aws/credentials files does not work well with the Amplify CLI, the CLI needs the permission to update the profiles in those files.

Why does it need that _when reading the files_? 

Yes, when it is _updating the files_, it certainly should ensure nice permissions, and it obviously needs to be able to modify them.  

Some reasons why I believe these calls shouldn't be there:
* From a very strict point of view of security: You need permission to write to these files, but even giving you permissions to modify the attributes of the files is questionable.
* From a developer point of view: Setting the permissions when reading, knowing that you will "soon" need them because probably you're going to write to them is pretty much a classical ToCToU race condition anyways. If I want to crash the Amplify CLI through this, I can very likely win that race.
* For a precedent: Observe how an ssh client would also be unhappy with the permissions on it's config files, but it will not change them -- it will tell you what's wrong.
* From a OS point of view: Observe that the _permissions_ on the file in my example are correct ("rw-" for owner), but that the ability to write to the file is restricted on a level you will not be able to change from the CLI code.

But, let's see whether we can find some middleground here. My minimal requirement is that the Amplify CLI doesn't crash when it cannot modify permissions, and I'm ok with not using certain functionality that requires writing to the files (for example `amplify configure`). I do want to see reasonable messages that a problem exists when I (accidentally/unintentionally/mistakenly) invoke these functionalities.

From your comments I understand your requirements seem to be that the CLI doesn't crash "too late" (at the point of actually writing), and rather checks as early as possible that everything is ok and will work. 

What about this then:
1. For reading and writing: Check the required permission, and only try to modify if it is missing (i.e. when trying to read check for read access, when trying to write check for write access). 
2. For reading and writing: If the "other" permission is missing, report that as a warning, but proceed. For instance when trying to read and there's a read, but not a write permission, report a warning that the write permission is missing, but proceed.

For my requirements: This would be more code to implement, but practically it would be working perfectly for me (because the permissions are actually what the CLI wants, I wouldn't even see the warning!).

Would this work for your requirements, or am I missing any?

jhockett (approved):
LGTM, cc @ammarkarachi 

The CLI has a more robust check for this that runs on every amplify command, but only attempts to change the profile's file mode if it's set incorrectly to default:
https://github.com/aws-amplify/amplify-cli/blob/75f5ace173a6b36b943e2110845e411a2cce5d6d/packages/amplify-cli/src/index.ts#L182

ankon:
@jhockett just to verify that I understand correctly, the check you point to tries to catch a situation where a user created the file with the default umask setting of 0o022 and a presumed default file mode of 0o666/`rw-rw-rw`, i.e. a file that ends up with `rw-r--r--`, and in that case it attempts to change the permissions to `rw-------`?

My file has permissions is 0o0700/`rwx------` already, so this then would not attempt any modification. That's good :)

But, given that the amplify CLI main parts already verify things, is it then not even more reasonable to not further touch the file _at all_ in the cloudformation provider?

jhockett:
> But, given that the amplify CLI main parts already verify things, is it then not even more reasonable to not further touch the file _at all_ in the cloudformation provider?

The provider needs to be able to create profiles if a new user wants one, I don't see why that should change.

Issue #5782, opened 2020-11-05
Allow skipping the AWS CLI profile creation in `configure`


ankon:
**Is your feature request related to a problem? Please describe.**

I'm having many AWS accounts to manage, and therefore already many profiles. For security reasons I have my AWS configuration (`~/.aws` directory) mounted read-only, and I prefer to not have arbitrary tools modify these. Instead _I_ manage the settings, and would just select the correct profile explicitly whenever using any AWS-related tool.

I'm right now following the AWS Amplify tutorial, and the `amplify configure` just did a couple of things I have issues with:
1. It expects to be fed the credentials. Given that this is a new tool (at least for me), and installed from npm without any specific version or anything, I rather not do that.
2. It also expects that it can modify the ~/.aws/credentials (and probably other files), and will crash in half-finished state (probably) when it can't:
    ~~~~
    Enter the access key of the newly created user:
    ? accessKeyId:  ********************
    ? secretAccessKey:  ****************************************
    This would update/create the AWS Profile in your local machine
    ? Profile Name:  amplify-tutorial
    EROFS: read-only file system, chmod '.../.aws/credentials'
    Error: EROFS: read-only file system, chmod '.../.aws/credentials'
        at Object.chmodSync (fs.js:1205:3)
        at Object.chmodSync (.../modules/nvs/node/12.19.0/x64/lib/node_modules/@aws-amplify/cli/node_modules/graceful-fs/polyfills.js:252:21)
        at makeFileOwnerReadWrite (.../modules/nvs/node/12.19.0/x64/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/src/system-config-manager.js:98:6)
        at Object.setProfile (.../modules/nvs/node/12.19.0/x64/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/src/system-config-manager.js:19:5)
        at Object.run (.../modules/nvs/node/12.19.0/x64/lib/node_modules/@aws-amplify/cli/node_modules/amplify-provider-awscloudformation/src/setup-new-user.js:113:25)
        at processTicksAndRejections (internal/process/task_queues.js:97:5)
    ~~~~

**Describe the solution you'd like**

1. Minimally: Handle the EROFS, by providing a "skip", or "try again" option.
2. Ideally: Allow me to enter the name of the profile to use instead of walking through creating a user.

[...]

PR #5784, opened 2020-11-05
feat: use existing profile if provided

ankon:
*Issue #, if available:*
Relates-to: https://github.com/aws-amplify/amplify-cli/issues/5782

*Description of changes:*

This is a _RFC_ for how an existing profile could be used. This will still do the fairly useless inquiries to work out which page in the console to open, which isn't really needed when a profile already exists.
I think it might actually make more sense to create a second provider (say "Existing profile", and then name the existing one "Create new IAM user"), to quickly step through things.

WDYT? Is a separate provider a better approach, or would you rather have a single provider that can handle both situations? 

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/setup-new-user.js
+++ b/packages/amplify-provider-awscloudformation/src/setup-new-user.js
@@ -49,75 +49,79 @@ async function run(context) {
   open(deepLinkURL, { wait: false }).catch(() => {});
   await context.amplify.pressEnterToContinue.run({ message: 'Press Enter to continue' });
 
-  context.print.info('Enter the access key of the newly created user:');
-  const accountDetails = await inquirer.prompt([
+  context.print.info('Configure the profile to use for the new user:');
+  const profileDetails = await inquirer.prompt([
     {
-      type: 'password',
-      mask: '*',
-      name: 'accessKeyId',
-      message: 'accessKeyId: ',
-      default: awsConfig.accessKeyId,
-      transformer: obfuscationUtil.transform,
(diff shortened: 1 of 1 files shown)


UnleashedMind:
@ankon Thank you for your contribution. 
IMHO, if there is already an IAM user whose credentials you want to use to configure the Amplify CLI through a profile, you can directly create a [profile](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html) in the `<home>/.aws/config` and `<home>/.aws/credentials` files with the user's credentials. The current awscloudformation provider will find it and allow you to select that profile. We don't need another provider. 

The  `setup-new-user` workflow is used by the `amplify configure` command or automatically invoked when the CLI cannot resolve aws configuration. It is meant to be used to conveniently set up new IAM user and create a profile on a fresh machine for new Amplify CLI users.

ankon:
> The setup-new-user workflow is used by the amplify configure command or automatically invoked when the CLI cannot resolve aws configuration. It is meant to be used to conveniently set up new IAM user and create a profile on a fresh machine for new Amplify CLI users.

Thanks, fair enough. As mentioned in the referenced issues, it does look more like a request to improve the specific tutorial I was following to make it clear that IFF I have a (suitable) user and profile available, THEN I can skip the `aws configure` call.

UnleashedMind:
@ankon 
Thanks, I will talk to our team and see how we can make this particular workflow clearly understood by our customers possibly through documentation or tutorials. So, I'd suggest to close this PR in favor of documentation or tutorials.

PR #5797, opened 2020-11-06
fix(amplify-appsync-simulator): utils.toJson false value

bboure:
*Description of changes:*

`$utils.toJson(false)` should return `false` (as a json). Currently, it returns `null`.

I also added some tests


attilah (approved):
Thanks for the contribution! LGTM!

Issue #1913, opened 2019-07-27
s3Trigger lambda function inline policy getting overwritten


onlybakam:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[official AWS Amplify Console forum](https://forums.aws.amazon.com/forum.jspa?forumID=314&start=0)

**Describe the bug**

1. the inline policy name for the S3TriggerBucketPolicy varies based on when the trigger is applied. when trigger is applied to existing bucket, "s3-trigger-lambda-execution-policy" is used. if not "amplify-lambda-execution-policy" is used. (see logic here: https://github.com/aws-amplify/amplify-cli/blob/54a8dbe8925a4e73358b03ba927267a2df328b78/packages/amplify-category-storage/provider-utils/awscloudformation/service-walkthroughs/s3-walkthrough.js#L418)
2. "amplify-lambda-execution-policy" is the name of the inline policy used for all trigger cases. if you try to grant access to multiples resources, the inline policy just gets overwritten. e.g.: if you want your s3Trigger lambda function to have access to the API, the inline policy granting access to the API ends up getting overwritten. (first the role is created by the function cfn template, then the role is updated by the storage cfn which can overwrite the existing inline function)

**To Reproduce**
1. amp add storage with trigger
2. amp add api
3. amp update function, grant API access
4. result is that storage cfn will overwrite the function cfn inline prolicy

this does not happen if you define the trigger after the bucket is created

Also note about "S3TriggerBucketPolicy": technically it is not a bucket policy since not applied to an S3 bucket; but to a role.

jkahn117:
Confirmed, just experienced same behavior.

Added Amplify Storage, the trigger function has GetObject, ListBucket, PutObject, and DeleteObject permissions on the bucket created.

Function needs access to the bucket and GraphQL API created by Amplify, so run `amplify function update` and select appropriate permissions for API category. Resulting policy covers AppSync only, S3 permissions are lost.

On attempting to add S3 permissions back by running `amplify function update` again and selecting Storage, receive an error message: "Error: Cannot add S3TriggerXXXXXXXX due to a cyclic dependency." Lambda environment variables are updated, but contain default values.

kaustavghosh06:
[...]

PR #5814, opened 2020-11-09
fix: use non-duplicate names for lambda execution role

toelke:
Hi all,

we have noticed that adding permissions to a lambda function that is triggered by another resource (e.g. an S3 bucket) will *remove* the permissions that lambda has on the triggering resource -- and you cannot re-add those permissions as that will introduce a cyclic dependency (the storage-stack depends on the function and the function-stack depends on the bucket). This is the behavior described in #1913 as well.

This behaviour is caused by using the same name `amplify-lambda-execution-policy` for both policies generated by the triggering resource and when adding permissions to the lambda.

I have solved that for now by changing the name of the policy generated by the triggering resource.

For now I have doubts about the following references to the name `amplify-lambda-execution-policy`:

* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/CreateAuthChallenge/cloudformation-templates/CreateAuthChallenge.json.ejs:175`
* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/CustomMessage/cloudformation-templates/CustomMessage.json.ejs:192`
* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/DefineAuthChallenge/cloudformation-templates/DefineAuthChallenge.json.ejs:168`
* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostAuthentication/cloudformation-templates/PostAuthentication.json.ejs:168`
* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostConfirmation/cloudformation-templates/PostConfirmation.json.ejs:175`
* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PreAuthentication/cloudformation-templates/PreAuthentication.json.ejs:168`
* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PreSignup/cloudformation-templates/PreSignup.json.ejs:182`
* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PreTokenGeneration/cloudformation-templates/PreTokenGeneration.json.ejs:168`
* `packages/amplify-category-auth/provider-utils/awscloudformation/triggers/VerifyAuthChallengeResponse/cloudformation-templates/VerifyAuthChallengeResponse.json.ejs:176`

Even stranger is the reference to `amplify-lambda-execution-policy-<%= props.triggerEventSourceMappings[i].modelName%>` to which I can find no resource actually creating it.

We think that this PR solves a *real* problem with amplify in an easy way:

image

Cheers,
Philipp

### DIFF
--- a/packages/amplify-category-predictions/provider-utils/awscloudformation/cloudformation-templates/s3-cloudformation-template.json.ejs
+++ b/packages/amplify-category-predictions/provider-utils/awscloudformation/cloudformation-templates/s3-cloudformation-template.json.ejs
@@ -306,7 +306,7 @@
 			],
             "Type": "AWS::IAM::Policy",
             "Properties": {
-                "PolicyName": "amplify-lambda-execution-policy",
+                "PolicyName": "amplify-lambda-execution-policy-prediction",
                 "Roles": [
                     {
                         "Ref": "function<%= props.adminTriggerFunction %>LambdaExecutionRole"
--- a/packages/amplify-category-storage/resources/cloudformation-templates/s3-cloudformation-template.json.ejs
+++ b/packages/amplify-category-storage/resources/cloudformation-templates/s3-cloudformation-template.json.ejs
@@ -369,7 +369,7 @@
             ],
             "Type": "AWS::IAM::Policy",
             "Properties": {
-                "PolicyName": "amplify-lambda-execution-policy",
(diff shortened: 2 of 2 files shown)


edwardfoyle (approved):
Hi @toelke thanks for the contribution! This change looks good to me!

PR #5384, opened 2020-09-23
Load session using standard boto3 process

pecigonzalo:
This changes the script to use the normal boto3 process of loading creds (file, introspective api, env vars) if no explicit credentials are passed.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5824, opened 2020-11-10
fix: backfill script - allow ak and sk arguments to be read when initiating lambda client

chantlong:
*Issue #, if available:*
No Issue.

*Description of changes:*
I use 2 different AWS Account IDs to manage by DEV / PROD environments.

Script would fail on the following conditions.
1. When I provide access key and secret key as an argument.
2. I am logged into my AWS CLI with DEV AWS Account.

On L31, when it initializes the client, it doesn't take my access key / secret key as arguments, so even if I am running the script on PROD, I would run to an error on L85 saying my DEV Lambda doesn't have to right to perform an action on a PROD thing.

```
lambda_response = client.invoke(
      FunctionName=lambda_f,
      Payload=records
  )
```

This fix will allow access key and secret key arguments to be read.

### DIFF
--- a/packages/graphql-elasticsearch-transformer/scripts/ddb_to_es.py
+++ b/packages/graphql-elasticsearch-transformer/scripts/ddb_to_es.py
@@ -28,7 +28,7 @@ def main():
     args.sk = args.sk or credentials.secret_key
     args.ak = args.ak or credentials.access_key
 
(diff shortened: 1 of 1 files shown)


ammarkarachi:
Closing in favor of https://github.com/aws-amplify/amplify-cli/pull/5384/

Issue #5358, opened 2020-09-18
modelgen bug with Has One @connection relation with an explicit field 


iartemiev:
**Describe the bug**
Related to this issue in JS: https://github.com/aws-amplify/amplify-js/issues/6560

`amplify codegen models` is incorrectly handling Has One relations. The `associatedWith` field is not being set to the field specified in the `fields` argument.

The following model:
```graphql
type Project @model {
  id: ID!
  name: String
  teamID: ID!
  team: Team @connection(fields: ["teamID"])
}
```

Gets modelgen'd to
```js
"Project": {
    "name": "Project",
    "fields": {
        "id": {
            "name": "id",
            "isArray": false,
            "type": "ID",
            "isRequired": true,
            "attributes": []
        },
        "name": {
            "name": "name",
            "isArray": false,
            "type": "String",
            "isRequired": false,
            "attributes": []
        },
        "teamID": {
            "name": "teamID",
            "isArray": false,
            "type": "ID",
            "isRequired": true,
            "attributes": []
        },
        "team": {
            "name": "team",
            "isArray": false,
            "type": {
                "model": "Team"
            },
            "isRequired": false,
            "attributes": [],
            "association": {
                "connectionType": "HAS_ONE",
                "associatedWith": "id"
            }
        }
    },
    "syncable": true,
    "pluralName": "Projects",
    "attributes": [
        {
            "type": "model",
            "properties": {}
        }
    ]
},
```

**Amplify CLI Version**
`4.29.2`

**To Reproduce**
Run `amplify codegen models` on the following schema:
```graphql
type Project @model {
  id: ID!
  name: String
  teamID: ID!
  team: Team @connection(fields: ["teamID"])
}

type Team @model {
  id: ID!
  name: String!
}
```

**Expected behavior**
I would expect that `team: Team @connection(fields: ["teamID"])` should result in:

```js
"team": {
    "name": "team",
    "isArray": false,
    "type": {
        "model": "Team"
    },
    "isRequired": false,
    "attributes": [],
    "association": {
        "connectionType": "HAS_ONE",
        "associatedWith": "teamID"
    }
}
```

AaronZyLee:
[...]

PR #5827, opened 2020-11-10
fix: @connection when field is given

akshbhu:
*Issue #5358 

*Description of changes:*

* added case for @coonection directive when field argument is present.

* added test for the same

### DIFF
--- a/packages/amplify-codegen-appsync-model-plugin/src/utils/process-connections.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/utils/process-connections.ts
@@ -164,6 +164,7 @@ export function processConnections(
             person: Person!
           }
           */
+
           return {
             kind: CodeGenConnectionType.BELONGS_TO,
             connectedModel: otherSide,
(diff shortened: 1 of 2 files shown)


packages/amplify-codegen-appsync-model-plugin/src/utils/process-connections.ts
```diff
@@ -181,12 +182,24 @@ export function processConnections(
...
+            case 1:
+              return {
+                kind: CodeGenConnectionType.BELONGS_TO,
+                connectedModel: otherSide,
+                isConnectingFieldAutoCreated,
+                targetName: connectionFields[0] || makeConnectionAttributeName(model.name, field.name),
+              };
+            default:
+              throw new Error('DataStore only support one key in field');
+          }
```
yuth, line 202: The check should be based on if the field is nullable or non-nullable. When an field is non-nullable, it beconmes belogns to and when it is nullable it becomes `hasOne`

yuth (changes requested):

packages/amplify-codegen-appsync-model-plugin/src/utils/process-connections.ts
akshbhu: yes , thats true, But What about the case when @connection directive has a field which is non nullable. for example 
```graphql
          type Person @model {
            licenseId: ID!
            license: License @connection(fields: ["licenseId"])
          }
          type License @model {
            id: ID!
            holder: Person! @connection
          }
```
Here licence Id is non nullable . Wont this mean every licence(LicenceID) belongs to a person?

yuth: Logically doesn't it make sense. For a license to exists, there needs to be a person. So a person has a license and a license belongs to a person

akshbhu:
Closing in favour of #6031

PR #5775, opened 2020-11-04
Merge flutter frontend to modelgen staging branch

AaronZyLee:
*Issue #, if available:*

*Description of changes:*
Merge the flutter frontend plugin to modelgen staging branch so that developers can add flutter type frontend and run test on modelgen.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #5849, opened 2020-11-12
feat(amplify-frontend-flutter)

haverchuck:
*Description of changes:*
Adds developer preview functionality for amplify-flutter projects.


packages/amplify-cli/scripts/post-install.js
```diff
@@ -25,6 +25,8 @@ console.log(chalk.green('Android Getting Started - https://docs.amplify.aws/star
 console.log(EOL);
 console.log(chalk.green('iOS Getting Started - https://docs.amplify.aws/start/q/integration/ios'));
 console.log(EOL);
+console.log(chalk.green('Flutter Getting Started - https://docs.amplify.aws/start/q/integration/flutter'));
```
renebrandel, line 28: Flutter (Developer Preview)

haverchuck: Done!

attilah (changes requested):
I reviewed the other Flutter PR, those comments should be addressed here then ;-)

haverchuck:
@attilah I'm assuming you mean this one:  https://github.com/aws-amplify/amplify-cli/pull/5775

attilah:
@haverchuck yes.

haverchuck:
@attilah I answerd your inline comments in @AaronZyLee 's  initial PR.  I made several of your requested revisions.

You also had two general comments:

> * Put all the Flutter related E2E tests into 1 test file within the E2E project to make separation easier later and don't spread across categories.
> * Reorg the plugin source code into `src` folder with `lib` output and add TypeScript passthrough compilation, like we have in other some packages like `amplify-provider-awscloudformation` as an example.

Regarding the first bullet, I decided not to do this because none of the other test files appear organzied by platform.  I think it would be better to reorganzie all of them if needed.

Regarding bullet 2 - as we discussed, this would potentially be a breaking change due to the need to upgrade the graphql-config library to version 3.  This upgrade would break our amplify-codegen package, since the new graphql-config libary uses "include/exclude" instead of "includes/excludes".

attilah (approved):
LGTM, thanks for the changes, as we discussed, I understand the TS concerns, we can address it later for all the packages in 1 go.

PR #5916, opened 2020-11-20
fix(amplify-category-auth): external auth enabled bugfix

elorzafe:
`externalAuthEnabled` was returning an incorrect value

*Issue #, if available:*
N/A
*Description of changes:*
Updating return value for `externalAuthEnabled` function


edwardfoyle (approved):

ericclemmons (approved):
Confirmed this resolved the bug I was experiencing locally with `amplify-dev push`.

You can see that `authProps.resourceName` is used above in line 137, 125 successfully.

PR #5918, opened 2020-11-20
fix: changing pinpoint policy

nikhname:
*Issue #, if available:*

*Description of changes:*
Change in pinpoint policy actions based on discussion

### DIFF
--- a/packages/amplify-category-analytics/provider-utils/awscloudformation/cloudformation-templates/pinpoint-cloudformation-template.json
+++ b/packages/amplify-category-analytics/provider-utils/awscloudformation/cloudformation-templates/pinpoint-cloudformation-template.json
@@ -285,7 +285,7 @@
           "Statement": [
             {
               "Effect": "Allow",
-              "Action": ["mobiletargeting:PutEvents", "mobiletargeting:UpdateEndpoint", "mobiletargeting:GetUserEndpoints"],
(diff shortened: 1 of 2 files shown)


ammarkarachi (approved):

palpatim:
@nikhname, @kaustavghosh06 -- This change has caused customer issues on the client side, because customers can no longer remove endpoints once they hit the 10 endpoint limit. What is the expectation for customers to be able to remove endpoints?

kayuapi:
+1, facing same issues as @palpatim

bakalar:
Having the same issue. Why was getUserEndpoints removed?

lawmicha:
this appears to be a corresponding JS library change https://github.com/aws-amplify/amplify-js/pull/7245
JS issue on client side after upgrading Amplify CLI https://github.com/aws-amplify/amplify-js/issues/7251

GeorgeBellTMH:
Can we please get this and the corresponding change in js reverted? This is causing all sorts of problems...

GeorgeBellTMH:
@ammarkarachi and @nikhname can you please give us some reasoning why this was done and what is being done to resolve this?

GeorgeBellTMH:
Are we just going to leave this broken?

thenderson55:
This is a serious issue that is causing a lot of problems.

PR #5953, opened 2020-11-24
fix: snapshot for dart modelgen

AaronZyLee:
*Issue #, if available:*

*Description of changes:*
Fix the snapshot for dart modelgen unit tests


AaronZyLee:
The reason is that there was a previous change about @auth not included in staging but in master. That's why unit tests fail

jhockett (approved):

PR #6004, opened 2020-11-29
fix(amplify-codegen-appsync-model-plugin): address feedback from flutter team(Nov 28)

AaronZyLee:
*Issue #, if available:*

*Description of changes:*

1. Change the location from `dart/models` to `lib/models`
2. Add exports in `ModelProvider` and change imports
3. Change Temporal Types for date, time and timestamp
4. Change time toString method
5. Change enum toString method
6. Add question mark to skip the null exception


akshbhu (commented):
LGTM. Can you also take a look @attilah ?

attilah (approved):

Issue #6067, opened 2020-12-04
Errors are not returned in Lambda Resolvers


half2me:
**Describe the bug**
Following the example here: https://docs.amplify.aws/cli/graphql-transformer/function I created a lambda resolver for one of my GraphQL queries. If I make an GraphQL query and my lambda function throws an error, it does not propagate to GraphQL. Instead, it just returns an empty response.

**Amplify CLI Version**
4.37.1

**To Reproduce**
1. Generate a lambda resolver: `amplify add function`
2. Add the `@function` directive to a GraphQL field
3. In the lambda resolver, use this code for an example:
```js
exports.handler = async (event) => {
  throw new Error("Foo");
}
```
4. Run `amplify mock`
If I just return some normal data, everything works fine, but as soon as I throw an error, my response is empty (GraphQL usually complains about missing fields at this point, because some of my fields are mandatory, and I'm not filling them out)

**Expected behavior**
Looking at the generated `InvokeFooDataSource.res.vtl` resolver mapping template, I see the following code:
```vtl
## [Start] Handle error or return result. **
#if( $ctx.error )
  $util.error($ctx.error.message, $ctx.error.type)
#end
$util.toJson($ctx.result)
## [End] Handle error or return result. **
```
I would expect `$ctx.error` to contain my error object, but it is always empty.

**Desktop (please complete the following information):**
 - OS: Mac
 - Node Version: v12.20.0

half2me:
@dabit3 this is the issue I mentioned on Discord

attilah:
@half2me I was able to reproduce it, #6096 fixes it.

half2me:
thanks @attilah, much appreciated! 😃

PR #6096, opened 2020-12-07
fix: #6067 - nodejs mock return error from lambda

attilah:
*Issue #, if available:*

fix: #6067

*Description of changes:*

This fixes the error scenarios with mock when Lambda resolvers are used, either with `throw` or the `callback('error')` way.


yuth (approved):
LGTM

PR #6145, opened 2020-12-10
fix(graphql-connection-transformer): store graphql float as java double

jamesonwilliams:
According to the [GraphQL spec](https://graphql.org/learn/schema/#scalar-types), a `Float` type is:
> Float: A signed double-precision floating-point value

The CLI maps GraphQL's `Float` to a `Double` in Swift, and a `double` in TypeScript, respectively. However, the codegen maps  the GraphQL type to __Java `Float`__, which is a single-precision value. In a perfect world, this would have always been the Java double-precision type, `java.lang.Double`.

Android customers have existing models in the field that use the `Float` we've generated for them. As a result, the Android library has been updated to support _both_. This compat support was released in Amplify Android 1.6.7 (See https://github.com/aws-amplify/amplify-android/pull/1040.)

After #6145, _Double_ will be generated for all _new_ models. The Java Float support in the Android library will eventually be deprecated.

Proper support for GraphQL `Float` is important since the other platforms may use it to store large values. When the Android client receives them over subscription or sync, it would crash - or worse - truncate and improperly store large values.

Resolves: https://github.com/aws-amplify/amplify-android/issues/1031

### DIFF
--- a/packages/amplify-codegen-appsync-model-plugin/src/scalars/index.ts
+++ b/packages/amplify-codegen-appsync-model-plugin/src/scalars/index.ts
@@ -4,7 +4,7 @@ export const JAVA_SCALAR_MAP: NormalizedScalarsMap = {
   ID: 'String',
   String: 'String',
   Int: 'Integer',
-  Float: 'Float',
+  Float: 'Double',
   Boolean: 'Boolean',
   AWSDate: 'java.util.Date',
   AWSDateTime: 'java.util.Date',


jamesonwilliams:
cc: @richardmcclellan @Amplifiyer

richardmcclellan (approved):
I'm assuming there are probably some tests that need to be updated too, right?

jamesonwilliams:
> I'm assuming there are probably some tests that need to be updated too, right?

Based on the CI outputs, there does not appear to be any coverage over this.

The `Readme.md` file [asks of a contributor](https://github.com/aws-amplify/amplify-cli/blob/master/Readme.md#developing):
> Before pushing code or sending a pull request, do the following:
> 
> - At the command line, run `npm run lint` at the top-level directory. This invokes lerna to check for lint errors in all of our packages.

There's a [note in the `CONTRIBUTING.md`](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests) about tests:

> Please ensure that your change still passes unit tests

SwaySway:
@jamesonwilliams Looks like it's intended to be a Double for new projects? Could you add a feature flag on this? It looks like it might be a breaking change as is.

jamesonwilliams:
> @jamesonwilliams Looks like it's intended to be a Double for new projects? Could you add a feature flag on this? It looks like it might be a breaking change as is.

Please see discussion in the PR comments

SwaySway (approved):

PR #6161, opened 2020-12-12
feat(amplify-provider-awscloudformation): add CFN logging

ammarkarachi:
Introduces CFN logging  for debug purposes. This was reverted as a part to test out a deployment issue.

Reverts aws-amplify/amplify-cli#6119

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
@@ -11,6 +11,8 @@ const providerName = require('../constants').ProviderName;
 const { formUserAgentParam } = require('./user-agent');
 const configurationManager = require('../configuration-manager');
 const { stateManager } = require('amplify-cli-core');
+const { fileLogger } = require('../utils/aws-logger');
+const logger = fileLogger('aws-cfn');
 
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -376,12 +391,15 @@ class CloudFormation {
 
   describeStack(cfnNestedStackParams, maxTry = 10, timeout = CFN_POLL_TIME) {
     const cfnModel = this.cfn;
+    const log = logger('describeStack.cfn.describeStacks', [cfnNestedStackParams]);
```
kaustavghosh06, line 394: Will cfnNestedStackParams be always defined? Do we need a check for it?

kaustavghosh06 (approved):

yuth:
Could you update the title and description of the PR. Now its double revert in title

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
ammarkarachi: I checked for that and `undefined` does not break the logic

PR #6175, opened 2020-12-14
Allow use of existing pipeline resolver for @function directive

half2me:
Move `dependsOn` to the append function instead of the resolver, to allow usage with existing pipeline resolvers.

*Description of changes:*
By moving the dependency injection out of the resolver's template and into the function's template, the `@function` directive should work nicely with existing pipeline resolvers not generated by the `@function` directive. It will just add the dependency. It should have no effect on pipeline resolvers generated with `@function`, but it would make it more interoperable.


danielleadams:
@half2me thanks for your patience here. This functionality has been moved to https://github.com/aws-amplify/amplify-category-api - would you mind opening a PR there if you are still looking for this to be merged? Thank you.

PR #6178, opened 2020-12-14
chore: convert configuration-manager to TS

jhockett:

### DIFF
--- a/packages/amplify-cli-core/src/index.ts
+++ b/packages/amplify-cli-core/src/index.ts
@@ -34,6 +34,7 @@ export type $TSContext = {
   usageData: $TSAny;
   runtime: $TSAny;
   pluginPlatform: IPluginPlatform;
+  newUserInfo?: $TSAny;
 };
 
 export type IPluginPlatform = {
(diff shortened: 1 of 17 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/make-id.ts
```diff
@@ -1,11 +1,8 @@
-export function makeId(n) {
-  if (!n) {
-    n = 5;
-  }
+export function makeId(n: number = 5) {
   let text = '';
   const possible = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789';
 
-  for (let i = 0; i < n; i += 1) {
+  for (let i = 0; i < n; ++i) {
```
akshbhu, line 5: Will this code face collisions in ID when n is small?

jhockett: Possibly. I just edited this file to fix the types for the time being.

packages/amplify-cli-core/src/index.ts
```diff
@@ -34,6 +34,7 @@ export type $TSContext = {
   usageData: $TSAny;
   runtime: $TSAny;
   pluginPlatform: IPluginPlatform;
+  newUserInfo?: $TSAny;
```
nikhname, line 37: Can this be more strongly typed?

jhockett: Not with ease. I don't fully know how it's being used throughout the CLI. This was added to fix TS compilation. We should refactor the functions that set this value to return the data directly instead mutating `context`, but that was out of scope for this PR.

nikhname (approved):
LGTM!

PR #6199, opened 2020-12-16
fix(amplify-codegen): excludes flutter projects from gql gen

haverchuck:
*Description of changes:*
Makes sure that flutter project customers are not asked about generating GraphQL statements and
types.

### DIFF
--- a/packages/amplify-codegen/src/callbacks/prePushAddCallback.js
+++ b/packages/amplify-codegen/src/callbacks/prePushAddCallback.js
@@ -2,7 +2,7 @@ const { normalizeInputParams } = require('../utils/input-params-manager');
 const constants = require('../constants');
 const askShouldGenerateCode = require('../walkthrough/questions/generateCode');
 const addWalkThrough = require('../walkthrough/add');
-const { isCodegenConfigured } = require('../utils');
+const { getFrontEndHandler, isCodegenConfigured } = require('../utils');
 const prePushUpdateCallback = require('./prePushUpdateCallback');
 const path = require('path');
 const { isDataStoreEnabled } = require('graphql-transformer-core');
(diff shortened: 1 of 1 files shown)


packages/amplify-codegen/src/callbacks/prePushAddCallback.js
```diff
@@ -16,11 +16,15 @@ async function prePushAddCallback(context, resourceName) {
 
   let shouldGenerateCode = false;
   if (context.exeInfo.inputParams) {
-    normalizeInputParams(context);
-    const inputParams = context.exeInfo.inputParams[constants.Label];
-    const yesFlag = context.exeInfo.inputParams.yes;
+    const frontend = getFrontEndHandler(context);
 
-    shouldGenerateCode = await determineValue(inputParams, yesFlag, 'generateCode', true, () => askShouldGenerateCode());
+    if (frontend !== 'flutter') {
```
akshbhu, line 21: Is this skip also required when codegen is already setup in this line: 13?

akshbhu (dismissed):
This PR evades gql Codegen during amplify push.
  
What about commands like ```amplify codegen statements```? Does this skip also required in such commands?

packages/amplify-codegen/src/callbacks/prePushAddCallback.js
haverchuck: To my knowledge the only way to configure codegen is to run the amplify codegen add command, which yields the following:
```
==> amplify-dev codegen add
flutter is not supported by codegen plugin
```

Therefor line 13 should never be executed in a flutter project, correct?

haverchuck:
> This PR evades gql Codegen during amplify push.
> 
> What about commands like `amplify codegen statements`? Does this skip also required in such commands?

@akshbhu I believe commands such as `amplify codegen statements` first require `amplify codegen add`, which  as noted above fails gracefully for Flutter projects.  So we should be ok.

edwardfoyle (approved):

jhockett (approved):

PR #6204, opened 2020-12-17
fix: skip check on delete and go ahead on force

ammarkarachi:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-cli/src/utils/team-provider-migrate.ts
+++ b/packages/amplify-cli/src/utils/team-provider-migrate.ts
@@ -43,9 +43,11 @@ function isInvalidEnvOrPulling(context: Context): boolean {
     return true;
   }
 
-  const isPulling = context.input.command === 'pull' || context.input.command === 'init' || context.input.command === 'env';
+  if (context.input.command) {
+    return ['pull', 'init', 'env', 'delete'].includes(context.input.command);
+  }
 
-  return isPulling;
+  return false;
 }
 
 function teamProviderInfoGetAuthResourceNameHasSecrets(): any | undefined {


packages/amplify-cli/src/utils/headless-input-utils.ts
```diff
@@ -29,4 +29,4 @@ export const readHeadlessPayload = async (): Promise<string> => {
   });
 };
 
-export const isYesFlagSet = (context: Context): boolean => context?.exeInfo?.inputParams?.yes;
+export const isYesFlagSet = (context: Context): boolean => context?.exeInfo?.inputParams?.yes || context?.exeInfo?.inputParams?.force;
```
kaustavghosh06, line 32: Do we need to set this or the change below is sufficient? Tehnically force flag is not the same as yes right?

ammarkarachi: Yeah you are right `--force` is only used for delete

packages/amplify-cli/src/utils/team-provider-migrate.ts
```diff
@@ -43,7 +43,11 @@ function isInvalidEnvOrPulling(context: Context): boolean {
     return true;
   }
 
-  const isPulling = context.input.command === 'pull' || context.input.command === 'init' || context.input.command === 'env';
+  const isPulling =
+    context.input.command === 'pull' ||
+    context.input.command === 'init' ||
+    context.input.command === 'env' ||
+    context.input.command === 'delete';
```
jhockett, line 50: nit: simplifies to

```ts
  const pullCommands = ['pull', 'init', 'env', 'delete'];
  return pullCommands.includes(context.input.command);
```

kaustavghosh06 (approved):
LGTM.

jhockett (approved):

Issue #6168, opened 2020-12-12
Cannot pull an existing backend since the latest version


maximelebastard:
**Describe the bug**
On an existing project, I want to pull an existing Amplify backend
I run `amplify pull --appId d26xxxxxx --envName dev`

**Expected:**
A new amplify/ directory is created with the backend informations

**Actual**
The CLI opens a Chrome browser and redirects me to the Admin UI (https://eu-west-1.admin.amplifyapp.com/admin/d26xxxxxx/dev/verify/).
I don't use Admin UI, it is not enabled - so I am completely blocked.

**Workaround**
Downgrading amplify-cli to 4.29.8 fixed the issue.

**Amplify CLI Version**
4.39.0

**Desktop (please complete the following information):**
Mac OS

abenzick:
I am experiencing this exact same issue.

jhockett:
Hi @maximelebastard @abenzick, was AdminUI enabled previously for this project? If not, can you double check it's not currently enabled?

I haven't been able to reproduce this yet.

abenzick:
Yes, it was enabled previously.

maximelebastard:
I may have enabled it a few minutes for a quick test.
For what I remember, I did not pull the project during that interval.

jhockett:
Try running `amplify logout --appId <AmplifyAppId>`. What's the output you see?

If the credentials exist, you will be asked if you want to logout from all sessions. Answer `y`. Then try pulling your app.

jhockett:
A fix has been merged and will be available in the next release. Thanks for the bug reports @maximelebastard @abenzick!

PR #6207, opened 2020-12-17
fix: #6168, update error messages, yarn.lock

jhockett:

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/attach-backend.js
+++ b/packages/amplify-provider-awscloudformation/src/attach-backend.js
@@ -10,7 +10,7 @@ const configurationManager = require('./configuration-manager');
 const { getConfiguredAmplifyClient } = require('./aws-utils/aws-amplify');
 const { checkAmplifyServiceIAMPermission } = require('./amplify-service-permission-check');
 const constants = require('./constants');
-const { doAdminCredentialsExist, isAmplifyAdminApp } = require('./utils/admin-helpers');
+const { isAmplifyAdminApp } = require('./utils/admin-helpers');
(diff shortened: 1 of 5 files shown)


packages/amplify-provider-awscloudformation/src/attach-backend.js
```diff
@@ -27,25 +27,18 @@ async function run(context) {
...
-          try {
-            await adminLoginFlow(context, appId, envName, res.region);
-          } catch (e) {
-            context.print.error(`Failed to authenticate: ${e.message || 'Unknown error occurred.'}`);
-          }
-        }
+  if (!useProfile && (!configLevel || configLevel === 'amplifyAdmin') && appId) {
+    const res = await isAmplifyAdminApp(appId);
+    isAdminApp = res.isAdminApp;
+    if (isAdminApp) {
```
kaustavghosh06, line 33: What's the backup option here - if a customer is using cloud9/EC2 instances and have admin app enabled? Can they still use the AWS profile instead?

jhockett: It's not ideal, but you can workaround by changing the `configLevel` to `"project"` in the `<root>/amplify/.config/local-aws-info.json` file.

jhockett: A larger PR to address fallback scenarios is coming, but I didn't want to hold this fix up on it.

kaustavghosh06: Ah! Got it.

kaustavghosh06 (approved):
LGTM.

PR #6257, opened 2020-12-24
feat(amplify-appsync-simulator): implement new utils

bboure:
Implements new utils as described in doc:
https://docs.aws.amazon.com/appsync/latest/devguide/resolver-util-reference.html#list-helpers-in-util-list
and
https://docs.aws.amazon.com/appsync/latest/devguide/resolver-util-reference.html#str-helpers-in-util-str

`$utils.list.sortList()`
`$utils.str.toUpper()`
`$utils.str.toLower()`
`$utils.str.toReplace()`
`$utils.str.normalize()`
`$util.math.roundNum()`
`$util.math.roundNum()`
`$util.math.maxVal()`
`$util.math.minVal()`
`$util.math.randomWithinRange()`
`$util.math.randomDouble()`

Note concerning `$utils.list.sortList()`:
I have honoured what the documentation states in the simulator:
> Only lists containing a maximum of 1000 objects are supported.

However, during my tests in a real AppSync API, it worked fine with > 1000 elements. (tested with > 3k). I will leave the limit, just in case.

### DIFF
--- a/packages/amplify-appsync-simulator/src/__tests__/velocity/util/math.test.ts
+++ b/packages/amplify-appsync-simulator/src/__tests__/velocity/util/math.test.ts
@@ -17,14 +17,26 @@ describe('$utils.math.round', () => {
   });
 });
 
-describe('$utils.str.minVal', () => {
+describe('$utils.math.minVal', () => {
   it('should get the min value', () => {
     expect(util.math.minVal(13.45, 45.67)).toEqual(13.45);
   });
 });
 
-describe('$utils.str.maxVal', () => {
+describe('$utils.math.maxVal', () => {
   it('get the max value', () => {
     expect(util.math.maxVal(13.45, 45.67)).toEqual(45.67);
+  });
+});
+
+describe('$utils.math.random', () => {
+  it('get a random value', () => {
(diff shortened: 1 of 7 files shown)


packages/amplify-appsync-simulator/src/__tests__/velocity/util/math.test.ts
```diff
@@ -0,0 +1,30 @@
+import { create } from '../../../velocity/util/index';
```
yuth, line 1: It would be nice to have tests for `randomWithinRange` and `randomDouble`

yuth (approved):
LGTM. Thank you for the change.

If you could you add tests for `randomDouble` and `randomWithinRange`, I will merge the change

bboure:
@yuth Thank you for your review.

> It would be nice to have tests for randomWithinRange and randomDouble

Because of the characteristics of these 2 methods, I did not test them on purpose. It is hard to test random values (non-deterministic).
Using mock would make no sense either.

Also, because these 2 functions are barely a wrapper of `Math.random` and `_.random`, I don't think it deserves spending time on testing.

That being said, I guess we could test they do indeed return "something".

PR #6277, opened 2020-12-29
feat: use default values in DynamoDB VTLs

ankon:
See https://github.com/aws/aws-appsync-community/issues/59 for a related discussion.

*Description of changes:*

Translate default values defined in `input` models into default values when generating DynamoDB VTLs.


ankon:
Before looking into fixing the CircleCI issues (I would need to figure out how to run mock_e2e_tests, and the test failure seems to be a problem in how the compiler is set up w.r.t. to the target platform), the first question is:

Do you think this is a viable approach to get default values into the templates?

Issue #6280, opened 2020-12-30
Some messages output "old-docs-url"


tacck:
**Describe the bug**
Some messages output "old-docs-url" like
"https://docs.amplify.aws/cli/graphql-transformer/directives#auth".

**Amplify CLI Version**
```
$ amplify -v
4.40.1
```

**To Reproduce**
`amplify add api`, use `API (GraphQL)`, then `amplify push`.

**Expected behavior**
Output "new-docs-url" like
"https://docs.amplify.aws/cli/graphql-transformer/auth".

**Screenshots**
```
$ amplify push
✔ Successfully pulled backend environment dev from the cloud.
y
Current Environment: dev

| Category | Resource name     | Operation | Provider plugin   |
| -------- | ----------------- | --------- | ----------------- |
| Api      | XXXXXXXXX         | Update    | awscloudformation |
| Function | XXXXXXXXXXXXXXXXX | No Change | awscloudformation |
| Auth     | XXXXXXXXXXXXXXXXX | No Change | awscloudformation |
? Are you sure you want to continue? Yes

The following types do not have '@auth' enabled. Consider using @auth with @model
         - Todo
Learn more about @auth here: https://docs.amplify.aws/cli/graphql-transformer/directives#auth

```

**Desktop (please complete the following information):**
 - OS: Mac
 - Node Version. You can use `node -v` to check the node version on your system
 ```
$ node -v
v12.4.0
```

PR #6281, opened 2020-12-30
fix: change to new docs url in some messages

tacck:
*Issue #, if available:*
#6280 

*Description of changes:*
Some messages output "old-docs-url" like
"https://docs.amplify.aws/cli/graphql-transformer/directives#auth".
So, modify it to "new-docs-url" like
"https://docs.amplify.aws/cli/graphql-transformer/auth".


jhockett (approved):
Thanks for the contribution @tacck! LGTM

PR #6317, opened 2021-01-05
fix: fixes Layer tests after removing nodeModules from #CurrentCloudbackend

akshbhu:
*Issue : fixes layer tests after removeing  Node modules from #CurCloudbackend

*Description of changes:*

* Added check to remove node_modules from Lambda functions


packages/amplify-cli/src/extensions/amplify-helpers/update-amplify-meta.ts
```diff
@@ -70,9 +73,9 @@ function moveBackendResourcesToCurrentCloudBackend(resources) {
   }
 }
 
-function removeNodeModulesDir(currentCloudBackendDir) {
+function removeNodeModulesDir(currentCloudBackendDirFunction) {
```
jhockett, line 76: Couple nitpicks:
I'd say leave the name as it was, because I first assumed you were passing a `function` based on the variable name, but I think you were trying to differentiate from layers.

Also, please add a type to the parameter - I'm assuming it's string, but it would be better if it's explicit

akshbhu: good Catch! , Thanks

jhockett (approved):
LGTM!

PR #6321, opened 2021-01-06
fix: apply tags on create and push nested stack

ammarkarachi:
*Issue #, if available:*

*Description of changes:*
The goal of this PR is to apply tags when the rootstack is initialized

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
@@ -350,7 +350,7 @@ export async function getResourceStatus(category?, resourceName?, providerName?,
     allResources = allResources.filter(resource => resource.providerPlugin === providerName);
   }
 
-  const tagsUpdated = compareTags(stateManager.getProjectTags(), stateManager.getCurrentProjectTags());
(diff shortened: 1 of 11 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
```diff
@@ -348,12 +349,8 @@ export async function getResourceStatus(category?, resourceName?, providerName?,
     resourcesToBeDeleted = resourcesToBeDeleted.filter(resource => resource.providerPlugin === providerName);
     allResources = allResources.filter(resource => resource.providerPlugin === providerName);
   }
-  let tagsUpdated = compareTags(stateManager.getProjectTags(), stateManager.getCurrentProjectTags());
 
-  // if tags updated but no resource to apply tags, ignore tags updated
-  if (allResources.filter(resource => resource.category === 'provider').length === 0) {
-    tagsUpdated = false;
-  }
+  const tagsUpdated = compareTags(stateManager.getProjectTags(), stateManager.getCurrentProjectTags());
```
attilah, line 353: No need for `compareTags` just use `_.isEqual`

attilah (approved):
LGTM, one change only to not to reimplement object comparison.

PR #6331, opened 2021-01-07
Replace black/white language with allow/block

j1mr10rd4n:
*Description of changes:*
After a lack of activity after reviewer comments in #5497, the original commits were cherry-picked onto the current master branch and the requested changes made here (reverting a potentially breaking change to lambda environment variables).

Additional instances of black/white terminology were found in comments and tests and also replaced.


attilah:
Thanks @j1mr10rd4n for taking the time to submit this PR overall looks good but I've a few additions:

1. Regarding the terminology we like to use allow/deny pair instead of allow/block.
2. The changes as-is could break existing auth trigger deployments if the triggers are not getting redeployed or customers are using  manually edited version of the generated environment variables.

For this reason parts of the code needs to be behind a feature flag, to preserve the functionality that using the previous terminology.

This requires a couple of steps, I can help you to make it or I can provide guidance on what steps are required.

- The feature flag must be registered [here](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-cli-core/src/feature-flags/featureFlags.ts#L564) with this snippet of code:

```js
    this.registerFlag('platform', [
      {
        name: 'useUpdatedTerminology',
        type: 'boolean',
      },
    ]);
```

- I think that the environment variable names `DOMAINWHITELIST` and `DOMAINBLACKLIST` can be updated to the new terminology, but needs manual patching during template generation, based on the value of the feature flag [here](https://github.com/aws-amplify/amplify-cli/blob/4e5f5c62287252ff59ecda050390e9b3c2b00146/packages/amplify-category-auth/src/provider-utils/awscloudformation/utils/trigger-flow-auth-helper.js#L31)

- The lambda templates has to be duplicated with the new variables and the old ones and the proper template must be selected based on the value of the feature flag.

Let me know if you require any other assistance in updating the PR if you like the team to take over the PR and make the changes.

j1mr10rd4n:
Thanks @attilah for the feedback. I've just been reading the [amplify-cli feature flags documentation](https://docs.amplify.aws/cli/reference/feature-flags) and think I understand what needs to be done. I'll have a go and update this PR.

renebrandel:
@j1mr10rd4n - Just checking in if you wanted to drive this PR to completion? Our team is happy to take it over as well.

cjihrig:
Closing in favor of #6919. Thank you for the PR though!

Issue #6192, opened 2020-12-16
Schema change error in a new environment


ayeganyan:
**Describe the bug**
Getting schema change error in the new environment

**Amplify CLI Version**
4.40.0

**To Reproduce**
1. Change key schema in schema.graphql
2. Add new environment
```
Do you want to use an existing environment? No
? Enter a name for the environment testa
Using default provider  awscloudformation
```
3. Push environment
```
amplify push --y
```
An error is reported:
```
Attempting to edit the key schema of the InvitationTable table in the Invitation stack. 
An error occurred during the push operation: Attempting to edit the key schema of the InvitationTable table in the Invitation stack. 
```
**Expected behavior**
User should be able to push any schema in new environment

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Node Version. v10.19.0

**Additional context**

ayeganyan:
Error exists even if I did not change anything 
In 4.29.8 as well

yuth:
@ayeganyan do you have an env with name `testa` already in the project? If there is an existing environment with the same name in `<project-root>/amplify/team-provider-info.json`, the CLI does not create an new environment. If you want to recreate the environment please run `amplify env remove <env-name>` and then add it back

ayeganyan:
That was completely new environment...
I tried with new environments several environment, and finally reverted all my changes I have made after last successful deployment, and removed auto-generated "#current-cloud-backend", amplify-meta.json, awscloudformation folders, so it seems issue occurred  because of environment info did not cleaned up from previous environment.

ayeganyan:
I am facing issue again, as you can see graphqlapi is in Create state:
```
amplify push --y
✔ Successfully pulled backend environment develop from the cloud.

Current Environment: develop

| Category | Resource name                  | Operation | Provider plugin   |
| -------- | ------------------------------ | --------- | ----------------- |
| Auth     | authentication                 | Create    | awscloudformation |
| Auth     | userPoolGroups                 | Create    |                   |
| Api      | graphqlapi                     | Create    | awscloudformation |
....
| Function | invitationQueueTrigger         | Create    | awscloudformation |
| Storage  | storage                        | Create    | awscloudformation |
| Queue    | postConfirmationQueue          | Create    | awscloudformation |
| Queue    | invitationQueue                | Create    | awscloudformation |

The following types do not have '@auth' enabled. Consider using @auth with @model
         - Policy
[...]

PR #6349, opened 2021-01-09
fix: stop sanity check when resource is in create status

akshbhu:
*Issue #6192 

*Description of changes:*

*  passing a list of sanity checks to be performed over api 

* added e2e test for the same

### DIFF
--- a/packages/amplify-e2e-tests/package.json
+++ b/packages/amplify-e2e-tests/package.json
@@ -18,10 +18,7 @@
   "private": true,
   "scripts": {
     "e2e": "npm run setup-profile && jest --verbose",
-    "setup-profile": "ts-node ./src/configure_tests.ts",
-    "build": "tsc",
-    "clean": "rimraf lib tsconfig.tsbuildinfo",
-    "watch": "tsc --watch"
+    "setup-profile": "ts-node ./src/configure_tests.ts"
   },
   "dependencies": {
     "amplify-cli-core": "1.18.0",
--- a/packages/amplify-e2e-tests/package.json
+++ b/packages/amplify-e2e-tests/package.json
@@ -18,7 +18,10 @@
   "private": true,
   "scripts": {
     "e2e": "npm run setup-profile && jest --verbose",
-    "setup-profile": "ts-node ./src/configure_tests.ts"
+    "setup-profile": "ts-node ./src/configure_tests.ts",
+    "build": "tsc",
+    "clean": "rimraf lib tsconfig.tsbuildinfo",
+    "watch": "tsc --watch"
   },
   "dependencies": {
     "amplify-cli-core": "1.18.0",
--- a/packages/amplify-provider-awscloudformation/src/transform-graphql-schema.ts
+++ b/packages/amplify-provider-awscloudformation/src/transform-graphql-schema.ts
@@ -323,6 +323,7 @@ export async function transformGraphQLSchema(context, options) {
     resources = resources.concat(allResources);
   }
   resources = resources.filter(resource => resource.service === 'AppSync');
+  const appSyncApiToBeUpdated = resourcesToBeUpdated.filter(resource => resource.service === 'AppSync');
 
   if (!resourceDir) {
     // There can only be one appsync resource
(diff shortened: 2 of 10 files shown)


edwardfoyle (commented):
Change looks fine to me but I question if we need to add an e2e test for this. Seems like a unit test could adequately cover this?

akshbhu:
Yeah  you are right I could just add a unittest for this. 

I included e2e as it involves  setting up adding a new environment to setup #current-cloudBackend (which we should change as we currently copy the current-cloud-backend when changing environment) and the calling function (transformgraphql schema ) is too long to be mocked (need few refractor I guess).

packages/amplify-e2e-tests/package.json
```diff
@@ -18,7 +18,10 @@
   "private": true,
   "scripts": {
     "e2e": "npm run setup-profile && jest --verbose",
-    "setup-profile": "ts-node ./src/configure_tests.ts"
+    "setup-profile": "ts-node ./src/configure_tests.ts",
+    "build": "tsc",
+    "clean": "rimraf lib tsconfig.tsbuildinfo",
+    "watch": "tsc --watch"
```
yuth, line 24: We dont need this as the tests are run with ts-jest

packages/amplify-provider-awscloudformation/src/transform-graphql-schema.ts
```diff
@@ -323,6 +324,9 @@ export async function transformGraphQLSchema(context, options) {
     resources = resources.concat(allResources);
   }
   resources = resources.filter(resource => resource.service === 'AppSync');
+  // check if api is in update status or create status
+  const appSyncApiToBeCreated: boolean =
```
yuth, line 328:
```suggestion
  const isNewAppSyncAPI: boolean =
```

packages/amplify-provider-awscloudformation/src/transform-graphql-schema.ts
```diff
@@ -470,6 +474,9 @@ export async function transformGraphQLSchema(context, options) {
     searchableTransformerFlag = true;
   }
 
+  const ff = new AmplifyCLIFeatureFlagAdapter();
+  const sanityCheckRulesList = getSanityCheckRulesFactory(appSyncApiToBeCreated, ff);
```
yuth, line 478:
```suggestion
  const sanityCheckRulesList = getSanityCheckRules(appSyncApiToBeCreated, ff);
```

packages/graphql-transformer-core/src/__tests__/util/amplifyUtils.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    expect(sanityCheckRules.diffRules.length).toBe(0);
+    expect(sanityCheckRules.projectRules.length).toBe(0);
+  });
+
+  test('sanitycheck rule list when api is in update status and ff enabled', () => {
+    const ff_mock = new AmplifyCLIFeatureFlagAdapter();
+    (<any>FeatureFlags.getBoolean).mockReturnValue(true);
+    const sanityCheckRules: SanityCheckRules = getSanityCheckRulesFactory(false, ff_mock);
+    expect(sanityCheckRules.diffRules.length).toBe(4);
+    expect(sanityCheckRules.projectRules.length).toBe(1);
```
yuth, line 21: Can we verify the rules themselves by name instead of the size

packages/graphql-transformer-core/src/__tests__/util/amplifyUtils.test.ts
```diff
@@ -0,0 +1,31 @@
...
+  });
+
+  test('sanitycheck rule list when api is in update status and no ff enabled', () => {
+    const ff_mock = new AmplifyCLIFeatureFlagAdapter();
+    (<any>FeatureFlags.getBoolean).mockReturnValue(false);
+    const sanityCheckRules: SanityCheckRules = getSanityCheckRulesFactory(false, ff_mock);
+    expect(sanityCheckRules.diffRules.length).toBe(6);
+    expect(sanityCheckRules.projectRules.length).toBe(2);
```
yuth, line 29: Same as above comment

packages/graphql-transformer-core/src/__tests__/util/amplifyUtils.test.ts
akshbhu: I tried this with snapshot, but they were prinitng [functions] instead of Names.
Is there a way to print their names?

packages/amplify-e2e-tests/package.json
akshbhu: This I did to do a yarn build in amplify-e2e-tests, so that dont have to run yarn build on all the packages.

yuth: We dont need to transpile typescript as jest will do it on the fly. If you have the transpiled version then jest will end up using it and you are forced to compile again

akshbhu: Ahh I got it , thnaks Yathi

yuth (approved):

PR #6500, opened 2021-01-28
fix: check for local env file and then fall back on exeinfo

ammarkarachi:
*Issue #, if available:*

*Description of changes:*
The local env file is not present when the root stack is initialized on an existing project. The envName and projectName are extracted from the exeInfo as a fall back

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/get-tags.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/get-tags.ts
@@ -1,14 +1,25 @@
-import { stateManager, Tag, HydrateTags } from 'amplify-cli-core';
+import { stateManager, Tag, HydrateTags, $TSAny } from 'amplify-cli-core';
 import { Context } from '../../domain/context';
 
 export function getTags(context: Context): Tag[] {
+  let tags: Tag[];
+  let envInfo: $TSAny;
+  let projectConfig: $TSAny;
   if (stateManager.isTagFilePresent()) {
-    return stateManager.getHydratedTags();
+    tags = stateManager.getProjectTags();
   } else {
-    const { envName } = context.exeInfo.localEnvInfo;
-    const { projectName } = context.exeInfo.projectConfig;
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/get-tags.ts
```diff
@@ -2,13 +2,24 @@ import { stateManager, Tag, HydrateTags } from 'amplify-cli-core';
...
+    tags = stateManager.getProjectTags();
   } else {
-    const { envName } = context.exeInfo.localEnvInfo;
-    const { projectName } = context.exeInfo.projectConfig;
-    return HydrateTags(initialTags, { envName, projectName });
+    tags = initialTags;
   }
+  if (stateManager.localEnvInfoExists() && stateManager.projectConfigExists()) {
+    envInfo = stateManager.getProjectConfig();
+    projectConfig = stateManager.getLocalEnvInfo();
```
jhockett, line 15: are theses variable assignments backwards?

from naming alone, I would expect:

```ts
    projectConfig = stateManager.getProjectConfig();
    envInfo = stateManager.getLocalEnvInfo();
```

ammarkarachi: Nice catch!

packages/amplify-cli/src/extensions/amplify-helpers/get-tags.ts
```diff
@@ -2,13 +2,24 @@ import { stateManager, Tag, HydrateTags } from 'amplify-cli-core';
 import { Context } from '../../domain/context';
 
 export function getTags(context: Context): Tag[] {
+  let tags: Tag[];
+  let envInfo: any;
+  let projectConfig: any;
```
jhockett, line 7: nit: use $TSAny from amplify-cli-core

jhockett (approved):
LGTM, just one minor nit

PR #6514, opened 2021-01-30
fix(amplify-appsync-simulator): expose errors from util.validate

### LINKED ISSUE #6068: $util.validate does not return a message
**Describe the bug**
When using `$util.validate()` in a VTL resolver, a message is never returned.
According to the docs: https://docs.aws.amazon.com/appsync/latest/devguide/resolver-util-reference.html
```vtl
$util.validate(boolean, String) : void
```
"_If the condition is false, throw a CustomTemplateException with the specified message._"

But in my case, the `String` argument is ignored.

**Amplify CLI Version**
4.37.1

**To Reproduce**
Run `amplify mock`
Use `$util.validate(false, "foo")` in a VTL resolver, it will throw an error with an empty message .
I also tried: `$util.validate(false, "foo", "bar")`
Which returns `bar` correctly for the type, but doesn't include `foo` as the message.

**Expected behavior**
The 2nd argument should be used as the error message.

**Screenshots**
Example output for `$util.validate(false, "foo", "bar")`
Screenshot 2020-12-04 at 22 44 52

 - OS: Mac
 - Node Version: v12.20.0

yuth:
Updated the simulator to expose util.errors

fix #6068

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/packages/amplify-appsync-simulator/src/velocity/index.ts
+++ b/packages/amplify-appsync-simulator/src/velocity/index.ts
@@ -57,9 +57,9 @@ export class VelocityTemplate {
     try {
       templateResult = this.compiler.render(context);
     } catch (e) {
-      const lastError = context.util.errors.pop();
+      const lastError = context.util.errors.length && context.util.errors[context.util.errors.length - 1];
       if (lastError && lastError instanceof ValidateError) {
-        return { result: lastError.data, errors: [...context.util.errors, lastError], isReturn: false, stash: context.ctx.stash.toJSON() };
+        return { result: lastError.data, errors: [...context.util.errors], isReturn: true, stash: context.ctx.stash.toJSON() };
       }
       return { result: null, errors: [...context.util.errors], isReturn: false, stash: context.ctx.stash.toJSON() };
     }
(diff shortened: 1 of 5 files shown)


packages/amplify-appsync-simulator/src/velocity/index.ts
```diff
@@ -57,6 +57,10 @@ export class VelocityTemplate {
     try {
       templateResult = this.compiler.render(context);
     } catch (e) {
+      const lastError = context.util.errors.length && context.util.errors[context.util.errors.length - 1];
```
attilah, line 60: could use: https://lodash.com/docs/4.17.15#last

packages/amplify-util-mock/src/__e2e__/utils/index.ts
```diff
@@ -44,6 +44,24 @@ export async function deploy(transformerOutput: any, client?: DynamoDB) {
   const simulator = await runAppSyncSimulator(config);
   return { simulator, config };
 }
+
+export async function reDeploy(
```
attilah, line 48: nit: redeploy

packages/amplify-appsync-simulator/src/velocity/util/errors.ts
```diff
@@ -36,10 +37,12 @@ export class TemplateSentError extends Error {
 export class Unauthorized extends TemplateSentError {
   constructor(gqlMessage, info: GraphQLResolveInfo) {
     super(gqlMessage, 'Unauthorized', {}, {}, info);
+    Object.setPrototypeOf(this, Unauthorized.prototype);
```
attilah, line 40: Q: are we using this prototype information elsewhere?

yuth: Without setting this `instanceof` will not work

PR #6530, opened 2021-02-01
fix(amplify-codegen): correct docFilePath in codegen add and configure

### LINKED ISSUE #4147: Graphql mutations, queries, subscriptions not created in specified directory.
**Describe the bug**
When going through the tutorial for amplify I'm asked about generating graphql files. 

```
? Choose the code generation language target javascript
? Enter the file name pattern of graphql queries, mutations and subscriptions graphql/**/*.js
? Do you want to generate/update all possible GraphQL operations - queries, mutations and subscriptions Yes
```

Here I specify the file pattern being `graphql/**/*.js`.

However the code generation tool writes these files to `src/graphql` instead.

The way I fixed this was to manually edit `.graphqlconfig.yml` and run `amplify codegen`.

**Amplify CLI Version**
4.18.1

### LINKED ISSUE #5835: Codegen can't generate code at designated path
**Describe the bug**
I've tried to generate graphql code like below

```bash
? Choose the code generation language target typescript
? Enter the file name pattern of graphql queries, mutations and subscriptions ./graphql/**/*.ts
? Do you want to generate/update all possible GraphQL operations - queries, mutations and subscriptions Yes
? Enter maximum statement depth [increase from default if your schema is deeply nested] 2
? Enter the file name for the generated code ./graphql/API.ts
? Do you want to generate code for your newly created GraphQL API Yes
✔ Generated GraphQL operations successfully and saved at src/graphql
```
it's expected to generate the code at `/graphql/*/ts` but the actual graphql code are generated at `src/graphql/*.ts`
Is it possible to generate the code other than /src?

**Amplify CLI Version**
4.26.1

**To Reproduce**
`amplify codegen models` with a path which doesn't include `/src`

**Expected behavior**
graphql code will be generated at `/graphql/*./ts` instead of `/src/graphql/*.ts`

 - OS: macOS 10.14.6
 - Node Version v12.16.1

### LINKED ISSUE #6468: Command 'amplify add codegen' don't follow configuration
**Describe the bug**
When setting up `codegen` described here: https://docs.amplify.aws/cli/graphql-transformer/codegen#workflows generated code is placed under different directory than pointed.

**Amplify CLI Version**
4.41.2

**To Reproduce**
1. Create and configure AppSync graphql schema
2. Export `schema.json` to your project directory
3. Run `amplify add codegen`
4. Enter this configuration:
```
amplify add codegen
? Choose the type of app that you're building javascript
? What javascript framework are you using react
? Choose the code generation language target typescript
? Enter the file name pattern of graphql queries, mutations and subscriptions aws/graphql/**/*.ts
? Do you want to generate/update all possible GraphQL operations - queries, mutations and subscriptions Yes
? Enter maximum statement depth [increase from default if your schema is deeply nested] 2
? Enter the file name for the generated code aws/API.ts
? Do you want to generate code for your newly created GraphQL API Yes
✔ Generated GraphQL operations successfully and saved at src/graphql
✔ Code generated successfully and saved in file aws/API.ts
```

**Expected behavior**
Generated GraphQL operations should be saved into `aws/graphql`

**Actual behavior**
Generated GraphQL operations should be saved into `src/graphql`

- OS: Linux Manjaro
- Node Version. v12.20.1

**Log output**
```
2021-01-24T09:47:40.071Z|info : amplify version core  {"version":true,"yes":false}
2021-01-24T09:48:10.804Z|info : amplify add codegen
```

AaronZyLee:
*Issue #, if available:*
fix #6468 
fix #5835 
fix #4147
*Description of changes:*
The generation path for doc files are now correctly configured by sync up with user's prompt glob expression in `includes` field. The fix applies to the following commands.

- `amplify codegen add`
image

- `amplify codegen configure`
image

### DIFF
--- a/packages/amplify-codegen/src/utils/getGraphQLDocPath.js
+++ b/packages/amplify-codegen/src/utils/getGraphQLDocPath.js
@@ -5,7 +5,6 @@ function getGraphQLDocPath(frontend, graphQLDirectory, includePathGlob) {
   if (frontend === 'android') {
     return join(graphQLDirectory, 'com/amazonaws/amplify/generated/graphql');
   }
-  //TODO
   return includePathGlob ? globParent(includePathGlob) : graphQLDirectory;
 }
 
--- a/packages/amplify-codegen/src/utils/getGraphQLDocPath.js
+++ b/packages/amplify-codegen/src/utils/getGraphQLDocPath.js
@@ -1,17 +1,11 @@
-const { join, dirname } = require('path');
+const { join } = require('path');
+const globParent = require('glob-parent');
 
-function getGraphQLDocPath(frontend, schemaLocation) {
-  const graphQLDirectory = dirname(schemaLocation);
-  let subDirectory;
-  switch (frontend) {
-    case 'android':
-      subDirectory = 'com/amazonaws/amplify/generated/graphql';
-      break;
-    default:
-      subDirectory = '';
+function getGraphQLDocPath(frontend, graphQLDirectory, includePathGlob) {
+  if (frontend === 'android') {
+    return join(graphQLDirectory, 'com/amazonaws/amplify/generated/graphql');
   }
-
-  return join(graphQLDirectory, subDirectory);
+  return includePathGlob ? globParent(includePathGlob) : graphQLDirectory;
 }
 
 module.exports = getGraphQLDocPath;
--- a/packages/amplify-codegen/src/utils/getGraphQLDocPath.js
+++ b/packages/amplify-codegen/src/utils/getGraphQLDocPath.js
@@ -5,6 +5,7 @@ function getGraphQLDocPath(frontend, graphQLDirectory, includePathGlob) {
   if (frontend === 'android') {
     return join(graphQLDirectory, 'com/amazonaws/amplify/generated/graphql');
   }
+  //TODO
   return includePathGlob ? globParent(includePathGlob) : graphQLDirectory;
 }
 
--- a/packages/amplify-codegen/src/walkthrough/add.js
+++ b/packages/amplify-codegen/src/walkthrough/add.js
@@ -57,6 +57,7 @@ async function addWalkThrough(context, skip = [], withoutInit, decoupleFrontend,
 
   const includePatternDefault = getIncludePattern(targetLanguage, schemaLocation);
   const includePathGlob = join(includePatternDefault.graphQLDirectory, '**', includePatternDefault.graphQLExtension);
+  const docsFilePathDefault = includePatternDefault.graphQLDirectory;
 
   if (!skip.includes('includePattern')) {
     answers.includePattern = await determineValue(inputParams, yesFlag, 'includePattern', [includePathGlob], () =>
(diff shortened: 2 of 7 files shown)


packages/amplify-codegen/src/utils/getGraphQLDocPath.js
```diff
@@ -1,17 +1,11 @@
...
-  let subDirectory;
-  switch (frontend) {
-    case 'android':
-      subDirectory = 'com/amazonaws/amplify/generated/graphql';
-      break;
-    default:
-      subDirectory = '';
+function getGraphQLDocPath(frontend, graphQLDirectory, includePathGlob) {
+  if (frontend === 'android') {
+    return join(graphQLDirectory, 'com/amazonaws/amplify/generated/graphql');
```
phani-srikar, line 6: Does this mean if the android users want to generate docs outside the graphQL directory, they cannot even if they have includePathGlob pointing to a different location?

packages/amplify-codegen/src/walkthrough/add.js
```diff
@@ -65,7 +66,7 @@ async function addWalkThrough(context, skip = [], withoutInit, decoupleFrontend,
   }
   if (!skip.includes('shouldGenerateDocs')) {
     answers.shouldGenerateDocs = await determineValue(inputParams, yesFlag, 'generateDocs', true, () => askShouldGenerateDocs());
-    answers.docsFilePath = getGraphQLDocPath(frontend, schemaLocation);
+    answers.docsFilePath = getGraphQLDocPath(frontend, docsFilePathDefault, answers.includePattern);
```
phani-srikar, line 69: Looks like we're using `includePattern` to determine the path to the generated docs. The walkthrough question that takes this info says `Enter the file name pattern of graphql queries, mutations and subscriptions`. As a developer, this sounds to me like I am being asked the path to existing docs if any. What is the purpose of `includePattern`? If it is to just used to store the generated docs, we might want to change it to `Enter the file name pattern for generated graphql queries, mutations and subscriptions`

phani-srikar (changes requested):
some developer experience related comments. Otherwise, much needed fix! Thanks.

packages/amplify-codegen/src/utils/getGraphQLDocPath.js
AaronZyLee: The android user generates code in Android Studio and uses different type generator. The docs file path is an input parameter for type generator. This part for android was previously hardcoded. Since it does not use amplify type codegen, changing it might fail in type generator. That's why I just keep it as hardcoded. We can fix it once we have an android user reporting issues. For this time, our customers are all from JS part.

packages/amplify-codegen/src/walkthrough/add.js
AaronZyLee: Yes this is a confusing part. The `includePattern` is an input param for type-gen. Originally CLI does not provide a question for the doc-gen output path. The current question is like assuming you have docs generated. However, from customers experience, I find that they tend to regard this question as the configure for doc-gen path. On the other hand, CLI does not give an option(question) to really change the `docFilePath` unless editting manually.

AaronZyLee: From my perspective, the question you mention basically does two tasks:
1. What is your doc-gen output path
2. What is doc path as an input for type-gen
The reason for combining these two is that amplify codegen automatically combines these two process. There are few use cases that you only use type-gen alone. 

I would say this code changes are sufficient to solve users issues. However, you are right that we should not include those two tasks in only one question. I suggest that this can be done in the codegen redesign since adding/removing questions need additional API reviews/document changes.

phani-srikar: Thanks for the clarification. Yeah we need to separate these two questions and I agree that this will fix the current issues and we can take it up again at later point.

packages/amplify-codegen/src/utils/getGraphQLDocPath.js
phani-srikar: We should mention this at-least in the PR description, saying something like: `android projects will still ignore the include file path and generate the docs in graphQL directory` to document this limitation.

phani-srikar (approved):

akshbhu (approved):
Discussed Offline LGTM!!

PR #6539, opened 2021-02-03
chore: update packages

yuth:
The PR contains the following changes
1. updade package react script
2. update eslint
3. update typescript eslint plugin
4. update .eslintrc to as result of breaking changes in eslint packages


attilah (approved):
Awesome, thanks for making this change! As we discussed offline make sure E2E test projects are compiling and it is good to go.

PR #6545, opened 2021-02-04
test: update migration tests to handle new text

jhockett:

### DIFF
--- a/packages/amplify-migration-tests/src/migration-helpers/init.ts
+++ b/packages/amplify-migration-tests/src/migration-helpers/init.ts
@@ -1,0 +1,72 @@
+import { addCircleCITags, getCLIPath, isCI, nspawn as spawn } from 'amplify-e2e-core';
+
+const defaultSettings = {
+  name: '\r',
+  envName: 'integtest',
+  editor: '\r',
+  appType: '\r',
+  framework: '\r',
+  srcDir: '\r',
+  distDir: '\r',
+  buildCmd: '\r',
+  startCmd: '\r',
+  useProfile: '\r',
+  profileName: '\r',
+  region: process.env.CLI_REGION,
+  local: false,
+  disableAmplifyAppCreation: true,
+};
+
+export function initJSProjectWithProfile(cwd: string, settings: Object, testingWithLatestCodebase = false) {
+  const s = { ...defaultSettings, ...settings };
+  let env;
+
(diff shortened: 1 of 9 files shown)


packages/amplify-migration-tests/src/migration-helpers/init.ts
```diff
@@ -0,0 +1,72 @@
...
+  buildCmd: '\r',
+  startCmd: '\r',
+  useProfile: '\r',
+  profileName: '\r',
+  region: process.env.CLI_REGION,
+  local: false,
+  disableAmplifyAppCreation: true,
+};
+
+export function initJSProjectWithProfile(cwd: string, settings: Object, testingWithLatestCodebase = false) {
```
SwaySway, line 20: nit: This helper could exist within `amplify-e2e-core`.

packages/amplify-migration-tests/src/migration-helpers/init.ts
```diff
@@ -0,0 +1,72 @@
...
+      .sendLine(s.srcDir)
+      .wait('Distribution Directory Path:')
+      .sendLine(s.distDir)
+      .wait('Build Command:')
+      .sendLine(s.buildCmd)
+      .wait('Start Command:')
+      .sendCarriageReturn()
+      .wait('Using default provider  awscloudformation');
+
+    if (testingWithLatestCodebase || !isCI()) {
```
SwaySway, line 54: We could use the same check used by `getCLIPath()` to be more consistent.

packages/amplify-migration-tests/src/migration-helpers/init.ts
jhockett: initJSProjectWithProfile already exists in `amplify-e2e-core`. I purposely moved this version to `amplify-migration-tests` because of the way testingWithLatestCodebase defaults to false by convention in other e2e helpers. I would have had to edit every single call to initJSProjectWithProfile in regular e2e tests otherwise.

packages/amplify-migration-tests/src/migration-helpers/init.ts
jhockett: `getCLIPath()` performs the isCI check internally, this is to make sure the behavior matches as expected

SwaySway (approved):
LGTM

PR #6572, opened 2021-02-09
fix: adding envName for intents

akshbhu:
*Issue #6347 

*Description of changes:*

* Added envName in interactions Lambda


edwardfoyle (approved):
Would be nice to have a test for this but I'm not sure how feasible that is with the current interactions implementation. Just to confirm though, were you able to set up the interactions category in another env and then checkout another env and it worked in that env too?

akshbhu:
Yeah, I have locally tested it on my account and it works. For the e2e test implementation, I tried that but in the implementation for this case, we need to select the already deployed Lambda function.

PR #6579, opened 2021-02-10
fix(graphql-auth-transformer): fix auth on non model types

### LINKED ISSUE #6557: Amplify push not updating cloud resources with GraphQL API auth changes
> **Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the
> [Amplify Console GitHub Issue Tracker](https://github.com/aws-amplify/amplify-console/issues)

**Describe the bug**
Some auth settings on my GraphQL schema are not generated correctly in the AppSync schema after a push. We have 2 AWS accounts (dev and prod), which both had just a recent deploy from the AWS Amplify console (set to track the `master` branch on our repo).

The autogenerated schemas that are displayed in the AppSync console are different. Some types are missing the @aws_iam tag, which is messing up the database access.

The only reason I can think of is that the master deploy was done from a different starting point on both accounts. Funny enough, both of them (prod and dev) are missing the tag in different places,  which makes me think that depending on the starting point, they skipped some changes.

The missing changes are:

- prod account: `type S3Object`,that is not a model, is not inheriting the `@aws_iam` tag when added as a field on a model that has it
- dev account: a model with auth rule `{ allow: owner, operations: [create, delete, update] }` use to also have `{ allow: private, operations: [read], provider: iam }`. The last was removed but the update is not showing up, which makes the `others` read operation invalid

I tried to add some changes on both environments and perform an `amplify push`, with no luck. Also redeployed `master` through the console. Something is making the auth changes to not be reflected in the schema. I even tried to add and revert the `{ allow: private, provider: iam }` lines, no luck.

**Amplify CLI Version**
4.41.2

**To Reproduce**
Add the a schema type `{ allow: private, provider: iam }` to the auth rules (which already are `@auth(rules: [{ allow: owner }])` and deploy the changes.

**Expected behavior**
AppSync should reflect the schema changes after an amplify push

- OS: `macOS Catalina 10.15.7`
- Node Version: `v15.7.0`

SwaySway:
*Issue #, if available:*
fix #6557

*Description of changes:*

allow appsync auth directives to merge on non model types which are referrenced by model types with
different auth providers

### DIFF
--- a/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
+++ b/packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
@@ -267,7 +267,7 @@ export class ModelAuthTransformer extends Transformer {
     ctx.mergeConditions(template.Conditions);
     this.updateAPIAuthentication(ctx);
     if (!ctx.metadata.has(AUTH_NON_MODEL_TYPES)) {
-      ctx.metadata.set(AUTH_NON_MODEL_TYPES, new Set<string>());
+      ctx.metadata.set(AUTH_NON_MODEL_TYPES, new Map<string, Set<string>>());
     }
   };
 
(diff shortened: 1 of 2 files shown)


akshbhu (approved):
LGTM!!

Issue #5229, opened 2020-09-02
@searchable directive ElasticsearchVersion: "6.2" is it possible to use version 7+


lenarmazitov:
Hello, when I use @searchable directive, amplify generates SearchableStack with ElasticsearchVersion: "6.2"
I understand that upgrade is not possible, but when creating new project is it possible to manage version to 7+
If so then is dynamodb trigger DbToEs will be generated to work with new version?

**Which Category is your question related to?**
amplify

**Amplify CLI Version** 
4.28.1

**What AWS Services are you utilizing?**
amplify, ElasticSearch

**Why?**
There are many opportunities to search word suggestions and use word vectors in version 7 of ElasticSearch

lenarmazitov:
Currently it is not easy to change version to 7.7
I had to make changes in code in /usr/local/lib/node_modules/@amplify-cli/...

Seems like currently DbToEs lambda trigger inserts, deletes and update data without any errors.

iuyt9003:
Any news regard this ? 
Want to update to 7 to have the native autocomplete elastic feature ...

Thanks!

iuyt9003:
Hey, I done it.

It's working, hope that Amplify support es 7.9.

But the way to do it,


Steps to upgrade elastic:

1) Update 2 times the elastic search version in the domain settings 
   1. 1) First update, from 6.2 to 6.8. 
   1. 2) After the first update done, you will need again to run update from 6.8 to 7.9
2) Change the amazon elastic put auto-generated function code. 
    Where the is const (DOC_TYPE = “doc”) to DOC_TYPE = “_doc” - This change is required by elastic search 7.X which changed 
    the doc type implementation.
3) Change the auto-generated search VTL request Query.search{TABLE_NAME}Searchs.req.vtl 
    instead of the $indexPath = "/assetsearch/doc/_search" into $indexPath = "/TABLE_NAME/_doc/_search" - the doc has been 
    changed to _doc which is the right one.
    
Celebrate elastic 7.9 :)

renebrandel:
Hi folks - we're planning on addressing this as part of our new GraphQL transformer. The approach is captured here in our RFC: https://github.com/aws-amplify/amplify-cli/issues/7546 love to hear your feedback!

josefaidt:
Hey y'all :wave: as noted in the comment above this will be captured in the linked RFC #7546 (under proposal 3), and new projects are now created using version 7.1

PR #6588, opened 2021-02-10
improvement: go listening for running process

Jhorlin:
*Issue #, if available:*
#5229 
*Description of changes:*
amplify-go-function-runtime-provider/localinvoke.ts will now listen for a running debug process.
If it finds one it will get the port number for that process and use it instead of invoking a new process.
This allows a developer to have a local debug process running while testing locally.

### DIFF
--- a/packages/amplify-app/src/index.js
+++ b/packages/amplify-app/src/index.js
@@ -443,7 +443,7 @@ async function createAmplifyHelperFiles(frontend) {
   }
 
   if (frontend === 'flutter') {
-    initializeAmplifyConfiguration(path.resolve('src'));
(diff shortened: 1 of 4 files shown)


cjihrig:
Thank you for the PR. I don't think this is the direction we should go in for a couple reasons:

1. This approach is specific to Go. There is a somewhat related issue for Node.js (https://github.com/aws-amplify/amplify-cli/issues/6894). I think we should try to come up with a more generic approach to debugging that works with all of the supported languages.
2. `netstat`'s flags aren't 100% consistent across platforms, so I think this approach could have compatibility issues.

I'm going to close this out, but if you'd like to help come up with a generic debugging approach, that would be welcome.

PR #6620, opened 2021-02-14
fix(amplify-appsync-simulator): expose other than GraphQLError

h-kishi:
*Issue #, if available:*

Since [version 1.25.1 of amplify-appsync-simulator](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-appsync-simulator/CHANGELOG.md#1251-2021-02-10), the behavior when an error occurs in a resolver has changed.
When I am using Lambda as DataSource and an error occurs in Lambda, the error type is Error instead of GraphQLError.
So, the error is filtered out from errors and I can not refer the error in `response.vtl`.
In a real environment, the error does not seem to be filtered out.

*Description of changes:*

I have removed the filter with GraphQLError and added a test to reproduce the issue.


h-kishi:
It seems that my approach was not appropriate, so I will close it and create a new pull request again.

PR #6621, opened 2021-02-14
fix(amplify-frontend-javascript) update aws-exports.js on new env creation

rajrajhans:
*Issue #, if available:* #6432 

*Description of changes:*

The issue was that running `amplify init` and `amplify push` to create a new environment would not update the `aws-exports.js` to match the newly created environment. (Note that it would set the current active env to the newly created one, but the `aws-exports.js` file would still have old endpoints and keys.)

The generation of `aws-exports.js` file is done [here](https://github.com/aws-amplify/amplify-cli/blob/0e5d85be780c800aad2322ebb2b5598187c97ae8/packages/amplify-frontend-javascript/lib/frontend-config-creator.js#L210), On the CLI side, `initialize-env.ts` calls `context.amplify.onCategoryOutputsChange` which in turn calls the `generateAWSExportsFile` function. The `generateAWSExportsFile` function takes in a config object and outputs it to the file.

So, the function was getting called properly, but [this line](https://github.com/aws-amplify/amplify-cli/blob/0e5d85be780c800aad2322ebb2b5598187c97ae8/packages/amplify-frontend-javascript/lib/frontend-config-creator.js#L85), i.e. `Object.assign(newAWSExports, customConfigs);` was causing the `newAWSExports` object values (which actually contain the config for the newly created env) were getting overridden by `customConfigs`, and `customConfigs` were the values that belonged to the older env. Thus, `aws-exports.js` ended up with the values that actually belonged to older env. 

To fix this, I have added a condition which does not execute the Object.assign line when the CLI command is "push" or "init", since those are the scenarios where the issue is occurring. But then even though it solves this particular issue, since it effectively ignores `customConfigs` on `push` and `init`, i am not sure what other repercussions it has, so i am including as much detail as possible so that someone experienced can help me out here.

Debug outputs (before Object.assign) on running `init` for `newAWSExports`, `cloudAWSExports`, `currentAWSExports` and `customConfigs`

image

As you can see, the `newAWSExports` object contains the correct object that should be written to file, but Object.assign overwrites that and includes the old one (in `customConfigs`) instead. Same thing happens with `push` as well. 

### DIFF
--- a/packages/amplify-frontend-javascript/lib/frontend-config-creator.js
+++ b/packages/amplify-frontend-javascript/lib/frontend-config-creator.js
@@ -82,7 +82,10 @@ async function createAWSExports(context, amplifyResources, cloudAmplifyResources
 
   const customConfigs = getCustomConfigs(cloudAWSExports, currentAWSExports);
 
-  Object.assign(newAWSExports, customConfigs);
+  if (context.input.command !== 'push' && context.input.command !== 'init') {
+    Object.assign(newAWSExports, customConfigs);
+  }
+
   generateAWSExportsFile(context, newAWSExports);
   return context;
 }
@@ -339,7 +342,8 @@ function getAPIGWConfig(apigwResources, projectRegion, configOutput) {
   };
 
   for (let i = 0; i < apigwResources.length; i += 1) {
-    if (apigwResources[i].output.ApiName && apigwResources[i].output.RootUrl) { // only REST endpoints contains this information
+    if (apigwResources[i].output.ApiName && apigwResources[i].output.RootUrl) {
+      // only REST endpoints contains this information
       apigwConfig.aws_cloud_logic_custom.push({
         name: apigwResources[i].output.ApiName,
         endpoint: apigwResources[i].output.RootUrl,


kaustavghosh06:
@rajrajhans The Object.assign() logic is in place for the exports and config files to allow for custom configurations and not overwrite them on each push. This PR would remove that logic and that is not desirable. Would it be possible to modify this so that for new env's the exports and config files only have the custom values and once push is done the Object.assign() logic kicks in?

rajrajhans:
Hi @kaustavghosh06 thank you for the review. I have some doubts regarding the changes suggested.

> Would it be possible to modify this so that for new env's the exports and config files only have the custom values and once push is done the Object.assign() logic kicks in?

So do you mean that instead of skipping the Object.assign() on **every** init and push, it should only skip for **init** and the **first push**, and for all later pushes, Object.assign() should take effect.

Can you please confirm if this is the intended behavior? 

Thanks

kaustavghosh06:
@rajrajhans Let me know if you need more input or help with this.

rajrajhans:
Hi @kaustavghosh06, my proposed changes, along with your suggestions, will solve this issue, but, upon further investigation, I realized that the root cause of the issue is that `cloudAWSExports` is not properly reflecting the value of previous env. If `cloudAWSExports` has what it is supposed to have, then `customConfigs` will strictly have only the custom configs, after which we wouldn't face this issue. 

So, I feel tackling the root cause will address this issue as well as any unforeseen scenarios. 

Upon going through old PRs, I came across #6496, which seems to tackle the root cause. I checked out that PR locally and tested whether it solves #6432 as well. It does. So, I feel my PR should be closed in favor of #6496.

Thanks.

PR #6622, opened 2021-02-14
add selectionSetList to pipeline function

kldeb:
*Issue #, if available:*
I didn't open an issue for this, but chatted on discord about it

*Description of changes:*
This change adds the selectionSetList from the ctx object to the payload so that it can be used in the lambda function. We could use this on our project so I modified the resolver my selection. I'm opening a PR because I think it's important enough a feature with no side effects that I can think of.

Think of the case where you need to make multiple calls to third party REST APIs to retrieve child data etc and you would like to do it async on the server rather than the client for two reasons I can think of: 

1. simplifies the front end for front end devs
2. reduce the number of round trips. 

The lambda resolver can now inspect in the selection set list and only make additional calls if the field is requested from the client. 


SwaySway (commented):
@kldeb 
Changes look good are you able to provide a unit/snapshot test checking the resolver output?

kldeb:
@SwaySway sure, can you point me to a good example to imitate?

packages/graphql-function-transformer/src/FunctionTransformer.ts
```diff
@@ -148,14 +148,15 @@ export class FunctionTransformer extends Transformer {
             source: ref('util.toJson($ctx.source)'),
             request: ref('util.toJson($ctx.request)'),
             prev: ref('util.toJson($ctx.prev)'),
+            selectionSetList: str('$ctx.info.selectionSetList'),
```
SwaySway, line 151: Can you make this a ref as well?

SwaySway (commented):
@kldeb 
Apologies for the late response, are you able to add the changes to this PR?

If so here is the `@http` unit tests for checking resolver output. 
https://github.com/aws-amplify/amplify-cli/blob/master/packages/graphql-http-transformer/src/__tests__/HttpTransformer.test.ts

PR #6661, opened 2021-02-18
fix(amplify-codegen-appsync-model-plugin): include isNullable in version hash

amhinson:
*Issue #, if available:*
Given a basic schema in a JavaScript project:
```graphql
type Post @model {
  id: ID!
  title: String
}
```
after running `amplify models codegen` for the first time, the `schema.js` file is populated as expected with a unique `version` hash based on the type and the fields.

However, after changing `title` to be required with `String!`, and then running `amplify models codegen` again, the `version` doesn't change.

My expectation is that a field's "required" attribute should be considered in this hash.

*Description of changes:*
- include `isNullable` when calculating the `version` hash.


akshbhu:
Hi @amhinson 

This package has been migrated to https://github.com/aws-amplify/amplify-codegen.
Can you reopen this PR to above repo as they are currently maintaining this package?

Issue #5287, opened 2020-09-10
Unable to add analytics in ap-northeast-2


jakemraz:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[Amplify Console GitHub Issue Tracker](https://github.com/aws-amplify/amplify-console/issues)

**Describe the bug**
can't create Pinpoint service using amplify-cli in ap-northeast-2 region

I executed 'amplify init' and created Amplify app in ap-northeast-2 region. It did work. I could see there is Amplify app in ap-northeast-2.
afterward, I executed 'amplify add notifications' but Pinpoint service was created in us-west-2 even though Amplify App is in ap-northeast-2 and region value is set to ap-northeast-2 in ~/.aws/config
(I don't know why Pinpoint services is created in us-west-2 even though there is pinpoint launched at ap-northeast-2)

I tried to create Pinpoint using amplify-cli 5 times, but everytime I created, it always is created in us-west-2

**Amplify CLI Version**
4.29.1

**To Reproduce**
set your region to ap-northeast-2 at ~/.aws/config

$ amplify add analytics
? Select an Analytics provider Amazon Pinpoint
? Provide your pinpoint resource name: remove
Adding analytics would add the Auth category to the project if not already added.
? Apps need authorization to send analytics events. Do you want to allow guests and unauthenticated users to send analyt
ics events? (we recommend you allow this when getting started) Yes
Successfully added auth resource locally.
Successfully added resource remove locally

Some next steps:
"amplify push" builds all of your local backend resources and provisions them in the cloud
"amplify publish" builds all your local backend and front-end resources (if you have hosting category added) and provisions them in the cloud

$ amplify console
https://ap-northeast-2.console.aws.amazon.com/amplify/home?region=ap-northeast-2#/dbobakoi9fape/YmFja2VuZA/dev

$ amplify push
✔ Successfully pulled backend environment dev from the cloud.

Current Environment: dev

| Category  | Resource name   | Operation | Provider plugin   |
| --------- | --------------- | --------- | ----------------- |
| Auth      | cognito03492cea | Create    | awscloudformation |
| Analytics | remove          | Create    | awscloudformation |
? Are you sure you want to continue? Yes
⠇ Updating resources in the cloud. This may take a few minutes...

UPDATE_IN_PROGRESS DeploymentBucket          AWS::S3::Bucket            Thu Sep 10 2020 10:29:26 GMT+0900 (Korean Standard Time)
[...]

PR #6676, opened 2021-02-19
fix(amplify-provider-awscloudformation): support ap-northeast

watilde:
Update region mapping of Pinpoint on ap-northeast.
Fixes: https://github.com/aws-amplify/amplify-cli/issues/5287

*Description of changes:*
As it's reported, `ap-northeast-1` and `ap-northeast-2` seem to be supported already and `amplify add analytics` should make Pinpoint resource in their region . Refs: [docs](https://docs.aws.amazon.com/general/latest/gr/pinpoint.html). I also heard the same issue from a developer that they needed to manually update the config file and I came up with this PR.

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/__tests__/utils/aws-pinpoint.test.ts
+++ b/packages/amplify-provider-awscloudformation/src/__tests__/utils/aws-pinpoint.test.ts
@@ -1,0 +1,26 @@
+const { getPinpointRegionMapping, getConfiguredPinpointClient } = require('../../aws-utils/aws-pinpoint');
+
+let region = 'non-existent-region';
+
+jest.mock('../../configuration-manager', () => ({
+  loadConfiguration: jest.fn(() => ({
+    region: region,
+  })),
+}));
+
+jest.mock('../../aws-utils/user-agent', () => ({
+  formUserAgentParam: jest.fn(),
+}));
+
+describe('getConfiguredPinpointClient', () => {
+  const regionMap = getPinpointRegionMapping();
+  test('return mapped region based on config', async () => {
+    const { config } = await getConfiguredPinpointClient();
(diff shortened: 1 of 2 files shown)


kaustavghosh06:
@watilde Could you make sure that this wouldn't break existing projects present in these regions when customers run `amplify update <category>`?

watilde:
Sure, I'll test if this affects existed resources!

watilde:
I just tested in my local and this changes does not seem to affect existed resources.

```console
$ amplify-dev update analytics
? Please select from one of the below mentioned services: Amazon Pinpoint
Update functionality not available for this service
```

cjihrig:
I think we will need to put region mapping changes like this behind a feature flag. I just performed the following test:

- `amplify-dev add notifications` where the Pinpoint app was mapped from `ap-northeast-1` to `us-west-2` (the current behavior). I enabled the SMS channel.
- Updated the mapping so that `ap-northeast-1` maps to `ap-northeast-1` (the behavior of this PR).
- `amplify-dev update notifications` and remove the SMS channel. This lead to an exception: `NotFoundException: Resource not found`.
- Restored the region mapping of `ap-northeast-1` to `us-west-2`.
- `amplify-dev update notifications` and remove the SMS channel. This time, the SMS channel was removed successfully.

cc: @kaustavghosh06

watilde:
Note - User in Twitter faced the same issue and we can fix by this PR.

Refs: https://twitter.com/otobou0320/status/1383665139241545729

watilde:
Note - We keep hearing requests for this change from developers in Japan. It'd be great if we can ship mine or @cjihrig's update.

cjihrig:
@watilde thanks for the PR. I'm going to close this in favor of https://github.com/aws-amplify/amplify-cli/pull/7024.

Issue #6450, opened 2021-01-21
Different error behavior for Amplify mock vs real API: "Cannot return null for non-nullable field"


Nxtra:
**Describe the bug**
Mock API throws error on non-nullable field of nested object when connection does not exist, real API works fine.

**Amplify CLI Version**
4.41.2

**To Reproduce**
* I use the following `schema.graphql`: 
```
type Writer @model @auth(rules: [{ allow: owner }]) {
  id: ID!
  name: String!
  addressLine: String
  blogs: [Blog] @connection(keyName: "byWriter", fields: ["id"])
}

type Blog
  @model
  @auth(rules: [{ allow: owner }])
  @key(name: "byWriter", fields: ["blogWriterId"]) {
  id: ID!
  name: String
  posts: [Post] @connection(keyName: "byBlog", fields: ["id"])
  blogWriterId: ID
  writer: Writer @connection
}

type Post
  @model
  @auth(rules: [{ allow: owner }])
  @key(name: "byBlog", fields: ["blogID"]) {
  id: ID!
  title: String!
  blogID: ID!
  blog: Blog @connection(fields: ["blogID"])
}
```

*  I run `amplify push --y` to deploy the API
* I start a local mock of the API: `amplify mock`
* Try to execute the following query locally: 
```
mutation createBlog {
  createBlog(input: {name: "Nick's blog"}) {
    id
    name
    posts {
      items {
        blogID
        id
        title
        owner
      }
    }
    writer {
      addressLine
      name
      owner
      id
    }
    owner
  }
}
```
This results in an error:
```
{
  "data": {
    "createBlog": {
      "id": "e8dba61a-1049-4212-9396-ac4d4c5e4123",
      "name": "Nick's blog",
      "posts": {
        "items": []
      },
      "writer": null,
      "owner": "Nxtra"
    }
  },
  "errors": [
    {
      "message": "Cannot return null for non-nullable field Writer.name.",
      "locations": [
        {
          "line": 30,
          "column": 7
        }
      ],
      "path": [
        "createBlog",
        "writer",
        "name"
      ]
    }
  ]
}
```
* Execute the same query in the AppSync console (cloud).
This gives me the following result (no error):
```
{
  "data": {
    "createBlog": {
      "id": "0fa3cb68-949b-499b-b308-7f54235fc469",
      "name": "Nick's blog",
      "posts": {
        "items": []
      },
      "writer": null,
      "owner": "nxtra"
    }
  }
}
```

[...]

PR #6711, opened 2021-02-23
fix(amplify-velocity-template): propagate null block return values

cjihrig:
*Issue #, if available:* #6450

*Description of changes:*
In the velocity template parser, some block statements can return `null` (for example `#return`). The use of the `||` operator was causing `null` to be replaced with the empty string. This commit allows the `null` values to be propagated back up the call stack.

Line 37 in `packages/amplify-velocity-template/src/compile/blocks.js` is the only change. Everything else was changed by automation.


edwardfoyle (approved):
LGTM!

Issue #6675, opened 2021-02-19
Add AppSync perms to lambda function


aprilmintacpineda:
>**Note**: If your feature-request is regarding the AWS Amplify Console service, please log it in the 
[official AWS Amplify Console forum](https://forums.aws.amazon.com/forum.jspa?forumID=314&start=0)

**Is your feature request related to a problem? Please describe.**

I am accessing the graphql api this way:

```js
const AWSAppSyncClient = require('aws-appsync').default;

// on the handler

const client = new AWSAppSyncClient({
  url: process.env.API_WRGDGRAPHQLAPI_GRAPHQLAPIENDPOINTOUTPUT,
  region: process.env.REGION,
  disableOffline: true,
  auth: {
    type: 'AWS_IAM',
    credentials: AWS.config.credentials
  }
});
```

Then I can do:

```js
client.mutate({ mutation, variables });
```

or 

```js
client.query({ quer, variables });
```

But the problem is, if I add the `graphql api` to the resource access list of the lambda function, it doesn't add the other necessary appsync permissions:

```json
{
  "Effect": "Allow",
  "Action": [
    "appsync:GraphQL"
  ],
```

But it needs to be:

```json
{
  "Effect": "Allow",
  "Action": [
    "appsync:Create*",
    "appsync:StartSchemaCreation",
    "appsync:GraphQL",
    "appsync:Get*",
    "appsync:List*",
    "appsync:Update*",
    "appsync:Delete*"
  ],
```

So I have to manually add this, which is a VERY BAD THING because when I do `update function` and change the api access, this will be rewritten, the best way is for amplify to offer a way to add these perms because amplify itself is modifying these files.

**Describe the solution you'd like**

Something like `amplify update function` then `AppSync` is one of the options OR add these perms by default when I give a function `API` access.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

wderezin:
The IAM role permission that amplify cli generate has changed.  I have confirm this was changed as early as `4.41.2` and confirmed it is still broke in `4.44.0` 

Currently the `amplify cli` generate function api permissions in `-cloudformation-template.json` as: 
```json
 {
              "Effect": "Allow",
              "Action": [
                "appsync:GraphQL"
              ],
              "Resource": [
                {
                  "Fn::Join": [
                    "",
                    [
                      "arn:aws:appsync:",
                      {
                        "Ref": "AWS::Region"
                      },
                      ":",
                      {
[...]

PR #6723, opened 2021-02-23
fix(amplify-category-api): mantain ff in iam api policy

SwaySway:
when using the ff for generating the appsync iam policy use the old format if the ff is empty

ref #6675

*Issue #, if available:*

*Description of changes:*


attilah (changes requested):
As the new permissions with real graphql api access does not have their associated e2e test, please add those to make sure whatever policies we generate, the types/mutations/queries will be accessible.

SwaySway:
> As the new permissions with real graphql api access does not have their associated e2e test, please add those to make sure whatever policies we generate, the types/mutations/queries will be accessible.

Updated unit tests to check for the new format and an additional format to check the old format is still generated.

SwaySway:
Discussed offline added e2e tests to check that lambda invoke works with FF enabled to `true` and `false`

edwardfoyle (approved):
LGTM. 👍  on the e2e tests

kylekirkby:
Can we please get this merged and released..? I’m having to do horrible work arounds to be able to access graphql inside a lambda function. This was working fine before :/

PR #6731, opened 2021-02-24
build: update package versions

yuth:
bumped up the package version and updated the import

*Issue #, if available:*

*Description of changes:*


edwardfoyle (commented):
can you also remove this dependency as part of this change: https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-function-plugin-interface/package.json#L20

It's causing a circular dependency and as fas as I can tell it's not used anywhere in the package

edwardfoyle (approved):

attilah (approved):
LGTM

PR #285, opened 2018-10-08
Fixing #183

mikeparisstuff:
*Issue #, if available:*

https://github.com/aws-amplify/amplify-cli/issues/183

*Description of changes:*

This PR fixes #183 and totally revamps the logic for all @auth checks. This also adds extensive new e2e testing suites in addition to all the existing tests that were already in place. The resolver codegen is now much cleaner and more efficient + it works with much more complex rule combinations.

I have also added more docs to this PR https://github.com/aws-amplify/docs/pull/12 that explains the upgraded functionality.

For example, this is from the tests:

```graphql
type AllThree
        @auth(rules: [
            {allow: owner},
            {allow: owner, ownerField: "editors"},
            {allow: groups, groups: ["Admin"]},
            {allow: groups, groups: ["Execs"]},
            {allow: groups, groupsField: "groups"},
            {allow: groups, groupsField: "alternativeGroup"}
        ])
        @model
    {
        id: ID!
        owner: String
        editors: [String]
        groups: [String]
        alternativeGroup: String
    }
```

This enables the following stories:

* The logged in user can CRUD objects of the AllThree type they are the **owner** of the object.
* The logged in user can CRUD objects of the AllThree type if they are one of a set of **editors** on the object.
* The logged in user can CRUD objects of the AllThree type if they are a member of the "Admin" group
* The logged in user can CRUD objects of the AllThree type if they are a member of the "Exec" group
* The logged in user can CRUD objects of the AllThree type if they are a member of any of the groups in **groups**
* The logged in user can CRUD objects of the AllThree type if they are a member of the group specified by **alternativeGroup**

As mentioned above this also implement multi-owner auth that was previously on the back log. This also improves the readability of the generated velocity code which now has helpful comments and better indentation. Here are the generated mapping templates that implement the auth checks for the above type:

**Query.getAllThree.response**

```vtl
## [Start] Static Group Authorization Checks **
#set( $userGroups = $ctx.identity.claims.get("cognito:groups") )
#set( $allowedGroups = ["Admin", "Execs"] )
#set($isStaticGroupAuthorized = $util.defaultIfNull($isStaticGroupAuthorized, false))
#foreach( $userGroup in $userGroups )
  #foreach( $allowedGroup in $allowedGroups )
    #if( $allowedGroup == $userGroup )
      #set( $isStaticGroupAuthorized = true )
    #end
  #end
#end
## [End] Static Group Authorization Checks **


## [Start] Dynamic Group Authorization Checks **
[...]

Issue #183, opened 2018-09-20
Resolver not picking up Cognito user groups


conana:
**Describe the bug**
Problem: Trying to use dynamic groups with @auth, the 'list' mutation returns no data.

**To Reproduce**
Using the default settings for Cognito User Pools as set up by amplify. Using the following model:

```
type TestAuth 
  @auth(rules: [
    {allow: groups, groupsField: "participants", mutations: [update, delete], queries: [get, list]} , 
    {allow: groups, groupsField: "watchers", mutations: [], queries: [get, list]}
  ])
  @model  
  {
    id: ID!
    content: String!
    participants: String
    watchers: String
  }
```

Amplify has generated the resolver for 'ListTestAuths' as follows:
```
## START: Filter Authorized Items. **
#set( $items = [] )
#foreach( $item in $ctx.result.items )
    #set( $isAuthorizedLocal = false )
      #set( $allowedGroups = $item.participants )
      #foreach( $userGroup in $userGroups )
    #if( $util.isList($allowedGroups) )
      #if( $allowedGroups.contains($userGroup) )
        #set( $isAuthorizedLocal = true )
      #end
    #end
    #if( $util.isString($allowedGroups) )
      #if( $allowedGroups == $userGroup )
        #set( $isAuthorizedLocal = true )
      #end
    #end
  #end
    #set( $isAuthorizedLocal = false )
      #set( $allowedGroups = $item.watchers )
      #foreach( $userGroup in $userGroups )
    #if( $util.isList($allowedGroups) )
      #if( $allowedGroups.contains($userGroup) )
        #set( $isAuthorizedLocal = true )
      #end
    #end
    #if( $util.isString($allowedGroups) )
      #if( $allowedGroups == $userGroup )
        #set( $isAuthorizedLocal = true )
      #end
    #end
  #end
  #if( $isAuthorized == true or $isAuthorizedLocal == true || $isAuthorized == true or $isAuthorizedLocal == true )
$util.qr($items.add($item))
  #end
#end
#set( $ctx.result.items = $items )
## END: Filter Authorized Items. **

$util.toJson($ctx.result)
```

Dynamo contains a single record as follows:

| "id (S)" | "__typename (S)" | "content (S)" | "createdAt (S)" | "participants (S)" | "updatedAt (S)" | "watchers (S)" |
| ---- | ---- | ---- | ---- | ---- | ---- | ---- |
[...]

PR #6761, opened 2021-03-01
Only apply $authCondition if $authCondition.expression is specified.

janpapenbrock:
Avoids errors when having an empty $authCondition object initialized.

*Issue #, if available:*
None.

*Description of changes:*

Changing the resolver condition to allow for initializing $authCondition above without necessarily setting $authCondition.expression (e.g. if there are no rules generating such an expression).

**The same change was already done for `makeUpdateResolver`** more than two years ago in this change: https://github.com/aws-amplify/amplify-cli/commit/faff0c3ca39900bcf215efa0cce832b7ccf0a83e#diff-d634883a373d0ee30a418eaa44857d7cefc79c7f3e1cc2d42da40157c3032aa8
I feel like the delete resolver may just have been overlooked.

The two-years-ago change relates to:
- https://github.com/aws-amplify/amplify-cli/issues/183
- https://github.com/aws-amplify/amplify-cli/pull/285

### DIFF
--- a/packages/graphql-dynamodb-transformer/src/resources.ts
+++ b/packages/graphql-dynamodb-transformer/src/resources.ts
@@ -761,7 +761,7 @@ export class ResourceFactory {
       RequestMappingTemplate: print(
         compoundExpression([
           ifElse(
-            ref(ResourceConstants.SNIPPETS.AuthCondition),
+            raw(`$${ResourceConstants.SNIPPETS.AuthCondition} && $${ResourceConstants.SNIPPETS.AuthCondition}.expression != ""`),
             compoundExpression([
               set(ref('condition'), ref(ResourceConstants.SNIPPETS.AuthCondition)),
               ifElse(


attilah:
@janpapenbrock thanks for the contribution, could you please make sure tests are passing by updating the snapshots?

SwaySway:
Hi @janpapenbrock - are you able update the snapshot tests as well?

cjihrig:
Closing due to lack of follow up.

janpapenbrock:
@cjihrig Does it really make sense to leave this bug unfixed? 

I did not follow up, because I'm not sure if it makes sense for me to get all the testing set up locally, when one of the maintainers can probably do it in a couple of minutes.

cjihrig:
@janpapenbrock unfortunately, we can't accept this patch as is because it causes the test suite to fail. That's why it was requested to update the tests as well.

PR #6768, opened 2021-03-02
Use encodeURIComponent over encodeURI

ebisbe:
AppSync utils uses `encodeURIComponent()` and `decodeURIComponent()` over `decodeURI` and `encodeURI`.

*Issue #, if available:* I haven't created the issue because I tested myself in the AppSync. 

*Description of changes:*

Using the context: 
```
{
    "identity": {
        "accountId": "id-accountId",
        "cognitoIdentityPoolId": "id-cognitoIdentityPoolId",
        "cognitoIdentityId": "id-cognitoIdentityId",
        "sourceIp": [
            "x.x.x.x"
        ],
        "username": "12314%40N32",
        "userArn": "my:user:arn"
    }
}
``` 

With the resolver: 
```
{
   ":userId": $util.dynamodb.toStringJson($util.urlDecode($context.identity.username))
}
```

You can see that AppSync decodes `12314%40N32` to `12314@N32`. 

It seems that
**decodeURI()**: It takes encodeURI(url) string so it cannot decoded characters (, / ? : @ & = + $ #)
**decodeURIComponent()**: It takes encodeURIComponent(url) string so it can decode these characters.

### DIFF
--- a/packages/amplify-appsync-simulator/src/velocity/util/general-utils.ts
+++ b/packages/amplify-appsync-simulator/src/velocity/util/general-utils.ts
@@ -14,10 +14,10 @@ export const generalUtils = {
     return jsStringEscape(value);
   },
   urlEncode(value) {
-    return encodeURI(value);
+    return encodeURIComponent(value);
   },
   urlDecode(value) {
-    return decodeURI(value);
+    return decodeURIComponent(value);
   },
   base64Encode(value) {
     // eslint-disable-next-line
--- a/packages/amplify-appsync-simulator/src/velocity/util/general-utils.ts
+++ b/packages/amplify-appsync-simulator/src/velocity/util/general-utils.ts
@@ -14,7 +14,10 @@ export const generalUtils = {
     return jsStringEscape(value);
   },
   urlEncode(value) {
-    return encodeURIComponent(value);
+    // Stringent in adhering to RFC 3986 ( except the asterisk that appsync ingores to encode )
+    return encodeURIComponent(value).replace(/[!'()]/g, function (c) {
+      return '%' + c.charCodeAt(0).toString(16).toUpperCase();
+    });
   },
   urlDecode(value) {
     return decodeURIComponent(value);
--- a/packages/amplify-appsync-simulator/src/__tests__/velocity/util/general-utils.test.ts
+++ b/packages/amplify-appsync-simulator/src/__tests__/velocity/util/general-utils.test.ts
@@ -87,3 +87,16 @@ describe('$utils.toJson', () => {
     expect(generalUtils.toJson(object)).toBe('false');
   });
 });
+
+describe('$util.urlEncode and $util.urlDecode following application/x-www-form-urlencoded specification', () => {
+  // Appsync does not encode the asterisk
+  const reservedChars = "!#$%&'()+,/:;=?@[]";
+  const encodedReservedChars = '%21%23%24%25%26%27%28%29%2B%2C%2F%3A%3B%3D%3F%40%5B%5D';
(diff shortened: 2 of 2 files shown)


cjihrig (commented):
Can you add a test please.

ebisbe:
@cjihrig I can add a test but that file has not test whatsoever and I'm not sure how do you want to handle that case. So what do would you like?

cjihrig:
> @cjihrig I can add a test but that file has not test whatsoever and I'm not sure how do you want to handle that case. So what do would you like?

It looks like that file is tested in `packages/amplify-appsync-simulator/src/__tests__/velocity/util/general-utils.test.ts`. Based on your original report, I think you'd want to add a test that calls `generalUtils.urlEncode()` and `generalUtils.urlDecode()` and verify that the values are correct with these changes.

ebisbe:
@cjihrig Following the https://en.wikipedia.org/wiki/Percent-encoding#The_application.2Fx-www-form-urlencoded_type spec it encodes all the reserved chars `!#$%&'()*+,/:;=?@[]` but [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) does not encode `-_.!~*'()` and they provide the function used in the test.

```
To be more stringent in adhering to RFC 3986 (which reserves !, ', (, ), and *), even though these characters have no formalized URI delimiting uses:

function fixedEncodeURIComponent(str) {
  return encodeURIComponent(str).replace(/[!'()*]/g, function(c) {
    return '%' + c.charCodeAt(0).toString(16);
  });
}
```

packages/amplify-appsync-simulator/src/velocity/util/general-utils.ts
```diff
@@ -14,10 +14,13 @@ export const generalUtils = {
     return jsStringEscape(value);
   },
   urlEncode(value) {
-    return encodeURI(value);
+    // Stringent in adhering to RFC 3986
+    return encodeURIComponent(value).replace(/[!'()*]/g, function (c) {
```
cjihrig, line 18: Can you please remove the asterisk here. AppSync does not appear to encode it. I applied your changes locally and encoded `'!#$%&'()*+,/:;=?@[]'`:

Mock output: `"%21%23%24%25%26%27%28%29%2A%2B%2C%2F%3A%3B%3D%3F%40%5B%5D"`
AppSync output: `"%21%23%24%25%26%27%28%29*%2B%2C%2F%3A%3B%3D%3F%40%5B%5D"`
Mock encoding output with these changes for

cjihrig (approved):
One small change, but LGTM.

packages/amplify-appsync-simulator/src/velocity/util/general-utils.ts
ebisbe: I'm not sure if you want me to move the `*` to the last place so anyone will see that we didn't forget about it. I'll leave a comment also.

Do you want me to rebase into 1 commit after is everything fine?

cjihrig: > I'm not sure if you want me to move the * to the last place so anyone will see that we didn't forget about it. I'll leave a comment also.

I think a comment is enough.

> Do you want me to rebase into 1 commit after is everything fine?

Rebasing would be ideal, but we can also squash on landing.

ebisbe: Sorry I meant squashing all commits into one. 

Ok. I'll leave a comment only then.

Issue #6779, opened 2021-03-03
Cannot finish amplify configure with Admin UI on Windows


watilde:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v14.13.0

### Amplify CLI Version

4.41.2

### What operating system are you using?

Windows, macOS

### Amplify Categories

auth

### Amplify Commands

pull

### Describe the bug

When I pull a backend setup by Admin UI, that shows a login page and makes an error when proceed. The error message was “Failed to communicate with the Amplify CLI”.

### Expected behavior

Should success login.

### Reproduction steps

1. Finish tutorial on https://sandbox.amplifyapp.com/start#datastore
2. Login AWS management console
3. Move to Admin UI
4. Enable Auth from Admin UI
5. Run displayed amplify pull
6. Open browser and login via Admin UI
6. See the error

### Log output

Let me put screenshot instead since the error is captured in browser.
![Error](https://user-images.githubusercontent.com/1716463/109735755-9a947280-7c06-11eb-8348-103b30ada878.JPG)

This issue is related to https://github.com/aws-amplify/amplify-adminui/issues/134. I debug the issue and turned out we do use `express` server in `amplify-provider-awscloudformation` to listen POST request from Admin UI login page. The default of hostname is `localhost`, `0.0.0.0` and `127.0.0.1` on macOS, but Ubuntu on WSL2 on Windows allows only `localhost` by default. Since `127.0.0.1` seems to be coded in Admin UI console, we can specify the hostname in CLI.

Refs: reported by a developer - https://ginpen.com/2021/03/02/failed-to-communicate-with-the-amplify-cli/

attilah:
@watilde Thanks for reporting the issue and also providing the PR, we look into it and it gets into the next release!

PR #6780, opened 2021-03-03
fix(amplify-provider-awscloudformation): specify 127.0.0.1 as hostname

watilde:
CLI uses express server in `amplify-provider-awscloudformation` to listen to POST request from Admin UI login page when `amplify configure`. The default of the hostname is localhost, 0.0.0.0 and 127.0.0.1 on macOS, but Ubuntu on WSL on Windows uses the only localhost by default. Since 127.0.0.1 seems to be coded in Admin UI, we should specify 127.0.0.1 as the hostname in CLI so that CLI can finish `amplify configure`.

Fixes: https://github.com/aws-amplify/amplify-cli/issues/6779
Fixes: https://github.com/aws-amplify/amplify-adminui/issues/134

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/__tests__/utils/admin-login-server.test.ts
+++ b/packages/amplify-provider-awscloudformation/src/__tests__/utils/admin-login-server.test.ts
@@ -30,7 +30,7 @@ jest.mock('express', () => {
 });
 
 describe('AdminLoginServer', () => {
-  test('run server with 127.0.0.1', async () => {
+  test('run server with 0.0.0.0', async () => {
     const adminLoginServer = new AdminLoginServer('appId', 'http://example.com', context_stub.print);
 
     await new Promise<void>(resolve => {
@@ -39,6 +39,6 @@ describe('AdminLoginServer', () => {
     });
     expect(useMock).toBeCalled();
     expect(postMock).toBeCalled();
-    expect(listenMock).toBeCalledWith(4242, '127.0.0.1');
+    expect(listenMock).toBeCalledWith(4242, '0.0.0.0');
   });
 });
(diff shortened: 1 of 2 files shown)


packages/amplify-provider-awscloudformation/src/utils/admin-login-server.ts
```diff
@@ -38,7 +40,13 @@ export class AdminLoginServer {
 
   public async startServer(callback: () => void) {
     await this.setupRoute(callback);
-    this.server = this.app.listen(this.getPort());
+    // Need to specify hostname for WSL
+    this.server = this.app.listen(this.getPort(), this.getHost());
+    this.serverLocalhost = this.app.listen(this.getPort());
+  }
```
jhockett, line 46: Why is a second server needed?

packages/amplify-provider-awscloudformation/src/__tests__/utils/admin-login-server.test.ts
```diff
@@ -0,0 +1,44 @@
...
+  build: jest.fn().mockResolvedValue({ rebuilt: true }),
+} as unknown) as jest.Mocked<FunctionRuntimeLifecycleManager>;
+
+const context_stub = ({
+  amplify: {
+    readBreadcrumbs: jest.fn().mockReturnValue({ pluginId: 'testPluginId' }),
+    loadRuntimePlugin: jest.fn().mockResolvedValue(runtimePlugin_stub),
+    updateamplifyMetaAfterBuild: jest.fn(),
+  },
+} as unknown) as jest.Mocked<$TSContext>;
```
jhockett, line 16: Only `context.print` method is passed to the `AdminLoginServer` constructor, there's no need to mock the extra parts of the `context` object.

jhockett (commented):
Thanks for the contribution @watilde!

yuth (approved):

packages/amplify-provider-awscloudformation/src/utils/admin-login-server.ts
watilde: This is to make diff smaller. We have tests that tries to access to localhost and I did not want to replace them all with 127.0.0.1

packages/amplify-provider-awscloudformation/src/__tests__/utils/admin-login-server.test.ts
watilde: Good point. We could make context smaller! Let me make follow up PR shortly to make the test simple. Thank you!

PR #6797, opened 2021-03-04
feat: added more explicit messaging

ammarkarachi:
*Issue #, if available:*

*Description of changes:*
Added explicit messaging for amplify app creation

### DIFF
--- a/packages/amplify-app/src/index.js
+++ b/packages/amplify-app/src/index.js
@@ -12,7 +12,7 @@ const { engines } = require('../package.json');
 const { initializeAwsExports } = require('amplify-frontend-javascript');
 const { initializeAmplifyConfiguration } = require('amplify-frontend-flutter');
 const { callAmplify } = require('./call-amplify');
-
+import Ora from 'ora';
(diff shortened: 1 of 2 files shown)


packages/amplify-app/src/index.js
```diff
@@ -423,8 +423,10 @@ async function createIosHelperFiles() {
     fs.writeFileSync(amplifyConfigFile, configJsonStr);
   }
 
+  console.log('Checking for existing amplify project...');
   if (fs.existsSync(path.join(amplifyDir, 'backend'))) {
     await addAmplifyFiles();
+    console.log('Generating Amplify configuration files');
```
jhockett, line 429: Since this will print after `await addAmplifyFiles();`, I suggest changing the text, change the order of the statement, or use an ora spinner:

```js
await addAmplifyFiles();
console.log('Generated Amplify configuration files');
```

```js
console.log('Generating Amplify configuration files');
await addAmplifyFiles();
```

```js
// Something like...
spinner.start('Generating Amplify configuration files...');
await addAmplifyFiles();
spinner.succeed();
```

PR #6800, opened 2021-03-05
test: check boolean values instead of truthy/falsy

jhockett:


edwardfoyle (approved):
revert the changes to config.yml (I realize the split test script did it, but it shouldn't be in this PR). Other than that, LGTM

PR #6803, opened 2021-03-05
fix(amplify-provider-awscloudformation): reverse asc events

### LINKED ISSUE #6578: Cloudformation messages are printed in the wrong order
**Describe the bug**

The CLI polls CF for status messages and getting it in chunks. Then we are printing it to the screen as we have received it, but it is in the wrong order, the latest entry is on the top:

image

**Amplify CLI Version**
4.41.2

**To Reproduce**
Push any project, observe the timestamps of the printed messages

**Expected behavior**
Print messages in correct order

watilde:
Fixes #6578

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
@@ -454,7 +454,8 @@ function formatOutputs(outputs) {
 }
 
 function showEvents(events) {
-  events = events.sort((a, b) => new Date(a.Timestamp) > new Date(b.Timestamp));
+  // CFN sorts the events by descending
+  events = events.reverse();
 
   if (events.length > 0) {
     console.log('\n');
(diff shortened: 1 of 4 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -454,7 +454,15 @@ function formatOutputs(outputs) {
 }
 
 function showEvents(events) {
-  events = events.sort((a, b) => new Date(a.Timestamp) > new Date(b.Timestamp));
+  events = events.sort((a, b) => {
```
attilah, line 457: I would not use sort based on timestamp because CFN already sorts the events for us, we just get it in reverse order, and if we get two events with the same timestamp we can hit an issue that we'll still print it in wrong order. Please update to just reverse the entries and it will be ok

attilah (dismissed):
Thanks for the contribution, I added a comment on the implementation, please update it and the it is good to be merged.

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
watilde: Thank you for your review! That is good to know. As that sounds to require domain knowledge, I left that as a comment in the implementation.

edwardfoyle (approved):

PR #7005, opened 2021-04-01
fix(graphql-key-transformer): optional id field if id not primary key

### LINKED ISSUE #6630: "Update" operation force input unnecessary fields
**Describe the bug**
it should only require the base table's primary key when running an update operation.

**To Reproduce**
```gql
type Review @model(subscriptions: { level: off })
@key(fields: ["owner", "serviceId"])
@key(name: "byService", fields: ["serviceId", "createdAt"], queryField: "listReviewsByService")
@key(name: "byStatus", fields: ["owner", "status", "createdAt"], queryField: "listReviewsByStatus")
@key(name: "byId", fields: ["id"], queryField: "listReviewsById")
{
    id: ID!
    serviceId: ID!
    owner: String!
    rating: Int
    title: String
    status: String
    createdAt: AWSDateTime!
}
```

↑gets this generated:
```gql
input UpdateReviewInput {
  id: ID!
  serviceId: ID!
  owner: String!
  rating: Int
  title: String
  status: String
  createdAt: AWSDateTime
}

input CreateReviewInput {
  id: ID
  serviceId: ID!
  owner: String!
  rating: Int
  title: String
  status: String
  createdAt: AWSDateTime
}

type Mutation {
  updateReview(input: UpdateReviewInput!, condition: ModelReviewConditionInput): Review
  ...
}
```

in the above example, the `id` filed is unnecessary in `UpdateReviewInput`, it should be the same as `CreateReviewInput`.

sharmaaditya570191:
#### Description of changes

The `graphql-key-transformer` was making the `id` field required in the schema built after running the command `amplify api gql-compile`. This happens in case when the `id` field is defined as a secondary index in the `graphql.schema`.

#### Issue #, if available

Fixes #6630 

Added a unit test for the case when `id` is defined as secondary index in `graphql` schema

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1077,14 +1077,6 @@ jobs:
     environment:
       TEST_SUITE: src/__tests__/plugin.test.ts
       CLI_REGION: ap-northeast-1
-  schema-iterative-update-locking-amplify_e2e_tests:
-    working_directory: ~/repo
-    docker: *ref_1
-    resource_class: large
-    steps: *ref_4
-    environment:
-      TEST_SUITE: src/__tests__/schema-iterative-update-locking.test.ts
-      CLI_REGION: ap-southeast-1
   migration-node-function-amplify_e2e_tests:
     working_directory: ~/repo
     docker: *ref_1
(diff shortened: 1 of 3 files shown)


sharmaaditya570191:
@SwaySway can you have a look please?

ammarkarachi:
CC: @SwaySway

SwaySway (commented):
Changes look good!
Can we also include this for the delete resolver as well when the primary key is different?

sharmaaditya570191:
> Can we also include this for the delete resolver as well when the primary key is different?

Sure.

sharmaaditya570191:
@SwaySway I checked the delete input in the build schema. It is getting generated properly. I think the problem was only with update input types.

SwaySway (approved):
LGTM

PR #7012, opened 2021-04-02
fix(amplify-provider-awscloudformation): add slow down on index check

SwaySway:
#### Description of changes

cfn is async to ddb gsi creation slowing the api call can wait for gsi operation to start before
checking a table with potentially wrong info

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/iterative-deployment/deployment-manager.ts
+++ b/packages/amplify-provider-awscloudformation/src/iterative-deployment/deployment-manager.ts
@@ -230,7 +230,7 @@ export class DeploymentManager {
     }
   };
 
-  private getTableStatus = async (tableName: string, region: string): Promise<boolean> => {
+  private getTableStatus = async (tableName: string): Promise<boolean> => {
     assert(tableName, 'table name should be passed');
 
     const response = await this.ddbClient.describeTable({ TableName: tableName }).promise();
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/src/iterative-deployment/deployment-manager.ts
```diff
@@ -239,13 +239,17 @@ export class DeploymentManager {
   };
 
   private waitForIndices = async (stackParams: DeploymentMachineOp) => {
-    if (stackParams.tableNames.length) console.log('\nWaiting for DynamoDB table indices to be ready');
+    if (stackParams.tableNames.length) {
+      console.log('\nWaiting for DynamoDB table indices to be ready');
+      // cfn is async to ddb gsi creation the api can return true before the gsi creation starts
+      await new Promise(res => setTimeout(res, 2000));
```
ammarkarachi, line 245: Is this timeout good for all cases?

cjihrig, line 245: This reduces the chance of encountering the race condition, but doesn't remove it, right?

ammarkarachi (approved):

packages/amplify-provider-awscloudformation/src/iterative-deployment/deployment-manager.ts
SwaySway: Overall it's good to have a wait time when tableNames are passed in. When tables are deleted it's not included in the list as it doesn't need a wait time.

SwaySway: Correct, though afaik there is no method to check that the request was made before checking table status. From testing I can say this does happen in very specific scenarios. So far I was only able to repro this by doing a multiple table GSI update where one of the tables had a GSI removed prior and on the next deployment a GSI was added back with same name.

PR #7023, opened 2021-04-06
chore(amplify-provider-awscloudformation): update netmask dependency

jhockett:
#### Description of changes

Update netmask dependency to version 2.0.2


cjihrig (approved):
The changes here LGTM. It looks like the old version is still in the dependency tree via another dependency.

PR #7024, opened 2021-04-06
fix(amplify-cli-core): add service mapping FFs

cjihrig:
#### Description of changes
This commit adds feature flags to enable services (pinpoint, etc.) in regions where support has been added. If support has not been added yet, the code continues to fall back to the existing region mapping.

#### Issue #, if available
https://github.com/aws-amplify/amplify-cli/issues/5287

Manual testing. For example, create a new project in the Tokyo region using a released version of the CLI. Using `amplify-dev`, add the SMS notification channel. Verify that a pinpoint app was created in us-west-2, and not ap-northeast-1. Delete the SMS channel and verify that no errors occur. Delete the app.

Create another app in Tokyo using `amplify-dev`. Repeat the process, but note that the pinpoint app is created in ap-northeast-1.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] Relevant documentation is changed or added (aws-amplify/docs#3596)


packages/amplify-cli-core/src/feature-flags/featureFlags.ts
```diff
@@ -612,5 +612,51 @@ export class FeatureFlags {
         defaultValueForNewProjects: true,
       },
     ]);
+
+    this.registerFlag('latestRegionSupport', [
```
jhockett, line 616: Since we may run into this again as we still map several regions, I think these feature flags might make more sense as version numbers instead of booleans. That way we don't have a latestlatestPinpointRegions flag in the future.

cjihrig: That's a good idea, thanks. When you say version numbers, are you just thinking an integer value that we increment each time the mappings change?

jhockett (approved):
LGTM

packages/amplify-cli-core/src/feature-flags/featureFlags.ts
kaustavghosh06: We usually use true and false for the feature-flags correct? Or is it 1 or 0?

packages/amplify-provider-awscloudformation/src/aws-predictions-regions.js
```diff
@@ -1,15 +1,22 @@
+const { FeatureFlags } = require('amplify-cli-core');
+const latestTranslateRegions = FeatureFlags.getNumber('latestRegionSupport.translate');
```
kaustavghosh06, line 2: for existing projects this value would be undefined correct? Just make sure that the ternary operations below are not impacted because of that :D

cjihrig: Shouldn't it pick up the `defaultValueForExistingProjects` value?

kaustavghosh06 (approved):

jhockett:
@cjihrig when you get a chance, please rebase and include a link to a docs PR that adds the feature flag to https://docs.amplify.aws/cli/reference/feature-flags/

cjihrig:
@jhockett rebased. Docs PR at https://github.com/aws-amplify/docs/pull/3596

PR #7035, opened 2021-04-07
fix(graphql-auth-transformer): fixes @auth directives for Admin UI

hsspain:
#### Description of changes

This fixes a scenario in with the Manage - Content section Admin UI no longer functioned in specific scenarios due to authorization errors from the AppSync API.

## Issue

If a user deployed an API with models that had less than full @auth operations directives (blank or ["create", "read", "update, "delete"]], the generated schema would not include the @aws_iam directive on the mutations as needed by the Manage - Content functionality.

For instance, if the schema looked like this:

```
type Post @model @auth(rules: [{allow: public, operations: [read, update]}]) {
      id: ID!
      title: String!
      createdAt: String
      updatedAt: String
  }
```

The transformed schema would have an update mutation that looked like this

`updatePost(input: CreatePostInput!): Post`

instead of what it needed to look like to authenticate properly

 `updatePost(input: CreatePostInput!): Post @aws_api_key @aws_iam`'

This would ultimately present itself with the Manage - Content section of the Admin UI failing to initialize DataStore or with failures to delete or update objects on the affected data model.

## Fix

This changes the order in which the directives are transformed, to add the requisite `iam` @auth rules to the underlying schema before the rules are split into operations.

Because the operation split had been happening before the `iam` rule was added, the mutations were ignored when adding proper directives to the transformed schema.

Unit tests were added for the affected scenarios. Use cases were also tested by using the development branch CLI to push schema.graphql changes directly to a previously broken app.  The Manage - Content section was verified to work.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are added (https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)


kaustavghosh06 (approved):

SwaySway (changes requested):
Doing this change will add this rule `{allow: 'private', provider: 'iam'}` into operation rules since the split happens after we do the amplify admin check. When passed into the mutation functions it'll add the directive and the policy into the auth role.

Example if you have the following schema
```gql
type Post @model @auth(rules: [
  { allow: private, operations: [read, update], provider: iam }]) {
  id: ID!
  title: String
  description: String
  createdAt: AWSDateTime
  updatedAt: AWSDateTime
}
```
It will generate a policy to allow private (in this case the auth role) access to create a post.

Can we do the admin check in `addDirectivesToOperation` and add a unit test to verify the policy generation? This way we don't create unwanted policies.

hsspain:
Don’t we need that policy to allow the IAM users of the Admin UI to manage content?

akshbhu:
> Doing this change will add this rule `{allow: 'private', provider: 'iam'}` into operation rules since the split happens after we do the amplify admin check. When passed into the mutation functions it'll add the directive and the policy into the auth role.
> 
> Example if you have the following schema
> 
> ```graphql
> type Post @model @auth(rules: [
>   { allow: private, operations: [read, update], provider: iam }]) {
>   id: ID!
>   title: String
>   description: String
>   createdAt: AWSDateTime
>   updatedAt: AWSDateTime
> }
> ```
> 
> It will generate a policy to allow private (in this case the auth role) access to create a post.
> 
> Can we do the admin check in `addDirectivesToOperation` and add a unit test to verify the policy generation? This way we don't create unwanted policies.

Won't this add the rule and policy if AdminUI is enabled? I think this is fine. 
Do we need to check for duplicates when adding this rule if AdminUI is enabled?

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -2265,7 +2285,10 @@ found '${rule.provider}' assigned.`,
     };
 
     // Get and validate the auth rules.
-    return getArg('rules', []) as AuthRule[];
+    const authRules = getArg('rules', []) as AuthRule[];
+
+    // All the IAM auth rules that are added using @auth directive need IAM policy to be generated. AuthRules added for AdminUI don't
+    return authRules.map(rule => (rule.provider === 'iam' ? { ...rule, generateIAMPolicy: true } : rule));
```
SwaySway, line 2291: Should we have this set to assume the transformer is creating the IAM Policy unless specified?

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1050,26 +1062,31 @@ operations will be generated by the CLI.`,
...
+   * @param updateSchema flag to indicate if protection needs to update the schema or just the resolver. When a connection is being protected
+   * the schema should not be updated
    */
   private protectGetQuery(
     ctx: TransformerContext,
     resolverResourceId: string,
     rules: AuthRule[],
     parent: ObjectTypeDefinitionNode | null,
     modelConfiguration: ModelDirectiveConfiguration,
+    updateSchema: boolean,
```
SwaySway, line 1076: Is there a reason we only need this on the GET query?

yuth: We use the same thing for connected models and that causes un desired update to schema.

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1176,6 +1195,7 @@ operations will be generated by the CLI.`,
     parent: ObjectTypeDefinitionNode | null,
     modelConfiguration: ModelDirectiveConfiguration,
     explicitOperationName: string = undefined,
+    updateSchema: boolean = true,
```
SwaySway, line 1198: nit: should include same description for `updateSchema` argument used in `protectGetQuery`

SwaySway (approved):
aside from nit LGTM!

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
yuth: Updated the doc string

PR #7053, opened 2021-04-10
fix(amplify-provider-awscloudformation): fix tests failing due to system-config-manager.js

rajrajhans:
#### Description of changes
Optional chaining operator is used in [packages/amplify-provider-awscloudformation/src/system-config-manager.js](https://github.com/aws-amplify/amplify-cli/blob/ebdaa5981004b86e2f84b94d25fc84b25325ee18/packages/amplify-provider-awscloudformation/src/system-config-manager.js#L294). However, since it is a JS file, and not a TS file, it isn't being processed by Babel and so some tests are failing ->

image

This PR converts `packages/amplify-provider-awscloudformation/src/system-config-manager.js` to a `.ts` file (as suggested by @cjihrig)

`yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


cjihrig:
Instead of adding more complexity, can we just change `system-config-manager.js` to a TypeScript file? Something like [this commit](https://github.com/cjihrig/amplify-cli/commit/dc4bd322fdffcfd17060e26e3a824845164e36e9), for example.

rajrajhans:
> 
> 
> Instead of adding more complexity, can we just change `system-config-manager.js` to a TypeScript file? Something like [this commit](https://github.com/cjihrig/amplify-cli/commit/dc4bd322fdffcfd17060e26e3a824845164e36e9), for example.

Oh, I didn't know that changing to `.ts` will fix it. Thanks.

packages/amplify-provider-awscloudformation/src/system-config-manager.ts
```diff
@@ -342,13 +341,4 @@ function getNamedProfiles() {
...
-
-module.exports = {
-  setProfile,
-  getProfiledAwsConfig,
-  getProfileCredentials,
-  getProfileRegion,
-  getNamedProfiles,
-  resetCache,
-};
+}
```
jhockett, line 344: nit: add a newline to the end of file

jhockett (approved):
1 tiny nitpick, but LGTM

PR #7098, opened 2021-04-15
fix(amplify-cli): remove redundant prompt #6535

### LINKED ISSUE #6535: "amplify env add" asks "Do you want to use an existing environment?"
When I enter `amplify env add` or `amplify env add <env name>` it first asks "Do you want to use an existing environment?" If I say yes, and select an environment, it will just use that environment and not create a new environment. I personally believe this is counter-intuitive. It should immediately just ask you for the name of the new environment. If I wanted to use an existing environment, I would run `amplify env checkout`.

Also, if you provide an environment name to `amplify env add` (e.g. `amplify env add stg`) it should be able to immediately add that environment and skip the environment name prompt.

**Amplify CLI Version**
4.41.2

**To Reproduce**
Simply create an Amplify project and enter `amplify env add` to add a new environment.

**Expected behavior**
It asks "Enter a name for the environment" and rest of the steps proceeds as expected for adding new environment. (selecting AWS profile, etc.)

- OS: Windows 10
- Node Version: v12.18.4

gitaalekhyapaul:
#### Description of changes

Improves the `amplify env add` workflow.
Changes made:

1. Skip the `Do you want to use an existing environment?` question when `amplify env add` is executed.

Screenshot from 2021-04-15 12-45-59

2. On passing environment name in `amplify env add <envName>`, the new environment is directly added.

Screenshot from 2021-04-15 12-47-02

#### Issue #, if available

Fixes #6535 

`yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
+++ b/packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
@@ -254,8 +254,9 @@ async function getEnvName(context: $TSContext) {
     await newEnvQuestion();
   } else {
     const allEnvs = context.amplify.getAllEnvs();
+    const envAddExec = checkEnvAddExec(context);
 
-    if (allEnvs.length > 0 && addNewEnv(context).exec === false) {
+    if (allEnvs.length > 0 && envAddExec === false) {
       if (await context.amplify.confirmPrompt('Do you want to use an existing environment?')) {
         const envQuestion: inquirer.ListQuestion = {
           type: 'list',
@@ -268,8 +269,8 @@ async function getEnvName(context: $TSContext) {
       } else {
         await newEnvQuestion();
       }
-    } else if (addNewEnv(context).exec === true && addNewEnv(context).envName) {
-      envName = addNewEnv(context).envName;
+    } else if (envAddExec === true && context.parameters.first) {
+      envName = context.parameters.first;
     } else {
       await newEnvQuestion();
     }
@@ -314,27 +315,14 @@ function getDefaultEditor() {
 }
 
 /**
- *Checks if `amplify env add` has been executed, and extract environment name if exists
+ *Checks if `amplify env add` has been executed
  * @param {$TSContext} context The Amplify context object
- * @returns `{exec: boolean, envName?: string | null}`
- *
- * `exec` denotes if `amplify env add` has been executed
- *
- * `envName` extracts environment name if provided
+ * @returns `boolean`
  */
-function addNewEnv(context): { exec: boolean; envName?: string | null } {
+function checkEnvAddExec(context): boolean {
   if (context.parameters.command === 'env' && context.parameters.array[0] === 'add') {
-    let envName = null;
-    if (context.parameters.first) {
-      envName = context.parameters.first;
-    }
-    return {
-      exec: true,
-      envName,
-    };
+    return true;
   } else {
-    return {
-      exec: false,
-    };
+    return false;
   }
 }
--- a/packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
+++ b/packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
@@ -315,14 +315,10 @@ function getDefaultEditor() {
 }
 
 /**
- *Checks if `amplify env add` has been executed
+ * Checks if `amplify env add` has been executed
  * @param {$TSContext} context The Amplify context object
  * @returns `boolean`
  */
 function checkEnvAddExec(context): boolean {
-  if (context.parameters.command === 'env' && context.parameters.array[0] === 'add') {
-    return true;
-  } else {
-    return false;
-  }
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -297,3 +299,29 @@ function getDefaultEditor() {
...
+/**
+ *Checks if `amplify env add` has been executed, and extract environment name if exists
+ * @param {$TSContext} context The Amplify context object
+ * @returns `{exec: boolean, envName?: string | null}`
+ *
+ * `exec` denotes if `amplify env add` has been executed
+ *
+ * `envName` extracts environment name if provided
+ */
+function addNewEnv(context): { exec: boolean; envName?: string | null } {
```
cjihrig, line 312: Can you please give this a different name. It currently seems like this function would add a new environment.

cjihrig, line 312: It seems like this function can just return a boolean (the value of `exec`), and the caller can use `context.parameters.first` to check for the environment name.

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -242,7 +242,7 @@ async function getEnvName(context) {
   } else {
     const allEnvs = context.amplify.getAllEnvs();
 
-    if (allEnvs.length > 0) {
+    if (allEnvs.length > 0 && addNewEnv(context).exec === false) {
```
cjihrig, line 245: Instead of calling `addNewEnv()` repeatedly, please store the result in a variable.

ammarkarachi:
@cjihrig

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -297,3 +300,16 @@ function getDefaultEditor() {
 
   return localEnvInfo.defaultEditor;
 }
+
+/**
+ *Checks if `amplify env add` has been executed
```
cjihrig, line 305:
```suggestion
 * Checks if `amplify env add` has been executed
```

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -297,3 +300,16 @@ function getDefaultEditor() {
...
+ * @param {$TSContext} context The Amplify context object
+ * @returns `boolean`
+ */
+function checkEnvAddExec(context): boolean {
+  if (context.parameters.command === 'env' && context.parameters.array[0] === 'add') {
+    return true;
+  } else {
+    return false;
+  }
```
cjihrig, line 314:
```suggestion
  return context.parameters.command === 'env' && context.parameters.array[0] === 'add'
```

edwardfoyle (changes requested):
The E2E test flows here (and maybe some other places too) need to be updated with this change
https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-e2e-tests/src/environment/env.ts

gitaalekhyapaul:
> The E2E test flows here (and maybe some other places too) need to be updated with this change
> https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-e2e-tests/src/environment/env.ts

@edwardfoyle I have been trying to execute the `env.test.ts`, but the test is failing with the following error code: `1`. As per the code, it is not a timeout error, since the timeout error code is `EXIT_CODE_TIMEOUT = 2` and not even any generic error since the error code for generic error is `EXIT_CODE_GENERIC_ERROR = 3`. I am not sure how to resolve this.

Is there any specific situations where the child procccess will return `1`?

Screenshot from 2021-05-11 13-36-50

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
gitaalekhyapaul: Resolved!

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
gitaalekhyapaul: Resolved!

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
gitaalekhyapaul: Resolved!

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
gitaalekhyapaul: Resolved!

gitaalekhyapaul: Resolved!

edwardfoyle (approved):

Issue #2227, opened 2019-09-09
Feature request: Custom env variables in lambdas


thedgbrt:
The update mentioned in [684](https://github.com/aws-amplify/amplify-cli/issues/684#issuecomment-497212548) allows lambdas to access env vars generated by other amplify resources. However we still need a way to add custom environment variables to a project. I have a lambda that interacts with an external service via an API Key, I need a simple way to declare it through the amplify-cli.

I tried setting it manually from the lambda page on the AWS Console. However if I later update the function permissions to access different resources via amplify (and therefore have amplify attach new environment variables), it erases it on the next push.

### Possible api for it
```
amplify add variable
? variablename **MyCustomEnvVariable**
? value **TheValue**
```

and then optin from the lambda:
```
amplify function update
? Please select the lambda function you would want to update **mylambda**
? Do you want to update permissions granted to this Lambda function to perform on other resources in your project?  **Yes**
? Select the category **variables**
? Select the variable name **MyCustomEnvVariable**
```

It would then become available in the lambda as process.env.MyCustomEnvVariable.

Secrets would need to be defined per `amplify env` environments. When switching environments, there could be a warning to mention that there are missing variables set up:
```
amplify env checkout staging
**warning: you haven't defined a value for MyCustomEnvVariable in this environment yet, make sure to call 'amplify variable update'***
Successfully initiated staging environment.
```

### Alternative considered
I considered AWS Secret manager as an alternative but:
1. it is not integrated into amplify, unnecessarily complexifying my workflow
2. it is overkill / expensive for simple API Keys that don't require advanced security

Please let me know if something like this is doable or not. If it can be considered by the amplify team, I can take responsibility for setting up an RFC and then speccing out the whole thing.

diego-palmeira:
+1
I need a way to add env variables to my lambdas using amplify and per environment (dev, pre, pro)
Thank you

jonoirwinrsa:
Not a built in solution but you can inject some values using bash + JQ in the CI

    jq '.Resources.LambdaFunction.Properties.Environment.Variables = .Resources.LambdaFunction.Properties.Environment.Variables + {"Your key": "your value"}' your-func-cloudformation-template.json > "$tmp" && mv "$tmp" your-func-cloudformation-template.json

dtelaroli:
+1

dansmith65:
+1

StefanV85:
+1

Lasim:
+1

[...]

PR #7101, opened 2021-04-15
feat(amplify-category-function): custom environment variable

watilde:
#### Description of changes
Integrate customization of environment variables for function.

##### Custom environment variable when `amplify function add`
Screen Shot 2021-04-19 at 0 21 40

##### Show list of environment variables when they do not exist
Screen Shot 2021-04-16 at 16 02 38

##### Show list of environment variables
Screen Shot 2021-04-16 at 16 03 42

##### Select operation when environment variable does not exist
Screen Shot 2021-04-16 at 15 48 57

##### Select operation when environment variables exist
Screen Shot 2021-04-16 at 15 21 12

##### Add new environment variable
Screen Shot 2021-04-16 at 15 20 54

##### Cannot add new environment variable with duplicated KEY
Screen Shot 2021-04-18 at 18 38 12

##### Edit existed environment variable
Screen Shot 2021-04-16 at 15 22 19

##### Cannot edit existed environment variable when it is not selected
Screen Shot 2021-04-18 at 18 50 24

##### Remove selected environment variable
Screen Shot 2021-04-16 at 15 27 16

#### Issue #, if available
https://github.com/aws-amplify/amplify-cli/issues/2227

##### 1. Add function
1. `amplify-dev function add`
2. Type `Yes` when cli asks `Do you want to configure advanced settings?`
3. Type `Yes` when cli asks `Do you want to configure environment variables for this function?`
4. Custom environment variables
5. Print `process.env` in `amplify/backend/<created function>/src/index.js`
6. Run `amplify mock function` and select `<created function>`

##### 2. Update function
1. `amplify-dev function update`
2. Select `Environment variables configuration`
3. Custom environment variables
4. Print `process.env` in `amplify/backend/<created function>/src/index.js`
5. Run `amplify mock function` and select `<created function>`

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
@@ -1,0 +1,176 @@
+import inquirer from 'inquirer';
+import _ from 'lodash';
+import { $TSContext } from 'amplify-cli-core';
+import {
+  validKey,
+  getStoredEnvironmentVariables,
+  setEnvironmentVariable,
+  deleteEnvironmentVariable,
+} from '../utils/environmentVariablesHelper';
+
+export const askEnvironmentVariableQuestions = async (context: $TSContext, resourceName: string): Promise<object> => {
+  const environmentVariables = getStoredEnvironmentVariables(context, resourceName);
+  const operation = await selectEnvironmentVariableQuestion(environmentVariables);
+  switch (operation) {
+    case 'add': {
+      const { newEnvironmentVariableKey, newEnvironmentVariableValue } = await addEnvironmentVariableQuestion(environmentVariables);
+      setEnvironmentVariable(context, resourceName, newEnvironmentVariableKey, newEnvironmentVariableValue);
+      return await askEnvironmentVariableQuestions(context, resourceName);
+    }
+    case 'update': {
+      const { newEnvironmentVariableKey, newEnvironmentVariableValue, targetedKey } = await updateEnvironmentVariableQuestion(
+        environmentVariables,
+      );
+      deleteEnvironmentVariable(context, resourceName, targetedKey);
+      setEnvironmentVariable(context, resourceName, newEnvironmentVariableKey, newEnvironmentVariableValue);
+      return await askEnvironmentVariableQuestions(context, resourceName);
+    }
+    case 'remove': {
+      const targetedKey = await removeEnvironmentVariableQuestion(environmentVariables);
+      deleteEnvironmentVariable(context, resourceName, targetedKey);
+      return await askEnvironmentVariableQuestions(context, resourceName);
+    }
+    case 'abort':
+    default:
+      const updatedEnvironmentVariables = getStoredEnvironmentVariables(context, resourceName);
+      return {
+        environmentMap: Object.keys(updatedEnvironmentVariables).reduce(
+          (acc, cur) => ({
+            ...acc,
+            [cur]: { Ref: _.camelCase(cur) },
+          }),
+          {},
+        ),
+      };
+  }
+};
+
+const selectEnvironmentVariableQuestion = async (environmentVariables: object): Promise<'add' | 'update' | 'remove' | 'abort'> => {
+  const { operation } = await inquirer.prompt([
+    {
+      name: 'operation',
+      message: 'Select what you want to do with environment variables:',
+      type: 'list',
+      choices: [
+        {
+          value: 'add',
+          name: 'Add new environment variable',
+        },
+        {
+          value: 'update',
+          name: 'Update existing environment variables',
+          disabled: _.size(environmentVariables) === 0,
+        },
+        {
+          value: 'remove',
+          name: 'Remove existing environment variables',
+          disabled: _.size(environmentVariables) === 0,
+        },
+        {
+          value: 'abort',
+          name: "I'm done",
+        },
+      ],
+    },
+  ]);
+
+  return operation;
+};
+
+const addEnvironmentVariableQuestion = async (
+  environmentVariables: object,
+): Promise<{ newEnvironmentVariableKey: string; newEnvironmentVariableValue: string }> => {
+  const { newEnvironmentVariableKey, newEnvironmentVariableValue } = await inquirer.prompt([
+    {
+      name: 'newEnvironmentVariableKey',
+      message: 'Enter the environment variable name:',
+      type: 'input',
+      validate: input => {
+        if (!validKey.test(input)) {
+          return 'You can use the following characters: a-z A-Z 0-9 _';
+        }
+        if (_.has(environmentVariables, input)) {
+          return `Key "${input}" is used already`;
+        }
+        return true;
+      },
+    },
+    {
+      name: 'newEnvironmentVariableValue',
+      message: 'Enter the environment variable value:',
+      type: 'input',
+      validate: input => {
+        if (input.length <= 2048) {
+          return 'The value must be 2048 characters or less';
+        }
+        return true;
+      },
+    },
+  ]);
+  return {
+    newEnvironmentVariableKey,
+    newEnvironmentVariableValue,
+  };
+};
+
+const updateEnvironmentVariableQuestion = async (
+  environmentVariables: object,
+): Promise<{ newEnvironmentVariableKey: string; newEnvironmentVariableValue: string; targetedKey: string }> => {
+  const { targetedKey } = await inquirer.prompt([
+    {
+      name: 'targetedKey',
+      message: 'Which environment variable do you want to update:',
+      type: 'list',
+      choices: Object.keys(environmentVariables),
+    },
+  ]);
+
+  const { newEnvironmentVariableKey, newEnvironmentVariableValue } = await inquirer.prompt([
+    {
+      name: 'newEnvironmentVariableKey',
+      message: 'Enter the environment variable name:',
+      type: 'input',
+      validate: input => {
+        if (!validKey.test(input)) {
+          return 'You can use the following characters: a-z A-Z 0-9 _';
+        }
+        if (_.has(environmentVariables, input) && input !== targetedKey) {
+          return `Key "${input}" is used already`;
+        }
+        return true;
+      },
+      default: targetedKey,
+    },
+    {
+      name: 'newEnvironmentVariableValue',
+      message: 'Enter the environment variable value:',
+      type: 'input',
+      validate: input => {
+        if (input.length <= 2048) {
+          return 'The value must be 2048 characters or less';
+        }
+        return true;
+      },
+      default: environmentVariables[targetedKey],
+    },
+  ]);
+
+  return {
+    newEnvironmentVariableKey,
+    newEnvironmentVariableValue,
+    targetedKey,
+  };
+};
+
+const removeEnvironmentVariableQuestion = async (environmentVariables: object): Promise<string> => {
+  const { targetedKey } = await inquirer.prompt([
+    {
+      name: 'targetedKey',
+      message: 'Which environment variable do you want to remove:',
+      type: 'list',
+      choices: Object.keys(environmentVariables),
+    },
+  ]);
+
+  return targetedKey;
+};
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
@@ -100,7 +100,7 @@ const addEnvironmentVariableQuestion = async (
       message: 'Enter the environment variable value:',
       type: 'input',
       validate: input => {
-        if (input.length <= 2048) {
+        if (input.length >= 2048) {
           return 'The value must be 2048 characters or less';
         }
         return true;
@@ -146,7 +146,7 @@ const updateEnvironmentVariableQuestion = async (
       message: 'Enter the environment variable value:',
       type: 'input',
       validate: input => {
-        if (input.length <= 2048) {
+        if (input.length >= 2048) {
           return 'The value must be 2048 characters or less';
         }
         return true;
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
@@ -14,10 +14,12 @@ import {
   resourceAccessSetting,
   cronJobSetting,
   lambdaLayerSetting,
+  environmentVariableSetting,
 } from '../utils/constants';
 import { category } from '../../../constants';
 import { getNewCFNParameters, getNewCFNEnvVariables } from '../utils/cloudformationHelpers';
 import { askExecRolePermissionsQuestions } from './execPermissionsWalkthrough';
+import { askEnvironmentVariableQuestions } from './environmentVariableWalkthrough';
 import { scheduleWalkthrough } from './scheduleWalkthrough';
 import { merge } from '../utils/funcParamsUtils';
 import { tryUpdateTopLevelComment } from '../utils/updateTopLevelComment';
@@ -31,6 +33,7 @@ import {
 } from '../utils/permissionMapUtils';
 import { JSONUtilities, stateManager } from 'amplify-cli-core';
 import { consolidateDependsOnForLambda } from '../utils/consolidateDependsOn';
+import { getStoredEnvironmentVariables } from '../utils/environmentVariablesHelper';
 
 /**
  * Starting point for CLI walkthrough that generates a lambda function
@@ -78,6 +81,11 @@ export async function createWalkthrough(
 
     // ask lambda layer questions and merge in results
     templateParameters = merge(templateParameters, await addLayersToFunctionWalkthrough(context, templateParameters.runtime));
+
+    // ask environment variable questions and merge in results
+    if (await context.amplify.confirmPrompt('Do you want to configure environment variables for this function?', false)) {
+      templateParameters = merge(templateParameters, await askEnvironmentVariableQuestions(context, templateParameters.functionName));
+    }
   }
 
   return templateParameters;
@@ -116,6 +124,18 @@ function provideInformation(context, lambdaToUpdate, functionRuntime, currentPar
     context.print.info('| Not configured');
     context.print.info('');
   }
+
+  // Provide environment variables
+  context.print.success('Environment variables:');
+  const storedEnvironmentVariables = getStoredEnvironmentVariables(context, lambdaToUpdate);
+  if (_.size(storedEnvironmentVariables) !== 0) {
+    _.forEach(storedEnvironmentVariables, (environmentVariableValue, environmentVariableKey) => {
+      context.print.info(`- ${environmentVariableKey}: ${environmentVariableValue}`);
+    });
+  } else {
+    context.print.info('- Not configured');
+  }
+  context.print.info('');
 
   // Provide lambda layer information
   context.print.success('Lambda layers');
@@ -296,6 +316,11 @@ export async function updateWalkthrough(context, lambdaToUpdate?: string) {
   // consolidate dependsOn as above logic is overwriting
   const projectMeta = stateManager.getMeta();
   functionParameters.dependsOn = consolidateDependsOnForLambda(projectMeta, functionParameters.dependsOn, lambdaToUpdate, selectedSettings);
+
+  if (selectedSettings.includes(environmentVariableSetting)) {
+    await askEnvironmentVariableQuestions(context, functionParameters.resourceName);
+  }
+
   return functionParameters;
 }
 
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
@@ -1,6 +1,6 @@
 import path from 'path';
 import _ from 'lodash';
-import { $TSContext, pathManager, JSONUtilities } from 'amplify-cli-core';
+import { $TSContext, stateManager, pathManager, JSONUtilities } from 'amplify-cli-core';
 import { categoryName, functionParametersFileName } from './constants';
 
 export const validKey = new RegExp(/^[a-zA-Z0-9_]+$/);
@@ -104,16 +104,16 @@ const setFunctionEnvVarReference = (resourceName: string, newReferences: object)
 };
 
 const getStoredKeyValue = (context: $TSContext, resourceName: string): object => {
-  const teamProviderInfoPath = pathManager.getTeamProviderInfoFilePath();
+  const projectPath = pathManager.findProjectRoot();
   const { envName } = context.amplify.getEnvInfo();
-  const teamProviderInfo = JSONUtilities.readJson(teamProviderInfoPath, { throwIfNotExist: false }) || {};
+  const teamProviderInfo = stateManager.getTeamProviderInfo(projectPath, { throwIfNotExist: false });
   return _.get(teamProviderInfo, [envName, 'categories', categoryName, resourceName], {});
 };
 
 const setStoredKeyValue = (context: $TSContext, resourceName: string, newKeyValue: object): void => {
-  const teamProviderInfoPath = pathManager.getTeamProviderInfoFilePath();
+  const projectPath = pathManager.findProjectRoot();
   const { envName } = context.amplify.getEnvInfo();
-  const teamProviderInfo = (JSONUtilities.readJson(teamProviderInfoPath, { throwIfNotExist: false }) as object) || {};
+  const teamProviderInfo = stateManager.getTeamProviderInfo(projectPath, { throwIfNotExist: false });
   _.set(teamProviderInfo, [envName, 'categories', categoryName, resourceName], newKeyValue);
-  JSONUtilities.writeJson(teamProviderInfoPath, teamProviderInfo);
+  stateManager.setTeamProviderInfo(projectPath, teamProviderInfo);
 };
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
@@ -46,9 +46,9 @@ export const setEnvironmentVariable = (
   newReferences[newEnvironmentVariableKey] = { Ref: cameledKey };
   newKeyValue[cameledKey] = newEnvironmentVariableValue;
 
-  updateStoredList(resourceName, newList);
-  updateStoredReferences(resourceName, newReferences);
-  updateStoredKeyValue(context, resourceName, newKeyValue);
+  setStoredList(resourceName, newList);
+  setFunctionEnvVarReference(resourceName, newReferences);
+  setStoredKeyValue(context, resourceName, newKeyValue);
 };
 
 export const deleteEnvironmentVariable = (context: $TSContext, resourceName: string, targetedKey: string): void => {
@@ -62,9 +62,9 @@ export const deleteEnvironmentVariable = (context: $TSContext, resourceName: str
   _.unset(newReferences, targetedKey);
   _.unset(newKeyValue, cameledKey);
 
-  updateStoredList(resourceName, newList);
-  updateStoredReferences(resourceName, newReferences);
-  updateStoredKeyValue(context, resourceName, newKeyValue);
+  setStoredList(resourceName, newList);
+  setFunctionEnvVarReference(resourceName, newReferences);
+  setStoredKeyValue(context, resourceName, newKeyValue);
 };
 
 const getStoredList = (resourceName: string): { cloudFormationParameterName: string; environmentVariableName: string }[] => {
@@ -75,7 +75,7 @@ const getStoredList = (resourceName: string): { cloudFormationParameterName: str
   return _.get(functionParameters, 'environmentVariableList', []);
 };
 
-const updateStoredList = (resourceName: string, newList: object): void => {
+const setStoredList = (resourceName: string, newList: object): void => {
   const projectBackendDirPath = pathManager.getBackendDirPath();
   const resourcePath = path.join(projectBackendDirPath, categoryName, resourceName);
   const functionParameterFilePath = path.join(resourcePath, functionParametersFileName);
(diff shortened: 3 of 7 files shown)


packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts
```diff
@@ -0,0 +1,148 @@
...
+const getStoredReferences = (context: any, resourceName: string) => {
+  const projectBackendDirPath = context.amplify.pathManager.getBackendDirPath();
+  const resourcePath = path.join(projectBackendDirPath, categoryName, resourceName);
+  const cfnFileName = `${resourceName}-cloudformation-template.json`;
+  const cfnFilePath = path.join(resourcePath, cfnFileName);
+  const cfnContent = JSONUtilities.readJson(cfnFilePath);
+  return _.get(cfnContent, ['Resources', 'LambdaFunction', 'Properties', 'Environment', 'Variables'], {});
+};
+
+const updateStoredReferences = (context, resourceName, newReferences) => {
```
edwardfoyle, line 32: I would combine this function with `updateStoredParameters` to be `updateCloudFormationTemplate` and have it update both the params and the env var refs. These configurations are tightly coupled and one should never be called without the other.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts
```diff
@@ -0,0 +1,148 @@
+import path from 'path';
```
edwardfoyle, line 1: I think for all of the functions in this file we only need "set" and "delete" operations (I would use "set" instead of "add" because if you try to add an env var with the same name it should overwrite the previous one, not add a new one. So basically I think the exported interface of this module should be something like:
```ts
export const setEnvVar = (name: string, value: string): void;

export const deleteEnvVar = (name: string);
```

Under the hood their should be non-exported functions for setting / removing the relevant config from team-provider-info, function-parameters and the cfn template. What you have is close to this, it just seems a bit over complicated.

edwardfoyle (commented):
Overall looking pretty good. I haven't pulled it and gone through the walkthrough yet but one thing I don't think I saw was any defensive logic for dealing with duplicate env var names. Left some other comments on general code structure below

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts
watilde: I agree that we may not need update function to export from this file. However, we should have `getStoredEnvironmentVariables` function which is used in 3 places. To not hard code `deploymentBucketName` and `s3Key`, we need a matching alrogithm([refs](https://github.com/aws-amplify/amplify-cli/blob/8166855fd74b66bb1bc394719d7ad8796c49a642/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts#L83-L99)) and it's actually a bit complicated.

JFYI - In addition, I also consider if we can have only `syncEnvVar` function to custom environment variables, but turned out it's complicated to not touch `deploymentBucketName` and `s3Key` for the same reson because it requires diff detection alrogthm.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts
watilde: Sounds good! Thank you for your comment :)

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,126 @@
...
+    },
+  ]);
+
+  return targetedKey;
+};
+
+export const askEnvironmentVariableQuestions = async (context, resourceName) => {
+  const environmentVariables = getStoredEnvironmentVariables(context, resourceName);
+  const operation = await selectEnvironmentVariableQuestion(environmentVariables);
+  if (operation === 'add') {
```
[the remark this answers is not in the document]
watilde: Sounds good with block statement! I usually avoid switch-case as it'd doesn't have scope, but find for this case.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts
watilde: I just added validations to take care overwriting. Please find the attached screenshots of `Cannot add new environment variable with duplicated KEY` and `Cannot edit existed environment variable when it is not selected` in the description to see their experiences.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,126 @@
+import inquirer from 'inquirer';
+import _ from 'lodash';
+import {
+  getStoredEnvironmentVariables,
+  updateEnvironmentVariable,
+  addEnvironmentVariable,
+  deleteEnvironmentVariable,
+} from '../utils/environmentVariables';
+
+const selectEnvironmentVariableQuestion = async environmentVariables => {
```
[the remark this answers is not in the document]
watilde: Totally agreed! I captured that in ToDo in the description of this PR. Thank you.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts
```diff
@@ -0,0 +1,148 @@
...
+    if (!_.has(storedKeyValue, cloudFormationParameterName)) return;
+
+    // 4. Take matched environment variables
+    environmentVariables[environmaneVariableName] = storedKeyValue[cloudFormationParameterName];
+  });
+
+  return environmentVariables;
+};
+
+export const addEnvironmentVariable = (context: any, resourceName: string, newEnvironmentVariable: object) => {
```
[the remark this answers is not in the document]
watilde: Sure!

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts
```diff
@@ -0,0 +1,148 @@
+import path from 'path';
+import _ from 'lodash';
+import { JSONUtilities } from 'amplify-cli-core';
+import { loadFunctionParameters } from './loadFunctionParameters';
+import { categoryName, functionParametersFileName } from '../utils/constants';
+
+const getStoredList = (context: any, resourceName: string) => {
+  const projectBackendDirPath = context.amplify.pathManager.getBackendDirPath();
+  const resourcePath = path.join(projectBackendDirPath, categoryName, resourceName);
+  const functionParameters = loadFunctionParameters(context, resourcePath);
```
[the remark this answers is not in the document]
watilde: Renamed!

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariables.ts
```diff
@@ -0,0 +1,148 @@
+import path from 'path';
+import _ from 'lodash';
+import { JSONUtilities } from 'amplify-cli-core';
+import { loadFunctionParameters } from './loadFunctionParameters';
+import { categoryName, functionParametersFileName } from '../utils/constants';
+
+const getStoredList = (context: any, resourceName: string) => {
+  const projectBackendDirPath = context.amplify.pathManager.getBackendDirPath();
```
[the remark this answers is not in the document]
watilde: Good to know, thank you! Done.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -77,6 +80,11 @@ export async function createWalkthrough(
 
     // ask lambda layer questions and merge in results
     templateParameters = merge(templateParameters, await addLayersToFunctionWalkthrough(context, templateParameters.runtime));
+
+    console.log(templateParameters);
```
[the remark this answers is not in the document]
watilde: Done!

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,126 @@
...
+  return targetedKey;
+};
+
+export const askEnvironmentVariableQuestions = async (context, resourceName) => {
+  const environmentVariables = getStoredEnvironmentVariables(context, resourceName);
+  const operation = await selectEnvironmentVariableQuestion(environmentVariables);
+  if (operation === 'add') {
+    const newEnvironmentVariable = await addEnvironmentVariableQuestion(environmentVariables);
+    addEnvironmentVariable(context, resourceName, newEnvironmentVariable);
+    await askEnvironmentVariableQuestions(context, resourceName);
```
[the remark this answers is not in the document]
watilde: Added `abort` handling in switch-case, thank you!

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,126 @@
...
+        },
+        {
+          value: 'abort',
+          name: "I'm done",
+        },
+      ],
+    },
+  ]);
+
+  return operation;
```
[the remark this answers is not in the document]
watilde: Done!

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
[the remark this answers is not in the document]
watilde: I added types to input and output for new functions.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export const setEnvironmentVariable = (context: $TSContext, resourceName: string, newEnvironmentVariable: object): void => {
+  const newList = getStoredList(resourceName);
+  const newReferences = getStoredReferences(resourceName);
+  const newKeyValue = getStoredKeyValue(context, resourceName);
+  _.each(newEnvironmentVariable, (value, key) => {
+    const cameledKey = _.camelCase(key);
+    newList.push({
+      cloudFormationParameterName: cameledKey,
+      environmaneVariableName: key,
```
cjihrig, line 40: Should this be `environmentVariableName` throughout the PR?

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,161 @@
...
+const addEnvironmentVariableQuestion = async (
+  environmentVariables: object,
+): Promise<{ newEnvironmentVariableKey: string; newEnvironmentVariableValue: string }> => {
+  const { newEnvironmentVariableKey, newEnvironmentVariableValue } = await inquirer.prompt([
+    {
+      name: 'newEnvironmentVariableKey',
+      message: 'Enter the environment variable name:',
+      type: 'input',
+      validate: input => {
+        if (input.match(/^[a-zA-Z0-9_]+$/) === null) {
```
cjihrig, line 84: Can you use `RegExp.prototype.test()` here since the result of `match()` is not used.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,161 @@
...
+          return 'Use only /^[a-zA-Z0-9_]+$/';
+        }
+        if (Object.keys(environmentVariables).includes(input)) {
```
cjihrig, line 87: I think you can use `Object.prototype.hasOwnProperty()` here.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,161 @@
...
+    },
+  ]);
+
+  const { newEnvironmentVariableKey, newEnvironmentVariableValue } = await inquirer.prompt([
+    {
+      name: 'newEnvironmentVariableKey',
+      message: 'Enter the environment variable name:',
+      type: 'input',
+      validate: input => {
+        if (input.match(/^[a-zA-Z0-9_]+$/) === null) {
```
cjihrig, line 124: Same comment about using `test()` here.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,161 @@
...
+          return 'Use only /^[a-zA-Z0-9_]+$/';
+        }
+        if (Object.keys(environmentVariables).includes(input) && input !== targetedKey) {
```
cjihrig, line 127: Same comment here about using `hasOwnProperty()` here.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+export const getStoredEnvironmentVariables = (context: $TSContext, resourceName: string): object => {
+  const storedList = getStoredList(resourceName);
+  const storedReferences = getStoredReferences(resourceName);
+  const storedKeyValue = getStoredKeyValue(context, resourceName);
+  const environmentVariables = {};
+
+  // Search for key and take matched
+  // environment variables based on 4 steps
+  storedList.forEach(({ environmaneVariableName, cloudFormationParameterName }) => {
+    // 1. Check if format is correct in function-parmeter.json
```
cjihrig, line 15:
```suggestion
    // 1. Check if format is correct in function-parameter.json
```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (!environmaneVariableName) return;
+    if (!cloudFormationParameterName) return;
+    // 2. Check if key exists in CFn template
+    if (!storedReferences[environmaneVariableName]) return;
+    if (!storedReferences[environmaneVariableName].Ref) return;
+    if (storedReferences[environmaneVariableName].Ref !== cloudFormationParameterName) return;
```
cjihrig, line 21: I think the first two lines here can be dropped if you use optional chaining in the third line.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,161 @@
...
+  environmentVariables: object,
+): Promise<{ newEnvironmentVariableKey: string; newEnvironmentVariableValue: string }> => {
+  const { newEnvironmentVariableKey, newEnvironmentVariableValue } = await inquirer.prompt([
+    {
+      name: 'newEnvironmentVariableKey',
+      message: 'Enter the environment variable name:',
+      type: 'input',
+      validate: input => {
+        if (input.match(/^[a-zA-Z0-9_]+$/) === null) {
+          return 'Use only /^[a-zA-Z0-9_]+$/';
```
edwardfoyle, line 85: give a friendlier error message like `Environment variable names can only contain alphanumeric characters and underscores`

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,161 @@
+import inquirer from 'inquirer';
+import _ from 'lodash';
+import { $TSContext } from 'amplify-cli-core';
+import { getStoredEnvironmentVariables, setEnvironmentVariable, deleteEnvironmentVariable } from '../utils/environmentVariablesHelper';
+
+export const askEnvironmentVariableQuestions = async (context: $TSContext, resourceName: string): Promise<object> => {
```
edwardfoyle, line 6: IMO this function should not actually modify any state. It just just return an object representing the changes such as:
```
{
  envVarsToSet: [
    {
      key: 'TEST',
      value: 'testvalue'
    },
    {
      key: 'ANOTHER',
      value: 'anothertest'
    }
  ],
  envVarsToDelete: [
    'DELETE_ME', 'ALSO_DELETE_ME'
  ]
}
```

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,161 @@
...
+          return `Key "${input}" is used already`;
+        }
+        return true;
+      },
+    },
+    {
+      name: 'newEnvironmentVariableValue',
+      message: 'Enter the environment variable value:',
+      type: 'input',
+      validate: input => input.length <= 2048,
```
edwardfoyle, line 97: Add an error message here that says the value must be 2048 chars or less

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
edwardfoyle: Should pull this test function into a helper function to be DRYer

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -0,0 +1,116 @@
+import path from 'path';
+import _ from 'lodash';
+import { $TSContext, pathManager, JSONUtilities } from 'amplify-cli-core';
+import { categoryName, functionParametersFileName } from './constants';
+
+export const getStoredEnvironmentVariables = (context: $TSContext, resourceName: string): object => {
```
edwardfoyle, line 6: this return type should be `Record<string, string>`. There are several other places where the `object` type is used that should be typed with this as well

edwardfoyle, line 6: IMO this function shouldn't do all the data validation it's doing. It should just read from function-parameters and team-provider-info as those files are the source of truth for the env vars. It doesn't need to cross reference the CFN file or check that the variable names exist. All of this should be handled when we write the files

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+  });
+
+  return environmentVariables;
+};
+
+export const setEnvironmentVariable = (context: $TSContext, resourceName: string, newEnvironmentVariable: object): void => {
+  const newList = getStoredList(resourceName);
+  const newReferences = getStoredReferences(resourceName);
+  const newKeyValue = getStoredKeyValue(context, resourceName);
+  _.each(newEnvironmentVariable, (value, key) => {
```
edwardfoyle, line 36: this doesn't need to be in a loop since newEnvironmentVariable always only contains one env var

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+const getStoredReferences = (resourceName: string): object => {
+  const projectBackendDirPath = pathManager.getBackendDirPath();
+  const resourcePath = path.join(projectBackendDirPath, categoryName, resourceName);
+  const cfnFileName = `${resourceName}-cloudformation-template.json`;
+  const cfnFilePath = path.join(resourcePath, cfnFileName);
+  const cfnContent = JSONUtilities.readJson(cfnFilePath, { throwIfNotExist: false }) || {};
+  return _.get(cfnContent, ['Resources', 'LambdaFunction', 'Properties', 'Environment', 'Variables'], {});
+};
+
+const updateStoredReferences = (resourceName: string, newReferences: object): void => {
```
edwardfoyle, line 93: this function also needs to set the CFN parameter

edwardfoyle, line 93: instead of relying on the caller of these functions to update newReferences properly, it would be better if this function was `setFunctionEnvVarReference = (resourceName: string, envVarName: string, cfnParameter: string)` and this function handled adding the ref and the parameter to the template directly

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -0,0 +1,116 @@
...
+
+const getStoredKeyValue = (context: $TSContext, resourceName: string): object => {
+  const teamProviderInfoPath = pathManager.getTeamProviderInfoFilePath();
+  const { envName } = context.amplify.getEnvInfo();
+  const teamProviderInfo = JSONUtilities.readJson(teamProviderInfoPath, { throwIfNotExist: false }) || {};
+  return _.get(teamProviderInfo, [envName, 'categories', categoryName, resourceName], {});
+};
+
+const updateStoredKeyValue = (context: $TSContext, resourceName: string, newKeyValue: object): void => {
+  const teamProviderInfoPath = pathManager.getTeamProviderInfoFilePath();
```
edwardfoyle, line 111: use `stateManager.getTeamProviderInfo()` and `stateManager.setTeamProviderInfo()`

edwardfoyle (changes requested):
Left some more code structure feedback but the biggest functional thing is the CFN parameters are not being added to the function CFN templates

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,161 @@
...
+    }
+    case 'remove': {
+      const targetedKey = await removeEnvironmentVariableQuestion(environmentVariables);
+      deleteEnvironmentVariable(context, resourceName, targetedKey);
+      return await askEnvironmentVariableQuestions(context, resourceName);
+    }
+    case 'abort':
+    default:
+      const updatedEnvironmentVariables = getStoredEnvironmentVariables(context, resourceName);
+      return Promise.resolve({
```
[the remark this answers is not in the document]
watilde: Right, I removed it. Thanks!

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
watilde: Sounds good. I'd use a similar message with what function name validtion shows then.
Screen Shot 2021-04-22 at 11 18 50

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
watilde: Right, I updated the key name in all places. Thank you :)

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
watilde: Right! Let me use `_.has` which is used in the other palces.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
watilde: That's nice! I updated the error message.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
watilde: Sounds good. I updated the argument to receive key and value and get rid of this loop.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
watilde: I moved this regex into helper and export as validKey. Thanks!

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
watilde: Sounds good! I updated. Thank you.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
watilde: Yes! I update to _.has in this.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
watilde: I think we still should have data validation here since developers expect shown list of editable environment variables should come to `process.env` and manual edition of CFn template can happen.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
watilde: Right, I updated. Thank you!

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
watilde: That's an option for sure, but requires more logics around `function Parameters` in CFn template to update at the same time. The current implementation follows `scheduleWalkthrough`'s way to update properties. Let's disscuss which options to take for this time.

Refs: https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/scheduleWalkthrough.ts#L73

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
watilde: As I commented that injection envVars into `process.env` acctually worked without CFN parameters. Does prameters expect to have all keys of environment variables?

watilde:
Rebased.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -77,6 +80,11 @@ export async function createWalkthrough(
 
     // ask lambda layer questions and merge in results
     templateParameters = merge(templateParameters, await addLayersToFunctionWalkthrough(context, templateParameters.runtime));
+
+    // ask environment variable questions and merge in results
+    if (await context.amplify.confirmPrompt('Do you want to configure environment variables for this function?', false)) {
+      templateParameters = merge(templateParameters, await askEnvironmentVariableQuestions(context, templateParameters.functionName));
```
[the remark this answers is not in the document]
watilde: Apparently, FunctionParameters variable exists only in `updateWalkthrough` and this is `createWalkthrough`.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
watilde: Sounds good. Updated!

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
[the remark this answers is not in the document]
edwardfoyle: FunctionParameters is the type, defined here: https://github.com/aws-amplify/amplify-cli/blob/eb9257eaee117d0ed53ebc23aa28ecd7b7510fa1/packages/amplify-function-plugin-interface/src/index.ts#L109

the type is used in both create and update

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
edwardfoyle: the schedule walkthrough does the entire walkthrough before writing anything to CFN though which we should follow here. It leads to better separation of the walkthrough logic from the state management logic

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
edwardfoyle: yeah that makes sense. in that case, I would recommend putting the validation in it's own function and calling it from here

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
edwardfoyle: it doesn't need the keys, it needs the values (which are in CFN parameters). When an env var is defined as:
```json
{
  "Resources": {
    "LambdaFunction": {
      "Properties": {
        "Environment": {
          "Variables": {
            "MY_ENV_VAR": {
              "Ref": "myEnvVar"
            }
          }
        }
      }
    }
  }
}
```
There must be a parameter in the template called "myEnvVar" that the ref is referencing:
```json
{
  "Parameters": {
    "myEnvVar": {
      "Type": "string"
    }
  }
}
```
Then when Amplify does the push, it has a mechanism for passing in the parameter value from `team-provider-info.json`

watilde:
I'm closing this PR as we re-architect the logic at https://github.com/aws-amplify/amplify-cli/pull/7166

PR #7117, opened 2021-04-16
fix: handle errors and provide better error message when adding data source

cjihrig:
#### Description of changes
This PR fixes two issues related to `amplify api add-graphql-datasource`:

1. Error handling is added so that the spinner does not keep the process alive in the case of an error. This is the first commit, and it wraps existing code in a `try...catch` block.
2. Additional error messaging is provided when the provided secret does not contain valid credentials for Aurora Serverless. This is the second commit.

#### Issue #, if available
#4384

Set up an Aurora Serverless cluster. Attempt to connect using a secret with both valid and invalid credentials.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.js
+++ b/packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.js
@@ -180,8 +180,8 @@ async function selectDatabase(context, inputs, clusterArn, secretArn, AWS) {
     // eslint-disable-next-line prefer-destructuring
     const records = dataApiResult.records;
 
-    for (let i = 0; i < records.length; i += 1) {
-      const recordValue = records[i][0].stringValue;
+    for (const record of records) {
+      const recordValue = record[0].stringValue;
       // ignore the three meta tables that the cluster creates
       if (!['information_schema', 'performance_schema', 'mysql'].includes(recordValue)) {
         databaseList.push(recordValue);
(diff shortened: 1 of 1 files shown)


packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.js
```diff
@@ -167,26 +167,38 @@ async function selectDatabase(context, inputs, clusterArn, secretArn, AWS) {
...
 
-  for (let i = 0; i < records.length; i += 1) {
-    const recordValue = records[i][0].stringValue;
-    // ignore the three meta tables that the cluster creates
-    if (!['information_schema', 'performance_schema', 'mysql'].includes(recordValue)) {
-      databaseList.push(recordValue);
+    // eslint-disable-next-line prefer-destructuring
+    const records = dataApiResult.records;
+
+    for (let i = 0; i < records.length; i += 1) {
```
jhockett, line 183: nit: I know this is how it was before the change, but can we simplify it?

```js
  for (const record of records) {
    const recordValue = record[0].stringValue;
```

jhockett (approved):
LGTM

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.js
cjihrig: Updated!

PR #7140, opened 2021-04-20
feat: automatically update dependent functions on layer version deploy

akshbhu:
#### Description of changes
* added  multi env changes for function
* Introduces new prompt for function to select layer latest version

* added unit tests for the same

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
@@ -350,7 +350,7 @@ export async function updateConfigOnEnvInit(context: $TSContext, resourceName: s
       envParams = await initTriggerEnvs(context, resourceParams, providerPlugin, envParams, srvcMetaData);
     }
 
-    if (resourceParams.lambdaLayers.length) {
+    if (Array.isArray(resourceParams.lambdaLayers) && resourceParams.lambdaLayers.length) {
       const envName = context.amplify.getEnvInfo().envName;
       const modifiedLambdaLayers: LambdaLayer[] = [];
       modifiedLambdaLayers.push(...convertProjectLayersToExternalLayers(resourceParams.lambdaLayers, envName));
@@ -365,7 +365,6 @@ export async function updateConfigOnEnvInit(context: $TSContext, resourceName: s
     const currentAmplifyMeta = stateManager.getCurrentMeta(projectPath);
     const amplifyMeta = stateManager.getMeta(projectPath);
     const currentCloudVersionHash: string = _.get(currentAmplifyMeta, [categoryName, resourceName, 'versionHash'], undefined);
-    //TODO: check if we require version hash when changing environment
     if (currentCloudVersionHash) {
       _.set(amplifyMeta, [categoryName, resourceName, 'versionHash'], currentCloudVersionHash);
     }
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
@@ -1,5 +1,11 @@
 import { $TSAny, $TSContext, JSONUtilities, open, pathManager, stateManager } from 'amplify-cli-core';
-import { FunctionParameters, FunctionTemplate, FunctionTriggerParameters, ProviderContext } from 'amplify-function-plugin-interface';
+import {
+  FunctionParameters,
+  FunctionTemplate,
+  FunctionTriggerParameters,
+  LambdaLayer,
+  ProviderContext,
+} from 'amplify-function-plugin-interface';
 import fs from 'fs-extra';
 import _ from 'lodash';
 import path from 'path';
@@ -19,6 +25,7 @@ import {
   saveMutableState,
   updateLayerArtifacts,
 } from './utils/storeResources';
+import { convertProjectLayersToExternalLayers, covertExternalLayersToProjectLayers } from './utils/convertLayersTypes';
 
 /**
  * Entry point for creating a new function
@@ -343,13 +350,22 @@ export async function updateConfigOnEnvInit(context: $TSContext, resourceName: s
       envParams = await initTriggerEnvs(context, resourceParams, providerPlugin, envParams, srvcMetaData);
     }
 
+    if (resourceParams.lambdaLayers.length) {
+      const envName = context.amplify.getEnvInfo().envName;
+      const modifiedLambdaLayers: LambdaLayer[] = [];
+      modifiedLambdaLayers.push(...convertProjectLayersToExternalLayers(resourceParams.lambdaLayers, envName));
+      modifiedLambdaLayers.push(...covertExternalLayersToProjectLayers(resourceParams.lambdaLayers, envName));
+      resourceParams.lambdaLayers = modifiedLambdaLayers;
+      JSONUtilities.writeJson(functionParametersPath, resourceParams);
+    }
+
     return envParams;
   } else if (isMultiEnvLayer(resourceName) && service === ServiceName.LambdaLayer) {
     const projectPath = pathManager.findProjectRoot();
     const currentAmplifyMeta = stateManager.getCurrentMeta(projectPath);
     const amplifyMeta = stateManager.getMeta(projectPath);
     const currentCloudVersionHash: string = _.get(currentAmplifyMeta, [categoryName, resourceName, 'versionHash'], undefined);
-
+    //TODO: check if we require version hash when changing environment
     if (currentCloudVersionHash) {
       _.set(amplifyMeta, [categoryName, resourceName, 'versionHash'], currentCloudVersionHash);
     }
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
@@ -360,10 +360,10 @@ const addLayerCFNParameters = (context: $TSContext, functionParameters: Partial<
     const resourceName = _.get(layer as ProjectLayer, ['resourceName'], null);
     if (resourceName) {
       const param: string = `function${resourceName}Arn`;
-      if (cfnContent.Parameters[`${param}`] === undefined) {
-        cfnContent.Parameters[`${param}`] = {
+      if (cfnContent.Parameters[param] === undefined) {
+        cfnContent.Parameters[param] = {
           Type: 'String',
-          Default: `${param}`,
+          Default: param,
         };
       }
     }
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
@@ -77,11 +77,6 @@ export async function createWalkthrough(
 
     // ask lambda layer questions and merge in results
     templateParameters = merge(templateParameters, await addLayersToFunctionWalkthrough(context, templateParameters.runtime));
-    if (templateParameters.lambdaLayers) {
-      const projectBackendDirPath = pathManager.getBackendDirPath();
-      const resourceDirPath = path.join(projectBackendDirPath, category, templateParameters.resourceName);
-      addLayerCFNParameters(context, templateParameters, resourceDirPath);
-    }
   }
 
   return templateParameters;
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
@@ -17,7 +17,6 @@ import {
 import { merge } from '../utils/funcParamsUtils';
 import { runtimeWalkthrough, templateWalkthrough } from '../utils/functionPluginLoader';
 import { convertLambdaLayerMetaToLayerCFNArray } from '../utils/layerArnConverter';
-import { layerAccountAccessPrompt } from '../utils/layerHelpers';
 import { loadFunctionParameters } from '../utils/loadFunctionParameters';
 import {
   fetchPermissionCategories,
@@ -353,10 +352,10 @@ export function migrate(context: $TSContext, projectPath: string, resourceName: 
   JSONUtilities.writeJson(cfnFilePath, newCfn);
 }
 
-const addLayerCFNParameters = (context, functionParameters: Partial<FunctionParameters>, resourceDirPath) => {
+const addLayerCFNParameters = (context: $TSContext, functionParameters: Partial<FunctionParameters>, resourceDirPath: string) => {
   const cfnFileName = `${functionParameters.resourceName}-cloudformation-template.json`;
   const cfnFilePath = path.join(resourceDirPath, cfnFileName);
-  const cfnContent: any = JSONUtilities.readJson(cfnFilePath);
+  const cfnContent = JSONUtilities.readJson<$TSAny>(cfnFilePath);
   functionParameters.lambdaLayers.forEach(layer => {
     const resourceName = _.get(layer as ProjectLayer, ['resourceName'], null);
     if (resourceName) {
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
@@ -17,6 +17,7 @@ import {
 import { merge } from '../utils/funcParamsUtils';
 import { runtimeWalkthrough, templateWalkthrough } from '../utils/functionPluginLoader';
 import { convertLambdaLayerMetaToLayerCFNArray } from '../utils/layerArnConverter';
+import { layerAccountAccessPrompt } from '../utils/layerHelpers';
 import { loadFunctionParameters } from '../utils/loadFunctionParameters';
 import {
   fetchPermissionCategories,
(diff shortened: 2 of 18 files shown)


packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambdaLayerWalkthrough.ts
```diff
@@ -1,6 +1,7 @@
 import { $TSContext, exitOnNextTick, ResourceDoesNotExistError } from 'amplify-cli-core';
 import inquirer, { InputQuestion } from 'inquirer';
 import _ from 'lodash';
+import uuid from 'uuid';
```
jhockett, line 4: I don't see this being used, remove if unnecessary

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
```diff
@@ -62,23 +63,34 @@ export const askLayerSelection = async (
...
 
+  let layerVersionArrPrompt: string[];
+
   for await (let layerName of layerSelections) {
-    const currentSelectionDefaults = filterProjectLayers(previousSelections).find(sel => sel.resourceName === layerName);
-    const currentVersion = currentSelectionDefaults ? currentSelectionDefaults.version.toString() : undefined;
     const layerVersions = await loadLayerDataFromCloud(context, layerName);
+    layerVersionArrPrompt = layerVersions.map(layerVersion => layerVersion.Version.toString());
+    if (!layerVersionArrPrompt.length) {
+      layerVersionArrPrompt.push('1'); // new latest version if no layers are pushed
```
jhockett, line 72: We cannot assume the version number, instead, skip the prompt and assume 'Always choose latest version'.

Versions cannot be pinned if they have not been deployed

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const covertExternalLayersToProjectLayers = (lambdaLayers: LambdaLayer[], envName: string): LambdaLayer[] => {
+  const modifiedLambdaLayers: LambdaLayer[] = [];
+  lambdaLayers.forEach(layer => {
+    if (layer.type === 'ExternalLayer' && layer.arn.hasOwnProperty(LAYER_ARN_KEY)) {
+      const layerArn = layer.arn[LAYER_ARN_KEY];
+      const layerNameWithEnv = layerArn.split(':')[layerArn.split(':').length - 2];
+      const layerVersion = parseInt(layerArn.split(':')[layerArn.split(':').length - 1], 10);
+      const layerName = layerNameWithEnv.split('-')[0];
+      const layerEnv = layerNameWithEnv.split('-')[1];
```
jhockett, line 37: Please add a comment with an example ARN so others reading this can have a better understanding

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -372,3 +352,26 @@ export function migrate(context: $TSContext, projectPath: string, resourceName:
 
   JSONUtilities.writeJson(cfnFilePath, newCfn);
 }
+
+const addLayerCFNParameters = (context, functionParameters: Partial<FunctionParameters>, resourceDirPath) => {
+  const cfnFileName = `${functionParameters.resourceName}-cloudformation-template.json`;
+  const cfnFilePath = path.join(resourceDirPath, cfnFileName);
+  const cfnContent: any = JSONUtilities.readJson(cfnFilePath);
```
jhockett, line 359: nit:
```ts
const cfnContent = JSONUtilities.readJson<$TSAny>(cfnFilePath);
```

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -372,3 +352,26 @@ export function migrate(context: $TSContext, projectPath: string, resourceName:
 
   JSONUtilities.writeJson(cfnFilePath, newCfn);
 }
+
+const addLayerCFNParameters = (context, functionParameters: Partial<FunctionParameters>, resourceDirPath) => {
```
jhockett, line 356: nit: add function parameter types

```ts
const addLayerCFNParameters = (context: $TSContext, functionParameters: Partial<FunctionParameters>, resourceDirPath: string) => {
```

packages/amplify-category-function/src/provider-utils/awscloudformation/index.ts
```diff
@@ -343,13 +350,22 @@ export async function updateConfigOnEnvInit(context: $TSContext, resourceName: s
...
+    }
+
     return envParams;
   } else if (isMultiEnvLayer(resourceName) && service === ServiceName.LambdaLayer) {
     const projectPath = pathManager.findProjectRoot();
     const currentAmplifyMeta = stateManager.getCurrentMeta(projectPath);
     const amplifyMeta = stateManager.getMeta(projectPath);
     const currentCloudVersionHash: string = _.get(currentAmplifyMeta, [categoryName, resourceName, 'versionHash'], undefined);
-
+    //TODO: check if we require version hash when changing environment
```
jhockett, line 368: We need to compare the contents of the layer folder to the last known layer version hash of the environment in order to correctly whether there are changes to the newly checked out environment. This is needed for `amplify status` and deploying the correct layer versions on push.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
attilah: Also map will return an empty array so do proper check on length value not just the presence of the property.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts
attilah: Move 'ExternalLayer' to a const value instead of string literal at multiple places

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
attilah: It is not nit ;-) please add!

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
```diff
@@ -62,23 +63,34 @@ export const askLayerSelection = async (
   const askArnQuestion = layerSelections.includes(provideExistingARNsPrompt);
   layerSelections = layerSelections.filter(selection => selection !== provideExistingARNsPrompt);
 
+  let layerVersionArrPrompt: string[];
```
attilah, line 66: this could be const and declared down in line 70 as not used before.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
```diff
@@ -62,23 +63,34 @@ export const askLayerSelection = async (
...
+
+    layerVersionArrPrompt.unshift(defaultlayerVersionPrompt);
     const layerVersionPrompt: ListQuestion = {
       type: 'list',
       name: 'versionSelection',
       message: versionSelectionPrompt(layerName),
-      choices: layerVersions.map(layerVersion => layerVersion.Version.toString()),
-      default: currentVersion,
+      choices: layerVersionArrPrompt,
+      default: defaultlayerVersionPrompt,
```
attilah: casing: defaultLayerVersionPrompt

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts
```diff
@@ -0,0 +1,52 @@
+import { ExternalLayer, LambdaLayer, ProjectLayer } from 'amplify-function-plugin-interface';
+import { convertProjectLayer } from './layerArnConverter';
+
+const LAYER_ARN_KEY = 'Fn::Sub';
+
+export const convertProjectLayersToExternalLayers = (lambdaLayers: LambdaLayer[], envName: string): LambdaLayer[] => {
```
attilah, line 8: based on the caller you can end up with `undefined` for the `lambdaLayers` parameter in case when a function does not have a layer, the caller has a `if (resourceName)` condition and if the code is not going in there then this parameter will be `undefined` that needs to be checked

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (layer.type === 'ProjectLayer' && layer.env !== envName && !layer.isLatestVersionSelected) {
+      const convertLayer: ExternalLayer = {
+        type: 'ExternalLayer',
+        arn: convertProjectLayer(layer, envName),
+      };
+      modifiedLambdaLayers.push(convertLayer);
+    } else {
+      const convertLayer: ProjectLayer = {
+        type: 'ProjectLayer',
+        resourceName: (layer as ProjectLayer).resourceName,
```
attilah, line 20: cast it once and just used the casted instance as this becomes verbose even for 2 usages.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts
```diff
@@ -0,0 +1,52 @@
...
+      };
+      modifiedLambdaLayers.push(convertLayer);
+    }
+  });
+  return modifiedLambdaLayers;
+};
+
+export const covertExternalLayersToProjectLayers = (lambdaLayers: LambdaLayer[], envName: string): LambdaLayer[] => {
+  const modifiedLambdaLayers: LambdaLayer[] = [];
+  lambdaLayers.forEach(layer => {
```
attilah, line 36: same as above, check `lambdaLayers`

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerArnConverter.ts
```diff
@@ -16,13 +19,18 @@ export const convertLambdaLayerMetaToLayerCFNArray = (input: LambdaLayer[], env:
   });
 };
 
-const convertProjectLayer = (layer: ProjectLayer, env?: string) => {
-  if (env) {
+export const convertProjectLayer = (layer: ProjectLayer, env?: string) => {
+  if (!layer.isLatestVersionSelected) {
+    if (env) {
```
attilah, line 24: I am not sure if we need to handle no environment like this as I am not convinced that when you get `envName` that would not fail, please check.

jhockett: The check is for is for legacy non-multienv layers. We should add a comment for clarity

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -372,3 +351,26 @@ export function migrate(context: $TSContext, projectPath: string, resourceName:
...
+  const cfnFilePath = path.join(resourceDirPath, cfnFileName);
+  const cfnContent = JSONUtilities.readJson<$TSAny>(cfnFilePath);
+  functionParameters.lambdaLayers.forEach(layer => {
+    const resourceName = _.get(layer as ProjectLayer, ['resourceName'], null);
+    if (resourceName) {
+      const param: string = `function${resourceName}Arn`;
+      if (cfnContent.Parameters[`${param}`] === undefined) {
+        cfnContent.Parameters[`${param}`] = {
+          Type: 'String',
+          Default: `${param}`,
```
jhockett, line 366: nit: 

```ts
`${param}`
``` 
can just be `param` since it's already a string

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
akshbhu: Added a check on Array to resolve this

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerArnConverter.ts
akshbhu: yeah , we are calling this function in update layer also I guess to support non-multienv layers

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts
akshbhu: This is coming from 
```    if (Array.isArray(resourceParams.lambdaLayers) && resourceParams.lambdaLayers.length) {```

Added an Array check on lambdaLayers so that no undefined is passed

File: packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts (hunk 11 above)
Comment (akshbhu) [reply packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts:39]: same as above

File: packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts (hunk 3 above)
Comment (akshbhu) [reply packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts]: done!

File: packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts (hunk 10 above)
Comment (akshbhu) [reply packages/amplify-category-function/src/provider-utils/awscloudformation/utils/convertLayersTypes.ts]: done

File: packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts [hunk 14]
```diff
@@ -63,23 +64,38 @@ export const askLayerSelection = async (
...
+        resourceName: layerName,
+        version: selectedVersion,
+        isLatestVersionSelected: isLatestVersionSelected,
+        env: context.amplify.getEnvInfo().envName,
+      });
+    } else {
+      lambdaLayers.push({
+        type: 'ProjectLayer',
+        resourceName: layerName,
+        version: 1,
```
jhockett, line 93: We cannot have a hardcoded version, as mentioned in previous comment

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
```diff
@@ -63,23 +64,38 @@ export const askLayerSelection = async (
...
+      const layerVersionPrompt: ListQuestion = {
+        type: 'list',
+        name: 'versionSelection',
+        message: versionSelectionPrompt(layerName),
+        choices: layerVersionArrPrompt,
+        default: defaultLayerVersionPrompt,
+      };
+      const versionSelection = (await inquirer.prompt(layerVersionPrompt)).versionSelection;
+      const isLatestVersionSelected = versionSelection.toString() === defaultLayerVersionPrompt ? true : false;
+      const selectedVersion = versionSelection.toString() === defaultLayerVersionPrompt ? layerVersionArrPrompt[1] : versionSelection;
```
jhockett, line 81: From what I can tell, this just pins the latest known version. We need to depend on the latest output of the layer CFN using `Fn::GetAtt`.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
akshbhu: Here we are alraedy checking if the layer exists using ,loadLayerData from cloud. 
This info needs to stored in functionsParameter.json when we are deploying layer for first time. After first deployment , we populate this information using loadLayerData.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
akshbhu: Yeah , for latest layer , we are using Ref and the parameters is populated using dependsOn section during amplify push.
The version used here is to save ```lambdaLayers``` params in ```function-parameters.json```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
jhockett: can we set the version to undefined or null in that case?

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
jhockett: Thanks for the clarification 👍. Can we store the `defaultLayerVersionPrompt` instead of the version number or will it break things?

akshbhu: done!

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
akshbhu: version is  set to default Prompt for new layers

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
akshbhu: done!

jhockett (approved):

PR #7166, opened 2021-04-23
feat(amplify-category-function): custom env var

watilde:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


edwardfoyle (commented):
Overall I like this approach better. It's more in line with the existing paradigm of how the category is organized. With a little more work to co-locate env var reading / writing logic I think this is the way to go

watilde:
@edwardfoyle Thank you for your comment! Then I'll close my previous PR and move here.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -0,0 +1,174 @@
...
+    }
+    case 'remove': {
+      const targetedKey = await removeEnvironmentVariableQuestion(environmentVariables);
+      delete environmentVariables[targetedKey];
+      return await askEnvironmentVariableQuestions(context, resourceName, environmentVariables);
+    }
+    case 'abort':
+    default:
+      return {
+        environmentMap: Object.keys(environmentVariables).reduce(
```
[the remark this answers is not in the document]
watilde: I think this is nessesary to detect difference in environment variables that user can edit.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -282,12 +299,23 @@ function createParametersFile(parameters: $TSObject, resourceName: string, param
...
-  return { ...parameters.mutableParametersState, ..._.pick(parameters, ['lambdaLayers']) };
+  let envVarList = { environmentVariableList: [] };
+  if (parameters.environmentVariables) {
+    _.each(parameters.environmentVariables, (val, key) => {
+      envVarList.environmentVariableList.push({
+        cloudFormationParameterName: _.camelCase(key),
+        environmentVariableName: key,
+      });
+    });
+  }
```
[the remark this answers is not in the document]
watilde: I took this logic into environment helper util and that helps to re-use the same logic in `function add` and `function update`.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -218,6 +219,22 @@ const createLayerParametersFile = (parameters: LayerParameters | StoredLayerPara
...
+export const saveTeamProviderInfo = (context: $TSContext, parameters: FunctionParameters | FunctionTriggerParameters) => {
+  if (parameters.environmentVariables) {
+    const projectPath = pathManager.findProjectRoot();
+    const { envName } = context.amplify.getEnvInfo();
+    const environmentVariables = {};
+    _.each(parameters.environmentVariables, (value, key) => {
+      environmentVariables[_.camelCase(key)] = value;
+    });
+    const teamProviderInfo = _.merge(
+      stateManager.getTeamProviderInfo(projectPath, { throwIfNotExist: false }),
```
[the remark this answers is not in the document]
watilde: I removed this function and separated the logic into the environment variable util.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/storeResources.ts
```diff
@@ -218,6 +219,22 @@ const createLayerParametersFile = (parameters: LayerParameters | StoredLayerPara
...
 };
 
+export const saveTeamProviderInfo = (context: $TSContext, parameters: FunctionParameters | FunctionTriggerParameters) => {
+  if (parameters.environmentVariables) {
+    const projectPath = pathManager.findProjectRoot();
+    const { envName } = context.amplify.getEnvInfo();
+    const environmentVariables = {};
+    _.each(parameters.environmentVariables, (value, key) => {
+      environmentVariables[_.camelCase(key)] = value;
+    });
```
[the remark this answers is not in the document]
watilde: I removed this function and separated the logic into the environment variable util.

edwardfoyle (commented):
Functionally these changes look good to me. Only thing is I'd like some unit tests for the environmentVariableWalkthrough file and the helpers file

watilde:
@edwardfoyle Thank you for your review. I added tests to make sure the new walkthrough and helper do not throw error.

watilde:
Rebased.

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -220,6 +230,108 @@ function getDefaultEnv(context: $TSContext): string | undefined {
   return undefined;
 }
 
+async function setEnvVar(context: $TSContext, currentEnvName: string, projectPath: string) {
```
edwardfoyle, line 233: all of the env var logic needs to be contained in the functions category package. The `amplify-cli` package shouldn't have knowledge of function env vars. The platform will call [this function](https://github.com/aws-amplify/amplify-cli/blob/23fcfa532b53366fd4d371f4913806cd078f723e/packages/amplify-category-function/src/index.ts#L105) on env initialization which is where the entry point to this logic should be located

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -220,6 +230,108 @@ function getDefaultEnv(context: $TSContext): string | undefined {
...
+
+  const envVarQuestion = async () => {
+    const envVarQuestion: inquirer.ListQuestion = {
+      type: 'list',
+      name: 'envVar',
+      message: 'You have custom function environment variables configured. What do you want to do?',
+      choices: [
+        {
+          value: 'carry',
+          name: 'Carry existing values over to the new environment (values can be updated later using amplify update function)',
```
edwardfoyle, line 252: nit:
```suggestion
          name: 'Carry existing values over to the new environment (values can be updated later using `amplify update function`)',
```

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -220,6 +230,108 @@ function getDefaultEnv(context: $TSContext): string | undefined {
...
+        },
+        {
+          value: 'update',
+          name: 'Update values now',
+        },
+      ],
+    };
+    if (context.parameters.options.quickstart) return;
+    const { envVar } = await inquirer.prompt(envVarQuestion);
+    if (envVar === 'carry') return;
```
edwardfoyle, line 262: how are the values in the new env being set in this case?

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -220,6 +230,108 @@ function getDefaultEnv(context: $TSContext): string | undefined {
...
+    if (envVar === 'update') await envVarSelectFunction();
+  };
+
+  const envVarSelectFunction = async () => {
+    const abortKey = uuid.v4();
+    const functionNameQuestion: inquirer.ListQuestion = {
+      type: 'list',
+      name: 'functionName',
+      message: 'Select the Lambda function you want to update values',
```
edwardfoyle, line 271: If the customer doesn't update values for a function / env var is the existing value carried over by default?

packages/amplify-cli/src/init-steps/s2-initProviders.ts
```diff
@@ -15,7 +17,10 @@ export async function initProviders(context) {
   });
 
   await sequential(initializationTasks);
-
+  // Carry environment variable values
+  if (Object.keys(categories).length > 0) {
```
edwardfoyle, line 21: seems like this should only modify the functions category in the object?

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -220,6 +230,108 @@ function getDefaultEnv(context: $TSContext): string | undefined {
...
+  };
+
+  const envVarSelectKey = async (functionName: string) => {
+    const envVars = functions[functionName];
+    const abortKey = uuid.v4();
+    const keyNameQuestion: inquirer.ListQuestion = {
+      type: 'list',
+      name: 'keyName',
+      message: 'Which environment variable do you want to update:',
+      choices: Object.keys(envVars)
```
edwardfoyle, line 294: we should list the actual env var name, not the CFN parameter name

edwardfoyle (changes requested):
the multi-env logic looks good but it needs to be moved into the functions package. The amplify-cli package shouldn't have functions-specific logic like this

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
watilde: Yes, it does. The parent function takes the original env vars from team-provider-info as default:
https://github.com/aws-amplify/amplify-cli/blob/ee3002458af30c7caad8197570d0ac92eb4d4b26/packages/amplify-cli/src/init-steps/s0-analyzeProject.ts#L238

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
watilde: The choices are from `team-provider-info.json` instead of CFN parameter names and that's what we care to carry over.

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
watilde: Then it goes to the next line:
https://github.com/aws-amplify/amplify-cli/blob/ee3002458af30c7caad8197570d0ac92eb4d4b26/packages/amplify-cli/src/init-steps/s0-analyzeProject.ts#L263

We have default values already. When users want to update one of the env vars, then it follows the below steps:
envVarSelectFunction => envVarSelectKey => envVarUpdateValue

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
watilde: @edwardfoyle We may want to keep having the logic as a part of init-steps because we cannot add a question to beginning of `env add` if we move the logic to `amplify-category-function`. `isNewEnv` is called after CFN workflow:

Screen Shot 2021-05-25 at 18 17 11

packages/amplify-cli/src/init-steps/s2-initProviders.ts
watilde: Agreed! I updated to modify only function instead of categories.

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
edwardfoyle: right, but the CFN parameter name is an implementation detail that is hidden from the customer. They should only know about the lambda function env var

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
watilde: I moved the logic into `amplify-category-function` :)

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
watilde: I replaced `team-provider-info` with the actual env vars! Thanks :)

edwardfoyle (changes requested):
Just pulled the PR to test it all out. One issue I discovered is configuring resource access permissions after configuring env vars blows all of the env var configuration away. To repro:
1. add a function with some env vars configured
2. add another function (all defaults is fine)
3. update the first function to have access to the second one using the resource access permissions flow
4. observe that all the env var config is gone

watilde:
@edwardfoyle Thank you for your comment! I could reproduce and added a fix patch. It needed to pass the existed env vars in update workflow.

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
```diff
@@ -7,8 +7,9 @@ export const askEnvironmentVariableQuestions = async (
   context: $TSContext,
   resourceName: string,
   environmentVariables: Record<string, string> = getStoredEnvironmentVariables(context, resourceName),
+  skip?: boolean,
```
edwardfoyle, line 10: nit: rename to `skipWalkthrough`

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
```diff
@@ -262,6 +262,8 @@ export async function updateWalkthrough(context, lambdaToUpdate?: string) {
   functionParameters.dependsOn = consolidateDependsOnForLambda(projectMeta, functionParameters.dependsOn, lambdaToUpdate, selectedSettings);
   if (selectedSettings.includes(environmentVariableSetting)) {
     merge(functionParameters, await askEnvironmentVariableQuestions(context, lambdaToUpdate));
+  } else {
+    merge(functionParameters, await askEnvironmentVariableQuestions(context, lambdaToUpdate, undefined, true));
   }
```
edwardfoyle: nit: can replace with:
```suggestion
  merge(functionParameters, await askEnvironmentVariableQuestions(context, lambdaToUpdate, undefined, !selectedSettings.includes(environmentVariableSetting)));
```

edwardfoyle (commented):
just a couple nits. I think once the push workflow is added this PR will be g2g

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/lambda-walkthrough.ts
watilde: Thank you! I integrated this into this branch :)

packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/environmentVariableWalkthrough.ts
watilde: Sounds good! I updated the variable name.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -147,10 +147,50 @@ export const askEnvironmentVariableCarryOut = async (context: $TSContext, curren
 
   await envVarQuestion();
   Object.keys(functions).forEach(functionName => {
-    setStoredKeyValue(context, functionName, functions);
+    setStoredKeyValue(context, functionName, functions[functionName]);
   });
 };
 
+export const askEnvironmentVariableToPush = async (context: $TSContext, currentEnvName: string, projectPath: string) => {
```
edwardfoyle, line 154: I'd rename this to something like `ensureEnvironmentVariableValues` because if everything is already defined, it's not going to ask anything

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -147,10 +147,50 @@ export const askEnvironmentVariableCarryOut = async (context: $TSContext, curren
...
+  });
+  const functions = teamProviderInfo[currentEnvName]?.categories?.function || {};
+  const functionNames = Object.keys(functions);
+  if (functionNames.length === 0) {
+    return;
+  }
+  context.print.info('');
+  context.print.info('Some Lambda function environment variables are defined but missing values.');
+  context.print.info('');
+  for (const [_, functionName] of functionNames.entries()) {
```
edwardfoyle, line 167: nit, can just iterate over the list directly:
```suggestion
  for (const functionName of functionNames) {
```

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -147,10 +147,50 @@ export const askEnvironmentVariableCarryOut = async (context: $TSContext, curren
...
+    const storedList = getStoredList(functionName);
+    const storedKeyValue = getStoredKeyValue(context, functionName);
+    for (const [_, item] of storedList.entries()) {
```
edwardfoyle, line 170: same here, can iterate over the list directly

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -147,10 +147,50 @@ export const askEnvironmentVariableCarryOut = async (context: $TSContext, curren
...
+      const envName = item.environmentVariableName;
+      const keyName = item.cloudFormationParameterName;
+      if (storedKeyValue.hasOwnProperty(keyName)) return;
+      const newValueQuestion: inquirer.InputQuestion = {
+        type: 'input',
+        name: 'newValue',
+        message: `Enter the missing environment variable value of ${envName} in ${functionName}:`,
+        validate: input => {
```
edwardfoyle, line 178: would it be easy to reuse the same validator from the other env var value prompt here?

edwardfoyle (commented):
Left a few comments but overall LGTM. Will pull it and test it out in a bit and leave another review with any findings then

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
```diff
@@ -0,0 +1,306 @@
...
+  context.print.info('');
+  context.print.info('Some Lambda function environment variables are defined but missing values.');
+  context.print.info('');
+  for (const [_, functionName] of functionNames.entries()) {
+    const storedList = getStoredList(functionName);
+    const storedKeyValue = getStoredKeyValue(context, functionName);
+    for (const [_, item] of storedList.entries()) {
+      const envName = item.environmentVariableName;
+      const keyName = item.cloudFormationParameterName;
+      if (storedKeyValue.hasOwnProperty(keyName)) return;
```
edwardfoyle, line 173: this should be a `continue` not a return

edwardfoyle (changes requested):
From testing it out here are the issues I found:
- `amplify env add --envName test --yes` should carry over existing env vars by default. Repro steps:
  - init new project and add a function with env vars
  - `amplify env add --envName test --yes`
  - above command should not prompt, but it still does
- The pre push workflow didn't execute properly. Repro steps:
  - init new project and add a function with env vars
  - git init
  - git checkout a new branch
  - add a new amplify env
  - update the function with a new env var
  - git commit the change
  - git checkout the original branch
  - git merge the other branch
  - amplify push and observe:
```
Some Lambda function environment variables are defined but missing values.

✔ Successfully pulled backend environment test from the cloud.

Current Environment: test

| Category | Resource name | Operation | Provider plugin   |
| -------- | ------------- | --------- | ----------------- |
| Function | envVarsTest   | Create    | awscloudformation |
? Are you sure you want to continue?
```
It appears that it detected there was a diff, but I wasn't prompted for new values. When selecting `yes` to the continue question, the push then fails with
```
CREATE_FAILED               functionenvVarsTest             AWS::CloudFormation::Stack Wed Jun 09 2021 11:35:20 GMT-0700 (Pacific Daylight Time) Parameters: [<envVarParamName>] must have values
```
I believe the cause of this is because of the `return` instead of `continue` that I mention below. Also, when fixing this, make sure that the behavior of `amplify push --yes` in this case is to fail with an error message specifying which env vars are missing values and to run `amplify push` interactively to fix it.

Also, we really need to add some unit tests and e2e tests to cover some of these cases beyond just the sanity check unit tests that are here. At a minimum we need comprehensive unit tests for `environmentVariablesHelper.ts` and `environmentVariablesWalkthrough.ts` and e2e tests that cover:
1. create function with env var, push and check that env var has correct value in the cloud
2. `amplify env add --envName test --yes` carries over env vars by default
3. pushing with a new env var prompts during push for a value. We can simulate this in the test by removing the value from `team-provider-info.json` without having to go through the whole git branch stuff
Obviously the more tests the better, but I think those are the big ones

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
watilde: Agreed on the better name you suggested! I'll update.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/environmentVariablesHelper.ts
watilde: Sounds good. I replaced return with continue :)

nubpro:
The yes flag should also recognize "--y" since amplify push recognize that too instead of just "--yes"

watilde:
Squashed, rebased and addressed @edwardfoyle 's comment.
+ Fixes `amplify env add --envName test --yes` case
+ Fixes git merge flow

edwardfoyle (approved):

Issue #4341, opened 2020-05-20
cognito trigger postConfirmation template causes `Invalid lambda function output : Invalid JSON` for async handler, works after removing the generated code in index.js


Albert-Gao:
CLI version:
4.20.0

I use Amplify CLI to generate a `postConfirmation` cognito trigger, 

```javascript
exports.handler = async (event) => {
  try {
    // doing some async stuff
  } catch (e) {
    console.log('error', JSON.stringify(e))
    throw e
  }
  console.log('event2', JSON.stringify(event))

  return event
}
```

Every time I `confirmSignUp`, I received a 400 code on the clientside `Invalid lambda function output: Invalid JSON`

Spent a whole night, then this morning I decide to remove all the code in the index.js generated by amplify-cli:

```javascript
exports.handler = (event, context, callback) => {
  const modules = process.env.MODULES.split(',');
  for (let i = 0; i < modules.length; i += 1) {
    const { handler } = require(`./${modules[i]}`);
    handler(event, context, callback);
  }
};
```

As the line here worries me: `handler(event, context, callback)`, as you can see it doesn't have the `await` tag when calling lambda, and I don't know why we need this code in the 1st place, it might be better to explain the `why` part at the top rather than the logic, the logic is straightforward to understand.

I completely removed this part, and moved my handler here, everything works like a charm now. 

As a rough guess, I think the current for loop doesn't compatible with `async` handler. 

Maybe we should just remove it, and replace it with whatever `blahblah.js` which is the main js the user needs to edit.

So:

1. to prevent things like this to happen
1. remove logic that is not user's concern unless it solves some problems (but i definitely don't need this when using serverless framework)
1. the user does not need to figure out which `.js` file to edit.

Thanks :)

attilah:
@Albert-Gao thanks for reporting we'll look into it.

UnleashedMind:
The Amplify CLI generated Lambda function for auth triggers uses "Non-async" handler instead of the async handler, for details, see [AWS Lambda doc](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html). 

It follow the module package folder structure
<img width="258" alt="Screen Shot 2020-08-14 at 3 00 30 PM" src="https://user-images.githubusercontent.com/6025263/90295912-04058d00-de3f-11ea-8b61-9c8feb84957c.png">

In which index.js is the entry point, and it has the exported handler
[...]

Issue #2735, opened 2019-11-14
Invalid lambda function output : Invalid JSON


danhumphrey:
**Describe the bug**

I'm using amplify auth and have a working model which uses a post confirmation hook with the following functionality:

- add user to group
- custom (added later)

My custom function is now adding the user to DynamoDB and doing some additional stuff so I no longer need to make use of the `add user to group` so I am using `amplify update auth` to remove this. 

After I `amplify push` and signup up and verify a user, I see the error 'Invalid lambda function output : Invalid JSON'

There are no errors reported in CloudWatch and my custom function has completed successfully. Also, the user is actually confirmed in the user pool.

Additionally, but not a high priority is that the PostConformation lambda still has the `GROUP` environment variable set after removing the `add user to group` functionality

**Amplify CLI Version**
3.17

**To Reproduce**

**1.** 
`amplify update auth`
What do you want to do? Walkthrough all the auth configurations
 Select the authentication/authorization services that you want to use: User Sign-Up, Sign-In, connected with AW
S IAM controls (Enables per-user Storage features for images or other content, Analytics, and more)
 Allow unauthenticated logins? (Provides scoped down permissions that you can control via AWS IAM) No
 Do you want to enable 3rd party authentication providers in your identity pool? No
 Do you want to add User Pool Groups? No
 Do you want to add an admin queries API? No
 Multifactor authentication (MFA) user login options: OFF
 Email based user registration/forgot password: Enabled (Requires per-user email entry at registration)
 Please specify an email verification subject: Your verification code
 Please specify an email verification message: Your verification code is {####}
 Do you want to override the default password policy for this User Pool? Yes
 Enter the minimum password length for this User Pool: 8
 Select the password character requirements for your userpool: (Press <space> to select, <a> to toggle all, <i>
to invert selection)Requires Lowercase, Requires Uppercase, Requires Numbers, Requires Symbols
 Specify the app's refresh token expiration period (in days): 15
 Do you want to specify the user attributes this app can read and write? Yes
 Specify read attributes: (Press <space> to select, <a> to toggle all, <i> to invert selection)Email
 Specify write attributes: (Press <space> to select, <a> to toggle all, <i> to invert selection)
 Do you want to enable any of the following capabilities?
 Do you want to use an OAuth flow? No
? Do you want to configure Lambda Triggers for Cognito? Yes
? Which triggers do you want to enable for Cognito (Press <space> to select, <a> to toggle all, <i> to invert se
[...]

PR #7180, opened 2021-04-25
fix(auth): add-to-group PostConfirmation Lambda should not throw

### LINKED ISSUE #7179: add-to-group PostConfirmation Lambda fails to return valid JSON with async exit

yarn

_No response_

### Amplify CLI Version

4.48.0

Ubuntu

### Amplify Categories

auth, function

### Amplify Commands

add, push

### Describe the bug

I ran into this by enabling the "Add User to Groups" functionality in `amplify add auth`. It is an async Lambda hook that is, according to [this page](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html), supposed to be able to just return a value or throw an error, but this is not the case.

I am opening this as an issue in this repo because I intend to send a PR in with a workaround so that this option is at least *more functional* than it is now, as it currently just breaks the sign-up process as the user confirms their email and then Cognito returns a code 400.

See: #2735  #4341

### Expected behavior

The add-to-group PostConfirmation Lambda should not fail.

### Reproduction steps

1. Enable "Add Users to Groups" in `amplify auth`
2. Deploy with `amplify push`
3. Sign ups will throw after a successful email confirmation, when PostConfirmation script fails to return valid JSON

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```


```

</details>

### Additional information

_No response_

ctjlewis:
#### Description of changes
The "Add Users to Groups" feature offered by `amplify auth` relies on an async Lambda function which, for some reason, will not actually resolve asynchronously, and instead, the PostConfirmation Lambda always throws 400 and breaks the sign-up flow. This PR adds an ugly workaround where the event is resolved with non-async `callback()`.

This is being submitted as a PR and not a draft because the current configuration actively breaks workflows, and merging this (or a modified version of it) ASAP is likely the best move. I'm struggling to get `yarn test` to run because of some issues with Cypress but I plan to circle back around, would appreciate reviews/edits to this PR.

#### Issue #, if available

#2735, #4341. Closes #7179.

I spent several hours trying to get the handler to asynchronously resolve to either a Promise or a value itself, but it did was unsuccessful. Only non-asynchronous `context.succeed()`, `context.done()`, and `callback()` allowed the Lambda to successfully complete, and `callback(null, event)` seemed most appropriate.  

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostConfirmation/function-template-dir/trigger-index.js
+++ b/packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostConfirmation/function-template-dir/trigger-index.js
@@ -1,14 +1,51 @@
-/*
-  this file will loop through all js modules which are uploaded to the lambda resource,
-  provided that the file names (without extension) are included in the "MODULES" env variable.
-  "MODULES" is a comma-delimmited string.
-*/
+/**
+ * @fileoverview
+ *
+ * This CloudFormation Trigger creates a Lambda function which waits for all
+ * other specified Lambdas (modules) to resolve, which should be located in the
+ * same directory as this file (./).
+ */
+
+/**
+ * The names of modules to load are stored as a comma-delimited string in the
+ * `MODULES` env variable.
+ */
 const moduleNames = process.env.MODULES.split(',');
+/**
+ * The array of imported modules.
+ */
 const modules = moduleNames.map(name => require(`./${name}`));
 
-exports.handler = (event, context, callback) => {
-  for (let i = 0; i < modules.length; i += 1) {
-    const { handler } = modules[i];
-    handler(event, context, callback);
-  }
+/**
+ * This handler is itself a Lambda function which iterates over all of the given
+ * modules and awaits them.
+ *
+ * @see https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html
+ *
+ * @param {object} event
+ *
+ * The event that triggered this Lambda.
+ *
+ * @param {object} context 
+ *
+ * The context for this Lambda. See:
+ * https://docs.aws.amazon.com/lambda/latest/dg/nodejs-context.html
+
+ * @param {function} callback 
+ *
+ * A deprecated way to send a response synchronously. See:
+ * https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html#nodejs-handler-sync
+ *
+ * @returns
+ *
+ * The CloudFormation Trigger event.
+ */
+exports.handler = async (event, context, callback) => {
+  /**
+   * Instead of naively iterating over all handlers, run them concurrently with
+   * `await Promise.all(...)`. This would otherwise just be determined by the
+   * order of names in the `MODULES` var.
+   */
+  await Promise.all(modules.map(module => module.handler(event, context, callback)));
+  return event;
 };
--- a/packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostConfirmation/add-to-group.js
+++ b/packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostConfirmation/add-to-group.js
@@ -1,6 +1,6 @@
 /* eslint-disable-line */ const aws = require('aws-sdk');
 
-exports.handler = async (event, context) => {
+exports.handler = async (event, context, callback) => {
   const cognitoidentityserviceprovider = new aws.CognitoIdentityServiceProvider({ apiVersion: '2016-04-18' });
   const groupParams = {
     GroupName: process.env.GROUP,
@@ -19,16 +19,21 @@ exports.handler = async (event, context) => {
     await cognitoidentityserviceprovider.createGroup(groupParams).promise();
   }
 
+  /**
+   * For some reason, naively returning a value / Promise or throwing an error
+   * will not work as it is specified in the Lambda docs:
+   * @see https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html
+   *
+   * This causes an "Invalid JSON" error which can, so far, only be worked
+   * around by resolving with deprecated non-async `callback`:
+   * @see https://github.com/aws-amplify/amplify-cli/issues/2735
+   * @see https://github.com/aws-amplify/amplify-cli/issues/4341
+   * @see https://github.com/aws-amplify/amplify-cli/issues/7179
+   */
   try {
     await cognitoidentityserviceprovider.adminAddUserToGroup(addUserParams).promise();
-    return {
-      statusCode: 200,
-      body: JSON.stringify(event)
-    }
+    callback(null, event);
   } catch (error) {
-    return {
-      statusCode: 500,
-      body: JSON.stringify(error)
-    }
+    callback(new Error(error));
   }
 };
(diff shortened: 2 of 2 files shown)


edwardfoyle (changes requested):
Hi @ctjlewis thanks for the contribution! The underlying issue here is that [this code](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostConfirmation/function-template-dir/trigger-index.js) does not handle the async function. I think just making that function async and adding await in front of the handler call will fix the issue.

That being said it looks like there is lots of inconsistency across the trigger templates using callbacks vs async/await and we should standardize on async/await. But that's beyond the scope of this PR.

ctjlewis:
Thank you so much for your comments @edwardfoyle, that was all I needed to figure out exactly what was happening here. The lack of async/await support in the CloudFormation trigger was definitely a thing, but the *real* issue was that the CloudFormation trigger **did not return an event**! Lambda threw 400 and said "invalid JSON" because *nothing* is not valid JSON!

The reason this worked with `context.succeed()` and `context.done()` was because it basically worked around that handler needing to return anything, and manually told the context what was happening.

A few things to review:
1. I added a ton of comments to the CloudFormation trigger and recommend they stay in. I'm not sure if that's feasible, but I left them in for now. This trigger is duplicated multiple times in each directory in `packages/amplify-category-auth/provider-utils/awscloudformation/triggers`, and I would recommend replacing all of them for the consistency you mentioned, as part of a separate PR.
2. I resolved the trigger with `await Promise.all(modules.map((module) => module.handler(...)))` instead of `for (const module of modules) await module.handler(...)`, and included reasoning in a comment. The reason race conditions don't seem to be a concern here is that all the modules will belong to the same step (here, `PostConfirmation`), and otherwise they'd just be executed sequentially in the order they're listed in `process.env.MODULES`, so might as well execute concurrently IMO.

cc @medelman17

ctjlewis:
Also set up [a repo](https://github.com/ctjlewis/aws-amplify-test) for easily testing `amplify-dev`, tested this on that project.

Not sure how to approach testing here, since these triggers are not already tested as far as I can tell.

packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostConfirmation/function-template-dir/trigger-index.js
```diff
@@ -1,14 +1,51 @@
...
+ * The CloudFormation Trigger event.
+ */
+exports.handler = async (event, context, callback) => {
+  /**
+   * Instead of naively iterating over all handlers, run them concurrently with
+   * `await Promise.all(...)`. This would otherwise just be determined by the
+   * order of names in the `MODULES` var.
+   */
+  await Promise.all(modules.map(module => module.handler(event, context, callback)));
+  return event;
```
edwardfoyle, line 50: Is there a reason for returning the event? Seems like just awaiting the promise and returning void would be better?

edwardfoyle (commented):
Just one comment but overall LGTM. Thanks for adding the comments!

packages/amplify-category-auth/provider-utils/awscloudformation/triggers/PostConfirmation/function-template-dir/trigger-index.js
ctjlewis: So, I forgot to really elaborate on this, I have been absolutely slammed for time, but I am pretty sure that is the real error here, was the generated top-level Lambda not returning anything.

I found that even with async/await implemented, I would get the JSON invalid error. After a few minutes, really getting a grasp on what's going on here, that top-level lambda is called by CloudFormation as I understand it, and that Lambda loads all the modules for that step (here, `PostConfirmation`). If that top-level Lambda passes CF back an event of `undefined`, it seems to throw the Invalid JSON error.

Basically, the theory is that even if I hadn't patched this with async/await, I could have added this return statement and patched that error, and the async stuff would have been running on its own out of order (not ideal) but the error would have disappeared.

We should look into this more IMO, and if I'm right, it implies that we should *really* go over all the Amplify templates, move them to async/await, and replace the `index-trigger.js` with this one right here.

Is my theory making sense? CloudFormation wants an object back from that trigger. If there's another layer in the stack after one of these Lambdas that doesn't return, that stack will get passed an invalid event (nothing) and throw "invalid JSON".

I would look into this more but as I said I am really out of time. Please let me know what you think and if you can confirm or unearth more here.

I tried to help someone in this thread with that trick, his trigger was totally synchronous so it seems consistent with these triggers not returning anything causing problems with CloudFormation, but it didn't solve his problem:

https://github.com/aws-amplify/amplify-cli/issues/4341#issuecomment-828140643

ctjlewis: Also, if we confirm this, I will add a mention of it to the `@fileoverview`.

ctjlewis: Consistent with this, the generated Lambda for Verification Link (which was the issue that fellow had in that thread) also 400s out despite mostly working. Case for broken `index-trigger.js` strong, basically every CloudFormation stack created by this CLI with more than one layer has been broken for some indeterminate amount of time it would seem.

image

ctjlewis: Confirmed, that updating the new `trigger-index.js` (with async/await support + returning event) *for each trigger in the stack* and updating the submodules to use modern syntax (`async` + return or throw), fixes this issue and also #4341.

ctjlewis: See #7219.

edwardfoyle (approved):

ctjlewis:
@edwardfoyle, before merging, please see @cjihrig's concerns in the other PR (which uses the same top-level template) about the async handler supporting the `callback` parameter.

We may want to remove support for this and mention it in the documentation before actually modifying these templates.

edwardfoyle:
@ctjlewis it almost seems like this PR is a subset of #7219 and should be added to that one? I also agree with @cjihrig that mixing async and callbacks is an anti-pattern and we should be consistent with the async pattern here

ctjlewis:
> @ctjlewis it almost seems like this PR is a subset of #7219 and should be added to that one? I also agree with @cjihrig that mixing async and callbacks is an anti-pattern and we should be consistent with the async pattern here

It basically is, but this was submitted first and then expanded to other templates with #7219.

And okay, that makes sense, but we will want to make sure the documentation clearly advises people that the `callback(...)` pattern will not longer work for Amplify-generated templates, which will be async and should return values directly.

We should refer + link to:
- [AWS Docs: Asynchronous invocation](https://docs.aws.amazon.com/lambda/latest/dg/invocation-async.html#invocation-async-destinations)
- [AWS Docs: AWS Lambda function handler in Node.js](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html)

SwaySway:
Closing this PR in favor of https://github.com/aws-amplify/amplify-cli/pull/7219

PR #7207, opened 2021-04-29
fix: 3 errors found during bugbash

attilah:
#### Description of changes

This PR fixes 3 error found during bugbash:
- filter out selected resource and Lambda Layers always from IAM access lists when access being added to functions
- When dependency error happens during resource remove don't print stacktrace
- Remove double resource prompt for `function remove`.

Manual testing

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/amplify-category-function/src/commands/function/remove.ts
+++ b/packages/amplify-category-function/src/commands/function/remove.ts
@@ -3,25 +3,35 @@ import { ServiceName } from '../../provider-utils/awscloudformation/utils/consta
 import { removeLayerArtifacts } from '../../provider-utils/awscloudformation/utils/storeResources';
 import { removeResource } from '../../provider-utils/awscloudformation/service-walkthroughs/removeFunctionWalkthrough';
 import { removeWalkthrough } from '../../provider-utils/awscloudformation/service-walkthroughs/removeLayerWalkthrough';
+import { $TSContext } from 'amplify-cli-core';
+
 const subcommand = 'remove';
 
 module.exports = {
   name: subcommand,
-  run: async context => {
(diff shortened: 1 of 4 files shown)


packages/amplify-category-function/src/commands/function/remove.ts
```diff
@@ -21,6 +21,8 @@ module.exports = {
       resourceName = resourceToBeDeleted;
     }
 
+    resourceName = response.resourceName;
```
jhockett, line 24: Since we're assigning `resourceName` here, we can remove the previous assignment as it will always be overwritten

packages/amplify-category-function/src/commands/function/remove.ts
```diff
@@ -29,8 +31,10 @@ module.exports = {
         }
       })
       .catch(err => {
-        context.print.info(err.stack);
-        context.print.error('An error occurred when removing the function resource');
+        if (err.stack) {
+          context.print.info(err.stack);
+          context.print.error('An error occurred when removing the function resource');
+        }
```
jhockett, line 37: Should we print the stack trace here?

attilah: If it is an unhandled error then sure, helps with troubleshooting, but this kind of handling was added in the main package as well to make sure user friendly error messages will not be followed by technical data like stack trace.

packages/amplify-category-function/src/commands/function/remove.ts
attilah: sideeffect of late night fixes, I update it and put it to else branch, good catch!

PR #7226, opened 2021-04-30
fix: show error message when project initalization fail

yuth:
#### Description of changes

The CLI throws ugly error if the initialization of project fails when the bucket limit is hit. This change shows list of errors instead of a stack trace
## Old error
```
ROLLBACK_COMPLETE amplify-initfail-dev-160404 AWS::CloudFormation::Stack Fri Apr 30 2021 16:04:20 GMT-0700 (Pacific Daylight Time)
⠏ Initializing project in the cloud...An error occurred when creating the CloudFormation stack
✖ Root stack creation failed
TypeError: Cannot read property 'Stacks' of null
    at processStackCreationData (/Users/yathiraj/.config/yarn/global/node_modules/amplify-provider-awscloudformation/src/initializer.js:108:44)
    at Object.run (/Users/yathiraj/.config/yarn/global/node_modules/amplify-provider-awscloudformation/src/initializer.js:80:7)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```
## New error

```
⠧ Initializing project in the cloud...

Following resources failed

Resource Name: amplify-initfail-dev-155719-unauthRole (AWS::IAM::Role)
Event Type: create
Reason: Resource creation cancelled


Resource Name: amplify-initfail-dev-155719-authRole (AWS::IAM::Role)
Event Type: create
Reason: Resource creation cancelled


Resource Name: DeploymentBucket (AWS::S3::Bucket)
Event Type: create
Reason: You have attempted to create more buckets than allowed (Service: Amazon S3; Status Code: 400; Error Code: TooManyBuckets; Request ID: 9PE5RBXKCAHYN2ZC; S3 Extended Request ID: YXRhl55q/zkp8k23YWOR89/ypoPAE2ZVmMbXGTphHdL63MM1aO/zF1CrR1yi9rkoln29v/8qDuI=; Proxy: null)


✖ Root stack creation failed
``` 

#### Issue #, if available
fix #7221

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7252, opened 2021-05-04
test(amplify-provider-awscloudformation): fix a test for system-config-manager that caused an error

fossamagna:
#### Description of changes

Fixed a test that caused an error in system-config-manager.test.ts

When I run `yarn test`, I get the following error:

```
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "Error: 'jsonString' argument missing or empty".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
FAIL src/__tests__/system-config-manager.test.ts
  ● Test suite failed to run

    Call retries were exceeded

      at ChildProcessWorker.initialize (../../node_modules/jest-worker/build/workers/ChildProcessWorker.js:193:21)
```

My environment:
- Node.js v16.0.0
- macOS Big Sur

The cause is as follows.
* The mock function is not returning the expected json string.
* Not waiting for the completion of the validation of the return value of the `getProfiledAwsConfig` function that returns a Promise.

I changed to wait for the completion of the promise. And set return value to mock function.

```
yarn test
```

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/__tests__/system-config-manager.test.ts
+++ b/packages/amplify-provider-awscloudformation/src/__tests__/system-config-manager.test.ts
@@ -23,12 +23,15 @@ describe('profile tests', () => {
     return true;
   });
 
-  it('should use credential_process defined in config file', () => {
+  it('should use credential_process defined in config file', async () => {
     fs_mock.readFileSync.mockImplementationOnce(() => {
       return '[profile fake]\noutput = json\nregion = us-fake-1\ncredential_process = fake credential process';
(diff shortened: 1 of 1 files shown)


cjihrig:
Hi @fossamagna. Thanks for the PR. It looks like these changes are already a part of https://github.com/aws-amplify/amplify-cli/pull/7226, which will be merged soon.

PR #7264, opened 2021-05-05
build: use @commitlint/cz-commitlint instead of cz-conventional-changelog

johnpc:
#### Description of changes

This pull request is a proposal that can be either taken or left. I got confused when I used `yarn commit` but no indication of valid `scopes` was provided  during that potion of interaction. This resulted in me using an invalid scope, failing the commit, and re-running `yarn commit` with a valid scope from the failure error message.

A lerna `scopes` enum (from `@commitlint/config-lerna-scopes`) is currently enforced by commitlint. However, `commitizen` is not configured to be aware of that list of valid scopes. This makes it possible to call `yarn commit` and result in the generated commit failing the lint rule:

cz-conventional-changelog mov

This pull request removes the dependency on `cz-conventional-changelog` and installs `@commitlint/cz-commitlint` as an alternative. `@commitlint/cz-commitlint` shares it's config with `commitlint.config.js`, which gives it access to `@commitlint/config-lerna-scopes` as selectable scopes without changing the interactive commit experience.

cz-commitlint mov

Further motivation is explained in https://github.com/conventional-changelog/commitlint/pull/2547

The downside of this change is that the scope of the change must be selected via arrow key-ing down to the scope you want. It might be faster for regular contributors to manually type their scope than scroll potentially very far down the list to find the desired scope.

If the current mechanism is preferred, feel free to close this PR. If the change is deemed valuable, let me know if anything else is needed to get it shipped.

I committed this very change on my fork in order to create this pull request

#### Checklist

- [x] PR description included

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,9 +81,11 @@
     "@babel/plugin-proposal-numeric-separator": "^7.10.4",
     "@babel/plugin-transform-modules-commonjs": "7.10.4",
     "@babel/plugin-transform-typescript": "7.10.5",
-    "@commitlint/cli": "^8.1.0",
-    "@commitlint/config-conventional": "^8.1.0",
-    "@commitlint/config-lerna-scopes": "^8.1.0",
+    "@commitlint/cli": "^12.1.2",
+    "@commitlint/config-conventional": "^12.1.2",
+    "@commitlint/config-lerna-scopes": "^12.1.2",
+    "@commitlint/cz-commitlint": "^12.1.2",
+    "@commitlint/prompt": "^12.1.2",
     "@types/jest": "^26.0.20",
     "@types/js-yaml": "^4.0.0",
     "@typescript-eslint/eslint-plugin": "^4.14.2",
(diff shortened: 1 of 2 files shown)


johnpc:
I’m not sure that cz-conventional-changelog offers a way to prevent asking for the scope. Additionally, having the scope in the commit message is useful for humans scrolling back through the repo history (beyond the changelog generation).

The version of cz-conventional-changelog currently being used also has a bug where it incorrectly enforces the combined character count of your commit type + scope + summary. It can cause commits to seem like they’re within the character limit, but once submitted and commitlint takes over, it fails as the top line is too long.

cjihrig:
Closing, as we discussed this during the meeting. We would like to make the committing process simpler and add validation around squashing if possible.

johnpc:
reopening - I think there was a misunderstanding that this wizard would remove the ability to author messages using `git commit`. If you use `git commit`, this PR does not impact your workflow. This PR simply fixes issues with the currently existing `yarn commit` wizard.

cjihrig:
@johnpc and I discussed this on Slack. I'm OK with the changes to `yarn commit` here.

cjihrig (approved):
LGTM with the merge conflicts resolved.

PR #7265, opened 2021-05-05
Sign in with Apple - add and update auth

letsbelopez:
#### Description of changes

Updated the --headless add and update auth flow to take Sign in with Apple in the headless payload and update the Cognito user pool.

I use a test web project and passed in the new SIWA headless payloads with the add and update commands:
`cat MyAuthTemplate.addauth.json | jq -c '.' | amplify add auth --headless`

I verified the changes in Cognito console.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Relevant documentation is changed or added (and PR referenced) The relevant headless documentation points to the headless schema package, which is updated in the PR.


edwardfoyle (approved):
One small nit, but LGTM

SwaySway (changes requested):
code lgtm - can you also add some tests to verify the add auth with siwa and headless pull with siwa works as expected?

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.js
```diff
@@ -280,14 +306,24 @@ function getRequiredParamsForHeadlessInit(projectType, previousValues) {
     if (previousValues.authProviders.includes('www.amazon.com')) {
       requiredParams.push('amazonAppId');
     }
+    if (previousValues.authProviders.includes('appleid.apple.com')) {
```
ammarkarachi, line 309: What's the URL like can we make sure we don't do an includes since it doesn't start with 'www'

letsbelopez: This is an array of [Cognito Identity Pool providers ](https://docs.aws.amazon.com/cognito/latest/developerguide/apple.html#sign-in-with-apple-cli-examples)

A few lines up you can see the facebook and google ones which, would have triggered the same warning, that's why I thought this was ok.

PR #7334, opened 2021-05-14
chore(gh actions): lock closed issues after inactivity

siegerts:
#### Description of changes

- chore(issue picker): add link to Discord server as an option for help
- chore(gh actions): add action to locked closed threads after 30 days of inactivity

#### Checklist

- [x] PR description included

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: I want help writing my Amplify application
+    url: https://discord.com/invite/amplify
+    about: Check out the `*-help` channels on Amplify's community Discord to ask your questions about building applications with the CLI or other Amplify libraries.
--- a/.github/workflows/lock.yaml
+++ b/.github/workflows/lock.yaml
@@ -1,0 +1,24 @@
+name: 'Lock Threads'
+
+on:
+  schedule:
+    - cron: '0 * * * *'
+
+jobs:
+  lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          issue-lock-inactive-days: '30'
+          issue-lock-reason: 'resolved'
+          issue-lock-comment: |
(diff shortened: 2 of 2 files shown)


siegerts:
This will run every hour since only 50 issues and pull requests will be handled at once. After that's processed, it schedule can be reduced.

cjihrig (commented):
This LGTM, thank you. Maybe we should consider forking https://github.com/dessant/lock-threads so we have full control over things.

I'm not going to "Approve" the PR yet. I want to make sure it doesn't accidentally get merged before others have a chance to weigh in.

attilah (approved):
LGTM, I like the idea of this kind of closure for issues and PRs.

.github/workflows/lock.yaml
```diff
@@ -0,0 +1,24 @@
+name: 'Lock Threads'
+
+on:
+  schedule:
+    - cron: '0 * * * *'
```
kaustavghosh06, line 5: Does this cron run daily?

.github/workflows/lock.yaml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  lock:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: dessant/lock-threads@v2
```
kaustavghosh06, line 11: Is this a reliable and widely used plugin which we can trust and use?

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: true
```
kaustavghosh06, line 1: Where would this issue template be used and is this relevant to this PR?

siegerts: It persists the current functionality (image below ) while also adding a link to Discord server as an option for help

image

.github/workflows/lock.yaml
siegerts: This will run every hour since only 50 issues and pull requests will be handled at once. After that's processed, the schedule can be reduced.

.github/workflows/lock.yaml
siegerts: Used in below repos - 
- [Jest](https://github.com/facebook/jest/blob/master/.github/workflows/lock.yml) 
- [Flutter ](https://github.com/flutter/flutter/blob/master/.github/lock.yml)
- [Next.js](https://github.com/vercel/next.js/blob/canary/.github/lock.yml)

kaustavghosh06 (approved):

PR #7452, opened 2021-06-03
fix: amplify-category-api case sensitivity failure

### LINKED ISSUE #6354: bug when 2 amplify api have the same name
**Describe the bug**
I created 2 api, one GraphQL api named "testBlog", another REST api named "testblog". in ./amplify/backend/api/ , there should have been 2 folders, one named "testblog", the other one named "testBlog", but only "testblog" appeared, which created errors when pushing to the cloud

**Amplify CLI Version**
4.41.0

**To Reproduce**
- create a new project
- use amplify to create 2 api, one GraphQL api named "testBlog", another REST api named "testblog".

**Expected behavior**
in ./amplify/backend/api/ , there should have been 2 folders, one named "testblog", the other one named "testBlog", but only "testblog" appeared.

johnpc:
#### Description of changes

This pull request addresses https://github.com/aws-amplify/amplify-cli/issues/6354 by throwing an error when an api will not be created properly due to a case sensitivity issue.

Amplify resource names are case-insensitive.

This pull request _only_ addresses the creation of APIs. It does not concern itself with other types of amplify resources.

**NOTE**: This commit uses `process.exit` directly from the helper util. I am not sure how to achieve this another way, but it feels wrong, so advice on that would be appreciated.

#### Issue #, if available

fix #6354

- unit tests added 
- `amplify-dev add api` attempted to create various combinations of valid and invalid amplify resources. Ex. `testblog` (succeeds) then `testBlog` (fails with error message), then `testblog` (update code path succeeds), then `test` (succeeds).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/provider-utils/awscloudformation/utils/check-case-sensitivity.ts
+++ b/packages/amplify-category-api/src/provider-utils/awscloudformation/utils/check-case-sensitivity.ts
@@ -4,25 +4,17 @@ import path from 'path';
 export const checkCaseSensitivityIssue = async (context: any, category: string, resourceName: string) => {
   const projectBackendDirPath = context.amplify.pathManager.getBackendDirPath();
 
-  const fsIsCaseSensitive = async () => {
-    const tempFileName = '/tmp/AMPLIFY-TEMP';
-    await fs.writeFile(tempFileName, 'deleteme');
-    const lowercaseVersionExists = fs.existsSync(tempFileName.toLowerCase());
-    await fs.unlink(tempFileName);
-    return !lowercaseVersionExists;
-  };
-
-  const caseSensitiveMatchExists = async () => {
+  const caseSensitivityConflict = async () => {
     const basePath = path.join(projectBackendDirPath, category);
     const filenames = await fs.readdir(basePath);
-    const matchExistsWhenAllLowercase = filenames.map(name => name.toLowerCase()).includes(resourceName.toLowerCase());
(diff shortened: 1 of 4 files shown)


packages/amplify-category-api/src/provider-utils/awscloudformation/utils/check-case-sensitivity.ts
```diff
@@ -0,0 +1,29 @@
+import fs from 'fs-extra';
+import path from 'path';
+
+export const checkCaseSensitivityIssue = async (context: any, category: string, resourceName: string) => {
+  const projectBackendDirPath = context.amplify.pathManager.getBackendDirPath();
+
+  const fsIsCaseSensitive = async () => {
+    const tempFileName = '/tmp/AMPLIFY-TEMP';
```
sachscode: Could this have issues when multiple amplify projects run on the same machine ? 
Does this file need to have a unique path ? e.g. /tmp/${procId} /AMPLIFY-TEMP to avoid side effects ?

sachscode (approved):
Looks good, one minor comment.

packages/amplify-category-api/src/provider-utils/awscloudformation/utils/check-case-sensitivity.ts
johnpc: oh - this file is actually no longer needed because we are going to treat all amplify projects the same regardless of whether the filesystem of their OS could technically support it.

PR #7497, opened 2021-06-10
chore: update AMPLIFY_PATH env var in config for migration tests

jhockett:


attilah (approved):
LGTM

nit: as a better solution later we could use the output of `npm config get prefix` to get the base for global.

PR #7508, opened 2021-06-11
chore(): removed plugin scan messages & initialize new version messages

renebrandel:
#### Description of changes

- removes the following messages but still prints errors if it encounters any:
```
renebrandel@renes-air Projects % amplify version
Initializing new Amplify CLI version...
Done initializing new version.
Scanning for plugins...
Plugin scan successful
```

- Tested it locally.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/plugin-manager.ts
+++ b/packages/amplify-cli/src/plugin-manager.ts
@@ -133,10 +133,8 @@ export function getAllPluginNames(pluginPlatform: PluginPlatform): Set<string> {
 }
 
 export async function scan(pluginPlatform?: PluginPlatform): Promise<PluginPlatform> {
-  print.info('Scanning for plugins...');
   try {
     const result = await scanPluginPlatform(pluginPlatform);
-    print.info('Plugin scan successful');
     return result;
   } catch (e) {
     print.error('Plugin scan failed.');
--- a/packages/amplify-cli/src/utils/post-install-initialization.ts
+++ b/packages/amplify-cli/src/utils/post-install-initialization.ts
@@ -8,7 +8,6 @@ import { print } from '../context-extensions';
  * It copies some assets needed by dependency packages to the .amplify folder
  */
 export const postInstallInitialization = async () => {
-  print.info(`Initializing new Amplify CLI version...`);
   // clean any previous libs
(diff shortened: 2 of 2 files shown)


packages/amplify-cli/src/utils/post-install-initialization.ts
```diff
@@ -8,7 +8,6 @@ import { print } from '../context-extensions';
  * It copies some assets needed by dependency packages to the .amplify folder
  */
 export const postInstallInitialization = async () => {
-  print.info(`Initializing new Amplify CLI version...`);
```
jhockett, line 11: Since `print` is no longer used, please remove the import

packages/amplify-cli/src/plugin-manager.ts
```diff
@@ -133,10 +133,8 @@ export function getAllPluginNames(pluginPlatform: PluginPlatform): Set<string> {
 }
 
 export async function scan(pluginPlatform?: PluginPlatform): Promise<PluginPlatform> {
-  print.info('Scanning for plugins...');
   try {
     const result = await scanPluginPlatform(pluginPlatform);
-    print.info('Plugin scan successful');
     return result;
   } catch (e) {
     print.error('Plugin scan failed.');
```
drochetti, line 140: since we're taking the time to improve the messages, if this fails what is the customer impact and/or recovery story here?

Let's say as a customer, I see this `Plugin scan failed.` in my console. What do I do? What are the consequences? Is there a troubleshooting (if so, can we link the URL here?)

palpatim: +1: Kind of fails our "Expose friendly error messages to the customer that assist with debugging" tenet

packages/amplify-cli/src/utils/post-install-initialization.ts
renebrandel: Done!

packages/amplify-cli/src/plugin-manager.ts
renebrandel: Discussed offline with other reviewers. The error rate is practically zero on this.

edwardfoyle (approved):

palpatim (approved):

PR #7538, opened 2021-06-17
fix: ensure REST API CFN outputs the API ID

cjihrig:
#### Description of changes
This commit updates the REST API migration logic to ensure that the API ID is a CFN Output, as the policy consolidation logic will rely on this during push.

Manual testing

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/utils/consolidate-apigw-policies.ts
+++ b/packages/amplify-provider-awscloudformation/src/utils/consolidate-apigw-policies.ts
@@ -248,8 +248,7 @@ function updateExistingApiCfn(context: $TSContext, api: $TSObject): void {
   const parameterJson = JSONUtilities.readJson(paramsFile, { throwIfNotExist: false }) ?? {};
 
   if (!cfn) {
-    // This should not happen. If it does, nothing else will work anyway so just bail out.
-    return;
+    throw new Error(`CloudFormation template missing for REST API ${resourceName}`);
   }
 
   const parameters = cfn.Parameters ?? {};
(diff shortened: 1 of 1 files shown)


ammarkarachi (approved):

packages/amplify-provider-awscloudformation/src/utils/consolidate-apigw-policies.ts
```diff
@@ -246,8 +246,14 @@ function updateExistingApiCfn(context: $TSContext, api: $TSObject): void {
   const apiParamsFile = path.join(resourceDir, API_PARAMS_FILE);
   const cfn: any = JSONUtilities.readJson(cfnTemplate, { throwIfNotExist: false }) ?? {};
   const parameterJson = JSONUtilities.readJson(paramsFile, { throwIfNotExist: false }) ?? {};
-  const parameters = cfn?.Parameters ?? {};
-  const resources = cfn?.Resources ?? {};
+
+  if (!cfn) {
+    // This should not happen. If it does, nothing else will work anyway so just bail out.
```
attilah, line 251: Can't we just throw an error to make sure a return here will not makes things bad? We have a similar throw for layers when a value MUST be present that further functionality depends on.

attilah (approved):
LGTM a nit about the return

packages/amplify-provider-awscloudformation/src/utils/consolidate-apigw-policies.ts
cjihrig: Sure, I can update it to throw.

cjihrig: Updated!

ammarkarachi: I think throwing returning would make the CLI proceed with pushing the category

Issue #7546, opened 2021-06-18
RFC: `@searchable` directive Enhancements


cjihrig:
*This is a Request For Comments (RFC). RFCs are intended to elicit feedback regarding a proposed change to the Amplify Framework. Please feel free to post comments or questions here.*

This document outlines a number of new features and improvements to the `@searchable` directive in Amplify CLI. The goal is to address some of the enhancement requests and bugs.

## Proposal 1: Backfilling improvements

Exposing the streaming function:

### Parameterize BatchSize and MaximumBatchingWindowInSeconds

**Github Issues**

* https://github.com/aws-amplify/amplify-cli/issues/4695

The `EventSourceMapping` parameter of the Lambda function that streams data from the DynamoDB table to AWS ElasticSearch assumes a batch size of one. This enables the Lambda to update documents as soon as possible in AWS ElasticSearch. This is slow if you are trying to migrate a large amount of data. Moving forward, the `BatchSize` and `MaximumBatchingWindowInSeconds` will be parametrized to be overridden as desired. The default values will remain the same as the current values.

```js
        "Searchable<ModelName>LambdaMapping": {
            "Type": "AWS::Lambda::EventSourceMapping",
            "Properties": {
             +  "BatchSize": {
             +      "Ref": "ElasticSearchStreamBatchSize" 
             +   },
             +  "MaximumBatchingWindowInSeconds": { 
             +      "Ref": "ElasticSearchStreamMaximumBatchingWindowInSeconds"
             +   },
                "Enabled": true,
                "EventSourceArn": {
                    "Fn::ImportValue": {
                        "Fn::Join": [
                            ":",
                            [
                                {
                                    "Ref": "AppSyncApiId"
                                },
                                "GetAtt",
                                "<ModelName>Table",
                                "StreamArn"
                            ]
                        ]
                    }
                },
                "FunctionName": {
                    "Fn::GetAtt": [
                        "ElasticSearchStreamingLambdaFunction",
                        "Arn"
                    ]
                },
                "StartingPosition": "LATEST"
            },
            "DependsOn": [
                "ElasticSearchStreamingLambdaFunction"
            ]
        },
```

### Sorting & Pagination

**Github Issues**

* https://github.com/aws-amplify/amplify-cli/issues/6729
* https://github.com/aws-amplify/amplify-cli/issues/4434
* https://github.com/aws-amplify/amplify-cli/issues/2822

[...]

PR #7556, opened 2021-06-18
feat(cli): upgrade elastic search version to 7.x

sundersc:
#### Description of changes
This change upgrades ElasticSearch from 6.x to 7.x.
[Link to Searchable RFC document](https://github.com/aws-amplify/amplify-cli/issues/7546)

#### Issue #, if available
NA.

- Manual test
- yarn test passed

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-searchable-transformer/src/cdk/create-es-domain.ts
+++ b/packages/amplify-graphql-searchable-transformer/src/cdk/create-es-domain.ts
@@ -10,7 +10,8 @@ export const createEsDomain = (stack: Construct, parameterMap: Map<string, CfnPa
   const { HasEnvironmentParameter } = ResourceConstants.CONDITIONS;
 
   const domain = new Domain(stack, ElasticsearchDomainLogicalID, {
-    version: ElasticsearchVersion.V7_7,
+    //version: ElasticsearchVersion.V7_7,
+    version: { version: "7.10" } as ElasticsearchVersion,
     ebs: {
       enabled: true,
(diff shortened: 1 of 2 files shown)


packages/amplify-graphql-searchable-transformer/src/cdk/create-es-domain.ts
```diff
@@ -10,7 +10,7 @@ export const createEsDomain = (stack: Construct, parameterMap: Map<string, CfnPa
   const { HasEnvironmentParameter } = ResourceConstants.CONDITIONS;
 
   const domain = new Domain(stack, ElasticsearchDomainLogicalID, {
-    version: ElasticsearchVersion.V6_2,
+    version: ElasticsearchVersion.V7_7,
```
cjihrig, line 13: We want to update to `V7_10`. Is this just a limitation of the version of the CDK we're using? If so, we can look into other workarounds or even updating the version of CDK (updating would likely need to be done across the entire project).

ammarkarachi: Apparently you would have to use a newer version of CDK to make that possible, which might not be trivial 

You can assign manual overrides on CDK
```js
 (domain.node.defaultChild as CfnDomain).elasticsearchVersion = "7.10"
```

Also go can you go through this if there is a path forward for upgrading to 7.10 without interruption

cjihrig (approved):

PR #7568, opened 2021-06-21
fix(cli): use more explicit amplify pragma in .gitignore

### LINKED ISSUE #7250: Amplify headless pull - do not update .gitignore
I have a setup with multiple frontends for one backend. I do a headless checkout of the amplify environment. 
`.gitignore` ignores contains `amplify/` to ignore the entire amplify directory.
When you run the headless checkout / pull. It updates the `.gitignore` file and the amplify dir is no longer ignored.

I want an option to indicate that amplify should keep its hands of my `.gitignore` file :) 

Now I always have to manually update the gitignore.

johnpc:
#### Description of changes

Previously, we used the #amplify pragma to show which block of .gitignore was managed by amplify.
This causes confusion for customers who think it is safe to add or remove lines this section of
their gitignore. The cli will blitz any changes to this section on amplify init or amplify pull. By
using a more explicit pragma, customers will realize they should not touch this section of the
gitignore.

docs https://github.com/aws-amplify/docs/pull/3330

#### Issue #, if available

fix #7250

created a new project and checked that .gitignore file was manipulated as expected

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/__tests__/extensions/amplify-helpers/git-manager.test.ts
+++ b/packages/amplify-cli/src/__tests__/extensions/amplify-helpers/git-manager.test.ts
@@ -9,7 +9,8 @@ const fsMock = fs as jest.Mocked<typeof fs>;
 fsMock.readFileSync.mockImplementation(() => 'test');
 fsMock.writeFileSync.mockImplementation();
 
-const amplifyMark = '#amplify';
+const amplifyMark = '#amplify-do-not-edit-begin';
+const amplifyEndMark = '#amplify-do-not-edit-end';
 
 const ignoreList = [
   'amplify/\\#current-cloud-backend',
@@ -32,7 +33,7 @@ const ignoreList = [
   '.secret-*',
 ];
 
-const toAppend = `${os.EOL + os.EOL + amplifyMark + os.EOL}${ignoreList.join(os.EOL)}`;
+const toAppend = `${os.EOL + os.EOL + amplifyMark + os.EOL}${ignoreList.join(os.EOL)}${os.EOL + amplifyEndMark + os.EOL}`;
 
 describe('git-manager', () => {
   beforeEach(() => {
--- a/packages/amplify-cli/src/__tests__/extensions/amplify-helpers/git-manager.test.ts
+++ b/packages/amplify-cli/src/__tests__/extensions/amplify-helpers/git-manager.test.ts
@@ -11,6 +11,7 @@ fsMock.writeFileSync.mockImplementation();
 
 const amplifyMark = '#amplify-do-not-edit-begin';
 const amplifyEndMark = '#amplify-do-not-edit-end';
+const legacyAmplifyMark = '#amplify';
 
 const ignoreList = [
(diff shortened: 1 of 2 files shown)


attilah:
Can you add more tests to cover legacy cases and such?

Also curious what would happen if I create a project with these changes in place but an older CLI runs and updates the fil? I think perhaps we need to preserve the legacy mark in case it was present?

johnpc:
> Can you add more tests to cover legacy cases and such?

 done.

> Also curious what would happen if I create a project with these changes in place but an older CLI runs and updates the file? 

In this case, as written, the gitignore file _will_ be updated to contain redundant lines from the `getGitIgnoreAppendString()`. Hopefully it is obvious to the author that those extraneous lines do not need to be checked in. Otherwise, the file will be cleaned next time an operation is run from a newer version of the cli. I'm not sure if there's a foolproof way around this.

> I think perhaps we need to preserve the legacy mark in case it was present?

Why should we preserve the legacy mark? I'm open to the idea but in my talks with @kaustavghosh06 we settled on rolling customers forward.

packages/amplify-cli/src/extensions/amplify-helpers/git-manager.ts
```diff
@@ -2,8 +2,12 @@ import * as fs from 'fs-extra';
 import * as os from 'os';
 import { LocalLogDirectory } from 'amplify-cli-logger';
 
-const amplifyMark = '#amplify';
+const amplifyMark = '#amplify-do-not-edit-begin';
+const amplifyEndMark = '#amplify-do-not-edit-end';
+const legacyAmplifyMark = '#amplify';
 const amplifyMarkRegExp = new RegExp(`^${amplifyMark}`);
+const amplifyEndMarkRegExp = new RegExp(`^${amplifyEndMark}`);
+const legacyAmplifyMarkRegExp = new RegExp(`^${legacyAmplifyMark}`);
```
jhockett, line 10: nit: I prefer "deprecated" over "legacy" (even though I've used it myself in the past)

packages/amplify-cli/src/__tests__/extensions/amplify-helpers/git-manager.test.ts
```diff
@@ -32,16 +34,22 @@ const ignoreList = [
...
+    fsMock.readFileSync.mockClear();
+    fsMock.appendFileSync.mockClear();
+    fsMock.existsSync.mockClear();
   });
   const gitIgnoreFilePath = 'testPath';
   test('appends files to an existing .gitignore', () => {
     fsMock.existsSync.mockImplementation(() => true);
+    fsMock.readFileSync.mockImplementation(() => 'amplify/');
     insertAmplifyIgnore(gitIgnoreFilePath);
+    expect(fsMock.writeFileSync.mock.calls[0][0]).toEqual('testPath');
```
jhockett, line 51: nit: use the `gitIgnoreFilePath` variable instead of hardcoding, here and elsewhere

jhockett (approved):
LGTM, just left a couple of nits

PR #7584, opened 2021-06-23
fix: correct featuer typo in gql transformer v2

cjihrig:
#### Description of changes
This commit fixes a typo in the transformer API.

This was fixed as part of a larger change in https://github.com/aws-amplify/amplify-cli/pull/7258, but reverted in https://github.com/aws-amplify/amplify-cli/pull/7578. We should fix this typo since it is part of the public API.

#### Description of how you validated changes
`yarn test`

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7590, opened 2021-06-23
Fix e2e tests for pluralization

johnpc:
#### Description of changes

When #7258 was merged, e2e tests began failing because of the new `improvePluralization` feature flag behavior during certain test runs.

We test the iterative update feature by renaming [the intitial Todo model](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-e2e-tests/schemas/iterative-push/change-model-name/initial-schema.graphql#L6) and [updating it to Todos](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-e2e-tests/schemas/iterative-push/change-model-name/final-schema.graphql#L6). Previously (before #7258), this created a resolver called `listTodoss` (because it incorrectly blindly appended `s`). With the pluralization fix, it creates a resolver called listTodos - which unfortunately already exists and therefore appsync throws a Only one resolver is allowed per field error (see [appsync bug](https://github.com/aws/aws-appsync-community/issues/128)).

This PR fixes the broken e2e tests by renaming the model from `Todo` to `Task` instead of to `Todos`, therefore eliminating the conflict.

#### Issue #, if available

#4224

- ran e2e tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


cjihrig:
I would prefer to land https://github.com/aws-amplify/amplify-cli/pull/7584 first so that the unrelated typo fix is not at risk of being reverted again.

edwardfoyle (approved):

PR #7597, opened 2021-06-23
ci: run e2e test artifact scanning

yuth:
#### Description of changes

Adding a additional step in E2E test to scan the test artifacts and if any of them are found, fail the build and delete the artifacts to prevent them from getting stored in CircleCI. The secret values scanned can be configured by setting up a comma separate environment variable `ENV_VAR_WITH_SECRETS` which contains the list of enviroment variables that has the secret values. For instance if the secret are stored in `MY_SECRET_ENV_1`, `MY_SECRET_ENV_2` then set `ENV_VAR_WITH_SECRETS` to `MY_SECRET_ENV_1, MY_SECRET_ENV_2`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -291,6 +291,7 @@ jobs:
             cd packages/amplify-migration-tests
             yarn run migration_v4.0.0 --maxWorkers=3 $TEST_SUITE
           no_output_timeout: 90m
+      - run: *scan_e2e_test_artifacts
       - store_test_results:
           path: packages/amplify-migration-tests/
       - store_artifacts:
@@ -314,6 +315,7 @@ jobs:
             cd packages/amplify-migration-tests
             yarn run migration_v4.28.2_nonmultienv_layers --maxWorkers=3 $TEST_SUITE
           no_output_timeout: 90m
+      - run: *scan_e2e_test_artifacts
       - store_test_results:
           path: packages/amplify-migration-tests/
       - store_artifacts:
@@ -337,6 +339,7 @@ jobs:
             cd packages/amplify-migration-tests
             yarn run migration_v4.52.0_multienv_layers --maxWorkers=3 $TEST_SUITE
           no_output_timeout: 90m
+      - run: *scan_e2e_test_artifacts
       - store_test_results:
           path: packages/amplify-migration-tests/
       - store_artifacts:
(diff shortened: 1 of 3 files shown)


.circleci/scan_artifacts.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+};
+
+const main = () => {
+  const envVarNameWithCredentialValues = (process.env.ENV_VAR_WITH_SECRETS || '').split(',').map(v => v.trim());
+  const values = envVarNameWithCredentialValues.map(v => process.env[v]).filter(Boolean);
+  if (values.length) {
+    const hasContent = hasMatchingContentInFolder(values);
+    if (hasContent) {
+      console.log('Scanning artifact has found secret value. Failing the build');
```
edwardfoyle, line 28: can we log which file / line contained the secret value?

.circleci/config.base.yml
```diff
@@ -31,6 +31,16 @@ clean_e2e_resources: &clean_e2e_resources
...
 
+scan_e2e_test_artifacts: &scan_e2e_test_artifacts
+  name: Scan And Cleanup E2E Test Artifacts
+  command: |
+    if ! yarn ts-node .circleci/scan_artifacts.ts; then
+      echo "Cleaning the repository"
+      git clean -fdx
+      exit 1
+    fi
+  when: always
```
edwardfoyle, line 42: I'm assuming this allows this step to run even if a previous step failed?

edwardfoyle: nvm, I can read: https://circleci.com/docs/2.0/configuration-reference/#the-when-attribute

edwardfoyle (approved):

.circleci/scan_artifacts.ts
attilah: I am in Ed with this, can help troubleshooting in a big way.

attilah (approved):

.circleci/config.base.yml
yuth: That is correct. Here is an example where the test failed and the artifact scanning still happened.
Screen Shot 2021-06-23 at 3 20 19 PM

PS: Artifact scan failure is done intentionally to test it

.circleci/scan_artifacts.ts
yuth: Line may be. 
I think you meant the secret env variable name instead of the value. Printing the secret value will it defeats the purpose. We can follow this change up with printing the file name and the secret ENV Var name in an upcoming change

PR #7621, opened 2021-06-27
fix(amplify-go-function-runtime-provider): localinvoke build error with Go 1.16

jasonxh:
#### Description of changes

With Go 1.16, If `localinvoke` was never built before, running `amplify mock function <go_function_name>` will fail with the following error:

```
Local invoker binary was not found, building it...
go: github.com/aws/aws-lambda-go@v1.15.0: missing go.sum entry; to add it:
	go mod download github.com/aws/aws-lambda-go
<go_function_name> failed with the following error:
Error: Command failed with exit code 1: go build main.go
```

This is due to a behavior change in [Go 1.16](https://golang.org/doc/go1.16#go-command):
> Build commands like go build and go test no longer modify go.mod and go.sum by default. Instead, they report an error if a module requirement or checksum needs to be added or updated (as if the -mod=readonly flag were used). Module requirements and sums may be adjusted with go mod tidy or go get.

The proposed fix is to simply run `go mod tidy` beforehand, to generate `go.sum`. Alternatively, we could add `go.sum` to source control, but I don't have context over whether it was left out on purpose.

#### Issue #, if available

N/A

```bash
# delete the already generated files to simulate new installation
rm -fr ~/.amplify/lib/amplify-go-function-runtime-provider/resources/localinvoke/{go.sum,main}

# verify go version is 1.16
go version

# build fails with the above error
amplify mock function <go_function_name>

# build succeeds with patched version
amplify-dev mock function <go_function_name>
```

Also validated the patch is backwards compatible with Go 1.13.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-go-function-runtime-provider/src/localinvoke.ts
+++ b/packages/amplify-go-function-runtime-provider/src/localinvoke.ts
@@ -26,6 +26,7 @@ const buildLocalInvoker = async (context: any) => {
   if (!fs.existsSync(localInvokeExecutablePath)) {
     // Build localinvoker
     context.print.info('Local invoker binary was not found, building it...');
+    executeCommand(['mod', 'tidy'], true, undefined, localInvokerDir);
     executeCommand(['build', MAIN_SOURCE], true, undefined, localInvokerDir);
   }
 


cjihrig (approved):

cjihrig:
Thank you for the contribution!

PR #7636, opened 2021-06-30
fix(amplify-appsync-simulator): string functions not called on items …

alichherawalla:
#### Description of changes
Calling string functions like replaceAll on children of the JavaArray returned from map.keySet in the resolvers does not work.

Adding an inline example where the issue is replicable locally, but it works as expected when actually deployed 

```
#foreach( $entry in $ctx.args.input.keySet() )
  #set( $regex = "([a-z])([A-Z]+)")
  #set( $replacement = "$1_$2")
  #set( $toSnake = $entry.replaceAll($regex, $replacement).toLowerCase() )
```

`yarn test` and manual testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-appsync-simulator/src/velocity/value-mapper/array.ts
```diff
@@ -1,3 +1,4 @@
+import { JavaString } from './string';
```
cjihrig, line 1: I don't think importing `JavaString` is the correct way to go here. The `mapper` argument passed to the constructor can do this for you.

packages/amplify-appsync-simulator/src/velocity/value-mapper/array.ts
```diff
@@ -7,6 +8,13 @@ export class JavaArray extends Array<any> {
       // splice sends a single object
       values = [values];
     }
+    values = values.map(value => {
+        if (typeof value === 'string' && !((value as any) instanceof JavaString)) {
+          return new JavaString(value);
+        }
+          return value;
+      });
```
cjihrig, line 16: Just a note here, but the indentation is off.

cjihrig (commented):
Left a couple comments. This also needs unit tests.

packages/amplify-appsync-simulator/src/velocity/value-mapper/array.ts
alichherawalla: hey, 

Thank you for taking the time to review this. 

Will take a look at how this can be done using the mapper, and will also add tests

alichherawalla: Hi @cjihrig, 

I've made the changes, please do let me know if there is anything else that would be required. 

Thank you for your time.

packages/amplify-appsync-simulator/src/velocity/value-mapper/map.ts
```diff
@@ -56,7 +57,7 @@ export class JavaMap {
   }
 
   keySet() {
-    return new JavaArray(Array.from(this.map.keys()), this.mapper);
+    return new JavaArray(Array.from(this.map.keys()).map(toJavaString), toJavaString);
```
cjihrig, line 60: Is this new `toJavaString()` function needed. Shouldn't `this.mapper` handle the conversion [here](https://github.com/aws-amplify/amplify-cli/blob/5c9172af5dbd911ce86823f32e774e4c89fd9f88/packages/amplify-appsync-simulator/src/velocity/value-mapper/mapper.ts#L27-L30)?

alichherawalla: From what I see and the tests that I've run, for the array values that are passed to the constructor, the mapper isn't called.

cjihrig: Sorry, let me rephrase my question. For this piece specifically:

```js
Array.from(this.map.keys()).map(toJavaString)
```

You are getting all of the keys and then mapping them to Java strings using the new `toJavaString()` function. That `map()` call is new. Instead of creating the new `toJavaString()` function, couldn't you pass the existing mapper there, which should handle the conversion via the code I linked to above?

alichherawalla: ah, I see your point! 

Let me make that change. Thanks for pointing that out

alichherawalla: @cjihrig,

In the interest of avoiding further back and forth. 

Is this what you mean? 
```
keySet() {
    return this.mapper(Array.from(this.map.keys()));
  }
```

alichherawalla: @cjihrig,

You good if we invoke `map` directly instead of using `this.mapper`, since `this.mapper` may or may not contain logic to convert a `string` to `JavaString`.

```
  keySet() {
    return map(Array.from(this.map.keys()));
  }
```

To add a bit more context, [this test](https://github.com/aws-amplify/amplify-cli/blob/5c9172af5dbd911ce86823f32e774e4c89fd9f88/packages/amplify-appsync-simulator/src/__tests__/velocity/value-mapper/map.test.ts#L65) will break if we use `this.mapper`

Apologies in case I misunderstood your suggestion.

cjihrig: The test suite passed for me with no modifications with this line:

```js
return new JavaArray(Array.from(this.map.keys()).map(this.mapper as any), this.mapper);
```

alichherawalla: just to double-check with you though, `this.mapper` may not always resolve a string to JavaString. 

For example, in the tests, we use `identityMapper`  which results in the newly added test, that expects a JavaArray with each element being a JavaString, to fail. 

Would there be a problem if we use this:

```
  keySet() {
    return map(Array.from(this.map.keys()));
  }
```

cjihrig: > just to double-check with you though, this.mapper may not always resolve a string to JavaString.

I think that is expected behavior. The mapper is responsible for converting input values to output values. The `identityMapper` just doesn't do any real mapping.

> Would there be a problem if we use this:

I think so. The reason for the `mapper` option is to customize the mapping behavior. Hard coding the mapper like that would break that. For your new test, you could use the real mapper though.

alichherawalla: The way I was thinking about it, is that the `keySet` function should always return a JavaArray < JavaString >, irrespective of the mapper that is passed to the JavaMap. 

Is there a case where the return value of the keySet would be dependant on the mapper of the JavaMap?

alichherawalla: Again, thank you so much for your time in reviewing this and patiently answering my questions.

packages/amplify-appsync-simulator/src/velocity/value-mapper/map.ts
```diff
@@ -56,7 +57,7 @@ export class JavaMap {
   }
 
   keySet() {
-    return new JavaArray(Array.from(this.map.keys()), this.mapper);
+    return map(Array.from(this.map.keys()));
```
cjihrig, line 60: Please update this to not use the default `map` function.

alichherawalla: done

packages/amplify-appsync-simulator/src/__tests__/velocity/value-mapper/map.test.ts
```diff
@@ -62,10 +68,16 @@ describe('JavaMap', () => {
 
   it('keySet', () => {
     const obj = { foo: 'Foo Value', bar: 'Bar Value' };
-    const map = new JavaMap(obj, identityMapper);
+    const map = new JavaMap(obj, mapper);
```
cjihrig, line 71: Can you revert the change here.

cjihrig (approved):
LGTM. Thank you for your contribution and for working through the comments.

Issue #4695, opened 2020-06-28
DynamoDB To ElasticSearch Function (DdbToEsFn-*) is VERY Slow


sacrampton:
**Describe the bug**

The nature of my solution is that I essentially write to DynamoDB and read from ElasticSearch - so its critical that writes to DynamoDB are written to ElasticSearch nearly instantaneously.

I have a utility to load data from a CSV file that calls the AppSync mutation and loads the data into DynamoDB. The DdbToEsFn then fires to send that data to ElasticSearch (provided the table is @searchable).

When looking at larger imports it was taking a very long time to get the data loaded where I could read it from ElasticSearch. I had just put this down to "AppSync must be slow". But I just noticed something when I went I loaded 40,437 records with an AppSync create mutation - these records loaded into DynamoDB VERY quickly. I didn't time it, but it was probably 90 seconds or so.

However, its now about 30 minutes later and only 3/4 of the records have made their way to ElasticSearch. So from this I can say that the AppSync create mutation is working really well and really performant - but the problem is with the Lambda function to stream the data to ElasticSearch.

**Amplify CLI Version**

4.21.3

**To Reproduce**

Bulk load a lot of data with an AppSync mutation at a searchable table and measure the time taken to load the data into DynamoDB and the time for that data to appear in ElasticSearch.

**Expected behavior**

I expect that data is streamed into ElasticSearch in almost the same time as it can be loaded into DynamoDB. That DynamoDB is nearly instantaneously streamed to ElasticSearch.

yuth:
@sacrampton I am trying to replicate what you're doing to see what is causing the lag in indexing the data. Could you clarify the following
1. When you load the CSV data do you run concurrent mutation in an thread pool
2. How long does it take for your script to load the data to AppSync using mutation
3. How long do you have to wait for Elastic search to catchup after the data is loaded using mutation
4. Could you share details how complex your types are so I can get some idea on the row size for each type that is being indexed

sacrampton:
Hi @yuth - thanks for taking a look at this. To answer your questions:

1. I believe we are loading via concurrent mutations (I'll double check with the developer)
2. As I mentioned I was seeing 40K records loading in under 2 minutes using the AppSync mutation. That is that the data had been loaded from the CSV file into DynamoDB.
3. I am waiting upwards of 30 minutes for those 40K records to appear in ElasticSearch.
4. My smallest type probably has 20 attributes and the largest has several hundred attributes.

Thanks

[...]

PR #7651, opened 2021-07-02
Create: BatchSize and MaximumBatchingWindow parameters, V2 streaming lambda

marcvberg:
#### Description of changes
This creates new parameters for BatchSize and MaximumBatchingWindow on the Lambda streaming function from DynamoDB to ElasticSearch for the GraphQL Transformer vNext. The default values are 100 records and 2 seconds, respectively

#### Issue #, if available
#4695 

I created a local project and validated that the streaming function generated for a searchable contained the default values from the new parameters

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

Don't see unit testing capability for creating parameters, ready to take another look if needed


cjihrig:
> The default values are 100 records and 2 seconds, respectively

I think we should keep the current defaults. By making them configurable, the values can be tweaked during something like backfilling, and then re-configured for "normal" operation.

marcvberg:
The case I would make for changing it is that it seems like it's not configurable at this point, we will need an additional change to pull in override parameters, and 100 is the default for DynamoDB EventSources already, we've just been overriding that.

Do we want to continue limiting it? The case I could see for it would be reducing the wait time for a single record to get to the streaming function, but increasing the batch size should bump throughput and cut cost by reducing API calls. Just want to consider benefits of both, or if there's a middle ground we think is better

cjihrig:
OK. Maybe default `maxBatchingWindow` to one second then?

marcvberg:
Sure, I can do that

cjihrig (approved):

PR #7665, opened 2021-07-05
ci: add support for e2e token rotation

johnpc:
#### Description of changes

Currently we use long-lived, manually generated access tokens to set up AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY environment variables allowing our tests to run. This is a bad practice - authorization tokens should be automatically rotated to minimize risk of the tokens landing in the wrong hands.

This PR sets up the Amplify CLI codebase to use temporary credentials during CI. Temporary STS credentials use an additional parameter, AWS_SESSION_TOKEN, to authorize the requests.

The token rotator itself is a lambda that runs every hour:

https://github.com/johnpc/amplify-e2e-token-rotation-lambda

This has been tested in CI on my fork and all tests pass:

https://app.circleci.com/pipelines/github/johnpc/amplify-cli/144/workflows/fb12f075-d881-4c92-8cf7-f8647afb909e

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -17,10 +17,7 @@ node12: &node12
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: ${AWS_ECR_ACCOUNT_URL}/amplify-cli-e2e-base-image-repo:latest
-      aws_auth:
-        aws_access_key_id: $ECR_ACCESS_KEY
-        aws_secret_access_key: $ECR_SECRET_ACCESS_KEY
+    - image: ${AWS_ECR_ACCOUNT_URL}/amplify-cli-e2e-base-image-repo-public:latest
   resource_class: large
 
 clean_e2e_resources: &clean_e2e_resources
--- a/packages/amplify-console-integration-tests/src/profile-helper.ts
+++ b/packages/amplify-console-integration-tests/src/profile-helper.ts
@@ -35,6 +35,7 @@ export function getConfigFromProfile() {
   return {
     accessKeyId: credentials[profileName].aws_access_key_id,
     secretAccessKey: credentials[profileName].aws_secret_access_key,
+    sessionToken: credentials[profileName].aws_session_token,
     region: config[configKeyName].region,
   };
 }
@@ -62,31 +63,33 @@ export function setupAWSProfile() {
   Object.keys(credentials).forEach(key => {
     const keyName = key.trim();
     if (profileName === keyName) {
-      credentials[key].aws_access_key_id = process.env.CONSOLE_AWS_ACCESS_KEY_ID;
-      credentials[key].aws_secret_access_key = process.env.CONSOLE_AWS_SECRET_ACCESS_KEY;
+      credentials[key].aws_access_key_id = process.env.AWS_ACCESS_KEY_ID;
+      credentials[key].aws_secret_access_key = process.env.AWS_SECRET_ACCESS_KEY;
+      credentials[key].aws_session_token = process.env.AWS_SESSION_TOKEN;
       isCredSet = true;
     }
   });
   if (!isCredSet) {
     credentials[profileName] = {
-      aws_access_key_id: process.env.CONSOLE_AWS_ACCESS_KEY_ID,
-      aws_secret_access_key: process.env.CONSOLE_AWS_SECRET_ACCESS_KEY,
+      aws_access_key_id: process.env.AWS_ACCESS_KEY_ID,
+      aws_secret_access_key: process.env.AWS_SECRET_ACCESS_KEY,
+      aws_session_token: process.env.AWS_SESSION_TOKEN,
     };
   }
 
-  process.env.CONSOLE_REGION = process.env.CONSOLE_REGION || testRegionPool[Math.floor(Math.random() * testRegionPool.length)];
+  process.env.CLI_REGION = process.env.CLI_REGION || testRegionPool[Math.floor(Math.random() * testRegionPool.length)];
   let isConfigSet = false;
   Object.keys(config).forEach(key => {
     const keyName = key.replace('profile', '').trim();
     if (profileName === keyName) {
-      config[key].region = process.env.CONSOLE_REGION;
+      config[key].region = process.env.CLI_REGION;
       isConfigSet = true;
     }
   });
   if (!isConfigSet) {
     const keyName = profileName === 'default' ? 'default' : `profile ${profileName}`;
     config[keyName] = {
-      region: process.env.CONSOLE_REGION,
+      region: process.env.CLI_REGION,
     };
   }
 
--- a/packages/amplify-console-integration-tests/src/profile-helper.ts
+++ b/packages/amplify-console-integration-tests/src/profile-helper.ts
@@ -77,19 +77,19 @@ export function setupAWSProfile() {
     };
   }
 
-  process.env.CLI_REGION = process.env.CLI_REGION || testRegionPool[Math.floor(Math.random() * testRegionPool.length)];
+  process.env.CONSOLE_REGION = process.env.CONSOLE_REGION || testRegionPool[Math.floor(Math.random() * testRegionPool.length)];
   let isConfigSet = false;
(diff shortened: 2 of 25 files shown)


edwardfoyle (commented):
Looks like `injectSessionToken()` is called in the chain.run() callback which will actually execute _after_ the CLI run completes. Not sure if this is intentional or not.

Depending on the timeout on the session token, maybe it makes sense to only call `injectSessionToken` in the `beforeAll` function for each test suite? I also worry about authors of future tests and test updates that now have to remember to add this function call. If there was a way to get this functionality "automatically" that would be ideal.

johnpc:
> Looks like injectSessionToken() is called in the chain.run() callback which will actually execute after the CLI run completes. Not sure if this is intentional or not.

This is intentional. What happens is `amplify init` creates the profile and sets the access key id / secret key, then the session token is injected after. I know for sure it works and is needed because if the contents of `injectSessionToken()` are commented out most tests fail.

> Depending on the timeout on the session token, maybe it makes sense to only call injectSessionToken in the beforeAll function for each test suite? I also worry about authors of future tests that now have to remember to add this function call. If there was a way to get this functionality "automatically" that would be ideal.

I think it makes more sense to only add it to the `init` methods already existing in amplify-e2e-core so that it _does not_ have to be duplicated by each test. Instead the injection occurs during execution of shared code and is therefore more automatic.

.circleci/config.base.yml
```diff
@@ -17,10 +17,7 @@ node12: &node12
 defaults: &defaults
   working_directory: ~/repo
   docker:
-    - image: ${AWS_ECR_ACCOUNT_URL}/amplify-cli-e2e-base-image-repo:latest
-      aws_auth:
-        aws_access_key_id: $ECR_ACCESS_KEY
-        aws_secret_access_key: $ECR_SECRET_ACCESS_KEY
+    - image: ${AWS_ECR_ACCOUNT_URL}/amplify-cli-e2e-base-image-repo-public:latest
```
yuth, line 20: We use different ENV variable to keep the credentials separate and have minimal permission for the ECR credentials. With the removal of this would it be possible to still keep them sepearate?

packages/amplify-console-integration-tests/src/profile-helper.ts
```diff
@@ -62,31 +63,33 @@ export function setupAWSProfile() {
...
-      aws_access_key_id: process.env.CONSOLE_AWS_ACCESS_KEY_ID,
-      aws_secret_access_key: process.env.CONSOLE_AWS_SECRET_ACCESS_KEY,
+      aws_access_key_id: process.env.AWS_ACCESS_KEY_ID,
+      aws_secret_access_key: process.env.AWS_SECRET_ACCESS_KEY,
+      aws_session_token: process.env.AWS_SESSION_TOKEN,
     };
   }
 
-  process.env.CONSOLE_REGION = process.env.CONSOLE_REGION || testRegionPool[Math.floor(Math.random() * testRegionPool.length)];
+  process.env.CLI_REGION = process.env.CLI_REGION || testRegionPool[Math.floor(Math.random() * testRegionPool.length)];
```
yuth, line 80: This environment variable is set separately as we have some custom allow list for our console e2e account to provision more Amplify App. This allow list is region specific and we run it in a region where there are not much traffic. If we want to reuse the `CLI_REGION`, then it should be overriden in the console e2e test

packages/amplify-e2e-core/src/configure/index.ts
```diff
@@ -66,6 +66,7 @@ export function amplifyConfigure(settings: AmplifyConfiguration): Promise<void>
       .wait('Successfully set up the new user.')
       .run((err: Error) => {
         if (!err) {
+          injectSessionToken();
```
yuth, line 69: The callback passed to `run` gets called after the command is executed. So I think the injectSessionToken should probably be called before `run`

packages/amplify-e2e-core/src/index.ts
```diff
@@ -31,6 +32,12 @@ export function isCI(): boolean {
   return process.env.CI && process.env.CIRCLECI ? true : false;
 }
 
+export function injectSessionToken() {
+  if (!fs.readFileSync(pathManager.getAWSCredentialsFilePath()).toString().includes('aws_session_token')) {
+    fs.appendFileSync(pathManager.getAWSCredentialsFilePath(), `aws_session_token=${process.env.AWS_SESSION_TOKEN}`);
```
yuth, line 37: The credential file uses `ini` format. We should be using `ini` package and update the file instead doing an `appendFileSync`

yuth (changes requested):
I suspect the tests are passing because we are using standard ENV variable and the SDK might be picking up the session token from the ENV variable. Could you change how we inject session token. Another option would be to update the profile credentials  with session token along with access and secret Key, when setting up the profile [here](https://github.com/aws-amplify/amplify-cli/blob/0c0adfb78350a192d4f44b722d6038b23c505527/packages/amplify-e2e-tests/src/configure_tests.ts#L3-L21)

.circleci/config.base.yml
johnpc: the ERC creds are still separate. unfortunately the circleci aws_auth config does not support session_token and cannot be forked. So the only way to do it is to make the repo public so that no auth is required upon pull.

packages/amplify-console-integration-tests/src/profile-helper.ts
johnpc: Ah, we can have multiple if we want. I was just hoping to clear up any redundancies.

packages/amplify-e2e-core/src/configure/index.ts
johnpc: as long as the access/secret keys get set and the file gets created before we call `run` that's fine with me. But it does work when implemented this way.

packages/amplify-e2e-core/src/index.ts
johnpc: :+1:

.circleci/config.base.yml
yuth: So the options we have are either
 - Use a long lived credentials which only allows pulling the image
 - Make the pull for repo public
I would rather make the repo public in that case and not be bothered with it. But for writing we will have to still use credentials

packages/amplify-e2e-core/src/configure/index.ts
yuth: If you're using the standard ENV variable name I think the SDK probably picks it up

packages/amplify-e2e-core/src/index.ts
```diff
@@ -31,6 +34,14 @@ export function isCI(): boolean {
...
 }
 
+export function injectSessionToken() {
+  const credentialsContents = ini.parse(fs.readFileSync(pathManager.getAWSCredentialsFilePath()).toString());
+  for (const key of Object.keys(credentialsContents)) {
+    credentialsContents[key].aws_session_token = process.env.AWS_SESSION_TOKEN;
+  }
+  fs.writeFileSync(pathManager.getAWSCredentialsFilePath(), ini.stringify(credentialsContents));
+}
+
```
yuth, line 44: I would prefer if this happened as part of setting up the profile after running the configure command to ensure the profile gets set correctly and done only once
https://github.com/aws-amplify/amplify-cli/blob/cd5d33aa822ceeb19a1af847d8c3eab0f1d10632/packages/amplify-e2e-tests/src/configure_tests.ts#L5-L17

.circleci/config.base.yml
johnpc: Yes - I've done that in this pull request: https://github.com/aws-amplify/amplify-cli/pull/7736

packages/amplify-e2e-core/src/configure/index.ts
johnpc: That probably is why it works this way. I think the amplify-cli codebase though reads from the config file, because these tests do not pass when I don't include this injection.

yuth (approved):

PR #7674, opened 2021-07-06
fix: add/update function when some LL are not yet migrated

jhockett:
#### Description of changes

Handle loading, and disabling of non-migrated LL in add/update function flows:

```
? Do you want to configure advanced settings? Yes
...
? Do you want to enable Lambda layers for this function? Yes
? Provide existing layers or select layers in this project to access from this function (pick up to 5): (Press <space> to select, <a> to toggle all, 
<i> to invert selection)
❯◯ Provide existing Lambda layer ARNs
 ◯ LayerA123
 - LayerB123 (Layer requires migration. Run "amplify function update" and choose this layer to migrate.)
```

Manual testing, yarn test, added migration test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
@@ -53,7 +53,7 @@ export const askLayerSelection = async (
     };
   }
 
-  const disabledMessage = 'Layer requires migration. Run "amplify update function" and choose this layer to migrate.';
+  const disabledMessage = 'Layer requires migration. Run "amplify function update" and choose this layer to migrate.';
   const currentResourceNames = filterProjectLayers(previousSelections).map(sel => (sel as ProjectLayer).resourceName);
   const choices = layerOptions.map(op => ({
     name: op,
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
@@ -39,8 +39,10 @@ export const askLayerSelection = async (
   const layerOptions = _.keys(functionMeta)
     .filter(key => functionMeta[key].service === ServiceName.LambdaLayer)
     .filter(key => {
-      // filter by compatible runtimes
-      return isRuntime(runtimeValue).inRuntimes(functionMeta[key].runtimes || getLayerRuntimes(key));
+      // filter by compatible runtimes - unless no runtimes are present for the given Lambda layer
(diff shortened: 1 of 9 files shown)


packages/amplify-category-function/src/provider-utils/awscloudformation/utils/addLayerToFunctionUtils.ts
```diff
@@ -49,9 +50,19 @@ export const askLayerSelection = async (
       askArnQuestion: true,
     };
   }
+
+  const disabledMessage = 'Layer requires migration. Run "amplify update function" and choose this layer to migrate.';
```
attilah, line 54: Can we use `amplify function update` as that's the by-design order, nevertheless it is working the other way around as well...

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerConfiguration.ts
```diff
@@ -24,7 +25,16 @@ export function getLayerConfiguration(layerName: string) {
 }
 
 export function getLayerRuntimes(layerName: string) {
-  return getLayerConfiguration(layerName).runtimes;
+  try {
+    return getLayerConfiguration(layerName).runtimes;
```
attilah, line 29: Can we do the proper checks here instead of `try..catch`? like check file existence, read it, look for runtimes property, etc...

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerMigrationUtils.ts
```diff
@@ -183,3 +183,20 @@ export function removeLayerFromTeamProviderInfo(envName: string, layerName: stri
...
   stateManager.setTeamProviderInfo(projectRoot, teamProviderInfo);
 }
+
+function readLegacyLayerParametersJson(layerDirPath: string) {
+  return JSONUtilities.readJson<LegacyLayerParametersJson>(path.join(layerDirPath, LegacyFilename.layerParameters));
+}
+
+function migrateAmplifyProjectFiles(layerName: string, latestLegacyHash: string, envName?: string) {
+  const projectRoot = pathManager.findProjectRoot();
+  removeLayerFromTeamProviderInfo(envName, layerName, projectRoot);
```
attilah, line 193: usually `projectRoot` is the first argument when we operate on state files.

attilah (approved):
LGTM in general, thanks for the migration test addition, added a few comments.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerConfiguration.ts
jhockett: I learned from @cjihrig not to do a file existence check, and instead use try/catch because it prevents a race condition.

I will add a check for the `runtimes` key though

Issue #7441, opened 2021-06-02
Trying to init an existing amplify repo of other creator 


Kevinberg:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### Amplify CLI Version

4.50.0

### What operating system are you using?

Mac

### Amplify Categories

Not applicable

### Amplify Commands

init

### Describe the bug

I am getting the following errors when I want to init an amplify project created by an other author:

⠦ Initializing your environment: devCould not initialize 'dev': Access Denied

Failed to get profile: Profile configuration is missing for: undefined

The steps I took:

npm install -g @aws-amplify/cli
$ amplify configure
amplify init --app https://github.com/Savinvadim1312/NetflixClone.git

? Do you want to use an existing environment? No
? Enter a name for the environment netflix
Using default provider  awscloudformation
? Select the authentication method you want to use: AWS profile

For more information on AWS Profiles, see:
https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html

? Please choose the profile you want to use default
? Select the authentication method you want to use: AWS profile
**Failed to get profile: Profile configuration is missing for: undefined**

_On a forum it said that you should add 2 files: , ~/.aws/config and ~/.aws/credentials_ to the root folder....

Currently these files are located here is that wrong? (if I delete them and try a new init then a new config and credentials are created there). 

/Users/kevinbergmeijer/.aws

### Expected behavior

Being able to launch this project by creating a new enviorment

### Reproduction steps

See above

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Additional information

.

siegerts:
Hi @Kevinberg :wave:, thanks for raising. Is this the same/duplicate of this issue https://github.com/aws-amplify/amplify-cli/issues/7428? If so, can you please consolidate and close the duplicate

Kevinberg:
[...]

PR #7682, opened 2021-07-07
fix: #7441 - init from git prompts for credentials twice

attilah:
#### Description of changes

- reference to `awsConfigInfo` with this name everywhere
- add `general` authType support when determining the auth type for a project
- added cleanup of Amplify state files after cloning a project to make sure other developers' resources are not leaking into the project (which later can lead to Forbidden or other errors during delete for example).
- guard spinner destruction when doing SSM operations

#### Issue #, if available

fixes: #7441 - init from git prompts for credentials twice

Manual testing of init and delete with various projects.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/system-config-manager.ts
+++ b/packages/amplify-provider-awscloudformation/src/system-config-manager.ts
@@ -13,7 +13,7 @@ const logger = fileLogger('system-config-manager');
 const credentialsFilePath = pathManager.getAWSCredentialsFilePath();
 const configFilePath = pathManager.getAWSConfigFilePath();
 
-export function setProfile(awsConfig: $TSAny, profileName: string) {
+export function setProfile(awsConfigInfo: $TSAny, profileName: string) {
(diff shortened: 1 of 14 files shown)


cjihrig (approved):
LGTM. Also verified locally that this fixes project deletion using the headless `configLevel: general` setting.

packages/amplify-provider-awscloudformation/src/system-config-manager.ts
```diff
@@ -13,7 +13,7 @@ const logger = fileLogger('system-config-manager');
 const credentialsFilePath = pathManager.getAWSCredentialsFilePath();
 const configFilePath = pathManager.getAWSConfigFilePath();
 
-export function setProfile(awsConfig: $TSAny, profileName: string) {
+export function setProfile(awsConfigInfo: $TSAny, profileName: string) {
```
jhockett, line 16: It's not apart of this change, but consider changing the type of `awsConfigInfo` to AwsSdkConfig

jhockett (approved):

PR #7695, opened 2021-07-09
test: add e2e test for geo

AaronZyLee:
#### Description of changes

Add e2e test cases for amplify geo category(map and place index). The commands are `add`, `update` and `remove`.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-tests/src/__tests__/geo_1.test.ts
```diff
@@ -0,0 +1,56 @@
+import {
```
phani-srikar, line 1: can we name the test files as `geo-add.test.ts` or equivalent to make it more clear?

phani-srikar (changes requested):
Nice set of e2e tests to cover the default walkthroughs. 
I have listed out some cases mostly related to changing the `isDefault` setting of maps/place-indices, that would be good to have. The default setting is important to cover in the sense that the Amplify Frontend Clients will use it to make all default SDK calls.

packages/amplify-e2e-tests/src/__tests__/geo_1.test.ts
AaronZyLee: Sure. The previous naming was just following the CLI's convention.

phani-srikar (approved):
LGTM! Found nothing of major concern.

PR #7698, opened 2021-07-09
Feature/amplify enhanced status

sachscode:
#### Description of changes
Feature description:
Amplify CLI's amplify status command shows the deployment state for Amplify resources (Create/Update/Delete/No Change). The command output only provides a high level view of the state of application resources.However, customers today can't see a detailed view of the pending changes. Customers especially lack clarity about changes to nested resources like GraphQL APIs, where a change may result in updates to multiple different AWS resources. Additional details can also inform customers about the security and financial implications of their next deployment. A detailed “diff view” on the terminal window is also helpful in generating useful data for debugging deployment failures.

CLI changes :
1. amplify status -v :  Introduced verbose mode status to display cloudformation-diff 
2. amplify status -v [ <category> ] : Introduced verbose mode with category to display only the cloudformation-diff for 
3. resources of a specific category or a list of categories.
   amplify status help : Introduced help for display of command syntax and examples

Implementation:
The "amplify status -v" flow , generates a visual diff  of the cloudformation templates in each resource subfolder in ${projectRoot}/amplify folder and displays it on the console.

known issues:
The amplify status -v feature currently doesn't show the difference in params.json. This will be added as part of subsequent updates to the "amplify status -v command".

1. Unit tests to trigger resource-status: showStatusTable() 
2. Manually tested for API, Storage, Function, Auth, Hosting, Analytics, Predictions, Notifications
    verified presence cloudformation-diff is displayed on add, update and delete..

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+          const absolutePath = path.join(fileFolder, templateFileName);
+          return absolutePath; //We support only one cloudformation template ( nested templates are picked )
+        }
+      }
+      return ""
+    }
+
+    //Load Cloudformation template from file.
+    //The filePath should end in json/yaml or yml
+    private loadCfnTemplate = async(filePath) => {
```
kaustavghosh06, line 232: Should this be a common helper code and I think we have a common helper function for loading cfn file? cc @attilah @yuth

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+          });
+        }
+        if (!templateDiff.isEmpty) {
+          cfnDiff.formatDifferences(stream || process.stderr, templateDiff);
+        }
+        return templateDiff.differenceCount;
+    }
+
+    //helper: reads file and deserializes
+    private loadStructuredFile = async(fileName: string, deserializeFunction) => {
```
kaustavghosh06, line 201: What does this function do? What is the deserialization function here? 
We have a common helper function to read/write files. Can we use that instead?

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+    const inputTypes = [ 'json', 'yaml', 'yml'] ; //check for existence of any one of the extensions.
+    for( let i = 0 ; i < inputTypes.length ; i++ ){
+      if ( fs.existsSync(`${filePath}.${inputTypes[i]}`) ){
+        return true;
+      }
+    }
+    return false;
+}
+
+//Interface to store template-diffs for C-R-U resource diffs
```
kaustavghosh06, line 265: For C-D-U? :)

packages/amplify-cli/src/index.ts
```diff
@@ -64,20 +64,37 @@ process.on('unhandledRejection', function (error) {
...
     let input = getCommandLineInput(pluginPlatform);
-
     // with non-help command supplied, give notification before execution
     if (input.command !== 'help') {
       // Checks for available update, defaults to a 1 day interval for notification
       notify({ defer: false, isGlobal: true });
     }
 
+    //Normalize status command options
+    if ( input.command == 'status'){
```
kaustavghosh06, line 94: Why do we need a special case for status? Can't we include this in the generic normalize parsing function?

packages/amplify-cli/src/commands/status.ts
```diff
@@ -1,10 +1,22 @@
...
-  await context.amplify.showHelpfulProviderLinks(context);
-  await showAmplifyConsoleHostingStatus(context);
+  const cliParams:CLIParams  = { cliCommand : context?.input?.command,
+                                 cliSubcommands: context?.input?.subCommands,
+                                 cliOptions : context?.input?.options }
+  const view = new ViewResourceTableParams( cliParams );
+  if ( context?.input?.subCommands?.includes("help")){
+     console.log( view.getStyledHelp())
+  } else {
+    await context.amplify.showStatusTable( view );
```
kaustavghosh06, line 13: Would this return gracefully when a wrong category input is passed here?

packages/amplify-cli/src/extensions/amplify-helpers/yaml-cfn.ts
```diff
@@ -0,0 +1,59 @@
+import * as yaml from 'yaml';
```
kaustavghosh06, line 1: Why do we need this file? Can this be re-used from somewhere?

packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
```diff
@@ -9,13 +9,17 @@ import { CLOUD_INITIALIZED, CLOUD_NOT_INITIALIZED, getCloudInitStatus } from './
 import { ServiceName as FunctionServiceName, hashLayerResource } from 'amplify-category-function';
```
kaustavghosh06, line 9: This file has grown way too huge - if possible can we look at splitting this file up (only if it makes sense to do so )

kaustavghosh06 (commented):
Still looking at it..but some initial comments

packages/amplify-cli/src/commands/status.ts
sachscode: Yes, If an incorrect category is provided, the CLI would print an empty summary table and no cloudformation-diff. We could update this to display an error message and a list of valid category types in subsequent PRs.

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+        this.cloudTemplate = {}; //requires file-access, hence loaded from async methods
+        //Note: All file names include full-path but no extension.Extension will be added later.
+        this.resourceFiles = {
+            localTemplateFile : path.normalize(path.join(this.localBackendDir, category, resourceName)),
+            cloudTemplateFile : path.normalize(path.join(this.cloudBackendDir, category, resourceName)),
+            //Paths using glob for Cfn file and Build file
+            localPreBuildCfnFile : this.globCfnFilePath( path.normalize(path.join(this.localBackendDir, category, resourceName))),
+            cloudPreBuildCfnFile : this.globCfnFilePath( path.normalize(path.join(this.cloudBackendDir, category, resourceName))),
+            localBuildCfnFile : this.globCfnFilePath( path.normalize(path.join(this.localBackendDir, category, resourceName, 'build'))),
+            cloudBuildCfnFile : this.globCfnFilePath( path.normalize(path.join(this.cloudBackendDir, category, resourceName, 'build'))),
```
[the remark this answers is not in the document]
sachscode: 1. Is this just for API? Can't this logic be handled in the this.globCfnFilePath function?
   The globCfnFilePath - when given a subdir-path fetches the cloudformation template in that subfolder. The intention was to reuse that function for other paths. We can indeed move it inside for brevity. 
2. localPreBuildCfnFile - Cloudformation which exists in backend-dir/category/resourceName . This is generated using the input params and amplify's template engine.
3. localBuildCfnFile  - Cloudformation which exists in backend-dir/category/resourceName/build folder. This is generated after a compilation step at the resource level. e.g GraphQL api has a build folder where the cfn gets generated after gql-compile.

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+      }
+    }
+
+    //helper: Convert cloudformation template diff data using CDK api
+    private printStackDiff = ( templateDiff, stream?: cfnDiff.FormatStream ) =>{
+        // filter out 'AWS::CDK::Metadata' resources from the template
+        if (templateDiff.resources ) {
+          templateDiff.resources = templateDiff.resources.filter(change => {
+            if (!change) { return true; }
+            if (change.newResourceType === 'AWS::CDK::Metadata') { return false; }
```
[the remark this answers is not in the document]
sachscode: CDK metadata is used by CDK for its bookkeeping , and the CDK view engine doesn't expect this information.

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
sachscode: Yes , can refactor to use the common function for reading files and applying deserialization.

packages/amplify-cli/src/extensions/amplify-helpers/yaml-cfn.ts
sachscode: Will replace this with yaml library call.

packages/amplify-cli/src/index.ts
```diff
@@ -64,20 +64,37 @@ process.on('unhandledRejection', function (error) {
   throw error;
 });
 
+function normalizeStatusCommandOptions( statusCommandOptions){
+  let options = statusCommandOptions;
+  //Normalize 'amplify status -v' to verbose, since -v is interpreted as 'version'
```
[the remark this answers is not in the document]
sachscode: Sure will do.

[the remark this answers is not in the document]
sachscode: --verbose  already works ( this is because the key used internally is 'verbose')

packages/amplify-cli/src/index.ts
sachscode: There were 2 data points for this decision
1. amplify -v flag maps to "version" of the library 
   If we were to support versioning ( and independent updating ) of categories in the future , we would need a "-v" option for each category, which would display, version, and the compatibility matrix.   
Status is special , since version information is provided already as part of the response (both summary and detailed).
2. We can definitely move this into a generic normalize function. The idea was to any specific behavior as a separate function , and to move them as specific implementations of an commandType interface .  
e.g <commandType>=><status>      => normalizeInput ( overridden by normalizeStatus ) 
                                    =><category> => normalizeInput ( overridden by category-specific

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
sachscode: Looks like there is a function readCFNTemplate in amplify-core . I will refactor to use that and remove this extraneous code.

packages/amplify-cli/src/commands/status.ts
```diff
@@ -1,10 +1,22 @@
+import { constructCloudWatchEventComponent } from "amplify-category-function/src/provider-utils/awscloudformation/utils/cloudformationHelpers";
```
akshbhu, line 1: where is this used?

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+export class ResourceDiff {
+    resourceName: string;
+    category : string;
+    provider : string;
+    service: string;
+    resourceFiles : ResourcePaths;
+    localBackendDir : string;
+    cloudBackendDir : string;
+    localTemplate : {
+                        [key: string]: any;
```
akshbhu, line 117: Can we have any as ```Template``` or some other type?

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+      //load resource template objects
+      this.localTemplate = await this.loadCfnTemplate(resourceTemplatePaths.localTemplatePath)
+      this.cloudTemplate = await this.loadCfnTemplate(resourceTemplatePaths.cloudTemplatePath);
+      const diff : cfnDiff.TemplateDiff = cfnDiff.diffTemplate(this.cloudTemplate, this.localTemplate );
+      return diff;
+    }
+
+    //helper: Select cloudformation file path from build folder or non build folder.
+    private getCfnResourceFilePaths = async() => {
+      return {
```
akshbhu, line 167: I think we have to return a promise here
return new Promise({}) for async/await fn

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+      //file-names are currently standardized around "cloudformation" not awscloudformation.
+      if ( provider === "awscloudformation"){
+        return CategoryProviders.CLOUDFORMATION
+      } else {
+        return provider;
+      }
+    }
+
+    //helper: Convert cloudformation template diff data using CDK api
+    private printStackDiff = ( templateDiff, stream?: cfnDiff.FormatStream ) =>{
```
akshbhu, line 184: any type for templateDiff?

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,295 @@
...
+    //The filePath should end in json/yaml or yml
+    private loadCfnTemplate = async(filePath) => {
+      if( filePath === ""){
+        return {}
+      }
+      const fileType = this.getFilePathExtension(filePath)
+      try {
+        //Load yaml or yml or json files
+        let providerObject = {};
+        if (fs.existsSync(filePath) ){
```
akshbhu, line 240: Instead of existsSync, Can we have fs.readFile() in try/catch to avoid multiple file system operations

packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
```diff
@@ -9,13 +9,17 @@ import { CLOUD_INITIALIZED, CLOUD_NOT_INITIALIZED, getCloudInitStatus } from './
...
 import { removeGetUserEndpoints } from '../amplify-helpers/remove-pinpoint-policy';
 import { pathManager, stateManager, $TSMeta, $TSAny, NotInitializedError } from 'amplify-cli-core';
+import { ViewResourceTableParams } from "amplify-cli-core/lib/cliViewAPI";
```
akshbhu, line 12: Can we export  ```ViewResourceTableParams``` index file so that we can directly import from 'amplify-cli-core'

akshbhu (changes requested):
Nice Work !!

Added a few minor comments , will continue to review more

packages/amplify-cli-core/src/cliViewAPI.ts
```diff
@@ -0,0 +1,67 @@
+//Use this file to store all types used between the CLI commands and the view/display functions
+// CLI=>(command-handler)==[CLI-View-API]=>(ux-handler/report-handler)=>output-stream
+import chalk from 'chalk';
+export interface CLIParams {
+    cliCommand : string;
+    cliSubcommands: string[]|undefined;
+    cliOptions : {[key:string] : any};
```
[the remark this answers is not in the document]
sachscode: This is definitely required. I can make this change as part of a subsequent PR, since the current code has some refactoring on resource-status.ts ( yet to be pushed ) besides the original feature in the PR. 
The problem with cliOptions is that it's  derived from `context.input.options`. In our code, we currently have context as any (this too will be solidified in subsequent PRs). Based on its usage in the code, it seems to be a generically defined type ( with the intent to pass any object from the command-line handler to the CLI engine. ) 
Hence it would be mitigate some risks by moving this change to the next PR

packages/amplify-cli-core/src/cliViewAPI.ts
```diff
@@ -0,0 +1,67 @@
...
+}
+//Resource Table filter and display params (params used for summary/display view of resource table)
+export class ViewResourceTableParams {
+    public command : string;
```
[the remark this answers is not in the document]
sachscode: should not be public, will fix.

packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
sachscode: will do

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
sachscode: The function is defined as async. This automatically wraps the return value into a promise.

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
sachscode: Will replace with cfnDiff.TemplateDiff

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
sachscode: used Template from cloudform-types

packages/amplify-cli/src/commands/status.ts
sachscode: Its not currently used. It was left from an earlier attempt to save cloudwatch-logs from amplify-push into a logfile and to surface them as error in status summary. removing this for now.

packages/amplify-cli-core/src/cliViewAPI.ts
```diff
@@ -0,0 +1,67 @@
...
+    public verbose : boolean; //display table in verbose mode
+    public help : boolean; //display help for the command
+    public categoryList : string[]|[] //categories to  display
+    public filteredResourceList : any //resources to *not* display - TBD define union of valid types
+    getCategoryFromCLIOptions( cliOptions : object ){
```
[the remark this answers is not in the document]
sachscode: Will addressed ( as above) as part of the PR to add types to CLI options

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
sachscode: The reason for adding existsSync was to detach template-path selection logic from file-load logic. In subsequent PRs, my intention is to remove existsSync and rely solely on the resource-provider-type + resource-state to infer path. (this will allow us to auto-calculate diffs and cache them whenever state changes (reactive/event-based), reducing response-times for bigger stacks).

packages/amplify-cli-core/src/index.ts
```diff
@@ -1,4 +1,6 @@
+import { ViewResourceTableParams, CLIParams } from './cliViewAPI';
 import { ServiceSelection } from './serviceSelection';
+export { ViewResourceTableParams, CLIParams };
```
akshbhu, line 3: already exported in line 26?

packages/amplify-cli/src/domain/amplify-toolkit.ts
```diff
@@ -258,6 +259,13 @@ export class AmplifyToolkit {
       this._showResourceTable || require(path.join(this._amplifyHelpersDirPath, 'resource-status')).showResourceTable;
     return this._showResourceTable;
   }
+
+  get showStatusTable(): any {
+    this._showStatusTable =
+    this._showStatusTable || require(path.join(this._amplifyHelpersDirPath, 'resource-status')).showStatusTable;
```
akshbhu, line 265: Can this be changed to await import?

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-data.ts
```diff
@@ -0,0 +1,536 @@
...
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import { ServiceName as FunctionServiceName, hashLayerResource } from 'amplify-category-function';
+import { removeGetUserEndpoints } from '../amplify-helpers/remove-pinpoint-policy';
+import { pathManager, stateManager, NotInitializedError } from 'amplify-cli-core';
+import { hashElement, HashElementOptions } from 'folder-hash';
+import { CLOUD_INITIALIZED, CLOUD_NOT_INITIALIZED, getCloudInitStatus } from './get-cloud-init-status';
+import * as resourceStatus from './resource-status-diff';
+import { IResourceDiffCollection } from './resource-status-diff';
+import { ViewResourceTableParams } from 'amplify-cli-core';
```
akshbhu, line 11: Can be inlcuded in line 7

packages/amplify-cli/src/commands/status.ts
```diff
@@ -1,13 +1,29 @@
...
+  const view = new ViewResourceTableParams( cliParams );
+  if ( context?.input?.subCommands?.includes("help")){
+     console.log( view.getStyledHelp())
+  } else {
+    try {
+      await context.amplify.showStatusTable( view );
+      await context.amplify.showHelpfulProviderLinks(context);
+      await showAmplifyConsoleHostingStatus(context);
+    } catch ( e ){
+      view.logErrorException(e);
```
akshbhu, line 18: Can we context.print.error here instead of console.log()?

packages/amplify-cli/src/commands/status.ts
```diff
@@ -1,13 +1,29 @@
...
+    }
+  }
 };
 
-async function showAmplifyConsoleHostingStatus(context) {
+async function showAmplifyConsoleHostingStatus( context) {
   const pluginInfo = context.amplify.getCategoryPluginInfo(context, 'hosting', 'amplifyhosting');
   if (pluginInfo && pluginInfo.packageLocation) {
-    const { status } = require(pluginInfo.packageLocation);
+    const { status } = await import(pluginInfo.packageLocation);
```
akshbhu, line 26: nice

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
```diff
@@ -0,0 +1,261 @@
...
+  allResources:any[],
+  tagsUpdated: boolean
+}
+
+//"CollateResourceDiffs" - Calculates the diffs for the list of resources provided.
+// note:- The mutationInfo may be used for styling in enhanced summary.
+export async function CollateResourceDiffs( resources , mutationInfo : StackMutationInfo  /* create/update/delete */ ){
+    const provider = CategoryProviders.CLOUDFORMATION;
+    let resourceDiffs : ResourceDiff[] = [];
+    for await (let resource of resources) {
```
akshbhu, line 255: const

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-view.ts
```diff
@@ -0,0 +1,36 @@
+import { getSummaryTableData, getResourceDiffs } from './resource-status-data';
+import * as resourceStatus from './resource-status-diff';
+import { getEnvInfo } from './get-env-info';
+import { print } from './print';
+import chalk from 'chalk';
+//view: displays resource-diff (cloudformation-diff, input parameters (pending))
+export async function viewResourceDiffs({ resourcesToBeUpdated, resourcesToBeDeleted, resourcesToBeCreated }) {
+  const resourceDiffs = await getResourceDiffs(resourcesToBeUpdated, resourcesToBeDeleted, resourcesToBeCreated);
+  for await (let resourceDiff of resourceDiffs.updatedDiff) {
```
akshbhu, line 9:
```suggestion
  for await (const resourceDiff of resourceDiffs.updatedDiff) {
```

Also for other loops

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-data.ts
```diff
@@ -0,0 +1,536 @@
...
+
+//helper to await results of all function calls
+//TODO: replace with 'await for of'
+async function asyncForEach(array, callback) {
+  for (let index = 0; index < array.length; ++index) {
+    await callback(array[index], index, array);
+  }
+}
+//TODO: replace with some library function
+function capitalize(str) {
```
akshbhu, line 534: Can we have this function in one file and exported , it seems same functions is in `resource-status-diff.ts`

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-data.ts
```diff
@@ -0,0 +1,536 @@
...
+  }
+
+  if (category !== undefined && !resourceName) {
+    // Create all the resources for the specified category in the cloud
+    resources = resources.filter(resource => resource.category === category);
+  }
+
+  // Check for dependencies and add them
+
+  for (let i = 0; i < resources.length; ++i) {
```
akshbhu, line 218: Can we changed it to ForEach() , just like in other functions if time permits?

akshbhu (commented):
Some minor nits , otherwise LGTM!!

We can also add a test case to check the verbose out matches as expected if its possible

packages/amplify-cli/src/domain/amplify-toolkit.ts
sachscode: This will require conversion of showStatusTable function into async . This is a different behavior compared to other functions in this class. For now I feel it's better to keep it as require and update it once for all functions in this class as part of performance update.

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-data.ts
sachscode: yes , will do!

packages/amplify-cli/src/commands/status.ts
sachscode: will do

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-diff.ts
sachscode: will do

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-data.ts
sachscode: moving it to print , since this is only used for view-functions and labels.

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-data.ts
sachscode: This seems to be a dual iteration for dependencies. will refactor this separately.

sachscode: This code was essentially one part of a 3 part split of the resource-status.ts file (legacy summary table), hence it didn't have some of the best practices which implemented in the earlier files. Have refactored it to use constants, common functions, for each among other things.

packages/amplify-cli/src/commands/status.ts
sachscode: Done

packages/amplify-cli/src/extensions/amplify-helpers/resource-status-data.ts
sachscode: Done

akshbhu (approved):
Changes looks good , just Resolve LGTM issues

PR #7700, opened 2021-07-10
fix: upgrade node default runtime to 14

michaelbrewer:
#### Description of changes

As we approach the EOL for node 12 security updates new
lambdas should be ideally be on node 14

- [Node release schedule](https://github.com/nodejs/Release#release-schedule)

Generate example stacks to verify node 14 version was used

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/amplify-e2e-tests/src/__tests__/migration/node.function.test.ts
```diff
@@ -44,7 +44,7 @@ describe('nodejs version migration tests', () => {
 
     fs.writeFileSync(projectConfigFileName, projectConfigContent, 'utf-8');
 
-    // Write back the nodejs version to 'node10.x' to test migration
+    // Write back the nodejs version to 'node12.x' to test migration
```
attilah, line 47: Leave it on 10.x, but change `nodejs12.x` to `nodejs14.x` in line 59.

attilah (changes requested):
thanks for the PR LGTM!

Except 1 change, I added a comment.

When node migration was added to the CLI, node14.x was not supported by CFN for custom lambda handlers.

packages/amplify-e2e-tests/src/__tests__/migration/node.function.test.ts
michaelbrewer: ok changes @attilah

attilah (approved):
LGTM

PR #7702, opened 2021-07-11
feat: Override functionality enabled for Root stack

akshbhu:
#### Description of changes

- Replaced the JSON template of root stack with cdk Constructs
- Added types for Rootstack to be exposed to Cx.
- added a build Command to generated CFN stacks with overrides
- Added Root stack override with amplify push

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
+import * as cdk from '@aws-cdk/core';
+import * as s3 from '@aws-cdk/aws-s3';
+import * as iam from '@aws-cdk/aws-iam';
+import * as path from 'path';
+import { prepareApp } from '@aws-cdk/core/lib/private/prepare-app';
```
yuth, line 5: We should not be depending on `private` part of CDK. This can change and we would not have any upgrade path when it does. Also I dont think they treat changes to private classes as breaking changes and could break CLI.

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+import { AmplifyRootStackResourceProps, AmplifyRootStackTemplateProps } from './types';
+import { JSONUtilities, pathManager } from 'amplify-cli-core';
+import { Template } from 'cloudform-types';
```
yuth, line 8: Why do we need `Template` type?

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+const CFN_TEMPLATE_FORMAT_VERSION = '2010-09-09';
+const ROOT_CFN_DESCRIPTION = 'Root Stack for AWS Amplify CLI';
+
+export class AmplifyRootStack extends cdk.Stack {
+  public templateObj: AmplifyRootStackTemplateProps;
+  constructor(scope: cdk.Construct, id: string) {
+    super(scope, id);
+    this.templateOptions.templateFormatVersion = CFN_TEMPLATE_FORMAT_VERSION;
+    this.templateOptions.description = ROOT_CFN_DESCRIPTION;
+    const CfnParameters: { [k: string]: cdk.CfnParameter } = {};
```
yuth, line 19:
```suggestion
    const CfnParameters: Record<string, cdk.CfnParameter> = {};
```

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    CfnParameters.authRoleName = new cdk.CfnParameter(this, 'AuthRoleName', {
+      type: 'String',
+      default: 'AuthRoleName',
+    });
+
+    CfnParameters.unauthRoleName = new cdk.CfnParameter(this, 'UnauthRoleName', {
+      type: 'String',
+      default: 'UnauthRoleName',
+    });
```
yuth, line 35: Is it necessary to store this in an object. My understanding was that when you create a CfnParameter, it automatically gets added to the stack. Specifically if this needs to be stored in an object

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+
+    cfnOutputs.authRoleArn = new cdk.CfnOutput(this, 'AuthRoleArn', {
+      value: cdk.Fn.getAtt('AuthRole', 'Arn').toString(),
+    });
+
+    cfnOutputs.unauthRoleArn = new cdk.CfnOutput(this, 'UnauthRoleArn', {
+      value: cdk.Fn.getAtt('UnauthRole', 'Arn').toString(),
+    });
+    this.templateObj = {};
```
yuth, line 105: How will we keep changes to Parameters or Outputs in sync? Is this `templateObj` allowed to be manipulated outside the class? If so why?

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+  'INIT',
+}
+
+type RootStackOptions = {
+  rootStackFileName: string;
+  event: CommandType;
+  overrideDir?: string;
+};
+
+/**
```
yuth, line 128: Could you elaborate

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
+import * as cdk from '@aws-cdk/core';
```
yuth, line 1: Could you split this into 3 files and so we can keep one class per file. For importing, you can create and `index.ts` and re-export it there

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+      // no override required
+      //apply init modifiers if any
+    }
+
+    if (this._rootStackOptions.event === CommandType.PUSH) {
+      // apply override here during push
+      //applyoverrides(stack.templateObj);
+      const projectPath = pathManager.findProjectRoot();
+      const overridePath = pathManager.getOverrideDirPath(projectPath, 'root');
+      const { overrideProps } = await import(path.join(overridePath, 'build', 'override.js'));
```
yuth, line 238: What happens if the file is missing?

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+  };
+
+  /**
+   * Generates Root stack Template
+   * @returns CFN Template
+   */
+  private generateRootStackTemplate = async (): Promise<Template> => {
+    const stack = new AmplifyRootStack(undefined as any, 'Amplify');
+    const stackOutputs = new AmplifyRootStackOutputs(undefined as any, 'AmplifyOutputs');
```
yuth, line 225: Why do you need this? Can't you get the output from the `AmplifyRootStack`

packages/amplify-cli-core/src/state-manager/pathManager.ts
```diff
@@ -35,6 +39,7 @@ export const PathConstants = {
   LocalAWSInfoFileName: 'local-aws-info.json',
   TeamProviderInfoFileName: 'team-provider-info.json',
   BackendConfigFileName: 'backend-config.json',
+  OverrideFileName: 'override.js',
```
yuth, line 42: I thought we were supporting CDK based overrides, and making use of typsecript. Has anything changed to use `.js` file

yuth (changes requested):

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+  private _rootStackOptions: RootStackOptions; // options to help generate  cfn template
+  private _command: CommandType;
+  constructor(options: RootStackTransformOptions, command: CommandType) {
+    this._resourceConfig = options.resourceConfig;
+    this._command = command;
+  }
+
+  public async transform(): Promise<Template> {
+    // parse Input data
+    this._rootStackOptions = await this.getInput(); // get RootStackOPtions from cli-inputs.json
```
attilah, line 186: `getInput` does not seems to read cli-input.json as the comment suggests it is more like `getRootStackOptions` or `getStackOptions` ?

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+      return buildConfig;
+    }
+  };
+
+  /**
+   * Generates Root stack Template
+   * @returns CFN Template
+   */
+  private generateRootStackTemplate = async (): Promise<Template> => {
+    const stack = new AmplifyRootStack(undefined as any, 'Amplify');
```
attilah, line 224: passing in `undefined` does not seems to be ok here as breaks the CDK graph and purpose. Can't we do it like we do in transformer V2, a CDK supported way via custom CdkApp and Custom Synthesizer?

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+    const stackOutputs = new AmplifyRootStackOutputs(undefined as any, 'AmplifyOutputs');
+    // merge the outputs in one to get the combined object
+    Object.assign(stack.templateObj, stackOutputs.templateObj);
+    if (this._rootStackOptions.event === CommandType.INIT) {
+      // no override required
```
attilah, line 229: leftover? no code in this block.

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+      //apply init modifiers if any
+    }
+
+    if (this._rootStackOptions.event === CommandType.PUSH) {
+      // apply override here during push
+      //applyoverrides(stack.templateObj);
```
attilah, line 235: leftover?

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
attilah, line 238: Why not use the const values and pathManager method?

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+      const projectPath = pathManager.findProjectRoot();
+      const overridePath = pathManager.getOverrideDirPath(projectPath, 'root');
+      const { overrideProps } = await import(path.join(overridePath, 'build', 'override.js'));
+      stack.templateObj = overrideProps(stack.templateObj);
```
attilah, line 239: this will fail if the `override.js` does not have a `overrideProps` export

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+    }
+
+    const cfnRootStack: Template = stack.toCloudFormation();
+    const cfnRootStackOutputs: Template = stackOutputs.toCloudFormation();
+    Object.assign(cfnRootStack.Outputs, cfnRootStackOutputs.Outputs);
+    return cfnRootStack;
+  };
+
+  private deployOverrideStacksToDisk = async (props: DeploymentOptions) => {
+    const rootFilePath = path.join(pathManager.getBackendDirPath(), 'awscloudformation', 'build', this._resourceConfig.stackFileName);
```
attilah, line 249: the full path could come from `pathManager` or at least the constants like build and the provider name.

packages/amplify-provider-awscloudformation/src/root-stack-builder/types.ts
```diff
@@ -0,0 +1,17 @@
+import * as cdk from '@aws-cdk/core';
+import * as s3 from '@aws-cdk/aws-s3';
+import * as iam from '@aws-cdk/aws-iam';
+
+export interface AmplifyRootStackTemplateProps {
+  Parameters?: { [k: string]: cdk.CfnParameter };
+  Resources?: AmplifyRootStackResourceProps;
+  Outputs?: { [k: string]: cdk.CfnOutput };
+  Mappinngs?: { [k: string]: cdk.CfnMapping };
```
attilah, line 9: `Mappings` typo, these should be `Record` types but not really needed based on previous comments

packages/amplify-provider-awscloudformation/src/root-stack-builder/types.ts
```diff
@@ -0,0 +1,17 @@
...
+  Conditions?: { [k: string]: cdk.CfnCondition };
+}
+
+export interface AmplifyRootStackResourceProps {
+  DeploymentBucket?: s3.CfnBucket;
+  authRole?: iam.CfnRole;
+  unauthRole?: iam.CfnRole;
```
attilah, line 16: mixed case props?

packages/amplify-provider-awscloudformation/src/utility-functions.js
```diff
@@ -6,6 +6,8 @@ const { Lex } = require('./aws-utils/aws-lex');
 const Polly = require('./aws-utils/aws-polly');
```
attilah, line 6: convert this file to TS if we adding new code to it.

attilah (changes requested):
I think that beside the code related comments I'd suggest to use a CDK supported way of generating stacks as I see some anti-patterns and dependence on internals of CDK which does not give me confidence that it will not break.

Also why `ext` is not based on CDK 2.0 RCs? Is there a specific reason?

packages/amplify-cli/amplify-plugin.json
```diff
@@ -2,6 +2,7 @@
   "name": "core",
   "type": "core",
   "commands": [
+    "build-overrides",
```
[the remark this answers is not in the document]
kaustavghosh06: We're not merging this to master - but to the ext branch. We'll have an API review closer to release (which is scheduled for later this year).

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+            Principal: {
+              Federated: 'cognito-identity.amazonaws.com',
+            },
+            Action: 'sts:AssumeRoleWithWebIdentity',
+          },
+        ],
+      },
+    });
+
+    CfnResources.unauthRole = new iam.CfnRole(this, 'UnauthRole', {
```
kaustavghosh06, line 62: We also add a custom lambda resource in the root-stack when you run `amplify add auth` and push. Is that accounted for or will that happen dynamically and cannot be overriden?

packages/amplify-provider-awscloudformation/src/__tests__/root-stack-builder/root-stack-builder.test.ts
```diff
@@ -0,0 +1,9 @@
+import { SynthUtils } from '@aws-cdk/assert';
+import { Stack } from '@aws-cdk/core';
+
+import * as rootStack from '../../root-stack-builder/root-stack-builder';
+
+test('Generated rootstack template', () => {
+  const stack = new rootStack.AmplifyRootStack(null, 'rootStack');
```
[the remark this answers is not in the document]
akshbhu: will continue add more

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
akshbhu: I want a type similar to template as we are using cfn modifiers in our cli which required that. I can make my own Template type too

packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -41,20 +40,37 @@ export async function run(context) {
       stackName,
     };
     const { amplifyAppId, verifiedStackName, deploymentBucketName } = await amplifyServiceManager.init(amplifyServiceParams);
+    await configurePermissionsBoundaryForInit(context);
```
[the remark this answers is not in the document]
akshbhu: This can be above or down but  before we upload Root stack. Its only updating Permission boundaries Arn to team provider info

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
```diff
@@ -0,0 +1,252 @@
...
+    // Reources
+
+    let CfnResources: AmplifyRootStackResourceProps = {};
+    CfnResources.DeploymentBucket = new s3.CfnBucket(this, 'DeploymentBucket', {
+      bucketName: CfnParameters['deploymentBucketName'].valueAsString,
+    });
+
+    CfnResources.DeploymentBucket.applyRemovalPolicy(cdk.RemovalPolicy.RETAIN);
+
+    CfnResources.authRole = new iam.CfnRole(this, 'AuthRole', {
```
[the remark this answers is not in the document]
akshbhu: I preferred L1 as using L2 can create other resources implicitly , for existing projects it might create problems , and also we are providing types as collection of L1 constructs.

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
akshbhu: In the Root stack, we have parameters and outputs with same logicalID as ```deployment bucketName```. While constructing cdk constructs , we have have resources in stack with same logical ID. To maintain consistency with existing projects , I created another constructs having duplicate logicalIDs present and merge the both templates after synthesizing it.

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
akshbhu: I am using templateObj object to be passed to overrideFn so that they can directly modify resources or ant new parameters, mapping , outputs as per their usecase.

This template Obj is now part of RootStackTransform class not the cdk-builder class(root-stack-builder.ts)

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
akshbhu: This one ill take over when we start overriding of Auth Category

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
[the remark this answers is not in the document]
attilah: Did you look into IAM L2 construct or is it just an assumption? There are a lot of value baked into L2 constructs shipped by the CDK team.

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
yuth: We should do all the processing/modification using this class rather than passing a cfn JSON object as the CDK class can provide lot more utility like tree walking

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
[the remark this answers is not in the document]
akshbhu: I have refractored the code , templateObj is now part of transform class which wraps the cdk resources and pass it to overrides for changes. After that we directly synthesize the overrided Constrcts.

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-builder.ts
akshbhu: I am doing all the processing inside AmplifyRootTransform class which has this lifecycle

getInput() -> generateRootStackResource() -> applyoverride -> synthesizeTemplate -> deploytoDisk()

this all are called under transform() function , which returns template type which is then used in the CFNmodifiers like(IAM permission boundary and S3BucketEncryption).

packages/amplify-cli/src/commands/build-overrides.ts
```diff
@@ -0,0 +1,27 @@
+/**
+ * Command to transform CFN with overrides
+ */
+const subcommand = 'build-overrides';
+
+module.exports = {
+  name: subcommand,
+  run: async context => {
```
sachscode, line 8: Do we have a type for Context ? Can we use that here ?

packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
```diff
@@ -49,6 +49,8 @@ export async function pushResources(
     }
   }
 
+  // building all CFN stacks here to get the resource Changes
+  context.amplify.executeProviderUtils(context, 'awscloudformation', 'buildOverrides', { forceCompile: true });
```
sachscode, line 53: please use constants here for awscloudformation and buildOverrides instead of strings

packages/amplify-provider-awscloudformation/src/root-stack-builder/root-stack-transform.ts
```diff
@@ -0,0 +1,264 @@
+import { AmplifyRootStackTemplate, Template } from './types';
+import { JSONUtilities, pathManager } from 'amplify-cli-core';
+import { AmplifyRootStack, AmplifyRootStackOutputs } from './root-stack-builder';
+import { RootStackSythesizer } from './stackSynthesizer';
+import { App } from '@aws-cdk/core';
+import * as cdk from '@aws-cdk/core';
+import { buildOverrideDir } from '../utils/override-skeleton-generator';
+
+export enum CommandType {
```
sachscode, line 9: Is this the amplify command? Could you please rename this with more information (AmplifyCLICommandType or AmplifyCommandType).

sachscode (approved):
Mostly LGTM. Some minor comments on naming and using constants instead of string. Please address them in subsequent PRs (since this PR is into a branch).

PR #7710, opened 2021-07-13
bug(amplify-cli):Solved Windows compatibility issue, made location pa…

### LINKED ISSUE #7708: Windows location path compatibility in amplify-cli command tests

yarn

14.0.0

### Amplify CLI Version

5.1.0

Windows

### Amplify Categories

Not applicable

### Amplify Commands

Not applicable

### Describe the bug

Since, the location path in different operating systems is defined with varied usage of forward and back slashes,
snapshot tests need to made OS-specific in order to avoid failure of test cases

Example below of a failing snapshot test, because of difference in backslashes, even though both represent same path location:
Screenshot 2021-07-09 143925

### Expected behavior

Pass as snapshot test cases successfully
Screenshot 2021-07-13 131912

### Reproduction steps

1. cd packages/amplify-cli
2. run yarn test in Windows

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```


```

</details>

### Additional information

_No response_

studpeps:
#### Description of changes
Solved failing snapshot-test case issues in command test by making all commands to run on OS-specific location paths
image

#### Issue #, if available
#7708

1. cd packages/amplify-cli
2. run yarn test in Windows

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/__tests__/commands/upgrade.pkg.test.ts
```diff
@@ -98,13 +109,14 @@ describe('run upgrade using packaged CLI', () => {
     // validate
     expect(fs_mock.move.mock.calls[0]).toMatchInlineSnapshot(`
       Array [
-        "homedir/bin/amplify-pkg-linux",
-        "homedir/bin/amplify",
+        "${path.posix.join('homedir', 'bin', 'amplify-pkg-linux')}",
```
cjihrig, line 112: If the new serializer is working properly, we shouldn't need the `posix` version anywhere - just `path.join()` should work.

packages/amplify-cli/src/__tests__/commands/upgrade.pkg.test.ts
```diff
@@ -147,24 +159,24 @@ describe('run upgrade using packaged CLI', () => {
     await run(context_stub_typed);
 
     // validate
-    expect(fs_mock.move.mock.calls[0]).toMatchInlineSnapshot(`
+      expect(fs_mock.move.mock.calls[0]).toMatchInlineSnapshot(`
```
cjihrig, line 162: The indentation looks off here and on a number of the lines below.

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
```diff
@@ -60,6 +60,18 @@ describe('uninstall packaged CLI on mac / linux', () => {
       value: originalPlatform,
     });
   });
+
+  expect.addSnapshotSerializer({
```
cjihrig, line 64: Can we put this in a separate file so we can reuse it more easily.

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
```diff
@@ -60,6 +60,18 @@ describe('uninstall packaged CLI on mac / linux', () => {
...
+    test(val) {
```
cjihrig, line 65: My understanding is that Jest will call `test()`, and if `test()` returns a match, then it will serialize the value using `print()`. In that case, I think `test()` should check if the value is a string (like it currently is), and `process.platform === 'win32'` (since we only have this problem on Windows).

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
```diff
@@ -60,6 +60,18 @@ describe('uninstall packaged CLI on mac / linux', () => {
...
+      return typeof val === 'string';
+    },
+    print(val) {
+      if (typeof val === 'string')
```
cjihrig, line 69: You should be able to remove this `if` statement.

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
```diff
@@ -60,6 +60,18 @@ describe('uninstall packaged CLI on mac / linux', () => {
...
+        return `"${val.replace(/\\/ig, '/')}"`;
```
cjihrig, line 70: The regular expression doesn't need the `i` flag for case insensitivity since it only checks slashes.

cjihrig, line 70: We shouldn't convert the slashes to forward slashes. We should convert to the proper number Windows slashes.

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
studpeps: It says that val has "undefined" type and this replace action won't be permitted, thus the if statement
yes, it should have gotten tested in test()

packages/amplify-cli/src/__tests__/commands/upgrade.pkg.test.ts
studpeps: My bad, sure will resolve this

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
studpeps:
```    // override process.platform
    Object.defineProperty(process, 'platform', {
      value: 'linux',
    });
```
Lines 101-104 ,override the platform to be linux ,thus path.join() will return with "/" joins but as Windows path location has "\\" , it could have been resolved by turning all "\\" into "/" when process.platform is not taken into consideration.

if process.platform==="win32" comes into play then since its being override by linux, it will go under no replacement 

So the best option I think would be not evaluate based on platform and try a standard of using "/" for all path joins, as it would be not bothered by any overrides and using path.posix.join() all paths in snapshots would have "/" and any windows location "\\" would be turned to "/"

So if any overrides are performed, regardless of that it will return standard "/" join path locations

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
studpeps: Oh yes, will remove case-sensitivity ```i``` flag

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
cjihrig: That is a TypeScript issue. You can try changing the function signature to `print(val: string)` here. If that doesn't work, you can force it by using `val as string` in the function body.

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
cjihrig: Instead of checking `process.platform`, can you check `originalPlatform === 'win32'` since that is the real value of `process.platform`.

packages/amplify-cli/src/snapshot-serializer.ts
```diff
@@ -0,0 +1,12 @@
+const originalPlatform = process.platform;
+// converts inputs to toMatchInlineSnapshot() function into supported format
+export function snapshotSerializer(){
+    expect.addSnapshotSerializer({
```
cjihrig, line 4: Can you update this file to use two space indentation please.

packages/amplify-cli/src/snapshot-serializer.ts
```diff
@@ -0,0 +1,12 @@
+const originalPlatform = process.platform;
+// converts inputs to toMatchInlineSnapshot() function into supported format
+export function snapshotSerializer(){
```
cjihrig, line 3: Let's export the serializer object as `windowsPathSerializer`, and then any tests that need to use it can call `expect.addSnapshotSerializer()`. It also allows us to potentially export more serializers from this file in the future.

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
```diff
@@ -106,7 +110,7 @@ describe('uninstall packaged CLI on windows', () => {
 
     expect(fs_mock.move.mock.calls[0]).toMatchInlineSnapshot(`
       Array [
-        "homedir/.amplify/bin/amplify.exe",
+        "${path.join('homedir','.amplify','bin','amplify.exe')}",
```
cjihrig, line 113:
```suggestion
        "${path.join('homedir', '.amplify', 'bin', 'amplify.exe')}",
```

packages/amplify-cli/src/snapshot-serializer.ts
```diff
@@ -0,0 +1,12 @@
+const originalPlatform = process.platform;
```
cjihrig, line 1: This file should go in `src/__tests__`. The main `src` directory is for application code.

cjihrig (commented):
Left a couple follow up comments, but this is looking much better!

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
```diff
@@ -4,6 +4,8 @@ import execa from 'execa';
 import * as fs from 'fs-extra';
 import { hideSync } from 'hidefile';
 import { setRegPendingDelete } from '../../utils/win-utils';
+import { windowsPathSerializer } from '../../../tests/amplify-helpers/testUtils/snapshot-serializer';
+import path from 'path';
```
cjihrig, line 8:
```suggestion
import * as path from 'path';
```

packages/amplify-cli/src/__tests__/commands/uninstall.pkg.test.ts
cjihrig: It looks like this requested change is still pending.

cjihrig:
I think this needs a rebase as well to address the build failure.

studpeps:
> I think this needs a rebase as well to address the build failure.

Why has the build failed, I am not sure?

cjihrig:
Actually, it looks like the failure is related to this PR:

> @aws-amplify/cli: error TS6059: File '/home/circleci/repo/packages/amplify-cli/tests/amplify-helpers/testUtils/snapshot-serializer.ts' is not under 'rootDir' '/home/circleci/repo/packages/amplify-cli/src'. 'rootDir' is expected to contain all source files.

cjihrig (approved):

PR #7717, opened 2021-07-13
fix(amplify-provider-awscloudformation): fix always prompting for ser…

### LINKED ISSUE #6391: Can't script ServerlessContainers for CI - always promts users input
**Describe the bug**
amplify-cli/packages/amplify-cli/sample-headless-scripts/headless_configure_project.sh script not working anymore. It prompts
`Do you want to enable container-based deployments? (y/N) `

but it used to work 1 month ago.

**Amplify CLI Version**
4.41.2

**To Reproduce**
to run the script

**Expected behavior**
no promts

**Screenshots**
image

Ubuntu
**Additional context**
Attempt like this:
`REACTCONFIG="{\
\"SourceDir\":\"src\",\
\"DistributionDir\":\"out\",\
\"BuildCommand\":\"npm run-script build\",\
\"StartCommand\":\"npm run-script start\",\
\"ServerlessContainers\":false\
}"`

not working either with the same result

Bentheburrito:
#### Description of changes

Fixes the `ServerlessContainers` option passed via script being ignored. Will only prompt with "Do you want to enable container-based deployments?" on first run when the `ServerlessContainers` option is not present. (See #6391 for context)

#### Issue #, if available

fix #6391

Manually tested with example script at `packages/amplify-cli/sample-headless-scripts/headless_configure_project.sh`.
`yarn test`
`yarn lint`

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/amplify-provider-awscloudformation/src/configuration-manager.ts
```diff
@@ -94,13 +94,14 @@ export async function configure(context: $TSContext) {
...
   const { config = {} } = context.exeInfo.projectConfig[frontend] || {};
-  // TODO: check for headless mode parameter to avoid the question
-  const { ServerlessContainers } = await prompt({
-    type: 'confirm',
-    name: 'ServerlessContainers',
-    message: 'Do you want to enable container-based deployments?',
-    default: config.ServerlessContainers === true,
-  });
+  const { ServerlessContainers } =
+    config.ServerlessContainers ??
```
cjihrig, line 98: Wouldn't this prevent changing your answer after the first time it's run?

cjihrig (commented):
> Manually tested with example script at `packages/amplify-cli/sample-headless-scripts/headless_configure_project.sh`

It doesn't look like that script includes anything related to serverless containers.

Also, I think this should include a new test to prevent regressions.

Bentheburrito:
> It doesn't look like that script includes anything related to serverless containers.

Yes, that was unclear, sorry. I used that script as a template, adding `\"ServerlessContainers\":true|false\` to REACTCONFIG, and changing `amplify` to `amplify-dev`.

Good idea, I'll write up a test now.

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
Bentheburrito: Not according to my tests - If `"ServerlessContainers:value" is present in the passed frontend config, it'll be used instead of the default value.

Bentheburrito: Well, it turns out I had the de-structuring wrong anyway, hah (or maybe this is what you were pointing out?) It should be:
```javascript
const serverlessContainers =
    config.ServerlessContainers ??
    (await prompt({
      type: 'confirm',
      name: 'ServerlessContainers',
      message: 'Do you want to enable container-based deployments?',
      default: config.ServerlessContainers === true,
    })).ServerlessContainers;
```
I'll push this fix up with the unit test

cjihrig: What I meant is that if I run `amplify-dev configure project` and select the Container-based deployments option, I can select if I want to enable it or not. If I run the same command again, I no longer have the option to enable/disable it. The problem is that `config.ServerlessContainers ??` only lets the prompt run if `ServerlessContainers` is `undefined`, which is only the case on the first run.

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
```diff
@@ -91,22 +91,25 @@ export async function configure(context: $TSContext) {
   }
 }
 
-async function enableServerlessContainers(context: $TSContext) {
+export async function enableServerlessContainers(context: $TSContext) {
   const frontend = context.exeInfo.projectConfig.frontend;
   const { config = {} } = context.exeInfo.projectConfig[frontend] || {};
-  // TODO: check for headless mode parameter to avoid the question
```
cjihrig, line 97: Are you able to rely on `context.exeInfo.inputParams.yes` to determine if you're running headlessly?

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
Bentheburrito: I see now, that makes sense. Thanks for the detailed explanation

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
Bentheburrito: I did some testing, and `context.exeInfo.inputParams.yes` seems to always be `true`, even if I run just `amplify-dev configure project` without any args passed via command line. There are of course some other args missing that might hint that we're not running headlessly, like `context.exeInfo.inputParams.javascript` being undefined. But is there a better way to tell if it's running headlessly/via script?

jhockett: Hi @Bentheburrito, we only want to use the --yes flag to determine whether this question is asked. We check for `context.exeInfo.inputParams.yes` in the code frequently. Are you certain this is always set to `true`?

Bentheburrito: Hi @jhockett! Yes, `context.exeInfo.inputParams.yes` seems to be set to `true` whenever the user changes [container or profile settings](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-cli/src/config-steps/c0-analyzeProject.ts#L50-L57). Using the `--yes` flag is the only way I could determine whether the user wants to avoid the prompt - though if I've missed a better way to check this, please let me know, and I'd be happy to make the changes! :slightly_smiling_face:

cjihrig: @Bentheburrito until the code you linked to gets fixed, you can try using `context?.input?.options?.yes` for the purposes of this PR.

sdstolworthy:
Hi, @Bentheburrito,

It looks there are merge conflicts with the target branch. Because there has not been recent activity on this pull request, we are closing it. Please review the contribution guidelines. Once these changes meet those requirements and the merge conflicts are resolved, feel free to re-open the pull request and tag @aws-amplify/amplify-cli, and we will take another look.

PR #7730, opened 2021-07-15
fix(cli): prevent re-prompt of info on `amplify pull`

johnpc:
#### Description of changes

 When the developer runs `amplify pull` and chooses "no" when prompted: "Do you plan on modifying this backend?", the cli should not prompt for all the project configuration settings when `amplify pull` is called again to receive updates. Since the entire `amplify/` dir is removed when this option is selected, we lose project configuration context.

This change keeps the `amplify/` dir around, only deleting the backend directories, and maintains the state of the project configuration.

#### Issue #, if available

N/A

Ran `amplify pull` multiple times for admin ui, profile, and access key authenticated apps

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/attach-backend.ts
+++ b/packages/amplify-cli/src/attach-backend.ts
@@ -30,7 +30,7 @@ export async function attachBackend(context: $TSContext, inputParams) {
     await generateFiles(context);
     await onSuccess(context);
   } catch (e) {
-    removeFolderStructure();
+    removeAmplifyFolderStructure();
     restoreOriginalAmplifyFolder();
 
     context.print.error('Failed to pull the backend.');
@@ -59,11 +59,11 @@ async function onSuccess(context: $TSContext) {
   await postPullCodegen(context);
 
   if (!inputParams.yes) {
-    const confirmKeepCodebase = context.exeInfo.existingLocalEnvInfo?.noUpdateBackend
+    const shouldKeepAmplifyDir = context.exeInfo.existingLocalEnvInfo?.noUpdateBackend
       ? !context.exeInfo.existingLocalEnvInfo.noUpdateBackend
       : await context.amplify.confirmPrompt('Do you plan on modifying this backend?', true);
 
-    if (confirmKeepCodebase) {
+    if (shouldKeepAmplifyDir) {
       if (stateManager.currentMetaFileExists()) {
         await initializeEnv(context, stateManager.getCurrentMeta());
       }
@@ -76,7 +76,7 @@ async function onSuccess(context: $TSContext) {
       context.print.info('');
     } else {
       stateManager.setLocalEnvInfo(process.cwd(), { ...context.exeInfo.localEnvInfo, noUpdateBackend: true });
-      removeFolderStructure(true);
+      removeAmplifyFolderStructure(true);
 
       context.print.info('');
       context.print.success(`Added backend environment config object to your project.`);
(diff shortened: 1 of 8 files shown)


packages/amplify-cli/src/attach-backend.ts
```diff
@@ -59,7 +59,9 @@ async function onSuccess(context: $TSContext) {
   await postPullCodegen(context);
 
   if (!inputParams.yes) {
-    const confirmKeepCodebase = await context.amplify.confirmPrompt('Do you plan on modifying this backend?', true);
+    const confirmKeepCodebase = context.exeInfo.existingLocalEnvInfo?.noUpdateBackend
```
sachscode, line 62: Can this logic be moved into a function, like checkBackendUpdated() or something to that effect?

packages/amplify-cli/src/attach-backend.ts
```diff
@@ -73,7 +75,8 @@ async function onSuccess(context: $TSContext) {
       context.print.info(`Run 'amplify pull' to sync future upstream changes.`);
       context.print.info('');
     } else {
-      removeFolderStructure();
+      stateManager.setLocalEnvInfo(process.cwd(), { ...context.exeInfo.localEnvInfo, noUpdateBackend: true });
+      removeFolderStructure(true);
```
sachscode, line 79: Can you name this more explicitly. remove{which}FolderStructure.

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
```diff
@@ -48,7 +48,7 @@ const defaultAWSConfig: AwsConfig = {
 };
 
 export async function init(context: $TSContext) {
-  if (!context.exeInfo.isNewProject && doesAwsConfigExists(context)) {
+  if (context.exeInfo.existingLocalEnvInfo?.noUpdateBackend || (!context.exeInfo.isNewProject && doesAwsConfigExists(context))) {
     return context;
```
sachscode, line 52: Why is normalization not required here?

packages/amplify-cli/src/attach-backend.ts
```diff
@@ -73,7 +75,8 @@ async function onSuccess(context: $TSContext) {
       context.print.info(`Run 'amplify pull' to sync future upstream changes.`);
       context.print.info('');
     } else {
-      removeFolderStructure();
+      stateManager.setLocalEnvInfo(process.cwd(), { ...context.exeInfo.localEnvInfo, noUpdateBackend: true });
```
sachscode, line 78: Here upUpdateBackend is set to true, should it instead use the result in confirmKeepCodeBase ?

sachscode (changes requested):
Changes look great, just a few minor questions.

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
johnpc: what do you mean by "normalization"?

packages/amplify-cli/src/attach-backend.ts
johnpc: Since this is in the `else` statement, `confirmKeepCodebase` will always be false. I could change the `true` to `!confirmKeepCodebase` but in my opinion that is unnecessarily obfuscating that we want to set `noUpdateBackend` to `true`.

sachscode (approved):
LGTM! 👍

packages/amplify-cli/src/commands/push.ts
```diff
@@ -41,6 +41,9 @@ async function syncCurrentCloudBackend(context: $TSContext) {
 export const run = async (context: $TSContext) => {
   try {
     context.amplify.constructExeInfo(context);
+    if (context.exeInfo.localEnvInfo.noUpdateBackend) {
+      throw new ConfigurationError('This local environment was configured to not publish backend updates.');
```
jhockett, line 45: How about "The local environment configuration does not allow backend updates."?

cc @renebrandel

jhockett (dismissed):
This change is significant enough to warrant an e2e test

PR #7761, opened 2021-07-20
fix: add auth config when hashing resource

SwaySway:
#### Description of changes

- include auth config when doing a glob hash of the api resource

#### Issue #, if available

- fix https://github.com/aws-amplify/amplify-cli/issues/4914

#### Checklist

- [x] PR description included
- [] `yarn test` passes


packages/amplify-cli/src/extensions/amplify-helpers/resource-status.ts
```diff
@@ -7,17 +7,29 @@ import { hashElement, HashElementOptions } from 'folder-hash';
 import { getEnvInfo } from './get-env-info';
 import { CLOUD_INITIALIZED, CLOUD_NOT_INITIALIZED, getCloudInitStatus } from './get-cloud-init-status';
 import { ServiceName as FunctionServiceName, hashLayerResource } from 'amplify-category-function';
+import { hashGQLResource } from 'amplify-provider-awscloudformation';
 import { removeGetUserEndpoints } from '../amplify-helpers/remove-pinpoint-policy';
 import { pathManager, stateManager, $TSMeta, $TSAny, NotInitializedError } from 'amplify-cli-core';
 
-async function isBackendDirModifiedSinceLastPush(resourceName, category, lastPushTimeStamp, hashFunction) {
+async function isBackendDirModifiedSinceLastPush(resourceName: string, categoryName: string, meta: $TSMeta) {
```
attilah, line 14: this logic was moved out for a reason in the layer's PR as it is not a dependent functions's role to have logic to decide about what hashing function to use, burying this logic inside will lead to hard to maintain+understand code, please move it out.

packages/amplify-provider-awscloudformation/src/upload-appsync-files.ts
```diff
@@ -1,16 +1,16 @@
-const fs = require('fs');
-const fsext = require('fs-extra');
-const path = require('path');
-
-const TransformPackage = require('graphql-transformer-core');
-const { S3 } = require('./aws-utils/aws-s3');
-const { fileLogger } = require('./utils/aws-logger');
+import fs from 'fs-extra';
```
attilah, line 1: update imports `* as path`, etc...

packages/amplify-provider-awscloudformation/src/upload-appsync-files.ts
```diff
@@ -1,16 +1,16 @@
...
+import path from 'path';
+import ora from 'ora';
+import crypto from 'crypto';
+import globby from 'globby';
+import { stateManager, JSONUtilities } from 'amplify-cli-core';
+import { uploadAPIProject } from 'graphql-transformer-core';
+import { S3 } from './aws-utils/aws-s3';
+import { fileLogger } from './utils/aws-logger';
+import { ProviderName as providerName } from './constants';
```
attilah, line 10: Don't alias the import as it adds to confusion when reading code, perhaps it is a sideeffect of refactor?

packages/amplify-provider-awscloudformation/src/upload-appsync-files.ts
```diff
@@ -28,7 +28,7 @@ function getProjectBucket(context) {
  * @param {*} allResources
  * @param {*} options
  */
-async function uploadAppSyncFiles(context, resourcesToUpdate, allResources, options = {}) {
+export async function uploadAppSyncFiles(context, resourcesToUpdate, allResources, options: any = {}) {
```
attilah, line 31: can we type it?

packages/amplify-provider-awscloudformation/src/upload-appsync-files.ts
```diff
@@ -189,9 +189,9 @@ async function uploadAppSyncFiles(context, resourcesToUpdate, allResources, opti
     if (!fs.existsSync(resourceBuildDir)) {
       return;
     }
-    const spinner = new ora('Uploading files...');
+    const spinner = ora('Uploading files...');
```
attilah, line 192: spinner needs to be in `try...finally` to make sure it will be destroyed and not causing a hang for the process

packages/amplify-provider-awscloudformation/src/upload-appsync-files.ts
```diff
@@ -227,21 +227,17 @@ async function uploadAppSyncFiles(context, resourcesToUpdate, allResources, opti
...
-    folders: {
-      exclude: ['build'],
-    },
-  };
-
-  const hashResult = await hashElement(directory, options);
-
-  return hashResult.hash;
+export async function hashGQLResource(directory: string, resourceName: string): Promise<string> {
+  const filePaths: string[] = await globby([path.join('**', '*')], {
```
attilah, line 231: glob pattern does not need to use path.join as it is always forward slashes under every OS.

naming: `directory` is `resourceDirectory` ?

attilah (changes requested):
Added a few comments. As a generic question about the changes, what happens with new CLI and unchanged APIs? will it signal an update to an unchanged API because of the hashing changes, right? are we ok with it?

PR #7764, opened 2021-07-20
Category graphql/searchable aggregate functions

sundersc:
#### Description of changes
Allows customer to use aggregate function on Elastic Search @searchable queries.
Supported aggregate functions include 
- terms
- min
- max
- avg
- sum

#### Issue #, if available
[Link to Searchable RFC document](https://github.com/aws-amplify/amplify-cli/issues/7546)

- Manual Test
- yarn test passes

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-searchable-transformer/src/generate-resolver-vtl.ts
```diff
@@ -83,15 +96,43 @@ export function responseTemplate(includeVersion = false) {
...
+          ]),
+        ),
+        qref('$aggResult.put("result", $aggResultValue)'),
+        qref('$aggregateValues.add($aggResult)'),
+      ]),
       toJson(
         obj({
           items: ref('es_items'),
-          total: ref('ctx.result.hits.total'),
+          total: ref('ctx.result.hits.total.value'),
```
SwaySway, line 133: Should it only be `ctx.result.hits.total.value` when using an aggregate search?

sundersc: No, it is not because of aggregate search. It looks like it changed in the latest open search update.

packages/amplify-graphql-searchable-transformer/src/definitions.ts
```diff
@@ -226,6 +226,58 @@ export function makeSearchableXSortInputObject(obj: ObjectTypeDefinitionNode): I
   };
 }
 
+export function makeSearchableAggregateTypeEnumObject(): EnumTypeDefinitionNode {
+  const name = graphqlName(`SearchableAggregateType`);
```
cjihrig, line 230:
```suggestion
  const name = graphqlName('SearchableAggregateType');
```

packages/amplify-graphql-searchable-transformer/src/__tests__/__snapshots__/amplify-graphql-searchable-transformer.tests.ts.snap
```diff
@@ -101,14 +101,49 @@ input SearchableEmployeeSortInput {
...
 }
 
+input SearchableEmployeeAggregationInput {
+  name: String!
+  type: SearchableAggregateType!
+  field: String!
+}
+
+enum SearchableAggregateType {
+  terms
```
cjihrig, line 111: Do we generate other enums with lowercase entries? I think enums are generally all capital letters, so we should do that here unless this was done for consistency. The RFC uses all caps entries.

packages/amplify-graphql-searchable-transformer/src/__tests__/__snapshots__/amplify-graphql-searchable-transformer.tests.ts.snap
```diff
@@ -357,14 +392,49 @@ input SearchablePostSortInput {
...
+  value: Float!
+}
+
+type SearchableAggregateBucketResult {
+  buckets: [SearchableAggregateBucketResultItem]
+}
+
+type SearchableAggregateBucketResultItem {
+  key: String!
+  doc_count: Int!
```
cjihrig, line 431: The snake case here looks inconsistent with the rest of the API. The RFC uses `count`.

cjihrig (commented):
Left a few comments, but this is looking good! Since this is new functionality, it would be good to include an E2E test or two for the queries.

packages/amplify-graphql-searchable-transformer/src/__tests__/__snapshots__/amplify-graphql-searchable-transformer.tests.ts.snap
sundersc: Yes, we use lower case in other places too. Example: SearchableSortDirection.
Also, OpenSearch expects this in lower case.

packages/amplify-graphql-searchable-transformer/src/__tests__/__snapshots__/amplify-graphql-searchable-transformer.tests.ts.snap
sundersc: Open search returns this field as doc_count. I think it is better to leave these properties in the same names as returned by Opensearch so it is easy to extend for other aggregate functions in the future.

packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformerV2.e2e.test.ts
```diff
@@ -0,0 +1,251 @@
...
+    const apiKey = getApiKey(finishedStack.Outputs);
+    expect(apiKey).toBeDefined();
+    expect(endpoint).toBeDefined();
+    GRAPHQL_CLIENT = new GraphQLClient(endpoint, { 'x-api-key': apiKey });
+
+    // Create sample mutations to test search queries
+    await createEntries();
+  } catch (e) {
+    console.error(e);
+    expect(true).toEqual(false);
```
cjihrig, line 124: I realize that this was copied from other tests, but I think we should just throw the original error.

packages/amplify-graphql-searchable-transformer/src/__tests__/__snapshots__/amplify-graphql-searchable-transformer.tests.ts.snap
cjihrig: How much effort would it be to perform the conversion?

cc: @renebrandel

cjihrig (approved):
LGTM with a small nit in the test and a question regarding the use of snake case.

packages/amplify-graphql-searchable-transformer/src/__tests__/__snapshots__/amplify-graphql-searchable-transformer.tests.ts.snap
sundersc: It would probably be 1-day effort. Importantly when we plan to support more OpenSearch aggregation methods in the future, we need to do conversion for different response data structures.

packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformerV2.e2e.test.ts
sundersc: Agree, I will update.

PR #7765, opened 2021-07-21
fix(graphql-auth-transformer): extend isOptional subscription owner check

lazpavel:
#### Description of changes
Extended the implementation for the logic that determines if the subscription owner argument is required to be not null and extracted the logic in its own method.

Manual testing with various combinations of owner type rules.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
```diff
@@ -1944,6 +1941,20 @@ operations will be generated by the CLI.`,
...
+  private isSubscriptionOwnerArgumentOptional(rules: AuthRule[]) {
+    const ownerRules = rules.filter(rule => rule.allow === OWNER_AUTH_STRATEGY);
+
+    // if there are multiple owner rules then the owner argument is optional
+    if (ownerRules.length > 1) {
+      return true;
+    }
+
+    // if there are any public, private, and/or static group rules then the owner argument is optional
+    return rules.find(
```
johnpc, line 1953: nit: maybe this function should return a boolean every time, so that it can return the value you want instead of having to do `this.isSubscriptionOwnerArgumentOptional(rules) ? false : true`?

johnpc (approved):
lgtm. you can address the nit or not your call :P

packages/graphql-auth-transformer/src/ModelAuthTransformer.ts
lazpavel: updated to Boolean, makes it cleaner, thanks!

PR #7779, opened 2021-07-22
feat(amplify-category-auth): use EnabledMFAs to only configure TOTP

ammarkarachi:
#### Description of changes
Added a way to only configure only TOTP. Configuring the user pool with any other attribute with 'phone_number' turn on SMS MFA

`yarn e2e src/__tests__/auth_*`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-auth/resources/cloudformation-templates/auth-template.yml.ejs
+++ b/packages/amplify-category-auth/resources/cloudformation-templates/auth-template.yml.ejs
@@ -182,11 +182,27 @@ Resources:
       AliasAttributes: !Ref aliasAttributes
       <% } %>
       MfaConfiguration: !Ref mfaConfiguration
+      <% if(props.useEnabledMfas && props.mfaConfiguration != 'OFF') {%>
+      EnabledMfas:
+        <% if(props.mfaTypes.includes('SMS Text Message') ||
+        props.autoVerifiedAttributes.includes('phone_number') ||
+        props.requiredAttributes.includes('phone_number')) {%> 
+        - SMS_MFA
+        <% } %>
+        <% if(props.mfaTypes.includes('TOTP')) {%>
+        - SOFTWARE_TOKEN_MFA
+        <% } %>
+      <% } %>  
+      <% if(!props.useEnabledMfas || 
+      (props.mfaConfiguration != 'OFF' && props.mfaTypes.includes('SMS Text Message')) || 
+      props.autoVerifiedAttributes.includes('phone_number') ||
+      props.requiredAttributes.includes('phone_number')) {%> 
       SmsVerificationMessage: !Ref smsVerificationMessage
       SmsAuthenticationMessage: !Ref smsAuthenticationMessage
       SmsConfiguration:
         SnsCallerArn: !GetAtt SNSRole.Arn
         ExternalId: <%=`${props.resourceNameTruncated}_role_external_id`%>
+      <% } %>
     <%if (props.mfaConfiguration != 'OFF') { %>
     DependsOn: SNSRole
     <% } %>
@@ -818,7 +834,7 @@ Resources:
     DependsOn: OAuthCustomResourceLogPolicy
   <% } %>
 
-  <%if (props.mfaConfiguration != 'OFF') { %>
+  <%if (!props.useEnabledMfas && props.mfaConfiguration != 'OFF') { %>
   # BEGIN MFA LAMBDA RESOURCES
   MFALambdaRole:
   # Created to execute Lambda which sets MFA config values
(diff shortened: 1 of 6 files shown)


packages/amplify-category-auth/resources/cloudformation-templates/auth-template.yml.ejs
```diff
@@ -182,12 +185,23 @@ Resources:
...
+      <% if(props.useEnabledMfas && props.mfaConfiguration != 'OFF') {%>
+      EnabledMfas:
+        <% if(configureSMS) {%> 
+        - SMS_MFA
+        <% } %>
+        <% if(props.mfaTypes.includes('TOTP')) {%>
+        - SOFTWARE_TOKEN_MFA
+        <% } %>
+      <% } %>  
+      <% if(!props.useEnabledMfas || configureSMS) {%>
```
SwaySway, line 197: Are we not using these messages when mfa is false and configureSMS is true?

ammarkarachi: The feature flag is to retain the current state of existing projects. which always keeps SMS configured

ammarkarachi: For new project we only configure the SMS when they start using phone_numbers as a required attribute or an auto verified attribute

packages/amplify-category-auth/resources/cloudformation-templates/auth-template.yml.ejs
```diff
@@ -1,4 +1,5 @@
 <% var autoVerifiedAttributes = props.autoVerifiedAttributes ? props.autoVerifiedAttributes.concat(props.aliasAttributes).filter((attr, i, aliasAttributeArray) => ['email', 'phone_number'].includes(attr) && aliasAttributeArray.indexOf(attr) === i) : [] %>
+<% var configureSMS = ((props.autoVerifiedAttributes && props.autoVerifiedAttributes.includes('phone_number')) || (props.mfaConfiguration != 'OFF' && props.mfaTypes && props.mfaTypes.includes('SMS Text Message')) || (props.requiredAttributes && props.requiredAttributes.includes('phone_number'))) %>
```
sachscode, line 2: This seems to be a complex check. could you please provide this as a prop.

packages/amplify-category-auth/resources/cloudformation-templates/auth-template.yml.ejs
```diff
@@ -818,7 +832,7 @@ Resources:
     DependsOn: OAuthCustomResourceLogPolicy
   <% } %>
 
-  <%if (props.mfaConfiguration != 'OFF') { %>
+  <%if (!props.useEnabledMfas && props.mfaConfiguration != 'OFF') { %>
```
sachscode, line 835: is it simple to convert double negation into isMFAConfigurationOn ?

packages/amplify-e2e-core/src/utils/sdk-calls.ts
```diff
@@ -116,6 +116,11 @@ export const getUserPool = async (userpoolId, region) => {
   return res;
 };
 
+export const getMFAConfiguration = async (userPoolId, region): Promise<CognitoIdentityServiceProvider.GetUserPoolMfaConfigResponse> => {
```
sachscode, line 119: please add types for userPoolId and Region.

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthrough-types.ts
```diff
@@ -27,6 +27,7 @@ export interface ServiceQuestionsBaseResult {
   userpoolClientReadAttributes: string[];
   userpoolClientWriteAttributes: string[];
   usernameCaseSensitive?: boolean;
+  useEnabledMfas?: boolean;
```
sachscode, line 30: Why is this an optional attribute ?  can it not be required, but set to false by default ?

sachscode (changes requested):
Changes look Good! Some minor changes requested to reduce complexity and improve readablity/debuggability

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthrough-types.ts
ammarkarachi: It's because the update flow might not necessarily have that option, since the value is read from FF

packages/amplify-category-auth/resources/cloudformation-templates/auth-template.yml.ejs
ammarkarachi: The MFAConfiguration can either be 'OPTIONAL or ON' that's why the !=

packages/amplify-category-auth/resources/cloudformation-templates/auth-template.yml.ejs
johnpc: any time complex logic starts developing in an .ejs file, I wonder if it has to be in here, or if it can be moved up a layer. I looked into it a little bit and it felt it like based on the way layers of above are extracted it feels easier to do it in here, even though it's not ideal.

ammarkarachi: I think there is an issue with providing it as a prop

sachscode: yeah I agree, looks like we can pass it in now. We could resolve this when moving away from ejs.

packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/utils/auth-defaults-appliers.test.ts
```diff
@@ -42,6 +42,7 @@ describe('update auth defaults applier', () => {
     } as ServiceQuestionsResult;
 
     const result = await getUpdateAuthDefaultsApplier({}, 'cognito-defaults.js', {} as ServiceQuestionsResult)(stubResult);
+    console.log(result);
```
sachscode, line 45: Do we need this here ?

ammarkarachi: Oh no I don't I was trying to debug the test

johnpc (approved):
lgtm!

sachscode (approved):

PR #7804, opened 2021-07-26
feat(amplify-util-mock): support pseudo parameters in environment variables when running `amplify mock function`

### LINKED ISSUE #6122: Add functionality to correctly resolve AWS:NoValue when using Amplify mock
I see that a lot of people have problems with environment variables for lambda functions when using amplify mock api.
Since mock doesn't support using for example **resolve:ssm** it translates it into the to the litteral value which works incorrectly when working locally. I tried using the **AWS:NoValue** instead to remove all env variables from the cloudformation template when using mock but that also translates in to en env variable with the value "AWS::NoValue"

I would like mock to be able to handle the pseudo parameter **AWS:NoValue** to correctly remove a property instead of translating it to a string value. This would enable me to declare my environment variables locally which is something I use when developing other applications and lambda functions.

My current solution is to read what ENV is running and use that to select "ENV_VAR_LOCAL" instead of "ENV_VAR" for all variables that are required to run the function locally. Solutions proposed in other issues either involve a separate env file or hardcoding values during development and reverting before pushing. Neither of those solutions work for me since they involve risking to committing sensitive information when pushing to GIT.

**Additional context**
[AWS - Pseudo parameter reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html)

Bentheburrito:
#### Description of changes

Environment variables can have now pseudo parameter values (i.e. AWS::NoValue) that will be substituted appropriately when running `amplify mock function`.

#### Issue #, if available

close #6122

- Manually tested with `amplify-dev`
  - Created lambda function in test app
  - Added environment variables with pseudo parameters like `AWS::NoValue` and `AWS::Region`
  - Configure lambda function to return the values of these environment variables
  - run `amplify-dev mock function`
- `yarn test`
- `yarn lint`

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/amplify-util-mock/src/CFNParser/intrinsic-functions.ts
+++ b/packages/amplify-util-mock/src/CFNParser/intrinsic-functions.ts
@@ -99,8 +99,6 @@ export function cfnRef(valNode, { params, resources }: CloudFormationParseContex
   }
 
   if (params.hasOwnProperty(key)) {
-    // Substitute any pseudo parameters with their mock value
-    if (params.hasOwnProperty(params[key])) params[key] = params[params[key]];
     return params[key];
   }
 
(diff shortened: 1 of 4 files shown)


cjihrig (approved):
The code changes LGTM. For some reason the build_test_deploy CI job didn't run.

PR #7817, opened 2021-07-28
fix: use improved pluralization in graphql transformer v2

cjihrig:
#### Description of changes
This commit updates GraphQL Transformer v2 to use the newly improved pluralization rules by default.

Manual and unit testing.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/utils.ts
+++ b/packages/amplify-graphql-index-transformer/src/utils.ts
@@ -1,4 +1,5 @@
 import { TransformerContextProvider } from '@aws-amplify/graphql-transformer-interfaces';
+import { plurality } from 'graphql-transformer-common';
 import { PrimaryKeyDirectiveConfiguration } from './types';
 
 export function lookupResolverName(config: PrimaryKeyDirectiveConfiguration, ctx: TransformerContextProvider, op: string): string | null {
@@ -26,7 +27,11 @@ export function lookupResolverName(config: PrimaryKeyDirectiveConfiguration, ctx
   }
 
   if (!resolverName) {
-    resolverName = `${op}${object.name.value}${op === 'list' ? 's' : ''}`;
+    if (op === 'list') {
+      resolverName = `${op}${plurality(object.name.value, true)}`;
+    } else {
+      resolverName = `${op}${object.name.value}`;
+    }
   }
 
   return resolverName;
--- a/packages/amplify-graphql-searchable-transformer/src/graphql-searchable-transformer.ts
+++ b/packages/amplify-graphql-searchable-transformer/src/graphql-searchable-transformer.ts
@@ -160,7 +160,7 @@ export class SearchableModelTransformer extends TransformerPluginBase {
     }
     const fieldName = searchFieldNameOverride
       ? searchFieldNameOverride
-      : graphqlName(`search${plurality(toUpper(definition.name.value), ctx.featureFlags.getBoolean('improvePluralization', true))}`);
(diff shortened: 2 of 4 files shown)


lazpavel (approved):

packages/amplify-graphql-index-transformer/src/utils.ts
```diff
@@ -26,7 +27,11 @@ export function lookupResolverName(config: PrimaryKeyDirectiveConfiguration, ctx
   }
 
   if (!resolverName) {
-    resolverName = `${op}${object.name.value}${op === 'list' ? 's' : ''}`;
+    if (op === 'list') {
+      resolverName = `${op}${plurality(object.name.value, true)}`;
+    } else {
+      resolverName = `${op}${object.name.value}`;
+    }
   }
```
lawmicha, line 35: This looks like a breaking change in the way, more details in 
before: 
- `type Wish` generates `listWishs` 
- `type Wishes` generates `listWishess`

after:
- `type Wish` generates `listWishes` 
- `type Wishes` generates `listWishes`

cjihrig:
@lawmicha all of the changes in this PR apply to v2 of the transformer which is still experimental and behind a feature flag.

packages/amplify-graphql-index-transformer/src/utils.ts
lawmicha: what is the current logic for generating the resolverName for conflict resolution enabled APIs for  sync queries?

packages/amplify-graphql-searchable-transformer/src/graphql-searchable-transformer.ts
```diff
@@ -160,7 +160,7 @@ export class SearchableModelTransformer extends TransformerPluginBase {
     }
     const fieldName = searchFieldNameOverride
       ? searchFieldNameOverride
-      : graphqlName(`search${plurality(toUpper(definition.name.value), ctx.featureFlags.getBoolean('improvePluralization', true))}`);
+      : graphqlName(`search${plurality(toUpper(definition.name.value), true)}`);
```
lawmicha, line 163: what is the point of the feature flag if once true, is not compatible with the iOS library. Let's say we saw this compatibility issue with the library

- Is there a larger issue here that the Amplify iOS library depends on how the APIs that are provisioned? There are many examples of this, such as DataStore uses a query called `"sync" + pluralName`, what happens when this is changed under a feature flag, and then eventually turned to true? All new provisionings won't work with the library anymore. Can the library introspect the GraphQL API at runtime instead of hardcoding specific names?

- Do we need the same feature flag in the library that corresponds to the CLI's feature flag so that developers can upgrade both at the same time?

lawmicha:
> @lawmicha all of the changes in this PR apply to v2 of the transformer which is still experimental and behind a feature flag.

Oh, that makes more sense, was the feature flag accidentally set to `true` then? see line 163 in this PR

cjihrig:
> Oh, that makes more sense, was the feature flag accidentally set to true then? see line 163 in this PR

That code bypasses the pluralization feature flag intentionally, but customers are not running v2 of the GraphQL transformer unless they opt in to a separate feature flag.

Have you ruled out https://github.com/aws-amplify/amplify-cli/commit/fc3ad0dd5a12a7912c59ae12024f593b4cdf7f2d and https://github.com/aws-amplify/amplify-cli/commit/1b0154060a416c11f187047e87219180fb6d06f0, which both landed before this PR and touch the transformer v1 code?

lawmicha:
Thanks, i'll take a look at those commits!

Issue #7842, opened 2021-07-30
amplify-dynamodb-simulator : Async callback was not invoked within the timeout 


studpeps:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

yarn

### If applicable, what version of Node.js are you using?

14.0.0

### Amplify CLI Version

5.1.0

### What operating system are you using?

Windows

### Amplify Categories

Not applicable

### Amplify Commands

Not applicable

### Describe the bug

Async function callback not reached within time-limit, causing test cases to fail
![Screenshot 2021-07-30 183348](https://user-images.githubusercontent.com/43947328/127658304-288914c3-1336-4ea1-81fa-bf726879d6cf.png)
![Screenshot 2021-07-30 183603](https://user-images.githubusercontent.com/43947328/127658323-3560b534-962c-4e5c-b8ae-bef60d902ce9.png)

### Expected behavior

![Screenshot 2021-07-30 182641](https://user-images.githubusercontent.com/43947328/127659571-d8f5edd3-d691-4600-8cb8-caed974521ff.png)
Callback revoked within time-limit

### Reproduction steps

1. cd packages/amplify-dynamodb-simulator
2. yarn test

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

PR #7843, opened 2021-07-30
fix(amplify-dynamodb-simulator):callback revoked within time-limit

studpeps:
#### Description of changes
Revoked async function within time-limit by calling the removeSync() function only for test cases requiring the same.
Since, one of the test case did not require sync to be broken right after as another test case was to run on previous test cases result,thus removeSync() was not to be called.

#### Issue #, if available
#7842 

1. cd packages/amplify-dynamodb-simulator
2. yarn test

Screenshot 2021-07-30 182641

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-dynamodb-simulator/__test__/index.test.js
+++ b/packages/amplify-dynamodb-simulator/__test__/index.test.js
@@ -7,9 +7,9 @@ jest.mock('amplify-cli-core', () => ({
   },
 }));
 
-describe('emulator operations', () => {
-  const dbPath = `${__dirname}/dynamodb-data/${process.pid}`;
-  // taken from dynamodb examples.
+let emulators;
+const dbPath = `${__dirname}/dynamodb-data/${process.pid}`;
+// taken from dynamodb examples.
   const dbParams = {
     AttributeDefinitions: [
       {
@@ -37,6 +37,7 @@ describe('emulator operations', () => {
     },
   };
 
+describe('emulator operations', () => {
   const ensureNoDbPath = () => {
     if (fs.existsSync(dbPath)) {
       fs.removeSync(dbPath);
@@ -46,7 +47,6 @@ describe('emulator operations', () => {
   beforeEach(ensureNoDbPath);
   afterEach(ensureNoDbPath);
 
-  let emulators;
   beforeEach(() => {
     emulators = [];
     jest.setTimeout(40 * 1000);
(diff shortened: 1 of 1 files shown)


cjihrig:
It's not clear to me how this fixes the failures in #7842.

- For the test that times out, does increasing the timeout fix the problem?
- For the test failing with `EBUSY` - does this test fail 100% of the time? It seems like something else has the file open and is preventing it from being deleted.

studpeps:
> It's not clear to me how this fixes the failures in #7842.
> 
> * For the test that times out, does increasing the timeout fix the problem?
> * For the test failing with `EBUSY` - does this test fail 100% of the time? It seems like something else has the file open and is preventing it from being deleted.

- No, increasing the time out doesn't fix the problem, it tests on different database samples and always fails on some database testing even when increased.
- Yes, it fails every time, did try restarting servers and operating system, but had no effect, don't know what it means by `EBUSY` here

  ```afterEach(ensureNoDbPath)```  throws an error only with the test case ``` 'should preserve state between restarts with dbPath'```

cjihrig:
> `afterEach(ensureNoDbPath)` throws an error only with the test case `'should preserve state between restarts with dbPath'`

In that test, does changing `emuTwo.terminate();` to `await emuTwo.terminate();` make any difference?

studpeps:
No, adding await doesn't have any effect, did try this out

cjihrig:
The `afterEach()` that performs the cleanup doesn't run for that test though, right (because the test is not in the same `describe()` block)?

I think the underlying issue is that the emulators aren't being cleaned up properly. On Windows, a file can't be deleted if another process has the file opened.

Another thing to try - can you combine the two `afterEach()` calls into a single `afterEach()` and make sure that the emulators are terminated before trying to remove the file. It's possible that Jest is currently running the `afterEach()` blocks in the order that they are defined in the code, which would mean trying to delete the file before terminating the emulator.

studpeps:
Yes, afterEach() was not being called for that test (in previous commit made)
Tried out a new way by combining both beforeEach() and afterEach() into one

packages/amplify-dynamodb-simulator/__test__/index.test.js
```diff
@@ -43,15 +43,16 @@ describe('emulator operations', () => {
...
   let emulators;
   beforeEach(() => {
+    ensureNoDbPath();
     emulators = [];
     jest.setTimeout(40 * 1000);
   });
-  afterEach(async () => await Promise.all(emulators.map(emu => emu.terminate())));
+
+  afterEach(async () => {
+    await Promise.all(emulators.map(emu => emu.terminate()));
```
cjihrig, line 54: I think we still want `ensureNoDbPath()` to be called after the emulators are terminated.

cjihrig (commented):
One more nit, and then I think this should be good to go.

packages/amplify-dynamodb-simulator/__test__/index.test.js
studpeps: For some reason, adding ensureNoDbPath() at the end causes an error:
` EBUSY: resource busy or locked, unlink 'C:\Users\Stuti\Desktop\amplify-cli\packages\amplify-dynamodb-simulator\__test__\dynamodb-data\2080\fake_us-fake-1.db`

cjihrig (approved):

PR #7879, opened 2021-08-04
fix(graphql-model-transformer): fixed input type field generation for enum types

lazpavel:
#### Description of changes
Fixed the input field type for enum types

#### Description of how you validated changes
- manual debug
- unit test added
- `yarn test` passes

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #7887, opened 2021-08-04
feat: add @index directive

cjihrig:
#### Description of changes
This PR adds a new `@index` directive to the GraphQL Transformer v2. It also ports the `@key` E2E tests to v2. It also includes the commit from #7879 which is a bug fix needed to get the test suite passing.

Manual, unit, and E2E testing.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/resolvers.ts
+++ b/packages/amplify-graphql-index-transformer/src/resolvers.ts
@@ -305,7 +305,6 @@ export function appendSecondaryIndex(config: IndexDirectiveConfiguration, ctx: T
   const keySchema = getDdbKeySchema(config);
   const attrDefs = attributeDefinitions(config, ctx);
   const primaryKeyPartitionKeyName = primaryKeyField?.name?.value ?? 'id';
-  const primaryKeyPartitionKeyType = attrDefs.find(attr => attr.attributeName === primaryKeyPartitionKeyName)?.attributeType ?? 'S';
   const partitionKeyName = keySchema[0]?.attributeName;
   const sortKeyName = keySchema?.[1]?.attributeName;
   const partitionKeyType = attrDefs.find(attr => attr.attributeName === partitionKeyName)?.attributeType ?? 'S';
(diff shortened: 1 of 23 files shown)


packages/amplify-graphql-index-transformer/src/resolvers.ts
```diff
@@ -272,3 +298,217 @@ function setQuerySnippet(config: PrimaryKeyDirectiveConfiguration, ctx: Transfor
...
+      projection: { projectionType: 'ALL' },
+      provisionedThroughput: cdk.Fn.conditionIf(ResourceConstants.CONDITIONS.ShouldUsePayPerRequestBilling, cdk.Fn.ref('AWS::NoValue'), {
+        ReadCapacityUnits: cdk.Fn.ref(ResourceConstants.PARAMETERS.DynamoDBModelTableReadIOPS),
+        WriteCapacityUnits: cdk.Fn.ref(ResourceConstants.PARAMETERS.DynamoDBModelTableWriteIOPS),
+      }),
+    });
+  }
+}
+
+function appendIndex(list: any, newIndex: any): any[] {
```
lazpavel, line 361: nit: consider using `$TSAny` instead of `any` if a type cannot be defined

packages/amplify-graphql-index-transformer/src/resolvers.ts
```diff
@@ -28,23 +36,26 @@ import {
...
 import { lookupResolverName } from './utils';
 
 export function replaceDdbPrimaryKey(config: PrimaryKeyDirectiveConfiguration, ctx: TransformerContextProvider): void {
   // Replace the table's primary key with the value from @primaryKey.
-  const { object } = config;
-  const table = getTable(ctx, object);
-  const tableAttrDefs = table.attributeDefinitions as any[];
-  const tableKeySchema = table.keySchema as any[];
+  const { field, object } = config;
+  const table = getTable(ctx, object) as any;
```
lazpavel, line 45: nit: consider using $TSAny instead of any if a type cannot be provided

lazpavel (approved):
LGTM, added a couple of nits for your consideration

packages/amplify-graphql-index-transformer/src/resolvers.ts
cjihrig: I didn't use `$TSAny` to try avoiding taking on a dependency on the CLI core since the Transformer can be used without the CLI.

lazpavel: makes sense, thanks

PR #7919, opened 2021-08-09
fix: added await to init invocation

xo28122000:
#### Description of changes

Added await to the async function init invocation.
Before this, the post execution events were being called before the axtual completion of the add env command. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/env/add.ts
+++ b/packages/amplify-cli/src/commands/env/add.ts
@@ -1,5 +1,6 @@
+import { $TSContext } from 'amplify-cli-core';
 import { run as init } from '../init';
 
-export const run = async context => {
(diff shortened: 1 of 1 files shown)


jhockett (approved):

packages/amplify-cli/src/commands/env/add.ts
```diff
@@ -1,5 +1,5 @@
 import { run as init } from '../init';
 
 export const run = async context => {
```
jhockett, line 3:
```suggestion
import { $TSContext } from 'amplify-cli-core';
import { run as init } from '../init';

export const run = async (context: $TSContext) => {
```

Issue #7300, opened 2021-05-10
"Profile configuration for 'default' is invalid" appears in latest 4.50.2 release on CI agent


Coteh:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

12.16.1

### Amplify CLI Version

4.50.2

### What operating system are you using?

Mac

### Amplify Categories

Not applicable

### Amplify Commands

init

### Describe the bug

We have a Jenkins server that's used for CI deployment of our Amplify projects. Today, I noticed an error I've never seen before:
```
Failed to get profile: Profile configuration for 'default' is invalid: missing aws_access_key_id, aws_secret_access_key
```
This does not make sense to me because I pass environment variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` directly to Amplify CLI and this has worked for me in the past. I did not need to configure profile at all.

On the machine, this is what's outputted from the `~/.aws/config` file for the default profile:

```
[default]
region = us-east-1
output = json
```

Also, this command does work for me locally, and I have the same values for default profile in `~/.aws/config`. However, I confirmed with the DevOps developers at my company that the AWS configuration on the CI machine was not altered in any way since the last successful build, so I think it's very likely there's an issue with Amplify itself.

My question is, is there a bug with Amplify, or must I now add complete profile information for the AWS profile to the agent? Please advise, thanks!

### Expected behavior

Amplify project is initialized on CI server and proceeds to build and deploy project

### Reproduction steps

1. Set `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to valid AWS credentials
2. Run `amplify init --yes`, passing in `--amplify` and `--providers` flags describing your Amplify project

ammarkarachi:
Hi @Coteh 
If you are already using env variables to setup the credentials. you can set the `useProfile` flag to false. The CLI should pick up the environment variables

[...]

PR #7933, opened 2021-08-11
Credential validation fix

marcvberg:
#### Description of changes

Our credential validation process did not account for the option to provide environment variables, this puts in a last second check in the case of missing access/secret keys to look at the environment

#### Issue #, if available

#7300

Ran unit tests and manually recreated a scenario of missing credentials but overriding with environment

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-provider-awscloudformation/src/system-config-manager.ts
```diff
@@ -297,10 +297,10 @@ export function getProfileCredentials(profileName: string) {
 
 function validateCredentials(credentials: $TSAny, profileName: string) {
   const missingKeys = [];
-  if (!credentials?.accessKeyId) {
+  if (!credentials?.accessKeyId && !process.env.AWS_ACCESS_KEY_ID) {
     missingKeys.push('aws_access_key_id');
   }
-  if (!credentials?.secretAccessKey) {
+  if (!credentials?.secretAccessKey && !process.env.AWS_SECRET_ACCESS_KEY) {
```
jhockett, line 303: I was under the impression that this function is only be called for validating profile creds. This logic should be skipped altogether if using env vars iirc

See line 89

marcvberg: The problem is that we have a customer who wants to use a profile which authenticates using the env var override, if we want to not support that flow we can close this PR and mark the issue as won't fix

lazpavel (approved):
LGTM

cjihrig:
@jhockett should this not land yet based on https://github.com/aws-amplify/amplify-cli/pull/7933#discussion_r686993258?

packages/amplify-provider-awscloudformation/src/system-config-manager.ts
marcvberg: My thought is that it might be a CICD thing where they're using contextual creds, say from EC2, and want to have a version controlled .aws/config file. It's a pretty specific use case but it's not invalid, I think it might make sense to allow something like this flow through some kind of escape hatch where we say hey do whatever you want and if it works through the AWS CLI, it's fine, but we won't give protections or validations.

I can understand not wanting to make the credential validation too complex because of all the permutations, but it's also probably not great for us to block alternative flows entirely. What's your feeling on it? Close the PR as a won't do, new escape hatch, or get some more feedback

jhockett:
> @jhockett should this not land yet based on [#7933 (comment)](https://github.com/aws-amplify/amplify-cli/pull/7933#discussion_r686993258)?



> @jhockett should this not land yet based on [#7933 (comment)](https://github.com/aws-amplify/amplify-cli/pull/7933#discussion_r686993258)?

Just getting a notification for this comment today somehow. Eventually, I think there's a longer term solution is needed to simplify the credential logic, but the changes in this PR LGTM.

jhockett (approved):

PR #7982, opened 2021-08-18
feat: graphql auth v2 add schemaChanges, iam policy generation, and query/read resolvers

SwaySway:
#### Description of changes
- Updated Get Resolver Logic in order to maintain to maintain auth vtl logic pre query.
  - updated snapshot tests to account for new get resolver logic
- Added Transform Schema logic
  -  Adding Owner Arguments on Subscription Operations
  - Adding AppSync Service Directives on Operations
  - Adding ownerfield in model type
- Resolver Logic for Reads
  - Get Resolver 
  - List Resolver
  - Field Resolver (if there is at least one role that is not allowed generate field resolver)
- IAM Policy Generation

### DIFF
--- a/packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
+++ b/packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
@@ -383,25 +383,44 @@ $util.toJson($UpdateItem)
   "Query.getTest.req.vtl": "## [Start] Get Request template. **
 #set( $GetRequest = {
   \\"version\\": \\"2018-05-29\\",
-  \\"operation\\": \\"GetItem\\"
+  \\"operation\\": \\"Query\\"
 } )
 #if( $ctx.stash.metadata.modelObjectKey )
-  #set( $key = $ctx.stash.metadata.modelObjectKey )
+  #set( $expression = \\"\\" )
+  #set( $expressionValues = {} )
+  #foreach( $item in $ctx.stash.metadata.modelObjectKey.entrySet() )
+    #set( $expression = \\"$expression$item.key = :$item.key AND \\" )
+    $util.qr($expressionValues.put(\\":$item.key\\", $item.value))
+  #end
+  #set( $expression = $expression.replaceAll(\\"AND $\\", \\"\\") )
+  #set( $query = {
+  \\"expression\\": $expression,
+  \\"expressionValues\\": $expressionValues
+} )
 #else
-  #set( $key = {
-  \\"id\\":   $util.dynamodb.toDynamoDB($ctx.args.id)
+  #set( $query = {
+  \\"expression\\": \\"id = :id\\",
+  \\"expressionValues\\": {
+      \\":id\\":     $util.dynamodb.toDynamoDBJson($ctx.args.id)
+  }
 } )
 #end
-$util.qr($GetRequest.put(\\"key\\", $key))
+$util.qr($GetRequest.put(\\"query\\", $query))
+#if( !$util.isNullOrEmpty($ctx.stash.get(\\"authFilter\\")) )
+  $util.qr($GetRequest.put(\\"filter\\", $ctx.stash.get(\\"authFilter\\")))
+#end
 $util.toJson($GetRequest)
 ## [End] Get Request template. **",
-  "Query.getTest.res.vtl": "## [Start] Get ResponseTemplate. **
-#if( $ctx.error )
-  $util.error($ctx.error.message, $ctx.error.type)
+  "Query.getTest.res.vtl": "## [Start] Get Response template. **
+#if( !$$ctx.result.items.isEmpty() && $ctx.result.scannedCount == 1 )
+  $util.toJson($ctx.result.items[0])
 #else
-  $util.toJson($ctx.result)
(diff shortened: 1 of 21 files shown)


packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
```diff
@@ -1031,25 +1062,44 @@ $util.toJson($UpdateItem)
...
+#if( !$util.isNullOrEmpty($ctx.stash.get(\\"authFilter\\")) )
+  $util.qr($GetRequest.put(\\"filter\\", $ctx.stash.get(\\"authFilter\\")))
+#end
 $util.toJson($GetRequest)
 ## [End] Get Request template. **",
-  "Query.getTest.res.vtl": "## [Start] Get ResponseTemplate. **
-#if( $ctx.error )
-  $util.error($ctx.error.message, $ctx.error.type)
+  "Query.getTest.res.vtl": "## [Start] Get Response template. **
+#if( !$$ctx.result.items.isEmpty() && $ctx.result.scannedCount == 1 )
```
cjihrig, line 1094: Is there an extra dollar sign here?

packages/amplify-graphql-model-transformer/src/resolvers/query.ts
```diff
@@ -16,26 +16,62 @@ import {
...
 
+const authFilter = methodCall(ref('ctx.stash.get'), str('authFilter'));
+
 /**
  * Generate get query resolver template
  */
 export const generateGetRequestTemplate = (): string => {
   const statements: Expression[] = [
-    set(ref('GetRequest'), obj({ version: str('2018-05-29'), operation: str('GetItem') })),
+    set(ref('GetRequest'), obj({ version: str('2018-05-29'), operation: str('Query') })),
```
cjihrig, line 32: I know we talked about this offline, but what was the specific reason for needing to switch the operation again?

packages/amplify-graphql-model-transformer/src/resolvers/query.ts
```diff
@@ -16,26 +16,62 @@ import {
...
     toJson(ref('GetRequest')),
   ];
 
   return printBlock('Get Request template')(compoundExpression(statements));
 };
 
+export const generateGetResponseTemplate = (): string => {
+  const statements: Expression[] = [
+    ifElse(
+      and([not(ref('$ctx.result.items.isEmpty()')), equals(ref('ctx.result.scannedCount'), int(1))]),
```
cjihrig, line 64:
```suggestion
      and([not(ref('ctx.result.items.isEmpty()')), equals(ref('ctx.result.scannedCount'), int(1))]),
```

packages/amplify-graphql-model-transformer/src/resolvers/query.ts
```diff
@@ -16,26 +16,62 @@ import {
...
+      toJson(ref('ctx.result.items[0]')),
+      compoundExpression([
+        iff(and([ref('$ctx.result.items.isEmpty()'), equals(ref('ctx.result.scannedCount'), int(1))]), ref('util.unauthorized()')),
```
cjihrig, line 67:
```suggestion
        iff(and([ref('ctx.result.items.isEmpty()'), equals(ref('ctx.result.scannedCount'), int(1))]), ref('util.unauthorized()')),
```

packages/amplify-graphql-model-transformer/src/resolvers/query.ts
SwaySway: The reason for this is so we can apply an auth filter as opposed to running auth resolver logic after the get.

packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
```diff
@@ -383,25 +383,44 @@ $util.toJson($UpdateItem)
...
-#else
-  #set( $key = {
-  \\"id\\":   $util.dynamodb.toDynamoDB($ctx.args.id)
+  #set( $expression = \\"\\" )
+  #set( $expressionValues = {} )
+  #foreach( $item in $ctx.stash.metadata.modelObjectKey.entrySet() )
+    #set( $expression = \\"$expression$item.key = :$item.key AND \\" )
+    $util.qr($expressionValues.put(\\":$item.key\\", $item.value))
+  #end
+  #set( $expression = $expression.replaceAll(\\"AND $\\", \\"\\") )
```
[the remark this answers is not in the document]
SwaySway: Yea we can, I just figured this would the shortest way to write it since we don't need to include the velocity count in the string due to it being a query expression.

packages/amplify-graphql-transformer-core/src/transformer-context/stack-manager.ts
```diff
@@ -38,6 +38,10 @@ export class StackManager implements StackManagerProvider {
     return this.rootStack;
   };
 
+  getRootStack = (): Stack => {
```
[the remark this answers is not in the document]
SwaySway: Ah right I missed that, changing the interface to display it as well.

cjihrig (approved):
There are still some LGTM bot issues to address.

SwaySway:
Merging this PR for now addressing NITs along with mutations.

PR #8035, opened 2021-08-25
fix(graphql-model-transformer): @model conflict resolution

lazpavel:
#### Description of changes
- implementation for graphql transformer V2 datastore resolvers

- added new tests
- ran `yarn test`
- manual testing by creating an app using `"useexperimentalpipelinedtransformer": true,`
- [circle ci](https://app.circleci.com/pipelines/github/lazpavel/amplify-cli/317/workflows/94594689-bad5-4bd2-8680-fb1df557b152)

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-transformer-interfaces/src/transform-host-provider.ts
```diff
@@ -18,11 +18,18 @@ import {
 } from './graphql-api-provider';
 import { IRole } from '@aws-cdk/aws-iam';
 
+export interface DynamoDbDataSourceOptions extends DataSourceOptions {
+  /**
+   * ServiceRole for the Amazon OpenSearch
```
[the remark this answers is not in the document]
lazpavel: this was the issue that we debugged with the appsync team, it looks that if we don't pass the serviceRole at the construction time, and assign one to the cfn object later, at the construction time one is created by default, this was causing the incorrect rights to write to the delta table. Will update the comment though

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -163,6 +183,7 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
       queries: {
         get: toCamelCase(['get', typeName]),
         list: toCamelCase(['list', plurality(typeName, true)]),
+        ...(ctx.resolvers.isProjectUsingDataStore() ? { sync: toCamelCase(['sync', plurality(typeName, true)]) } : {}),
```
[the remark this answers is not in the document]
lazpavel: updated

packages/amplify-graphql-transformer-core/src/transform-host.ts
```diff
@@ -19,6 +19,7 @@ import { IRole } from '@aws-cdk/aws-iam';
 import { InlineTemplate, S3MappingFunctionCode } from './cdk-compat/template-asset';
 import { toCamelCase } from 'graphql-transformer-common';
 import { GraphQLApi } from './graphql-api';
+import { DynamoDbDataSourceOptions } from '@aws-amplify/graphql-transformer-interfaces/src/transform-host-provider';
```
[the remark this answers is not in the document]
lazpavel: updated

SwaySway (approved):
aside from nits lgtm

packages/amplify-graphql-transformer-core/src/config/transformer-config.ts
```diff
@@ -1,3 +1,53 @@
...
+  name: string;
+  issuerUrl: string;
+  clientId?: string;
+  iatTTL?: number;
+  authTTL?: number;
+};
+
 // Sync Config
 export const enum ConflictHandlerType {
   Optimistic = 'OPTIMISTIC_CONCURRENCY',
```
SwaySway, line 53: nit: Can you rename the enums to uppercase, since most of the other enums used in the cli are uppercased as well for consistency

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -213,109 +231,24 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
 
       // Update the field with auto generatable Fields
       this.addAutoGeneratableFields(ctx, type);
+
+      if (this.options.SyncConfig) {
+        this.addModelSyncFields(ctx, type);
+      }
```
SwaySway, line 237: we should use the datastore check here if we can (note: the interfaces will need this function as well)
```suggestion
      if (ctx.isProjectUsingDataStore()) {
        this.options.SyncConfig = SyncUtils.getSyncConfig(ctx as TransformerContextProvider, def!.name.value);
        this.addModelSyncFields(ctx, type);
      }
```

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -163,6 +176,9 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
       queries: {
         get: toCamelCase(['get', typeName]),
         list: toCamelCase(['list', plurality(typeName, true)]),
+        ...((ctx as TransformerContextProvider).isProjectUsingDataStore()
+          ? { sync: toCamelCase(['sync', plurality(typeName, true)]) }
+          : undefined),
```
SwaySway, line 181: We add the function on all interfaces the context uses we won't need the type casting.

packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
```diff
@@ -199,6 +215,8 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
     this.ensureModelSortDirectionEnum(ctx);
     for (const type of this.typesWithModelDirective) {
       const def = ctx.output.getObject(type)!;
+      this.options.SyncConfig = SyncUtils.getSyncConfig(ctx as TransformerContextProvider, def!.name.value);
```
SwaySway, line 218: see comment below on using datastore check

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -180,6 +182,28 @@ export class TransformerResolver implements TransformerResolverProvider {
             const tableName = this.datasource.ds.dynamoDbConfig?.tableName;
             dataSource = `$util.qr($ctx.stash.put("tableName", "${tableName}"))`;
           }
+
+          const syncConfig = SyncUtils.getSyncConfig(context, this.typeName);
+          if (syncConfig) {
```
SwaySway, line 187: use datastore check here as well
```suggestion
          if (ctx.isProjectUsingDataStore()) {
            const syncConfig = SyncUtils.getSyncConfig(context, this.typeName) as SyncConfig;
```

SwaySway (changes requested):
A few more comments on using datastore check.
Can we also update the flag in searchable to use the value if datastore is enabled?
https://github.com/aws-amplify/amplify-cli/blob/50f07efcf7c5663b6fd123d300ec473377d03abe/packages/amplify-graphql-searchable-transformer/src/graphql-searchable-transformer.ts#L80

cjihrig (approved):

Issue #8010, opened 2021-08-23
Removing all resource access permission is not possible in a function


cfbo:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### Amplify CLI Version

5.3.0

### What operating system are you using?

Windows

### Amplify Categories

function

### Amplify Commands

update

### Describe the bug

I want to remove a permission from a function that was previously added (to api), but when I remove it and press enter I get the error in the second screenshot.. 

Current status:
![image](https://user-images.githubusercontent.com/7568421/130475665-1788386b-65f3-434b-b030-8467e31dd764.png)

Need to change like that:
![image](https://user-images.githubusercontent.com/7568421/130475497-d457c75b-e569-4d88-ba20-a6850a8412ba.png)

Looks like an amplify bug I guess? 

### Expected behavior

I should be able to remove a previously added permission via CLI.

### Reproduction steps

1. add api A
2. add function B
3. amplify update function B > add resource permission to api A
4. push
5. amplify update function B > remove permission to api B >>>> not possible due validation "You must select at least one category"

josefaidt:
Hey @cfbo :wave: thanks for filing this! I was able to successfully reproduce with the steps provided and am unable to remove access permissions to my API resource. Marking this as a bug 🙂

PR #8052, opened 2021-08-27
fix(amplify-category-function): allow user to update function without length validation

danielleadams:
#### Description of changes
When updating a function, removes to category length validation so that a user can remove all category access.

#### Issue #, if available
Fixes https://github.com/aws-amplify/amplify-cli/issues/8010

I tested this locally by duplicating the Issue steps, and it works as requested now. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/service-walkthroughs/execPermissionsWalkthrough.ts
@@ -202,7 +202,6 @@ const selectCategories = (choices: DistinctChoice<any>[], currentPermissionMap: 
   name: 'categories',
   message: 'Select the categories you want this function to have access to.',
   choices,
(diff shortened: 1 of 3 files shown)


danielleadams:
Upon discussion with product, this is a bugfix.

attilah (changes requested):
Could you add an E2E test that:
- creates an API
- creates a Lambda
- adds permission to Lambda to access API
- push
- remove all permissions
- push
- verify pushed cfn file to not to have the access permissions between resources defined.

This way we will cannot have a regression around the full removal of permissions.

danielleadams:
@attilah this is ready for re-review. thanks!

attilah (approved):
Thanks for the test! LGTM!

danielleadams:
Moving to draft until I have time to fix the test in the rebase.

sachscode (approved):

cjihrig (approved):

PR #8057, opened 2021-08-28
fix(amplify-category-function): 'filter' of undefined when removing layers

quan-ntekka:
#### Description of changes

When metadata for function doesn't have dependsOn property, then amplify function remove a layer would throw an error 'Cannot read property 'filter' of undefined' ... on layerHelpers.ts:248:33

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


jhockett (approved):

jhockett:
Thanks for the fix @ntekka!

PR #8124, opened 2021-09-08
feat(amplify-category-api): prompt api key creation on amplify push

danielleadams:
#### Description of changes
When an API key is either expired or not present AND global sandbox mode is enabled, `amplify push` will prompt the user to create an API key for use on the GraphQL API.

The flow has been tested from the `enable-sandbox-mode` feature branch. There are also unit tests that test the functionality of the code.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-api/src/provider-utils/awscloudformation/cfn-api-artifact-handler.ts
```diff
@@ -129,6 +129,30 @@ class CfnApiArtifactHandler implements ApiArtifactHandler {
     printApiKeyWarnings(this.context, oldConfigHadApiKey, authConfigHasApiKey(authConfig));
   };
 
+  updateArtifactsWithoutCompile = async (request: UpdateApiRequest): Promise<void> => {
```
SwaySway, line 132: is there a reason we would not want to compile?

SwaySway (commented):
for prompts can we use `amplify-prompts`?

packages/amplify-category-api/src/provider-utils/awscloudformation/cfn-api-artifact-handler.ts
danielleadams: Originally it was so that the compiler didn't run twice on amplify push when updating the API Key, but I think we can consolidate this work with the fix I am doing on the transformer branch. I'll update this.

danielleadams: @SwaySway I removed the `updateArtifactsWithoutCompile` and am just using `updateArtifacts` for now.

SwaySway: > Originally it was so that the compiler didn't run twice on amplify push when updating the API Key, but I think we can consolidate this work with the fix I am doing on the transformer branch. I'll update this.

Ah that makes sense, maybe we can add a flag into the update artifacts to not run gql-compile. I think there might be some value in add that, though I'd imagine we'd need to review this.

SwaySway (approved):

PR #8155, opened 2021-09-10
fix(amplify-category-geo): pricing plan, data provider,  JS plugin name updates

phani-srikar:
#### Description of changes

- Remove deprecated `context.print` and use `amplify-prompts` package for console outputs.
- Update Pricing Plan Walkthrough to include a clickable "learn more" option, also update wordings
- Correct the naming of Data provider to `HERE` in place index walkthrough
- Change the plugin name for JS frontend to `amazon_location_service`

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/8096

unit tests and e2e tests are modified and locally run

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-geo/src/commands/geo/add.ts
+++ b/packages/amplify-category-geo/src/commands/geo/add.ts
@@ -25,7 +25,7 @@ export const run = async(context: $TSContext) => {
     if (error.message) {
       printer.error(error.message);
     }
-    printer.info('');
+    printer.blankLine();
     if (error.stack) {
       printer.info(error.stack);
     }
--- a/packages/amplify-category-geo/src/commands/geo/add.ts
+++ b/packages/amplify-category-geo/src/commands/geo/add.ts
@@ -3,6 +3,7 @@ import { category } from '../../constants';
 import { supportedServices } from '../../supportedServices';
 import { $TSContext } from 'amplify-cli-core';
 import { addResource } from '../../provider-controllers';
+import { printer } from 'amplify-prompts';
 
 export const name = 'add';
 
@@ -10,11 +11,11 @@ export const run = async(context: $TSContext) => {
   const { amplify } = context;
 
   try {
-    context.print.warning(previewBanner);
+    printer.warn(previewBanner);
     const result: {service: string, providerName: string} = await amplify.serviceSelectionPrompt(context, category, supportedServices, chooseServiceMessageAdd);
 
     if (result.providerName !== provider) {
-      context.print.error(`Provider ${result.providerName} not configured for this category`);
+      printer.error(`Provider ${result.providerName} not configured for this category`);
       return;
     }
 
@@ -22,13 +23,13 @@ export const run = async(context: $TSContext) => {
 
   } catch (error) {
     if (error.message) {
-      context.print.error(error.message);
+      printer.error(error.message);
     }
-    context.print.info('');
+    printer.info('');
     if (error.stack) {
-      context.print.info(error.stack);
+      printer.info(error.stack);
     }
-    context.print.error('There was an error adding the geo resource');
+    printer.error('There was an error adding the geo resource');
     context.usageData.emitError(error);
     process.exitCode = 1;
   }
--- a/packages/amplify-category-geo/src/provider-controllers/map.ts
+++ b/packages/amplify-category-geo/src/provider-controllers/map.ts
@@ -68,7 +68,7 @@ export const removeMapResource = async (
   await amplify.removeResource(context, category, resourceToRemove)
     .catch((err: $TSAny) => {
       if (err.stack) {
-        printer.info(err.stack);
+        printer.error(err.stack);
         printer.error(`An error occurred when removing the geo resource ${resourceToRemove}`);
       }
 
--- a/packages/amplify-category-geo/src/provider-controllers/map.ts
+++ b/packages/amplify-category-geo/src/provider-controllers/map.ts
@@ -6,6 +6,7 @@ import { convertToCompleteMapParams, MapParameters } from '../service-utils/mapP
 import { $TSAny, $TSContext } from 'amplify-cli-core';
 import { printNextStepsSuccessMessage, setProviderContext } from './index';
 import { ServiceName } from '../service-utils/constants';
+import { printer } from 'amplify-prompts';
 
 export const addMapResource = async (
   context: $TSContext
@@ -20,8 +21,7 @@ export const addMapResource = async (
 
   await createMapResource(context, completeParameters);
 
-  const { print } = context;
-  print.success(`Successfully added resource ${completeParameters.name} locally.`);
+  printer.success(`Successfully added resource ${completeParameters.name} locally.`);
   printNextStepsSuccessMessage(context);
   return completeParameters.name;
 };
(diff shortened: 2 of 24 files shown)


packages/amplify-category-geo/src/commands/geo/add.ts
```diff
@@ -3,32 +3,33 @@ import { category } from '../../constants';
...
 
     return await addResource(context, result.service);
 
   } catch (error) {
     if (error.message) {
-      context.print.error(error.message);
+      printer.error(error.message);
     }
-    context.print.info('');
+    printer.info('');
```
edwardfoyle, line 28: nit: there's a `printer.blankLine()` that could be used instead of this

packages/amplify-category-geo/src/provider-controllers/map.ts
```diff
@@ -69,8 +68,8 @@ export const removeMapResource = async (
   await amplify.removeResource(context, category, resourceToRemove)
     .catch((err: $TSAny) => {
       if (err.stack) {
-        context.print.info(err.stack);
-        context.print.error(`An error occurred when removing the geo resource ${resourceToRemove}`);
+        printer.info(err.stack);
```
edwardfoyle, line 71: nit: maybe put the stack under the error message?

packages/amplify-category-geo/src/service-walkthroughs/mapWalkthrough.ts
```diff
@@ -79,12 +80,12 @@ export const mapAdvancedWalkthrough = async (context: $TSContext, parameters: Pa
     // const includePricingPlan = await geoServiceExists(ServiceName.Map) || await geoServiceExists(ServiceName.PlaceIndex);
     const includePricingPlan = false;
     const currentPricingPlan = parameters.pricingPlan ? parameters.pricingPlan : await getGeoPricingPlan();
-    context.print.info('Available advanced settings:');
-    context.print.info('- Map style & Map data provider (default: Streets provided by Esri)');
+    printer.info('Available advanced settings:');
+    printer.info('- Map style & Map data provider (default: Streets provided by Esri)');
     if (includePricingPlan) {
-        context.print.info(`- Map pricing plan (current: ${currentPricingPlan})`);
+        printer.info(`- Map pricing plan (current: ${currentPricingPlan})`);
```
edwardfoyle, line 86: nit: use `formatter.list()` to print a list of values

packages/amplify-category-geo/src/service-walkthroughs/placeIndexWalkthrough.ts
```diff
@@ -79,13 +80,13 @@ export const placeIndexAdvancedWalkthrough = async (context: $TSContext, paramet
...
     const currentPricingPlan = parameters.pricingPlan ? parameters.pricingPlan : await getGeoPricingPlan();
-    context.print.info('Available advanced settings:');
-    context.print.info('- Search data provider (default: Esri)');
-    context.print.info('- Search result storage location (default: no result storage)');
+    printer.info('Available advanced settings:');
+    printer.info('- Search data provider (default: Esri)');
+    printer.info('- Search result storage location (default: no result storage)');
     if (includePricingPlan) {
-        context.print.info(`- Map pricing plan (current: ${currentPricingPlan})`);
+        printer.info(`- Map pricing plan (current: ${currentPricingPlan})`);
```
edwardfoyle, line 87: same here, can use `formatter.list()` (and anywhere else a list is printed)

packages/amplify-category-geo/src/service-walkthroughs/resourceWalkthrough.ts
```diff
@@ -29,22 +30,28 @@ export async function pricingPlanWalkthrough<T extends ResourceParameters>(
...
         name: 'pricingPlanBusinessType',
         message: 'Are you tracking commercial assets for your business in your app?',
         choices: pricingPlanBusinessTypeChoices,
-        default: pricingPlan === PricingPlan.RequestBasedUsage ? false : true
+        default: pricingPlan === PricingPlan.RequestBasedUsage ? PricingPlan.RequestBasedUsage : 'Unknown'
     };
 
-    const pricingPlanBusinessTypeChoice = ((await inquirer.prompt([pricingPlanBusinessTypePrompt])).pricingPlanBusinessType as boolean);
-    if (pricingPlanBusinessTypeChoice === false) {
+    let pricingPlanBusinessTypeChoice = ((await inquirer.prompt([pricingPlanBusinessTypePrompt])).pricingPlanBusinessType);
```
edwardfoyle, line 48: would be nice to migrate to `prompter.pick()` for this prompt and the ones below. Should be something like:
```suggestion
    let pricingPlanBusinessTypeChoice = await prompter.pick('Are you tracking commercial...', pricingPlanBusinessTypeChoices)
```

packages/amplify-category-geo/src/service-walkthroughs/mapWalkthrough.ts
phani-srikar: Nice. Will use it

packages/amplify-category-geo/src/service-walkthroughs/resourceWalkthrough.ts
phani-srikar: I did try to adapt the new `prompter` and faced some issues with `pick`.

Without using `prompter.pick`:

Screen Shot 2021-09-13 at 9 54 16 AM
And when I hit cancel(`ctrl+z`), it graciously terminates
Screen Shot 2021-09-13 at 9 54 44 AM

Using `prompter.pick`:
Screen Shot 2021-09-13 at 9 51 02 AM
The cursor is at the wrong place which causes the terminal to be left in weird state when we hit `ctrl+z`. 
Screen Shot 2021-09-13 at 9 51 30 AM
Moreover, the options displayed are cleaner with `>` in front of them and a message `use arrow keys`. 

Another DX related issue. When we use array of name-value pairs in pick, we need to specify the index as default. This might need additional effort to compute the index as currently we specify the value as default, not the index. 

I would like to come back to this in a separate PR when these issues are resolved.

edwardfoyle: What's your use case for stopping the process? I tried out stopping and resuming the process when in the middle of an inquirer prompt and there's an error there as well. If you just want to exit the process, ctrl+c will work.

Re: the `Use arrow keys` suggestion, yes that's good feedback, we can add that. However, I would still recommend using prompter.pick() and when that update lands you'll get it for free

As for specifying the index rather than the value of the initial selection, this was an intentional choice because if the type of the values in the list are numbers it's ambiguous if you intend to select a value or an index. Furthermore, since the list value type can be anything, it may not be sufficient to compare values using `===` Since an array of values must already be provided, it should be trivial to pass in an initial index of `selectionList.indexOf(defaultValue)`

phani-srikar: The use-case for stopping the current flow is when I am not sure of the options to choose and might want to add the resource at a later time. The process is stopped successfully but the output is not cleared graciously as you can see from the screenshot.

edwardfoyle (approved):

packages/amplify-category-geo/src/service-walkthroughs/resourceWalkthrough.ts
edwardfoyle: @phani-srikar  Right, but as far as I can tell, the current prompts that use inquirer don't support resuming the process in the middle of a prompt either. If you want to stop the process and restart it later, why not use ctrl+c?

PR #8195, opened 2021-09-16
feat: add @manytomany directive

cjihrig:
#### Description of changes
This PR adds a new `@manyToMany` directive.

Note to simplify reviewing this, the first six of the seven commits are #8035, so only the last commit needs reviewed here.

Manual, unit, and E2E testing.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-relational-transformer/src/graphql-many-to-many-transformer.ts
+++ b/packages/amplify-graphql-relational-transformer/src/graphql-many-to-many-transformer.ts
@@ -1,0 +1,208 @@
+import { DirectiveWrapper, InvalidDirectiveError, TransformerPluginBase } from '@aws-amplify/graphql-transformer-core';
+import {
+  TransformerContextProvider,
+  TransformerPrepareStepContextProvider,
+  TransformerSchemaVisitStepContextProvider,
+  TransformerTransformSchemaStepContextProvider,
+  TransformerValidationStepContextProvider,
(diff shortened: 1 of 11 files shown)


packages/amplify-graphql-relational-transformer/src/graphql-many-to-many-transformer.ts
```diff
@@ -0,0 +1,208 @@
...
+    } as ManyToManyDirectiveConfiguration);
+
+    validateModelDirective(args);
+    args.connectionFields = [];
+
+    if (!isListType(definition.type)) {
+      throw new InvalidDirectiveError(`@${directiveName} must be used with a list.`);
+    }
+
+    addDirectiveToRelationMap(this.relationMap, args);
```
lazpavel, line 73: nit: is there a reason `addDirectiveToRelationMap` is not a type method and avoid passing the `this.relationMap`

lazpavel (approved):
small nit, but LGTM

packages/amplify-graphql-relational-transformer/src/graphql-many-to-many-transformer.ts
cjihrig: No strong reason it couldn't be one, but my reasoning was:

1. It makes code easier to move around if it's not an instance method.
2. I didn't want to make it part of the public API. Our transpilation process doesn't actually make private things private as far as I can tell.

PR #8214, opened 2021-09-17
fix(test): v2 searchable aggregate test

sundersc:
Fix for Searchable v2 flaky test.

### DIFF
--- a/packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformerV2.e2e.test.ts
+++ b/packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformerV2.e2e.test.ts
@@ -128,7 +128,6 @@ afterAll(async () => {
 });
 
 test('query for aggregate scalar results', async () => {
-  const expectedValue = 60;
   const searchResponse = await GRAPHQL_CLIENT.query(
     `query {
       searchPosts(aggregates: [{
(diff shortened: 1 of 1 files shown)


lazpavel (approved):
LGTM

packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformerV2.e2e.test.ts
```diff
@@ -149,6 +151,15 @@ test('query for aggregate scalar results', async () => {
     {},
   );
   expect(searchResponse).toBeDefined();
+
+  let searchPosts = searchResponse.data.searchPosts.items;
+  let expectedValue = searchPosts[0].ups;
```
yuth, line 156: We should not be relying on the first result and compare that value. It should be a know value (before hand) like you had it earlier. Can we just wait for all the data to be propagated before running this test

yuth (changes requested):

packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformerV2.e2e.test.ts
sundersc: Not relying on the first value, running the searchPosts query and calculating MIN value from the results and verifying it against the aggregate result.

I agree that waiting for the data to be propagated is the best solution, will work on it in a separate PR.

sundersc:
Followed a different approach to make sure the data propagation is complete.
https://github.com/aws-amplify/amplify-cli/pull/8251

PR #8235, opened 2021-09-21
env name in sandbox directive fix

danielleadams:
#### Description of changes
Changes the checking of env name for sandbox mode to only use the GraphQL schema to store env name. This PR:
- runs the API Key check at push resources at the `amplify-provider-awscloudformation` level, removes the messaging from the `amplify-cli` level and removes checking of current env name from transformer level.
- adds a flag to disable GraphQL compile when artifacts are being updated, so `amplify push` does not run compile twice.

This PR was opened against the `@auth` feature branch, but it can be redirected to `master` in order to simplify merging of the feature branch.

Unit tests were moved from their previous place and ran a demo with `amplify-dev`. Tests that were in place to check e2e and resolver output haven't broken.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


danielleadams:
Closing since this was merged into the feature branch.

PR #8251, opened 2021-09-22
fix(Searchable): confirm ES data propagate

sundersc:
Fix for flaky searchable v2 E2E test.


packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformerV2.e2e.test.ts
```diff
@@ -60,8 +60,32 @@ const createEntries = async () => {
...
+          items {
+            id
+          }
+        }
+      }`,
+      {},
+    );
+    currentRetryCount += 1;
+    waitInMilliseconds = waitInMilliseconds * 2;
+  } while (searchResponse.data.searchPosts?.items?.length < expectedCount && currentRetryCount <= maxRetryCount);
```
SwaySway, line 86: nit: can we use equality here?

SwaySway (approved):
aside from nits lgtm

packages/graphql-transformers-e2e-tests/src/__tests__/SearchableModelTransformerV2.e2e.test.ts
sundersc: If the data is not propagated yet, the count would be less than the expected count, so we can loop until the count is equal to the records count.

PR #8276, opened 2021-09-25
Root stack override

akshbhu:
#### Description of changes

- Root stack builder using cdk

- Manual testing
- unit tests for root stack transform classes

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/commands/build-override.ts
```diff
@@ -23,6 +23,7 @@ export const run = async (context: $TSContext) => {
     if (categoryName && resourceName) {
       filteredResources = resourcesToBuild.filter(resource => resource.category === categoryName && resource.resourceName === resourceName);
     }
+    await context.amplify.invokePluginMethod(context, 'awscloudformation', undefined, 'transformResourceWithOverrides', [context]);
```
kaustavghosh06, line 26: What is the third param which is undefined here?

packages/amplify-cli/src/extensions/amplify-helpers/root-stack-status.ts
```diff
@@ -0,0 +1,36 @@
+import { pathManager } from 'amplify-cli-core';
+import { nestedStackFileName } from 'amplify-provider-awscloudformation';
+import { hashElement, HashElementOptions } from 'folder-hash';
+import * as fs from 'fs-extra';
+
+export function getHashForRootStack(dirPath, files?: string[]) {
+  const options: HashElementOptions = {
+    folders: { exclude: ['.*', 'node_modules', 'test_coverage', 'dist', 'build'] },
```
kaustavghosh06, line 8: the build will have the cfn.json file..so should this folder be excluded?

packages/amplify-provider-awscloudformation/src/commands/awscloudformation/override.ts
```diff
@@ -3,17 +3,28 @@
...
+    const srcPath = path.normalize(path.join(__dirname, '..', '..', '..', 'resources', 'overrides-resource'));
+    // empty build dir for old CFN files
+    // no need for rollback since these files will be autogenerated after push
+    const rootStackBuildPath = path.join(destPath, 'build');
+    if (fs.existsSync(rootStackBuildPath)) {
+      fs.emptyDirSync(rootStackBuildPath);
+    }
+    await generateOverrideSkeleton(context, srcPath, destPath);
+  } else {
+    printer.info('Overrides are not enabled for Root stack . Turn overrides.project = true in cli.json');
```
kaustavghosh06, line 28: Project level overrides is currently not turned on. In cli.json file please include the following:
```
{
  override: {
     project: true
  }
}
```

packages/amplify-provider-awscloudformation/src/commands/awscloudformation/override.ts
```diff
@@ -3,17 +3,28 @@
...
-  const srcPath = path.normalize(path.join(__dirname, '..', '..', '..', 'resources', 'overrides-resource'));
-
-  await generateOverrideSkeleton(context, srcPath, destPath);
+    const destPath = path.normalize(path.join(backendDir, 'awscloudformation'));
+    const srcPath = path.normalize(path.join(__dirname, '..', '..', '..', 'resources', 'overrides-resource'));
+    // empty build dir for old CFN files
+    // no need for rollback since these files will be autogenerated after push
+    const rootStackBuildPath = path.join(destPath, 'build');
+    if (fs.existsSync(rootStackBuildPath)) {
+      fs.emptyDirSync(rootStackBuildPath);
```
kaustavghosh06, line 24: Why do we need to empty this? Can't we just override this? I'm a bit scared of having destroy operations of customer folders in code.

packages/amplify-provider-awscloudformation/src/index.ts
```diff
@@ -162,4 +167,5 @@ module.exports = {
   updateEnv,
   uploadHooksDirectory,
   transformResourceWithOverrides,
+  nestedStackFileName,
```
kaustavghosh06, line 170: Why export twice? I think we're exporting it above as well. Can we export in only place in this file please?

packages/amplify-provider-awscloudformation/src/root-stack-builder/types.ts
```diff
@@ -0,0 +1,31 @@
...
+  unauthRole?: iam.CfnRole;
+
+  addCfnParameter(props: cdk.CfnParameterProps, logicalId: string): void;
+  addCfnOutput(props: cdk.CfnOutputProps, logicalId: string): void;
+  addCfnMapping(props: cdk.CfnMappingProps, logicalId: string): void;
+  addCfnCondition(props: cdk.CfnConditionProps, logicalId: string): void;
+  addCfnResource(props: cdk.CfnResourceProps, logicalId: string): void;
+}
+
+export interface Template {
```
kaustavghosh06, line 17: don't we have the same type defined in core? can we re-use instead of re-defining here again?

packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -165,13 +173,37 @@ function cloneCLIJSONForNewEnvironment(context) {
...
     context = await storeCurrentCloudBackend(context);
     await storeArtifactsForAmplifyService(context);
     await uploadHooksDirectory(context);
   }
   return context;
 }
 
+export const storeRootStackTemplate = async (context: $TSContext, template?: Template) => {
+  // generate template again as the folder structure was not created when root stack was initiaized
+  if (template === undefined) {
```
kaustavghosh06, line 188: when would this be the case?

packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -165,13 +173,37 @@ function cloneCLIJSONForNewEnvironment(context) {
...
+    template = await transformRootStack(context);
+    await prePushCfnTemplateModifier(template);
+  }
+
+  // RootStack deployed to backend/awscloudformation/build
+  const projectRoot = pathManager.findProjectRoot();
+  const rootStackBackendBuildDir = pathManager.getRootStackDirPath(projectRoot);
```
kaustavghosh06, line 195: wouldn't you need to append the "build" path to the rootstckdir path?

packages/amplify-provider-awscloudformation/src/root-stack-builder/stackSynthesizer.ts
```diff
@@ -0,0 +1,45 @@
+import { ISynthesisSession, Stack, LegacyStackSynthesizer } from '@aws-cdk/core';
```
kaustavghosh06, line 1: Should this file be in core? I'm assuming every category would need this?

kaustavghosh06: also change to stack-synthesizer.ts

packages/amplify-provider-awscloudformation/src/utility-functions.js
```diff
@@ -71,6 +71,7 @@ module.exports = {
     for (const resource of options.resourcesToBuild) {
       await transformResourceWithOverrides(context, resource);
     }
+    await transformResourceWithOverrides(context);
```
kaustavghosh06, line 74: why the repeat call? It's being called above too in the loop.

packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -47,18 +47,24 @@ import { preProcessCFNTemplate } from './pre-push-cfn-processor/cfn-pre-processo
...
 
 const spinner = ora('Updating resources in the cloud. This may take a few minutes...');
-const rootStackFileName = 'rootStackTemplate.json';
-const nestedStackFileName = 'nested-cloudformation-stack.yml';
+
 const optionalBuildDirectoryName = 'build';
 const cfnTemplateGlobPattern = '*template*.+(yaml|yml|json)';
 const parametersJson = 'parameters.json';
+export const rootStackFileName = 'rootStackTemplate.json';
+export const nestedStackFileName = FeatureFlags.getBoolean('overrides.project')
```
kaustavghosh06, line 65: even without feature flag - can we change this file to 'root-cloudformation-stack.json' please? It won't be a breaking. change as this is a runtime file and not checked in.

packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -222,7 +230,7 @@ export async function run(context: $TSContext, resourceDefinition: $TSObject) {
 
         // generate nested stack
         const backEndDir = pathManager.getBackendDirPath();
-        const nestedStackFilepath = path.normalize(path.join(backEndDir, providerName, nestedStackFileName));
+        const nestedStackFilepath = path.normalize(path.join(backEndDir, providerName, rootStackFileName));
```
kaustavghosh06, line 233: change var name to rootStackFilePath?

packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -266,6 +274,11 @@ export async function run(context: $TSContext, resourceDefinition: $TSObject) {
 
         try {
           await updateCloudFormationNestedStack(context, nestedStack, resourcesToBeCreated, resourcesToBeUpdated);
+          if (FeatureFlags.getBoolean('Overrides.project')) {
+            storeRootStackTemplate(context, nestedStack);
+            // if the only root stack updates, function is called with empty resources . this fn copies amplifyMeta and backend Config to #current-cloud-backend
+            context.amplify.updateamplifyMetaAfterPush([]);
```
kaustavghosh06, line 280: why do we need this?

packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -667,12 +680,11 @@ async function updateCloudFormationNestedStack(
...
 ) {
   const backEndDir = pathManager.getBackendDirPath();
-  const nestedStackFilepath = path.normalize(path.join(backEndDir, providerName, nestedStackFileName));
-
-  JSONUtilities.writeJson(nestedStackFilepath, nestedStack);
-
-  const transformedStackPath = await preProcessCFNTemplate(nestedStackFilepath);
-
+  const projectRoot = pathManager.findProjectRoot();
+  const nestedStackFilePath = path.join(pathManager.getRootStackDirPath(projectRoot), nestedStackFileName);
```
kaustavghosh06, line 684: change var to rootStackFilePath

packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -822,15 +834,39 @@ async function formNestedStack(
   serviceName?: string,
   skipEnv?: boolean,
 ) {
-  const initTemplateFilePath = path.join(__dirname, '..', 'resources', rootStackFileName);
-  const nestedStack = JSONUtilities.readJson<Template>(initTemplateFilePath);
+  let rootStack;
+  if (FeatureFlags.getBoolean('overrides.project')) {
```
kaustavghosh06, line 838: Don't think we even need this check for FF for root-stack since right now its a transient runtime file which is not checked in. So with this change, we would be placing this root-stack in a build sub-folder and uploading it to current-cloud..which is fine since it's not a breaking change.

packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -822,15 +834,39 @@ async function formNestedStack(
...
+      AuthRoleName: rootStack.Resources.AuthRole.Properties.RoleName,
+      UnauthRoleName: rootStack.Resources.UnauthRole.Properties.RoleName,
+    };
+    // sanitize this data if needed
+    for (const key of Object.keys(metaToBeUpdated)) {
+      if (typeof metaToBeUpdated[key] === 'object' && 'Ref' in metaToBeUpdated[key]) {
+        delete metaToBeUpdated[key];
+      }
+    }
+    // update amplify meta with updated root stack Info
```
kaustavghosh06, line 855: we would need to update team-provider-info.json too?

packages/amplify-cli/src/commands/build-override.ts
akshbhu: this is service parameter, invoke plugin requires it in its implementation

packages/amplify-cli/src/extensions/amplify-helpers/root-stack-status.ts
akshbhu: this for build inside build folder when we apply cfnmodifiers

packages/amplify-provider-awscloudformation/src/commands/awscloudformation/override.ts
akshbhu: I it fine to keep the old cfn file in their project when they migrate root stack?

packages/amplify-provider-awscloudformation/src/initializer.ts
akshbhu:
```
  getRootStackDirPath = (projectPath: string): string => {
    return this.constructPath(projectPath, [
      PathConstants.AmplifyDirName,
      PathConstants.BackendDirName,
      PathConstants.ProviderName,
      PathConstants.CfnStacksBuildDirName,
    ]);
  };
```

last is the build folder

packages/amplify-provider-awscloudformation/src/root-stack-builder/stackSynthesizer.ts
akshbhu: it depends since storage isnt using it and API also

packages/amplify-provider-awscloudformation/src/utility-functions.js
akshbhu: reapeat call is for root stack when no resources are provided

packages/amplify-provider-awscloudformation/src/push-resources.ts
akshbhu: if there is only root stack override , then copying backend resources to current cloud backend, if resources are there then it will get copied again

packages/amplify-provider-awscloudformation/src/push-resources.ts
akshbhu: ill check this out in migration test for this

packages/amplify-provider-awscloudformation/src/push-resources.ts
akshbhu: Nice catch

packages/amplify-provider-awscloudformation/src/index.ts
akshbhu: this wasnt working for me when importing it , that's why I exported it twice, 
will check in next commit for this

packages/amplify-provider-awscloudformation/src/initializer.ts
akshbhu: one is init case -> no template is there since we are generating it

one is push case -> as we are adding nested resources on root stack

packages/amplify-cli/src/commands/build-override.ts
kaustavghosh06: I think it's category and then resource-name which is expected. Just make a comment out here for maintainability as to why we pass in undefined in this case.

packages/amplify-cli/src/extensions/amplify-helpers/root-stack-status.ts
kaustavghosh06: but including this would exclude the top level build and the inner build folder correct?

packages/amplify-provider-awscloudformation/src/commands/awscloudformation/override.ts
kaustavghosh06: It would be safer to just remove the file and not the entire folder.

packages/amplify-provider-awscloudformation/src/root-stack-builder/stackSynthesizer.ts
kaustavghosh06: But auth is using it correct? We should standardize if we should use _toCloudFormation or this method.

packages/amplify-provider-awscloudformation/src/initializer.ts
kaustavghosh06: getRootStackDirPath shouldn't append the build dir or re-name the function to "getRootStackBuildDirPath"

packages/amplify-provider-awscloudformation/src/initializer.ts
kaustavghosh06: so template is always undefind? Both in init and push? Then why do we need the condition?

akshbhu: For push template is defined with nested stack resources as we are adding in formNestedStack function

for init its not defined yet , we ned to create it using transform

packages/amplify-provider-awscloudformation/src/utility-functions.js
akshbhu: if resource is passed , then it will build only that resource
if not , then it will build root stack

packages/amplify-cli/src/extensions/amplify-helpers/root-stack-status.ts
akshbhu: I am getting root stack changes inside build also

packages/amplify-provider-awscloudformation/src/root-stack-builder/stackSynthesizer.ts
kaustavghosh06: Discussed offline that we'd be using synthesizer and consolidating it in one place as a part of the auth PR (which is coming up next).

packages/amplify-cli/src/extensions/amplify-helpers/root-stack-status.ts
akshbhu: do a build when there is a difference in override.ts file

kaustavghosh06 (approved):

PR #8277, opened 2021-09-25
fix(amplify-appsync-simulator): none data source type definition and add test

tinovyatkin:
#### Description of changes

This PR fixes type definition for NONE data source - in type definition we were expecting `None` string while in the actual implementation it was `NONE` string.

To fix that error an enum was added and used internally, and string (produced from enum by TS 4.1) is also allowed externally.

Added test covers touched functionality and a little bit more (see codecov comment below).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are or added

### DIFF
--- a/packages/amplify-appsync-simulator/src/__tests__/__helpers__/appsync-client.ts
+++ b/packages/amplify-appsync-simulator/src/__tests__/__helpers__/appsync-client.ts
@@ -1,0 +1,104 @@
+/**
+ * AppSync GraphQL client for testing AmplifyAppSyncSimulator
+ */
+
+import * as http from 'http';
+import type { GraphQLError } from 'graphql';
+import { AmplifyAppSyncSimulator, AmplifyAppSyncSimulatorAuthenticationType } from '../../';
+import jwt from 'jsonwebtoken';
+
+/**
+ * Minimal gql tag just for syntax highlighting and Prettier while writing client GraphQL queries
+ */
+export const gql = (chunks: TemplateStringsArray, ...variables: unknown[]): string =>
+  chunks
+    .reduce((accumulator, chunk, index) => `${accumulator}${chunk}${index in variables ? variables[index] : ''}`, '')
+    .replace(/^\s+|\s$/g, '');
+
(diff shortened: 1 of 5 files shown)


tinovyatkin:
@yuth can you please take a look? It's a real bug preventing us writing tests in TypeScript with NONE data source. Tests also never hurts.

packages/amplify-appsync-simulator/src/__tests__/__helpers__/appsync-client.ts
```diff
@@ -0,0 +1,104 @@
...
+        },
+        'mockSecret',
+        { issuer: `https://cognito-idp.mock-region.amazonaws.com/mockUserPool` },
+      );
+      break;
+
+    case AmplifyAppSyncSimulatorAuthenticationType.AWS_IAM:
+    default:
+      // doing just enough to cheat amplify-appsync-simulator/src/utils/auth-helpers/current-auth-mode.ts
+      headers.Authorization = 'AWS4-HMAC-SHA256-mock-signature';
```
yuth, line 70: There has been [slight update to Authorization header](https://github.com/aws-amplify/amplify-cli/blob/b3ca83b9bec986f0fd525d46738b277eb93e4384/packages/amplify-appsync-simulator/src/utils/auth-helpers/helpers.ts#L44-L62) is validated. A [single AccessKeyId](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-appsync-simulator/src/index.ts#L29) is used to determine if the current request gets auth or unAuth role.

I don't see this being problem now as the schema that is under test does not use IAM auth but this is something we might run into later

yuth (approved):
Thank you for the PR. One nitpicky comment about IAM auth

packages/amplify-appsync-simulator/src/__tests__/__helpers__/appsync-client.ts
tinovyatkin: Thank you, improved here!

PR #8291, opened 2021-09-28
feat(graphql-default-value-transformer): implemented default value directive

lazpavel:
#### Description of changes
- implemented transformer for @default directive to allow setting default values in the graphql schema
 
- added new tests
- ran `yarn test`
- manual testing by creating an app using `"useexperimentalpipelinedtransformer": true,`

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -0,0 +1,136 @@
...
+  try {
+    new URL(x);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
+
+const validateAwsIpAddress = (x: string) => {
+  const IPV4_REGEX =
```
cjihrig, line 108: Node's `net` module has `isIP()`, `isIPv4()`, and `isIPv6()`. Maybe you could leverage those instead?

packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
```diff
@@ -0,0 +1,162 @@
...
+    throw new InvalidDirectiveError('The @default directive may only be added to object definitions annotated with @model.');
+  }
+}
+
+function assertFieldTypeAndDefaultValueTypeMatch(
+  ctx: TransformerSchemaVisitStepContextProvider,
+  config: DefaultValueDirectiveConfiguration,
+): void {
+  // if field type is non-nullable, ensure value is not null
+  if (config.value == null) {
```
cjihrig, line 144: Unless you're using `==` to check for both `null` and `undefined` simultaneously, please stick with `===`. You're using `==` elsewhere in this PR.

packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
```diff
@@ -0,0 +1,162 @@
...
+      return null;
+    }
+
+    return ctx.resolvers.getResolver(objectName, resolverLogicalId) ?? null;
+  };
+
+  private addSnippetToResolverSlot = (resolver: TransformerResolverProvider, snippets: string[]): void => {
+    const res = resolver as any;
+    res.addToSlot(
+      'postAuth',
```
cjihrig, line 104: Should default values happen before auth?

cjihrig:
Also, can you add an e2e test.

packages/amplify-graphql-default-value-transformer/src/validators.ts
lazpavel: this is the same validation we do in the app-sync-simulator, wanted to keep consistent, if you still think it's better to update I can do that

cjihrig: I'm not sure why it's like that in the simulator, but I definitely would prefer to leverage built in functionality than to ship our own regular expressions for this.

packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
```diff
@@ -0,0 +1,166 @@
...
+      this.directiveMap.set(parent.name.value, []);
+    }
+
+    this.directiveMap.get(parent.name.value)!.push(config);
+  };
+
+  generateResolvers = (ctx: TransformerContextProvider): void => {
+    const context = ctx as TransformerContextProvider;
+
+    for (let typeName of this.directiveMap.keys()) {
```
cjihrig, line 60: Can this be `const` instead of `let`?

packages/amplify-graphql-default-value-transformer/src/graphql-default-value-transformer.ts
```diff
@@ -0,0 +1,166 @@
...
+      const snippets: string[] = [];
+      for (const config of this.directiveMap.get(typeName)!) {
+        const fieldName = config.field.name.value;
+
```
cjihrig, line 64: Please remove the empty line.

packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -0,0 +1,132 @@
+import { isValidNumber } from 'libphonenumber-js';
+import  * as net from 'net';
```
cjihrig, line 2:
```suggestion
import * as net from 'net';
```

packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -0,0 +1,132 @@
...
+const validateInt = (x: string) => !Number.isNaN(parseInt(x, 10));
+const validateFloat = (x: string) => !Number.isNaN(parseFloat(x));
+const validateAwsDate = (x: string) => validateDate(x);
+const validateAwsTime = (x: string) => validateTime(x);
+const validateAwsDateTime = (x: string) => validateDateTime(x);
+const validateAwsTimestamp = (x: string) => !Number.isNaN(parseInt(x, 10));
+const validateAwsPhone = (x: string) => isValidNumber(x);
+
+const validateTime = (time: string): boolean => {
+  const TIME_REGEX = /^([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])(\.\d{1,})?(([Z])|([+|-]([01][0-9]|2[0-3]):[0-5][0-9]))$/;
```
cjihrig, line 14: Can you move this out of the function so the same regex object can be reused.

packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -0,0 +1,132 @@
...
+  return TIME_REGEX.test(time);
+};
+
+const leapYear = (year: number): boolean => {
+  return (year % 4 === 0 && year % 100 !== 0) || year % 400 === 0;
+};
+
+const validateDate = (date: string): boolean => {
+  const RFC_3339_REGEX = /^(\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01]))$/;
```
cjihrig, line 23: Same comment here.

packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -0,0 +1,132 @@
...
+      }
+      break;
+  }
+
+  return true;
+};
+
+const validateDateTime = (dateTime: string): boolean => {
+  const RFC_3339_REGEX =
+    /^(\d{4}-(0[1-9]|1[012])-(0[1-9]|[12][0-9]|3[01])T([01][0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9]|60))(\.\d{1,})?(([Z])|([+|-]([01][0-9]|2[0-3]):[0-5][0-9]))$/;
```
cjihrig, line 58: And here

packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -0,0 +1,132 @@
...
+
+  // Validate the structure of the date-string
+  if (!RFC_3339_REGEX.test(dateTime)) {
+    return false;
+  }
+
+  // Check if it is a correct date using the javascript Date parse() method.
+  const t = Date.parse(dateTime);
+  // eslint-disable-next-line
+  if (t !== t) {
```
cjihrig, line 68: Is this a check for `NaN`? If so, we should use `Number.isNaN()`.

packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -0,0 +1,132 @@
...
+  // Split the date-time-string up into the string-date and time-string part.
+  // and check whether these parts are RFC 3339 compliant.
+  const index = dateTime.indexOf('T');
+  const date = dateTime.slice(0, index);
+  const time = dateTime.slice(index + 1);
+  return validateDate(date) && validateTime(time);
+};
+
+const validateBoolean = (x: string): boolean => {
+  const BOOL_REGEX = /^(true|false)$/i;
```
cjihrig, line 80: I'll stop calling out the regexes that can be hoisted out of the function. Just don't hoist anything with a `g` flag.

packages/amplify-graphql-default-value-transformer/src/validators.ts
```diff
@@ -0,0 +1,132 @@
...
+  try {
+    new URL(x);
+    return true;
+  } catch (e) {
+    return false;
+  }
+};
+
+const validateAwsIpAddress = (x: string) => {
+  return net.isIPv4(x) || net.isIPv6(x);
```
cjihrig, line 109:
```suggestion
  return net.isIP(x) !== 0;
```

cjihrig (approved):
Left a few more comments. Not a huge fan of duplicating all of this functionality from the npm package, but other than that, LGTM

PR #8364, opened 2021-10-05
chore: ddb walkthrough refactor and override tests

kaustavghosh06:
tests tied to this PR - https://github.com/aws-amplify/amplify-cli/pull/8325


packages/amplify-e2e-tests/package.json
```diff
@@ -44,7 +44,8 @@
     "yargs": "^15.1.0"
   },
   "devDependencies": {
-    "ts-node": "^8.9.0"
+    "ts-node": "^8.9.0",
+    "jest": "^27.2.4"
```
jhockett, line 48: The CLI depends on jest "^26.6.3" in the root package.json. We should keep these consistent.

jhockett: While we're at it, we have 3 different ts-node versions defined in package.json. This could at least be updated to use `8.10.2`

akshbhu (approved):
LGTM!!

sachscode (approved):
LGTM

packages/amplify-e2e-tests/package.json
kaustavghosh06: Done.

packages/amplify-e2e-tests/overrides/override-storage-ddb.ts
```diff
@@ -0,0 +1,6 @@
+export function overrideProps(props: any) {
```
jhockett, line 1: nit: use a better type for `props`. Is there a CDK or Cloudform type we can use? If not, can at least use $TSObject

```suggestion
import { $TSObject } from 'amplify-cli-core';

export function overrideProps(props: $TSObject) {
```

packages/amplify-e2e-tests/src/__tests__/storage.test.ts
```diff
@@ -1,7 +1,10 @@
+import { JSONUtilities } from 'amplify-cli-core';
```
jhockett, line 1: if possible, we should rebase against the main branch before making these changes. The storage tests have been split up into 3 files.

packages/amplify-e2e-tests/overrides/override-storage-ddb.ts
kaustavghosh06: We'd be exporting these prop types to a new package - amplify-cli-helpers (which would be published in the future to npm)

Issue #8351, opened 2021-10-04
Amplify env checkout falls out of sync when a resource is updated in another environment


phani-srikar:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v14

### Amplify CLI Version

6.1.1

### What operating system are you using?

MacOS

### Amplify Categories

auth

### Amplify Commands

env

### Describe the bug

**Although `auth` category resource is used as an example, the following issue is applicable to any category resource that stores updatable configuration in `amplify-meta.json`.** 
I have 2 environments in my Amplify JS App (say `dev` and `test`). I add and push an Auth resource in my `dev` env with default configuration. Now, I checkout the `test` environment using `amplify env checkout test`, update the Auth resource's MFA configuration (for example) and push the modified resource in `test` environment. 
Now when I switch back to the `dev` environment using `amplify env checkout dev`, `amplify status` shows that the Auth resource has updates to push (I confirm this is correct as the `parameters.json` file for that resource has updated configuration inputs to the stack). However, the `amplify-meta` file does not have the updated resource configuration. Hence, after I run `amplify push` in `dev`, the Auth resource stack provisions the updated stack but neither `aws-exports` nor `amplify-meta` show the correct updated configuration. 

Digging a little into why `env checkout` does not fetch the resource configuration updates into `amplify-meta`, I see that there is a file called `backend-config` which is not updated during `amplify update <category>` commands. `amplify env checkout` might be using that file to initialize the `amplify-meta` during env switching, hence making it fall out of sync. 

### Expected behavior

Both `amplify-meta` and `aws-exports` should reflect the changes in resource configuration made in another environment after running `amplify push`. 

### Reproduction steps

1. `amplify init` in a JS app - choose env name as `dev`. Run `amplify push`.
2. `amplify env add test`, `amplify env checkout test` followed by `amplify push` in `test` environment. 
3. Switch back to dev environment using `amplify env checkout dev`.
[...]

Issue #8286, opened 2021-09-27
Amplify push in a new environment is not updating the `aws-exports` file


phani-srikar:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v14

### Amplify CLI Version

6.0.1

### What operating system are you using?

MacOS

### Amplify Categories

Not applicable

### Amplify Commands

add, env, push

### Describe the bug

Running `amplify push` in a new environment after `amplify env add <env-name>` does not update the `aws-exports` file.

### Expected behavior

`amplify push` should update the `aws-exports` file.

### Reproduction steps

1. Run `amplify init` in a JS project, followed by `amplify add storage`
2. Run `amplify push` and check the `aws-exports` file is updated correctly.
3. Add a new environment using `amplify env add test` and run `amplify push`
4. The `aws-exports` file is not updated with correct bucket name, it looks like "aws_user_files_s3_bucket": "<name>-dev", instead of "aws_user_files_s3_bucket": "<name>-test".

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

josefaidt:
Hey @phani-srikar :wave: thanks for raising this! I was able to successfully reproduce by following the provided steps. Marking this as a bug 🙂 

As a workaround we can run `amplify env checkout <current-env-name>` to update the `aws-exports` file.

lazpavel:
Unable to reproduce with latest CLI. Please reopen if you still experience this issue after upgrading.

PR #8373, opened 2021-10-06
Geo category plugin - support for additional regions

phani-srikar:
#### Description of changes

Convert the Amazon Location Service CFN native resource constructs to CFN Custom resource constructs.
This adds support for regions that are not supported by Amazon Location Service but supported by Amplify CLI.

Increase stability for e2e tests:
`amplify-prompts` automatically makes the selection for the user when there is only one choice in the selection set.
This could interfere with the `nspawn` process chain of commands defined in the e2e test setup.
We are avoiding such cases by having multiple geo resources in the e2e tests.

Tested with JS sample app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-geo/src/__tests__/service-stacks/__snapshots__/mapStack.test.ts.snap
+++ b/packages/amplify-category-geo/src/__tests__/service-stacks/__snapshots__/mapStack.test.ts.snap
@@ -10,6 +10,14 @@ Object {
     },
   },
   "Outputs": Object {
+    "Arn": Object {
+      "Value": Object {
+        "Fn::GetAtt": Array [
+          "CustomMap",
+          "MapArn",
+        ],
+      },
+    },
     "Name": Object {
       "Value": Object {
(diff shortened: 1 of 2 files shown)


phani-srikar:
> Added some comments and change requests.
> 
> Is `dependsOn` managed between `geo` and `auth` resources?
> 
> Also can we have multi-environment and amplify console pull e2e tests as well to make sure it will function correctly?

We do not need a `dependsOn` relationship with Auth as none of Geo stacks currently reference the Auth category outputs. 
We noticed a couple of issues as noted in https://github.com/aws-amplify/amplify-cli/issues/8351 and https://github.com/aws-amplify/amplify-cli/issues/8286 for multi-env flows. We are in the process of working around them and adding multi-env e2e tests in this sprint. However, CLI additionally, runs the current e2e tests in regions that weren't previously supported by Geo and this way we can be sure that these changes are good in addition to the manual testing we did using sample apps.

PR #8542, opened 2021-10-25
ci: prevent setup from timing out during yarn installation

johnpc:
#### Description of changes

`yarn` installation takes a good amount of time, and CCI's continuation API only gives us a short window while continuations are valid. See example: https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/3482/workflows/0c801c30-0e20-4eb4-8f61-39140494ba20/jobs/60831

```
{"message":"Key has expired."}
```

This PR only installs the minimum dependencies necessary to run `split-e2e-tests` to speed up the continuation process and prevent key expiration.

Ran these commands locally, then `circleci config validate .circleci/generated_config.yml`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


edwardfoyle (commented):
could we add these dependencies to a `scripts/package.json` file and then cd into that dir and run yarn before running the script? That seems like a less obfuscated way to maintain these deps

ammarkarachi (approved):

PR #8620, opened 2021-11-01
feat: add get project info helper to overrides helper package

kaustavghosh06:
Adding getProjectInfo() helper function to the helpers package which would provide override and custom resource functions access to:
1. Project name
2. Current env name

This will enable customers to have conditional logic in code(override or custom-resources) per env.

### DIFF
--- a/packages/amplify-cli-overrides-helper/src/helpers/project-info.ts
+++ b/packages/amplify-cli-overrides-helper/src/helpers/project-info.ts
@@ -1,0 +1,14 @@
+import { stateManager } from 'amplify-cli-core';
+import { AmplifyProjectInfo } from '../types';
+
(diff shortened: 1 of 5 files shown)


akshbhu (approved):
LGTM

packages/amplify-cli-overrides-helper/src/helpers/project-info.ts
```diff
@@ -0,0 +1,14 @@
+import { stateManager } from 'amplify-cli-core';
+import { AmplifyProjectInfo } from '../types';
+
+export function getProjectInfo(): AmplifyProjectInfo {
```
akshbhu, line 4: nit: 
```suggestion
export const  getProjectInfo = ( ): AmplifyProjectInfo  => {
```

PR #8660, opened 2021-11-04
test: add @http migration test

cjihrig:
#### Description of changes
This commit adds a test that pushes an `@http` schema with gql transformer v1, updates the transformer to v2, and pushes again.


lazpavel (approved):
LGTM, does it make sense to add at lease some queries in the test?

cjihrig:
> does it make sense to add at lease some queries in the test?

I didn't because that would make the test rely on amazon.com being available. The GraphQL E2E tests actually deploy a server to communicate with, but it seemed like overkill (and duplication) for the purposes of this test.

PR #8695, opened 2021-11-05
feat: overrides uniformity accross all the categories

kaustavghosh06:
1. Rename "overrideProps" to "override"
2. Rename "props" parameter to "resources"
3. Remove "return props" requirement and test that it doesn't crash
4. "✔ Do you want to edit override.ts file now? (y/N)" should default to "true"
5. Support for: amplify build -> from amplify build-override
6. Changes are applied to all categories
7. use vm2 uniformly


akshbhu (approved):
Looks Good!

PR #8703, opened 2021-11-06
fix(cli): rewire override project to override rootStack internally

ammarkarachi:
Test passing

 PASS  src/__tests__/configure-project.test.ts (153.965 s)
  amplify configure project tests
    ✓ should update the project to use access keys (12781 ms)
    ✓ should update the project to use a profile (12998 ms)
    ✓ should update the project to remove a profile (12539 ms)
    ✓ should update the project to add access keys when configLevel is general (12626 ms)
    ✓ should update the project to use a profile (12653 ms)

 PASS  src/__tests__/containers-api.test.ts (2270.977 s)
  amplify api add
  amplify api add
    ✓ init project, enable containers and add multicontainer api (1002251 ms)
    ✓ init project, enable containers and add multicontainer api push, edit and push (1259656 ms)

 PASS  src/__tests__/custom_policies_container.test.ts (914.739 s)
  ✓ should init and deploy a api container, attach custom policies to the Fargate task (912354 ms)

### DIFF
--- a/packages/amplify-cli/src/input-manager.ts
+++ b/packages/amplify-cli/src/input-manager.ts
@@ -10,7 +10,7 @@ export function getCommandLineInput(pluginPlatform: PluginPlatform): Input {
   /* tslint:disable */
   if (result.argv && result.argv.length > 2) {
     let index = 2;
-
+    aliasArgs(result.argv);
     // pick up plugin name, allow plugin name to be in the 2nd or 3rd position
     const pluginNames = getAllPluginNames(pluginPlatform);
 
@@ -192,4 +192,10 @@ export function verifyInput(pluginPlatform: PluginPlatform, input: Input): Input
   }
 
   return result;
+}
+
+function aliasArgs(argv: string[]) {
+  if (argv.length >= 4 && argv[2] === 'override' && argv[3] === 'project') {
+    argv[3] = 'root';
+  }
 }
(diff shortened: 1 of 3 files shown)


sachscode (approved):

packages/amplify-e2e-core/src/categories/api.ts
```diff
@@ -719,7 +719,7 @@ export function rebuildApi(projDir: string, apiName: string) {
 export function addRestContainerApiForCustomPolicies(projectDir: string, settings: { name: string }) {
   return new Promise<void>((resolve, reject) => {
     spawn(getCLIPath(), ['add', 'api'], { cwd: projectDir, stripColors: true })
-      .wait('Please select from one of the below mentioned services:')
+      .wait('Select from one of the below mentioned services:')
```
kaustavghosh06, line 722: I think we reverted this change? Can you confirm once

packages/amplify-cli/src/input-manager.ts
```diff
@@ -193,3 +193,9 @@ export function verifyInput(pluginPlatform: PluginPlatform, input: Input): Input
 
   return result;
 }
+
+function aliasArgs(argv: string[]) {
+  if (argv.length >= 4 && argv[2] === 'override' && argv[3] === 'project') {
+    argv[3] = 'rootStack';
```
kaustavghosh06, line 199: you can just use "root" - that should work as is since it's an alias already defined.

packages/amplify-e2e-core/src/categories/api.ts
ammarkarachi: I see it in extOverrides

packages/amplify-cli/src/input-manager.ts
ammarkarachi: I thought root is for something else if it's only for override then I can change it

packages/amplify-e2e-core/src/categories/api.ts
ammarkarachi: image

kaustavghosh06 (approved):

PR #8709, opened 2021-11-07
feat: consolidate dependent resource types in backend/

kaustavghosh06:
1. Moved "amplify-dependent-resources-ref.d.ts" to amplify/backend/types folder instead of having multiple copies in each of the custom resource folder.
2. Generate "amplify-dependent-resources-ref.d.ts" as a part of amplify build and amplify push commands.

### DIFF
--- a/packages/amplify-category-custom/src/utils/dependency-management-utils.ts
+++ b/packages/amplify-category-custom/src/utils/dependency-management-utils.ts
@@ -115,7 +115,7 @@ export function addCDKResourceDependency(
   dependentResources: AmplifyDependentResourceDefinition[],
 ) {
   const dependsOn: AmplifyDependentResourceDefinition[] = [];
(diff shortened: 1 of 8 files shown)


packages/amplify-category-custom/src/utils/dependency-management-utils.ts
```diff
@@ -115,7 +115,7 @@ export function addCDKResourceDependency(
   dependentResources: AmplifyDependentResourceDefinition[],
 ) {
   const dependsOn: AmplifyDependentResourceDefinition[] = [];
-  const dependentParameters: $TSObject = {};
+  const dependentParameters: any = {};
```
akshbhu, line 118: nit : can revert to $TSObject or $TSAny

akshbhu (approved):
LGTM

PR #8714, opened 2021-11-07
fix: override http datasource logical IDs

cjihrig:
#### Description of changes
This commit overrides the logical IDs of http datasources
created by the gql v2 transformer to prevent issues when
migrating from v1. It also renames the `@http` stack to be
consistent with v1.


lazpavel (approved):
LGTM

PR #8716, opened 2021-11-08
feat: change override.ts file to override.ts.sample in resources/ for pkg CLI to work

kaustavghosh06:
change override.ts file to override.ts.sample in resources/ for pkg CLI to work


packages/amplify-cli-core/src/overrides-manager/override-skeleton-generator.ts
```diff
@@ -94,6 +94,6 @@ export const generateTsConfigforProject = (backendDir: string, srcResourceDirPat
   // ensure build dir path
   fs.ensureDirSync(path.join(destDirPath, 'build'));
   const resourceTsConfigFileName = path.join(destDirPath, 'build', 'tsconfig.resource.json');
-  fs.writeFileSync(overrideFileName, fs.readFileSync(path.join(srcResourceDirPath, 'override.ts')));
+  fs.writeFileSync(overrideFileName, fs.readFileSync(path.join(srcResourceDirPath, 'override.ts.sample')));
```
akshbhu, line 97: Do we need  strip off sample right when we copy this to customers folder?

kaustavghosh06: Nope overrideFileName is override.ts

akshbhu (approved):

PR #8719, opened 2021-11-08
fix: remove depending on conditional output variables from stacks

kaustavghosh06:


packages/amplify-category-custom/src/utils/dependency-management-utils.ts
```diff
@@ -58,7 +58,17 @@ export function getResourceCfnOutputAttributes(category: string, resourceName: s
...
+      let outputsWithoutConditions: any = {};
+
+      for (const key in allOutputs) {
+        if (!allOutputs[key]['Condition']) {
+          // Filter out outputs which are conditional to avoid deployment failures
+          outputsWithoutConditions[key] = allOutputs[key];
+        }
+      }
+
+      return Object.keys(outputsWithoutConditions) as [string?];
```
akshbhu, line 71: shoudnt this be string[]?

akshbhu (approved):

PR #8736, opened 2021-11-08
fix: move @model params to root stack and fix ds logical id

SwaySway:
#### Description of changes
- moved `@model` params to root stack
- updated sync table logical id
- removed table encryption should be using the default with SSE enabled but it's setting it to false. CFN still sets it to true so removing it to match v1 behavior.
- updated migration unit tests

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
+++ b/packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
@@ -1089,4 +1089,84 @@ describe('ModelTransformer: ', () => {
 
     expectFieldsOnInputType(updateTodoInput!, ['name']);
   });
+
+  it('should add the model parameters at the root sack', () => {
+    const validSchema = `type Todo @model {
+      name: String
+    }`;
+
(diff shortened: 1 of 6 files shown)


packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
```diff
@@ -1089,4 +1089,51 @@ describe('ModelTransformer: ', () => {
 
     expectFieldsOnInputType(updateTodoInput!, ['name']);
   });
+
+  it('should the model parameters at the root sack', () => {
```
cjihrig, line 1093: Typo? "should add the model..."

cjihrig (approved):

packages/amplify-graphql-model-transformer/src/__tests__/model-transformer.test.ts
SwaySway: whoops updating the test name

cjihrig (approved):

PR #8752, opened 2021-11-09
fix: update custom cdk seleton templates + format override skeleton file

kaustavghosh06:
1. Formats override skeleton ts file
2. Add more details and examples in cdk sample to include env agnostic examples
3. Rename cdk-stack.ts to cdk-stack.ts.sample in resource/ for pkg CLI to bundle it as an asset
4. Add a type for AmplifyResourceProps which are passed to the custom resources function


packages/amplify-cli-extensibility-helper/package.json
```diff
@@ -35,7 +35,7 @@
     "@aws-cdk/aws-lambda": "~1.124.0",
     "@aws-cdk/aws-iam": "~1.124.0",
     "@aws-amplify/amplify-category-custom": "1.0.0",
-    "amplify-cli-core": "1.31.1"
+    "amplify-cli-core": "~1.124.0"
```
sachscode, line 38: Why are we relying on an older version of cli-core ?

sachscode (commented):
LGTM

sachscode (approved):

PR #7092, opened 2021-04-14
fix: removing sts call for usage data

ammarkarachi:
#### Description of changes
Removed sts call for usage-data 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #7090, opened 2021-04-14
Being repeatedly asked for which Auth method I want to use for each push


PeteDuncanson:
### Before opening, please confirm:

I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

yarn

### Amplify CLI Version

4.48

### What operating system are you using?

Windows

### Amplify Categories

Not applicable

### Amplify Commands

push

### Describe the bug

I was logged in using AdminUI but I had a permissions issue. So I had to logout and insteead authenticate with Amplify using IAM keys. Now when every I When I do a push I get asked which Authentication method I want to use. This is a bit of a drag.

### Expected behavior

Just use the last setting from the last push?

### Reproduction steps

Authenticate with AdminUI 
`amplify logout` and select the nukeclear option of "log me out of everything"
`amplify init`
follow the walk through to use `AWS Access Keys`
`amplify push`
Get the menu asking which auth method you want to use, this appears everytime.

jhockett:
Hi @PeteDuncanson, can you share your `<project-root>/amplify/.config/local-aws-info.json`?

PeteDuncanson:
`{
  "localdev": {
    "configLevel": "amplifyAdmin"
  }
}`

Not much is it... :)

PeteDuncanson:
@jhockett seems its showing me that menu for most commands (update, add, etc) not just push if that helps.

PeteDuncanson:
Had a dig around with a team mate comparing configs and seems I'm missing a few lines. Added this in there instead:

{
  "localdev": {
    "configLevel": "project",
    "useProfile": false,
    "awsConfigFilePath": "C:\\Users\\pete\\.amplify\\awscloudformation\\6wH9YnAEwC"
  }
}

And now I no longer get asked for auth choices which is great. However...the question still remains...how did it get in that state in the first place? :)

kaustavghosh06:
@PeteDuncanson We've fixed the issue with the latest version of the CLI - v4.49.0. Please let us know if you're still seeing this with the latest version of the CLI. 
[...]

PR #8754, opened 2021-11-09
feat: add accountid and arbitrary additional info in usagedata

johnpc:
#### Description of changes

Previously, we sent `accountId` along with all usageData reports. This was reverted (https://github.com/aws-amplify/amplify-cli/pull/7092) in an effort to mitigate https://github.com/aws-amplify/amplify-cli/issues/7090. 

However, I have re-implemented this change and I am unable to reproduce the customer issue that caused us to revert (again, https://github.com/aws-amplify/amplify-cli/issues/7090)

This PR re-adds the accountId to our usageData statistics. It also adds a new property to our usage data, `record` that allows us to include arbitrary additional data fields to our payloads.

- Wrote a plugin that uses the new `addRecord` functionality
- Attempted to reproduce the failure described in https://github.com/aws-amplify/amplify-cli/issues/7090, was unable to
- Interacted with projects in a sandbox environment, including without any credentials set in `~/.aws/credentials`. No errors or unexpected auth prompts.
- Interacted with projects in a non-sandbox, authenticated environment. No errors or unexpected auth prompts.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
+++ b/packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
@@ -18,6 +18,7 @@ export class UsageDataPayload {
   isCi: boolean;
   accountId: string;
   projectSetting: ProjectSettings;
+  record: Record<string, any>[];
   constructor(
     sessionUuid: string,
     installationUuid: string,
@@ -28,6 +29,7 @@ export class UsageDataPayload {
     accountId: string,
     project: ProjectSettings,
     inputOptions: InputOptions,
+    record: Record<string, any>[],
   ) {
     this.sessionUuid = sessionUuid;
     this.installationUuid = installationUuid;
@@ -43,6 +45,7 @@ export class UsageDataPayload {
     this.isCi = ci.isCI;
     this.projectSetting = project;
     this.inputOptions = inputOptions;
+    this.record = record;
     if (error) {
       this.error = new SerializableError(error);
     }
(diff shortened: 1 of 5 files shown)


packages/amplify-provider-awscloudformation/src/index.ts
```diff
@@ -159,6 +161,7 @@ module.exports = {
   createDynamoDBService,
   resolveAppId,
   loadConfigurationForEnv,
+  getAccountId,
```
jhockett, line 164: I don't think we need to export this twice.

jhockett (commented):
Changes themselves LGTM, just some comments on the import/export

@johnpc can you ensure this works when using authentication through the Admin UI?

packages/amplify-provider-awscloudformation/src/index.ts
johnpc: Both exports were required when I called it via 

```
const awsConfigInfo = await context.amplify.invokePluginMethod(
  context,
  "awscloudformation",
  undefined,
  "getAccountId",
  [context, environmentName, appId]
);
```

But with `context.amplify.executeProviderUtils` I don't think it's necessary anymore.

ammarkarachi (approved):

packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
```diff
@@ -18,6 +18,7 @@ export class UsageDataPayload {
   isCi: boolean;
   accountId: string;
   projectSetting: ProjectSettings;
+  record: Record<string, any>[];
```
sachscode, line 21: Is this a record from accountID to log data ? Could you pl make it explicit with comments

sachscode (commented):
LGTM, just one nit

PR #8771, opened 2021-11-10
Ext overrides4

kaustavghosh06:
1) build resources on pull even with no override flag passed
2) update package versions of ext helper in skeleton package.json of resources

### DIFF
--- a/packages/amplify-category-auth/resources/overrides-resource/package.json
+++ b/packages/amplify-category-auth/resources/overrides-resource/package.json
@@ -8,7 +8,7 @@
       "test": "echo \"Error: no test specified\" && exit 1"
     },
     "dependencies": {
-      "@aws-amplify/cli-extensibility-helper": "latest"
+      "@aws-amplify/cli-extensibility-helper": "1.0.1-ext19.0"
     },
     "devDependencies": {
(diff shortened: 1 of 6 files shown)


packages/amplify-provider-awscloudformation/src/initialize-env.ts
```diff
@@ -114,4 +116,21 @@ export async function run(context: $TSContext, providerMetadata: $TSMeta) {
...
+
+  const { allResources } = await context.amplify.getResourceStatus();
+
+  allResources.forEach(resourceCreated => {
+    resourcesToBuild.push({
+      service: resourceCreated.service as string,
+      category: resourceCreated.category as string,
+      resourceName: resourceCreated.resourceName as string,
+    });
+  });
```
jhockett, line 132: nit: This simplifies with `.map()`

```suggestion
  const { allResources } = await context.amplify.getResourceStatus();

  const resourcesToBuild: IAmplifyResource[] = allResources.map(resource => ({
    service: resource.service as string,
    category: resource.category as string,
    resourceName: resource.resourceName as string,
  }));
```

jhockett (approved):

PR #8794, opened 2021-11-11
feat: exclude amplify/backend/awscloudformation from gitignore path

kaustavghosh06:
1.exclude amplify/backend/awscloudformation from gitignore path
2.add spinner when building resources

### DIFF
--- a/packages/amplify-cli/src/input-manager.ts
+++ b/packages/amplify-cli/src/input-manager.ts
@@ -4,6 +4,8 @@ import { constants } from './domain/constants';
 import { PluginPlatform } from './domain/plugin-platform';
 import { getPluginsWithName, getAllPluginNames } from './plugin-manager';
 import { InputVerificationResult } from './domain/input-verification-result';
+import { pathManager, stateManager } from 'amplify-cli-core';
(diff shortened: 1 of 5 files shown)


packages/amplify-cli/src/input-manager.ts
```diff
@@ -197,5 +199,10 @@ export function verifyInput(pluginPlatform: PluginPlatform, input: Input): Input
 function aliasArgs(argv: string[]) {
   if (argv.length >= 4 && argv[2] === 'override' && argv[3] === 'project') {
     argv[3] = 'root';
+
+    // Also update gitignore to latest list - mainly to exclude amplify/backend/awscloudformation dir from .gitingore for older projects
+    const { projectPath } = stateManager.getLocalEnvInfo();
+    const gitIgnoreFilePath = pathManager.getGitIgnoreFilePath(projectPath);
```
akshbhu, line 205: will this include build files also inside awscloudformation?

akshbhu (approved):
just one question

packages/amplify-cli/src/input-manager.ts
kaustavghosh06: Nope. build/ folders are excluded. Checked on it.

PR #8798, opened 2021-11-11
fix: add alphanumeric validation on custom resource-name

kaustavghosh06:
resource-name needs to be alphanumeric conforming to amplify resource naming patterns,

### DIFF
--- a/packages/amplify-category-custom/src/utils/common-questions.ts
+++ b/packages/amplify-category-custom/src/utils/common-questions.ts
@@ -1,4 +1,4 @@
-import { prompter } from 'amplify-prompts';
+import { alphanumeric, prompter } from 'amplify-prompts';
 import { v4 as uuid } from 'uuid';
 import { customDeploymentOptions } from '../utils/constants';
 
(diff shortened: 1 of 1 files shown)


packages/amplify-category-custom/src/utils/common-questions.ts
```diff
@@ -11,6 +11,9 @@ export async function customResourceNameQuestion(): Promise<string> {
   const [shortId] = uuid().split('-');
   const defaultResourceName = `customResource${shortId}`;
 
-  const resourceName = await prompter.input('Provide a name for your custom resource', { initial: defaultResourceName });
+  const resourceName = await prompter.input('Provide a name for your custom resource', {
+    initial: defaultResourceName,
+    validate: alphanumeric(),
```
ammarkarachi, line 16: Shouldn't this be 
```suggestion
    validate: alphanumeric,
```

ammarkarachi: Never mind it's a curry function

ammarkarachi (approved):

PR #8817, opened 2021-11-11
fix: create pipeline resolvers for @http v2

cjihrig:
#### Description of changes
This commit updates `@http` v2 to create pipeline resolvers instead of unit resolvers. This allows `@http` to be used with other directives such as `@auth`.

This commit also adds missing auth information to `@http` v2 and adds a missing `package.json` dependency to `@function` v2.

### DIFF
--- a/packages/amplify-graphql-http-transformer/src/graphql-http-transformer.ts
+++ b/packages/amplify-graphql-http-transformer/src/graphql-http-transformer.ts
@@ -1,5 +1,13 @@
-import { DirectiveWrapper, MappingTemplate, TransformerContractError, TransformerPluginBase } from '@aws-amplify/graphql-transformer-core';
+import {
+  DirectiveWrapper,
+  IAM_AUTH_ROLE_PARAMETER,
+  IAM_UNAUTH_ROLE_PARAMETER,
+  MappingTemplate,
+  TransformerContractError,
+  TransformerPluginBase,
+} from '@aws-amplify/graphql-transformer-core';
 import { TransformerContextProvider, TransformerSchemaVisitStepContextProvider } from '@aws-amplify/graphql-transformer-interfaces';
+import { AuthorizationType } from '@aws-cdk/aws-appsync';
 import * as cdk from '@aws-cdk/core';
 import {
   DirectiveNode,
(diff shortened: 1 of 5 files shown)


packages/amplify-graphql-http-transformer/src/graphql-http-transformer.ts
```diff
@@ -288,11 +324,23 @@ function createResolver(stack: cdk.Stack, dataSourceId: string, context: Transfo
           ref('util.qr($util.appendError($ctx.result.body, $ctx.result.statusCode))'),
         ),
       ),
+      `${functionId}.res.vtl`,
```
SwaySway, line 327: nit: Can we use the naming convention of `TypeName.FieldName.DataSource.request/response.vtl`?
```suggestion
      `${config.resolverTypeName}.${config.resolverFieldName}.DataResolver.res.vtl`,
```

packages/amplify-graphql-http-transformer/src/graphql-http-transformer.ts
```diff
@@ -288,11 +324,23 @@ function createResolver(stack: cdk.Stack, dataSourceId: string, context: Transfo
...
+    stack,
+  );
+
+  return context.api.host.addResolver(
+    config.resolverTypeName,
+    config.resolverFieldName,
+    MappingTemplate.inlineTemplateFromString(printBlock('Stash resolver specific context.')(compoundExpression(requestTemplate))),
+    MappingTemplate.s3MappingTemplateFromString(
+      '$util.toJson($ctx.prev.result)',
       `${config.resolverTypeName}.${config.resolverFieldName}.res.vtl`,
```
SwaySway, line 339: nit: since this is the last appsync function can we use
```suggestion
      `${config.resolverTypeName}.${config.resolverFieldName}.after.res.vtl`,
```

SwaySway (approved):
lgtm aside from nit

packages/amplify-graphql-http-transformer/src/graphql-http-transformer.ts
cjihrig: This one isn't an AppSync function - it's the resolver mapping template. I don't think we use `.after.res.vtl` anywhere else in the codebase.

PR #8843, opened 2021-11-13
fix: append apiKey if global auth is enabled and its not default auth

SwaySway:
#### Description of changes
- add global auth check on index resolvers
- for the following scenarios apply the appsync `@aws_api_key` directive when global auth is enabled on non auth models and their extended queries (ex. `@index`, `@searchable`, etc...)
- add sandbox flag into context
- update connection field function with ability to pass directive node array

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-searchable-transformer/src/definitions.ts
+++ b/packages/amplify-graphql-searchable-transformer/src/definitions.ts
@@ -8,6 +8,8 @@ import {
   DocumentNode,
   EnumTypeDefinitionNode,
   EnumValueDefinitionNode,
+  DirectiveNode,
+  NamedTypeNode,
 } from 'graphql';
 import {
   graphqlName,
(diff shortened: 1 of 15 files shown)


packages/amplify-graphql-searchable-transformer/src/definitions.ts
```diff
@@ -40,8 +46,13 @@ const STRING_CONDITIONS = ID_CONDITIONS;
 const INT_CONDITIONS = ['ne', 'gt', 'lt', 'gte', 'lte', 'eq', 'range'];
 const FLOAT_CONDITIONS = ['ne', 'gt', 'lt', 'gte', 'lte', 'eq', 'range'];
 const BOOLEAN_CONDITIONS = ['eq', 'ne'];
-import assert from 'assert';
 
+export const AGGREGATE_TYPES = [
+  'SearchableAggregateResult',
+  'SearchableAggregateScalarResult',
+  'SearchableAggregateBucketResult',
+  'SearchableAggregateBucketResultItem',
```
sundersc, line 54: should we add input aggregate types too? I don't think it is necessary, just wanted to confirm.

sundersc (approved):
LGTM. Just check whether we should add aggregate input types as well to the list.

packages/amplify-graphql-searchable-transformer/src/definitions.ts
SwaySway: they are only necessary for the return types they cannot be applied for inputs

cjihrig (approved):

danielleadams (approved):

PR #8859, opened 2021-11-14
Modify update custom flow + update cdk skeleton sample

kaustavghosh06:
1. `amplify update custom` flow improvements to allow deselection of all other Amplify resource dependencies.
2. Update cdk sample skeleton app to not use cdk.fn.join as it's not needed + fix one typo + make comments better

### DIFF
--- a/packages/amplify-category-custom/src/utils/dependency-management-utils.ts
+++ b/packages/amplify-category-custom/src/utils/dependency-management-utils.ts
@@ -177,10 +177,6 @@ function addDependsOnToResource(category: string, resourceName: string, dependsO
 }
 
 export async function addCFNResourceDependency(context: $TSContext, customResourceName: string) {
-  if (!(await prompter.yesOrNo('Do you want to access Amplify generated resources in your custom CloudFormation file?', false))) {
-    return;
-  }
-
   const selectResourcesInCategory = (
     choices: DistinctChoice<any>[],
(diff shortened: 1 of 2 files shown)


sachscode (approved):

packages/amplify-category-custom/src/utils/dependency-management-utils.ts
```diff
@@ -216,6 +212,16 @@ export async function addCFNResourceDependency(context: $TSContext, customResour
     });
   }
 
+  const isExistingResources = Object.keys(existingDependentResources).length > 0;
+
+  if (
```
sachscode, line 217: Looks like this is a Typo?

packages/amplify-category-custom/src/utils/dependency-management-utils.ts
```diff
@@ -216,6 +212,16 @@ export async function addCFNResourceDependency(context: $TSContext, customResour
...
+    !(await prompter.yesOrNo('Do you want to access Amplify generated resources in your custom CloudFormation file?', isExistingResources))
```
renebrandel, line 218:
```suggestion
    !(await prompter.yesOrNo('Do you want to access Amplify-generated resources in your CloudFormation template?', isExistingResources))
```

kaustavghosh06: Done

PR #8953, opened 2021-11-19
fix: address initial comments

akshbhu:
#### Description of changes

Address comments of this PR : https://github.com/aws-amplify/amplify-cli/pull/8916

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-category-api/src/index.ts
```diff
@@ -267,7 +267,7 @@ export async function transformCategoryStack(context: $TSContext, resource: $TSO
     if (canResourceBeTransformed(resource.resourceName)) {
       const backendDir = pathManager.getBackendDirPath();
       const overrideDir = path.join(backendDir, resource.category, resource.resourceName);
-      const isBuild = await buildOverrideDir(backendDir, overrideDir).catch(error => {
+      const isBuild = buildOverrideDir(backendDir, overrideDir).catch(error => {
```
jhockett, line 270: I think it's safer to await the promise here, no?

akshbhu: I guess its the old way since we don't need both async/await and .then/catch

jhockett: The code isn't using `.then()` so I think you're sending a promise, not a boolean to the overrideFlag below. I think using full async/await (like below) and wrap in a try/catch would be better

akshbhu: ill change to previous since as its preset in all overridden categories

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.ts
```diff
@@ -464,7 +464,8 @@ export const updateWalkthrough = async (context): Promise<UpdateApiRequest> => {
   if (await checkAppsyncApiResourceMigration(context, resourceName, true)) {
     await context.amplify.invokePluginMethod(context, 'awscloudformation', undefined, 'compileSchema', [context, { forceCompile: true }]);
   } else {
-    printer.warn(`The GraphQL API is using transformer version 1. Run 'amplify migrate api' to upgrade to transformer version 2.`);
+    printer.error('Update Operations works on migrated projects. Run amplify update api and opt for migration');
```
jhockett, line 467:
```suggestion
    printer.error('Update operations only work on migrated projects. Run "amplify update api" and opt for migration.');
```

jhockett (approved):

PR #8958, opened 2021-11-19
feat: REST API extensibility

jhockett:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-category-api/.npmignore
```diff
@@ -1,6 +1,7 @@
 **/__mocks__/**
 **/__tests__/**
 ./src
+!resources/overrides-resource/tsconfig.json
```
akshbhu, line 4: is this not under cloudformation folder?

packages/amplify-category-api/src/provider-utils/awscloudformation/cdk-stack-builder/apigw-stack-transform.ts
```diff
@@ -0,0 +1,216 @@
...
+    const resourceDirPath = pathManager.getResourceDirectoryPath(undefined, category, this.resourceName);
+    fs.ensureDirSync(resourceDirPath);
+
+    const buildDirPath = path.join(resourceDirPath, PathConstants.BuildDirName);
+    fs.ensureDirSync(buildDirPath);
+
+    stateManager.setResourceParametersJson(undefined, AmplifyCategories.API, this.resourceName, this.cfnInputParams);
+
+    const cfnFilePath = path.resolve(path.join(buildDirPath, `${this.resourceName}-cloudformation-template.json`));
+    return writeCFNTemplate(this.cfn, cfnFilePath);
```
akshbhu, line 214:
```suggestion
    return await writeCFNTemplate(this.cfn, cfnFilePath);
```

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthrough-types/aPIGateway-user-input-types.ts
```diff
@@ -0,0 +1,37 @@
+/**
```
akshbhu, line 1: naming of this file seems improper

akshbhu (approved):
Looks good

packages/amplify-category-api/.npmignore
jhockett: it is indeed, good catch!

packages/amplify-category-api/src/provider-utils/awscloudformation/cdk-stack-builder/apigw-stack-transform.ts
jhockett: `await` here isn't needed because we're returning a promise, so the caller will handle the awaiting.

jhockett: `await` here isn't needed because the function is returning a promise, so the caller will handle the awaiting.

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthrough-types/aPIGateway-user-input-types.ts
jhockett: It's intentional, it's the service name 'API Gateway' with no spaces and the first character to lower case, because that's what the cli-core code is looking for

PR #9030, opened 2021-11-22
feat: add amplify status api -acm Model

cjihrig:
This commit adds a command for viewing the ACM for @auth v2.

### DIFF
--- a/packages/amplify-cli/src/commands/status.ts
+++ b/packages/amplify-cli/src/commands/status.ts
@@ -1,6 +1,6 @@
 import * as path from 'path';
 import * as fs from 'fs-extra';
-import { FeatureFlags, ViewResourceTableParams, CLIParams, $TSAny, $TSContext, pathManager, stateManager } from 'amplify-cli-core';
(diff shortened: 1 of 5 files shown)


packages/amplify-cli/src/commands/status.ts
```diff
@@ -30,3 +39,40 @@ async function showAmplifyConsoleHostingStatus(context) {
     }
   }
 }
+
+async function showApiAuthAcm(context) {
+  const transformerVersion = FeatureFlags.getNumber('graphQLTransformer.transformerVersion');
```
lazpavel, line 44: nit: should this use [getTransformerVersion](https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-provider-awscloudformation/src/transform-graphql-schema.ts#L638-L647) in case only the usePipelinedResolver... flag is set?

renebrandel (approved):

lazpavel (approved):
LGTM, added a nit

sachscode (approved):

Issue #9075, opened 2021-11-24
Cannot run mock after the migration to graphql v2


funes79:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v17.1.0

### Amplify CLI Version

7.5.1

### What operating system are you using?

Ubuntu WSL

### Amplify Categories

auth, api

### Amplify Commands

Not applicable

### Describe the bug

```
$ amplify mock
GraphQL schema compiled successfully.

Edit your schema at /home/tomas/test/amplify/backend/api/graphql/schema.graphql or place .graphql files in a directory at /home/tomas/test/amplify/backend/api/graphql/schema
Failed to start API Mock endpoint Error: Error:Parse error on resolvers/Post.postsByCategory.auth.1.req.vtl:12:
Lexical error on line 13. Unrecognized text.
... = ($authorClaim == $$ctx.args.author) )
-----------------------^
```

### Expected behavior

Running mock server on port 20002

### Reproduction steps

1. amplify migrate api
2. amplify mock

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line
type Post
  @model
  @auth(
    rules: [
      { allow: owner, ownerField: "author" }
      { allow: private, operations: [read] }
      { allow: public, operations: [read] }
    ]
  ) {
  id: ID!
  category: String
    @index(
      queryField: "postsByCategory"
      name: "postCatIndex"
      sortKeyFields: ["author"]
    )
  content: String!
  author: String
  comments: [Comment] @hasMany
}

type Comment
  @model
  @auth(
    rules: [
      { allow: owner, ownerField: "author" }
      { allow: private, operations: [read] }
      { allow: public, operations: [read] }
    ]
  ) {
  id: ID!
  post: Post @belongsTo
  message: String!
  author: String
}


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

lazpavel:
[...]

PR #9077, opened 2021-11-24
fix: remove extra $ in vtl template

cjihrig:
AppSync handles duplicate $s, but mock does not.

Fixes: https://github.com/aws-amplify/amplify-cli/issues/9075


lazpavel (approved):
LGTM

Issue #9097, opened 2021-11-26
DEPLOY FAILING! Error: Cannot find module 'provider-utils/awscloudformation/index' after update to Amplify 7.5.2


georgeleeventures:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v10.23.2

### Amplify CLI Version

7.5.2

### What operating system are you using?

Amazon Linux, Cloud9

### Amplify Categories

Not applicable

### Amplify Commands

pull, push

### Describe the bug

I am unable to do any deployments via a push or a pull after upgrading my amplify to the latest 7.5.2.

When I run amplify pull on my Amazon Linux Cloud9, I get the following error:
```
Error: Cannot find module 'provider-utils/awscloudformation/index'
Require stack:
- /home/ec2-user/.nvm/versions/node/v10.23.2/lib/node_modules/@aws-amplify/cli/node_modules/@aws-amplify/amplify-category-api/lib/index.js
- /home/ec2-user/.nvm/versions/node/v10.23.2/lib/node_modules/@aws-amplify/cli/lib/initialize-env.js
- /home/ec2-user/.nvm/versions/node/v10.23.2/lib/node_modules/@aws-amplify/cli/lib/pull-backend.js
- /home/ec2-user/.nvm/versions/node/v10.23.2/lib/node_modules/@aws-amplify/cli/lib/commands/pull.js
- /home/ec2-user/.nvm/versions/node/v10.23.2/lib/node_modules/@aws-amplify/cli/lib/index.js
- /home/ec2-user/.nvm/versions/node/v10.23.2/lib/node_modules/@aws-amplify/cli/bin/amplify
    at Promise.resolve.then (/home/ec2-user/.nvm/versions/node/v10.23.2/lib/node_modules/@aws-amplify/cli/node_modules/@aws-amplify/amplify-category-api/lib/index.js:128:80)
[...]

PR #9105, opened 2021-11-26
fix: use __dirname instead of '.' in import path

cjihrig:
I haven't been able to reproduce #9097, but this may be related. Even if this does not address #9097, I think we should make this change.

Fixes: https://github.com/aws-amplify/amplify-cli/issues/9097


cjihrig:
Also worth noting: `path.join('.', 'foo')` evaluates to `'foo'`, which is not a relative path like we need here.

jhockett (approved):

Issue #9160, opened 2021-12-01
Amplify Status should display GQL version information


Straubulous:
### Is this feature request related to a new or existing Amplify category?

api

### Is this related to another service?

no

### Describe the feature you'd like to request

Amplify now supports GraphQL version1 and GraphQL version 2. There are also a lot of feature flags that change the way amplify behaves.  Amplify Status (or at least with the -v) should display information about the configuration including which version of GQL transformer & the state of the feature flags.

### Describe the solution you'd like

Amplify Status - V displays GQL version, feature flag settings, and other configuration information (like package versions)

### Describe alternatives you've considered

Amplify version displays the current version, but I think a version block in the status -v output would be useful too

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request
- [ ] ⚠️ This feature might incur a breaking change

PR #9202, opened 2021-12-03
fix: add gql transformer version to status output

cjihrig:
Fixes: https://github.com/aws-amplify/amplify-cli/issues/9160

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/display-helpful-urls.js
+++ b/packages/amplify-provider-awscloudformation/src/display-helpful-urls.js
@@ -3,6 +3,7 @@ const chalk = require('chalk');
 const { BannerMessage, stateManager, FeatureFlags } = require('amplify-cli-core');
 const { fileLogger } = require('./utils/aws-logger');
 const { SNS } = require('./aws-utils/aws-sns');
+const { getTransformerVersion } = require('./transform-graphql-schema');
 const { printer } = require('amplify-prompts');
 
 const logger = fileLogger('display-helpful-urls');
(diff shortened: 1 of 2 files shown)


danielleadams (changes requested):
Would it be possible to add a unit test for showGraphQLTransformerVersion?

cjihrig:
@danielleadams tests added. Please take a look.

danielleadams (approved):

packages/amplify-provider-awscloudformation/src/display-helpful-urls.js
```diff
@@ -278,7 +280,24 @@ function showGraphQLTransformerMigrationMessage() {
...
+    .filter(([_, apiResource]) => apiResource.service === 'AppSync')
+    .map(([name]) => name).length;
+
+  if (!hasGraphqlApi) {
+    return;
+  }
+
+  const transformerVersion = await getTransformerVersion(context);
+
+  context.print.info(chalk`GraphQL transformer version: ${transformerVersion}`);
```
jhockett, line 296: nit: use `printer` from amplify-prompts

`chalk` without `()` looks like a typo possibly?

```js
chalk`GraphQL transformer version: ${transformerVersion}`
```

cjihrig: > nit: use printer from amplify-prompts

I originally did, but changed it for easier mocking for the requested unit tests.

> chalk without () looks like a typo possibly?

That is the style used throughout that file.

jhockett: TIL that is valid JS.

PR #9216, opened 2021-12-04
fix: reference table by logical id from cloudformation template outputs

alex-vladut:
#### Description of changes

I came across this issue that sometimes the Outputs of the Cloudformation Templates generated based on Amplify GraphQL transformers are not referencing the right resource within the template. The resource name includes a hash suffix (generated by the CDK according to the references in this StackOverflow issue https://stackoverflow.com/a/59070762), while the reference here is using the table name. To make sure it covers this case when the logical ID is not the same as the resource name passed when the Table is created, I think it makes sense to always reference through the Table's `logicalId`. Let me know please if you think this change makes sense.

e.g. the resource would be defined like this:

```
  "Resources": {
    "UserTableBD4BF69E": {
      "Type": "AWS::DynamoDB::Table",
       ....
```

while the outputs will be:

```
  "Outputs": {
    "GetAttUserTableName": {
      "Description": "Your DynamoDB table name.",
      "Value": {
        "Ref": "UserTable" <--- no resource "UserTable" defined in this template
      },
      ....
```

**Note:** the issue in my case was caused by a AWS CDK version mismatch, but still I think it is a good idea to reference the resources by their `logicalId` to avoid eventual problems.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
+++ b/packages/amplify-graphql-model-transformer/src/graphql-model-transformer.ts
@@ -1230,17 +1230,17 @@ export class ModelTransformer extends TransformerModelBase implements Transforme
     const streamArnOutputId = `GetAtt${ModelResourceIDs.ModelTableStreamArn(def!.name.value)}`;
     // eslint-disable-next-line no-new
     new cdk.CfnOutput(stack, streamArnOutputId, {
-      value: cdk.Fn.getAtt(tableLogicalName, 'StreamArn').toString(),
+      value: cdk.Fn.getAtt(cfnTable.logicalId, 'StreamArn').toString(),
       description: 'Your DynamoDB table StreamArn.',
-      exportName: cdk.Fn.join(':', [context.api.apiId, 'GetAtt', tableLogicalName, 'StreamArn']),
+      exportName: cdk.Fn.join(':', [context.api.apiId, 'GetAtt', cfnTable.logicalId, 'StreamArn']),
     });
 
     const tableNameOutputId = `GetAtt${tableLogicalName}Name`;
     // eslint-disable-next-line no-new
     new cdk.CfnOutput(stack, tableNameOutputId, {
-      value: cdk.Fn.ref(tableLogicalName),
+      value: cdk.Fn.ref(cfnTable.logicalId),
       description: 'Your DynamoDB table name.',
-      exportName: cdk.Fn.join(':', [context.api.apiId, 'GetAtt', tableLogicalName, 'Name']),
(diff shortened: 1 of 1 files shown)


cjihrig:
Is there a currently broken scenario using the CLI that you can provide us that is fixed by this patch?

alex-vladut:
I will try to put a small repo together for reproducing this issue. This is probably only related to the way I use the Graphql transformers as I don't use the Amplify CLI (instead I call the `GraphqlTransform` directly, which has a dependency on AWS CDK). And the issue was due to some NPM package version mismatch with AWS CDK libraries. But still, even though it won't change the current behaviour of the Amplify CLI, I think it is better to reference the resources by their `logicalId` because AWS CDK could add this hash suffix to ensure resource name uniqueness. Anyways, I'll have a look at this over the next couple of days and see if I can get a simple example easy to reproduce.

alex-vladut:
Hey @cjihrig I added a simple GitHub repo here to reproduce this issue: https://github.com/alex-vladut/amplify-graphql-transformer.

A `schema.graphql` was added [here](https://github.com/alex-vladut/amplify-graphql-transformer/blob/main/schema.graphql) and the transformation could be performed by running the command `npm run transform` which runs [this script](https://github.com/alex-vladut/amplify-graphql-transformer/blob/main/transform.js). This script will run the Graphql transformation and will write the Cloudformation template to disk. And the resulting stack [User](https://github.com/alex-vladut/amplify-graphql-transformer/blob/main/lib/build/stacks/User.json) is not valid because the User DynamoDB table resource is named `UserTableBD4BF69E`, but then in the Outputs it is referenced as `UserTable` e.g.

```
    "GetAttUserTableStreamArn": {
      "Description": "Your DynamoDB table StreamArn.",
      "Value": {
        "Fn::GetAtt": [
          "UserTable",
          "StreamArn"
        ]
      },
```

Let me know if you need any other details. Thanks for your help!

cjihrig:
@alex-vladut I created a GraphQL API via the Amplify CLI and used the schema you provided. I was able to build the schema and successfully push it via CFN. I looked at the `GetAttUserTableStreamArn` output in the CFN console, and the value is there and looks correct. In my `User` CFN stack, the DynamoDB table is defined as `UserTable` with no suffix. It also looks like `User` CFN stack you linked to is also ~400 lines larger than what I got out of the CLI.

I'm hesitant to accept this change until we can confirm that it is needed.

alex-vladut:
Got it, I think it is up to you then, feel free to close the PR 👍 . Thanks for taking the time to look into it, appreciate it

cjihrig:
I'll close this for now, but if you come up with a bug reproduction, please reach out.

PR #9230, opened 2021-12-06
fix: prevent A <-> B @hasOne or @hasMany relationships with DS

cjihrig:
If DataStore is enabled, Model A cannot have a `@hasOne` or `@hasMany` relationship with Model B if Model B also has a `@hasOne` or `@hasMany` relationship with Model A.

### DIFF
--- a/packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-has-many-transformer.test.ts
+++ b/packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-has-many-transformer.test.ts
@@ -1,6 +1,6 @@
 import { IndexTransformer, PrimaryKeyTransformer } from '@aws-amplify/graphql-index-transformer';
 import { ModelTransformer } from '@aws-amplify/graphql-model-transformer';
-import { GraphQLTransform, validateModelSchema } from '@aws-amplify/graphql-transformer-core';
+import { ConflictHandlerType, GraphQLTransform, validateModelSchema } from '@aws-amplify/graphql-transformer-core';
 import { Kind, parse } from 'graphql';
 import { BelongsToTransformer, HasManyTransformer, HasOneTransformer } from '..';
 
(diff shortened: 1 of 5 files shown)


danielleadams (approved):

lazpavel (approved):

packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-has-many-transformer.test.ts
```diff
@@ -646,3 +646,50 @@ test('validates VTL of a complex schema', () => {
...
+      project: {
+        ConflictDetection: 'VERSION',
+        ConflictHandler: ConflictHandlerType.AUTOMERGE,
+      },
+    },
+    transformers: [new ModelTransformer(), new HasOneTransformer(), new HasManyTransformer()],
+  });
+
+  expect(() => transformer.transform(inputSchema)).toThrowError(
+    `Blog and Post cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead.`,
```
raphkim, line 693: this makes it sound like you need to switch both to `@belongsTo`, which is another invalid case. I think we need to explicitly specify that only one of them needs to be switched to `@belongsTo`

raphkim (commented):
I think it would be helpful to suggest `@manyToMany` as a valid alternative to `@hasMany` <-> `@hasMany`

packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-has-many-transformer.test.ts
cjihrig: Please provide the error message you would like displayed.

iartemiev (approved):
Discussed this on call - LGTM!

raphkim (approved):

lawmicha (approved):

packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-has-many-transformer.test.ts
iartemiev:
```suggestion
    `Blog and Post cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead. See https://docs.amplify.aws/cli/graphql/data-modeling/#belongs-to-relationship`,
```
```suggestion
    `Blog and Post cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead.`,
```

packages/amplify-graphql-relational-transformer/src/__tests__/amplify-graphql-has-one-transformer.test.ts
```diff
@@ -350,3 +350,97 @@ test('creates has one relationship with composite sort key.', () => {
...
+      project: {
+        ConflictDetection: 'VERSION',
+        ConflictHandler: ConflictHandlerType.AUTOMERGE,
+      },
+    },
+    transformers: [new ModelTransformer(), new HasOneTransformer(), new HasManyTransformer()],
+  });
+
+  expect(() => transformer.transform(inputSchema)).toThrowError(
+    `Post and Blog cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead.`,
```
iartemiev, line 418:
```suggestion
    `Blog and Post cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead. See https://docs.amplify.aws/cli/graphql/data-modeling/#belongs-to-relationship`,
```

packages/amplify-graphql-relational-transformer/src/utils.ts
```diff
@@ -171,3 +171,37 @@ function getIndexName(directive: DirectiveNode): string | undefined {
...
+      getBaseType(field.type) === modelType &&
+      field.directives!.some(directive => {
+        return directive.name.value === 'hasOne' || directive.name.value === 'hasMany';
+      })
+    );
+  });
+
+  if (hasUnsupportedConnectionFields) {
+    throw new InvalidDirectiveError(
+      `${modelType} and ${relatedType.name.value} cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead.`,
```
iartemiev, line 204:
```suggestion
      `${modelType} and ${relatedType.name.value} cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead. See https://docs.amplify.aws/cli/graphql/data-modeling/#belongs-to-relationship`,
```

cjihrig: Suggestion applied.

iartemiev:
@cjihrig the expected error messages in the unit tests need to be updated as well, right?

cjihrig:
@iartemiev we'll see how the CI run goes, but it should be fine according to https://jestjs.io/docs/expect#tothrowerror

PR #9272, opened 2021-12-09
fix(amplify-util-mock): correct mock env variable name to match name …

loganhersh:
#### Description of changes

Updates the mock static environment variable AWS_LAMBDA_LOG_GROUP_STREAM_NAME to match the runtime environment variable AWS_LAMBDA_LOG_STREAM_NAME. The naming mismatch can lead to unexpected errors while mocking or at runtime if the function expects the env variable to have a value.

There is no indication that the current variable name was ever used except in this mocking context.

#### Issue #, if available

No existing issue.

Compared value of process.env from mock execution with value of process.env from deployed execution. Mock env variable names now match and error is not thrown in code that relies on the existence of AWS_LAMBDA_LOG_STREAM_NAME.

Executed yarn test.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


edwardfoyle (approved):

loganhersh:
Any idea why this failed? It doesn't seem to be related to the change.

PR #9281, opened 2021-12-10
test(geo): add geo populate e2e tests

AaronZyLee:
#### Description of changes

Add e2e tests for `amplify geo populate`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-tests/src/__tests__/geo-populate.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    }
+    deleteProjectDir(projRoot);
+  });
+
+  it('should throw error if there is no geofence collection provisioned', async () => {
+    await initJSProjectWithProfile(projRoot, {});
+    await addAuthWithDefault(projRoot);
+    await updateAuthAddUserGroups(projRoot, ['admin']);
+    await addGeofenceCollectionWithDefault(projRoot);
+    await expect(populateGeofencesWithDefault(projRoot)).rejects.toThrowError();
```
phani-srikar, line 34: Can we move the steps from `initJS...` until `addGeofenceCollection...` to the test setup as we are only testing the populate workflow in all the tests.

packages/amplify-e2e-tests/src/__tests__/geo-populate.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+  it('should upload a valid GeoJSON file with auto-generated root level ID to provisioned geofence collection', async () => {
+    await initJSProjectWithProfile(projRoot, {});
+    await addAuthWithDefault(projRoot);
+    await updateAuthAddUserGroups(projRoot, ['admin']);
+    await addGeofenceCollectionWithDefault(projRoot);
+    await amplifyPushWithoutCodegen(projRoot);
+    await expect(populateGeofencesWithDefault(projRoot, { geoJSONFileName: 'valid-root-level-id.json' })).resolves.not.toThrow();
```
phani-srikar, line 42: We might need to also verify that the collection now has the geofences in it by using `https://docs.aws.amazon.com/location-geofences/latest/APIReference/API_ListGeofences.html`. You can do this as an enhancement in a different PR if you would like to.

phani-srikar (changes requested):
One change requested and other can be handled in a different PR.

packages/amplify-e2e-tests/src/__tests__/geo-populate.test.ts
AaronZyLee: I am considering compress the similar tests into one case where we provision once and upload multiple geo json files, which will save much time on total run time

phani-srikar: That might make debugging difficult in case something fails. I would say as a best practice test one specific case in a test.

AaronZyLee: Update: The cases are still separate but only has one provision run

phani-srikar (approved):

Issue #9335, opened 2021-12-17
How to model hierarchical/recursive data models?


alex-vladut:
>**Note**: If your question is regarding the AWS Amplify Console service, please log it in the 
[AWS Amplify Console repository](https://github.com/aws-amplify/amplify-console/issues)

**Which Category is your question related to?**
`API`

**Amplify CLI Version** 

You can use `amplify -v` to check the amplify cli version on your system

`7.6.4`

**What AWS Services are you utilizing?**

AppSync, DynamoDB, Cognito, Lambda etc.

**Provide additional details e.g. code snippets**

I would like to model the following hierarchical relationship representing a simple file/directory structure:

```
type Document @model {
  id: ID! @primaryKey
  name: String!
  parentId: ID
  parent: Document @hasOne(fields: ["parentId"])
  createdAt: AWSDateTime!
  updatedAt: AWSDateTime!
}
```

This used to work fine with the previous transformer's version, but now I get an error when running `amplify push`:

```
✖ An error occurred when pushing the resources to the cloud
🛑 An error occurred during the push operation: Document and Document cannot refer to each other via @hasOne or @hasMany when DataStore is in use. Use @belongsTo instead. See https://docs.amplify.aws/cli/graphql/data-modeling/#belongs-to-relationship
```

Is there any other way suggested now to model this kind of relationships? Thanks for your help!

yucomds:
I've got the same problem.

in my case I have to create a model called Category. Each Category can have multiple child Categories.

The strange thing is that until yesterday the backend was working correctly. Yesterday suddenly the backend build went over and gave me the same error reported in this issue.
Why this change?
Is there another way this structure can be modeled?

cjihrig:
Thanks for reporting this. I've opened https://github.com/aws-amplify/amplify-cli/pull/9336 to fix the issue. Until that is released, you can use v7.6.3 as a workaround.

PR #9336, opened 2021-12-17
fix: support recursive @hasOne/@hasMany with DataStore

cjihrig:
This commit allows recursive @hasOne and @hasMany relationships
on the same model when DataStore is enabled.

Fixes: https://github.com/aws-amplify/amplify-cli/issues/9335


lazpavel (approved):
LGTM

PR #9355, opened 2021-12-20
fix: update amplify configure to use AdministratorAccess-Amplify

cjihrig:
This commit updates the IAM policy used by amplify configure to
be one controlled by the Amplify team.

Refs: https://github.com/aws-amplify/amplify-cli/pull/7571


cjihrig:
@kaustavghosh06 would you mind taking a look?

pavellazar (approved):
LGTM

attilah (approved):

kaustavghosh06 (approved):
LGTM.
Let's please add documentation as well for how customers can add additional policies if required for custom resources or overrides if required. 
cc @renebrandel

kaustavghosh06:
Also, it would be good to investigate if there is a way to open up the current IAM user in the IAM console to add any additional IAM policies when required straight from the CLI.

PR #9325, opened 2021-12-15
fix: rest api override CloudFormation parameters

jhockett:
#### Description of changes
Add optional `value` argument to enable storing parameter values in parameters.json in the build folder.

Proposed updated function signature: 
```ts
addCfnParameter(props: cdk.CfnParameterProps, logicalId: string, value?: any): void;
```

#### Issue #, if available
https://github.com/aws-amplify/amplify-cli/issues/9260
https://github.com/aws-amplify/amplify-cli/issues/9221

#### Description of how you validated changes
manual verification, e2e test passes locally, `yarn test` passes for api category package

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #9465, opened 2022-01-08
Fix APIGW parameters.json getting ignored and overwritten

nathan-knight:
#### Description of changes
`apgw-stack-transform` previously never loaded `parameters.json` and then would overwrite it at the end, resulting in any parameters defined there getting lost and causing the definitions to not be used when they are needed in the push state (`push-resources.ts` loads the `parameters.json` file, since it was deleted, the additional parameters were previously dropped). Now it loads the parameters and writes them back to disk.

I tested them against a project I am working on and it resolved the issue I was encountering (parameters.json was getting deleted and I could not define parameters).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes (there were several tests not working on Windows due to different path separators)

### DIFF
--- a/packages/amplify-category-api/src/provider-utils/awscloudformation/cdk-stack-builder/apigw-stack-transform.ts
+++ b/packages/amplify-category-api/src/provider-utils/awscloudformation/cdk-stack-builder/apigw-stack-transform.ts
@@ -62,7 +62,7 @@ export class ApigwStackTransform {
   }
 
   generateCfnInputParameters() {
-    this.cfnInputParams = {};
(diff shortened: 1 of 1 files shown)


jhockett:
Hi @nathan-knight, we released a [PR](https://github.com/aws-amplify/amplify-cli/pull/9325) to make setting key value pairs in the `parameters.json` file from the `override.ts` file much easier.

We don't want to load the values from a build artifact because it would prevent the `override.ts` file from being a single source of truth. Closing this pull request.

Issue #9370, opened 2021-12-22
Dynamic Groups authorization check not applied on field level in GraphQL transformer v2


alex-vladut:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm install -g @aws-amplify/cli

### If applicable, what version of Node.js are you using?

v14.16.0

### Amplify CLI Version

7.6.5

### What operating system are you using?

MacOS

### Amplify Categories

api

### Amplify Commands

Not applicable

### Describe the bug

As I gave it a try and tested more in depth the authorization rules applied through GraphQL transformer v2, I cam across a couple of issues in regard to dynamic groups authorization rules:

1) The dynamic groups based authorization logic doesn't seem to be properly applied on field level and, as a result, the resolver returns `null` when the user is actually allowed to access this field. 

I know there was this related issue https://github.com/aws-amplify/amplify-cli/issues/9222 which was fixed with this PR https://github.com/aws-amplify/amplify-cli/pull/9264. Possibly the changes applied on the query may need to be applied as well on the [field resolvers](https://github.com/lazpavel/amplify-cli/blob/60977996d23f6f57e9e95f0f774401792ec4ca41/packages/amplify-graphql-auth-transformer/src/resolvers/field.ts) as far as I could understand this issue. The generated field level resolver `UserMembership.organisation` has this authorization section, so as I understand it, it will simply check the claim string, rather than parsing it to be a list as expected:

```
#if( $util.authType() == "User Pool Authorization" )
  #if( !$isAuthorized )
    #set( $primaryRole0 = $util.defaultIfNull($ctx.identity.claims.get("tenants"), "___xamznone____") )
    #if( (!$util.isNull($ctx.source.organisationId)) && ($ctx.source.organisationId == $primaryRole0) )
      #set( $isAuthorized = true )
      $util.qr($ctx.stash.put("authFilter", null))
    #else
      #if( !$isAuthorized && $util.isNull($ctx.stash.authFilter) )
[...]

PR #9466, opened 2022-01-09
fix: parse dynamic groups custom claim to list

alex-vladut:
#### Description of changes

With this PR the custom claims specified for the dynamic groups will be parsed into a list so that it allows multiple dynamic groups to be specified and to fix the functionality which was working in the GraphQL Transformer v1. Currently it will either return items the user shouldn't even be allowed to access, or will thrown an `Unauthorized` error when accessing the items they should be allowed to view.
The custom claim included in the JWT could either be a string:

```json
{
  "tenants": "05949c86-6c1a-4f13-8bb5-bdbee322f0d6",
  ...
}
```

or a list:

```json
{
  "tenants": "[\"05949c86-6c1a-4f13-8bb5-bdbee322f0d6\",\"c812425e-77f3-4362-aac1-67da9aa30c3f\",\"b3c3ab8a-c89b-4690-884b-82a22e3f64d5\"]",
  ...
}
```

At the moment, the generated mutation for creating an entity, for example, doesn't parse the custom claim from JSON, but then it is assuming it to be a list and as a results it ends up in the wrong comparison:

```
#if( $util.authType() == "User Pool Authorization" )
  #if( !$isAuthorized )
    #set( $groupEntity0 = $util.defaultIfNull($ctx.args.input.organisationId, null) )
    #set( $groupClaim0 = $util.defaultIfNull($ctx.identity.claims.get("tenants"), []) )
    #set( $groupAllowedFields0 = [] )
    #foreach( $userGroup in $groupClaim0 )
      #if( $groupEntity0 == $userGroup )
        #if( !$groupAllowedFields0.isEmpty() )
          $util.qr($allowedFields.addAll($groupAllowedFields0))
        #else
          #set( $isAuthorized = true )
          #break
        #end
      #end
    #end
  #end
#end
```

The change I made in this PR the logic was added to first parse the custom claim into a list so that it could be iterated on and compared against the entity id:

```
#if( $util.authType() == "User Pool Authorization" )
  #if( !$isAuthorized )
    #set( $groupEntity0 = $util.defaultIfNull($ctx.args.input.organisationId, null) )
    #set( $groupClaim0 = $util.defaultIfNull($ctx.identity.claims.get("tenants:admin"), []) )
    #if( $util.isString($groupClaim0) ) <----------------------- the logic added starts here
      #if( $util.isList($util.parseJson($groupClaim0)) )
        #set( $groupClaim0 = $util.parseJson($groupClaim0) )
      #else
        #set( $groupClaim0 = [$groupClaim0] )
      #end
    #end <--------------------------------------------------- and ends here
    #set( $groupAllowedFields0 = [] )
    #foreach( $userGroup in $groupClaim0 )
      #if( $groupEntity0 == $userGroup )
        #if( !$groupAllowedFields0.isEmpty() )
          $util.qr($allowedFields.addAll($groupAllowedFields0))
        #else
          #set( $isAuthorized = true )
          #break
        #end
      #end
    #end
  #end
#end
```

In addition to that, there will be other changes necessary for fixing the resolvers generated for getting an entity by ID, index queries and field resolvers. I wanted to get some early feedback on these changes, an unless there is already some work in progress to fix this by the core Amplify team, I may be able to give it a try at some point and fix those cases in another PR as well. Thanks for your help!

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/9370

I applied the changes, then ran `yarn build` from the `amplify-cli` project and the copied the built files into the `node_modules/@aws-amplify/cli` package. Then I just ran `amplify push` on an actual project to test that the authorization rules are now applied correctly and it all worked well.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-graphql-auth-transformer/package.json
+++ b/packages/amplify-graphql-auth-transformer/package.json
@@ -42,6 +42,7 @@
     "md5": "^2.3.0"
   },
   "devDependencies": {
+    "@aws-amplify/graphql-index-transformer": "0.8.6",
     "@aws-amplify/graphql-searchable-transformer": "0.10.6",
     "@types/node": "^12.12.6"
   },
--- a/packages/amplify-graphql-auth-transformer/src/__tests__/group-auth.test.ts
+++ b/packages/amplify-graphql-auth-transformer/src/__tests__/group-auth.test.ts
@@ -1,5 +1,6 @@
 import { AuthTransformer } from '../graphql-auth-transformer';
 import { ModelTransformer } from '@aws-amplify/graphql-model-transformer';
+import { IndexTransformer } from '@aws-amplify/graphql-index-transformer';
 import { GraphQLTransform } from '@aws-amplify/graphql-transformer-core';
 import { ResourceConstants } from 'graphql-transformer-common';
 import { AppSyncAuthConfiguration } from '@aws-amplify/graphql-transformer-interfaces';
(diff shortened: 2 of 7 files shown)


danielleadams:
HI @alex-vladut - thank you for the contribution! Do you mind adding unit tests to this PR? There are unit tests in the Auth Transformer's `__tests__` directory that can serve as guidance.

alex-vladut:
Hey @danielleadams I added some tests to check for the expected claim parsing logic to be present. Please take a look when you get the time and let me know if that looks fine to you. Thanks!

johnpc (approved):

n0rb:
@alex-vladut unfortunately this fix doesn't seem to be working when the groupsField is set to "id"
the generated resolver vtl is different than if the groupsField is set to any other field than "id", it does not contain the new code with isList, parseJson etc and the authFilter will look like this 
`
{
        "or": [
          {
            "id": {
              "eq": "[\"fleet1\",\"fleet2\"]"
            }
          }
        ]
      } `

snippet of generated resolver auth req vtl
```
 #if( !$util.isNull($ctx.args.id) )
    #set( $idClaim = $util.defaultIfNull($ctx.identity.claims.get("custom:fleets"), "___xamznone____") )
    #if( $util.isString($ctx.args.id) )
      #set( $idCondition = ($idClaim == $ctx.args.id) )
    #else
      #set( $idCondition = ($idClaim == $util.defaultIfNull($ctx.args.id.get("eq"), "___xamznone____")) )
    #end
    #if( $idCondition )
      #set( $isAuthorized = true )
      $util.qr($ctx.stash.put("authFilter", null))
    #end
  #else
    $util.qr($primaryFieldMap.put("id", $util.defaultIfNull($ctx.identity.claims.get("custom:fleets"), "___xamznone____")))
  #end
  #if( !$isAuthorized && !$primaryFieldMap.isEmpty() )
    #if( $util.isNull($ctx.args.id) )
      #set( $authFilter = $util.defaultIfNull($ctx.stash.get("authFilter").get("or"), []) )
      #foreach( $entry in $primaryFieldMap.entrySet() )
        #set( $filterMap = {} )
        #if( $util.isList($entry.value) )
          $util.qr($filterMap.put($entry.key, { "in": $entry.value }))
        #else
          $util.qr($filterMap.put($entry.key, { "eq": $entry.value }))
        #end
        $util.qr($authFilter.add($filterMap))
      #end
      $util.qr($ctx.stash.put("authFilter", { "or": $authFilter }))
    #else
```

alex-vladut:
Hi @n0rb thanks for looking into it. As I mentioned in the other PR here https://github.com/aws-amplify/amplify-cli/pull/9535 my plan was to address this problem in multiple PRs with this one focusing on mutations and list queries. I am not very familiar with the codebase and I'm still learning how everything fits together so this is why I try to perform these changes in a more gradual manner. Let me know if you have other suggestions for better addressing it, thanks

endoodev1:
Hey @alex-vladut, thank you so much for making those changes to the code base as those feature are already mentioned in the docs but it seems as they have not been working. 
I'm now also running in the issue mentioned by @n0rb as we have a model which has dynamic group auth enabled and uses _groupsField: "id"_. 
```graphql
type Location @model 
@auth(rules: [ 
  { allow: groups, groups: ["Admin"], operations: [create, read, update, delete] } 
  { allow: groups, groupsField: "id", groupClaim: "la", operations: [read, update, delete] } 
]) {
  id: ID!
}
```

Because we are using the id field, the auth rules are not created correctly as they are coming from another function _generateAuthOnModelQueryExpression_ and there is no toJSON and group parsing. Is there any chance that this bug will be fixed soon as our auth logic is relying on this idea?
Moreover it would be great to mention this issue in the documentation that _groupsField:'id'_ is not supported at the moment. cc @danielleadams (I'm not sure you are the right contact person but you helped in #9535)

PR #9485, opened 2022-01-10
chore: use caret codegen for tag release

AaronZyLee:
#### Description of changes
Use caret version codegen in tag `amplify-codegen-e2e-tests`. This will be used by codegen repo e2e tests only.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/package.json
+++ b/packages/amplify-cli/package.json
@@ -51,7 +51,7 @@
     "amplify-category-xr": "3.2.12",
     "amplify-cli-core": "2.4.7",
     "amplify-cli-logger": "1.1.0",
-    "amplify-codegen": "2.26.22",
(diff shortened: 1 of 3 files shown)


alharris-at (approved):

alharris-at:
For posterity, we are rolling this out to allow our e2e tests to use an unpinned version of our own libraries, otherwise verdaccio tries to pull the pinned version, and when we publish locally as part of testing, we increment to a new patch version, so our tests aren't pointing at our own code.

lazpavel (approved):
LGTM

cjihrig:
This approach seems like it will require a decent amount of maintenance to ensure codegen is testing with the latest CLI code. Is it possible to use something like dependency resolutions when running your test suite to force the version to what you want?

PR #9486, opened 2022-01-10
fix: clean up missing and unused GraphQL v2 dependencies

cjihrig:
This commit cleans up the dependencies in the GraphQL Transformer
v2. The changes are based on the output of depcheck.


lazpavel (approved):
LGTM if builds and tests succeed

PR #9562, opened 2022-01-19
chore: add test schema library package

cjihrig:
This commit adds the initial version of a new package, `@aws-amplify/graphql-schema-test-library`, which is intended to contain a collection of schemas that can be shared among teams for common validation.

cc: @alharris-at. Are there any specific schemas you'd like to see included here? I'd like to balance including schemas that codegen can test and finds useful, while not duplicating the CLI's full test suite.

### DIFF
--- a/packages/amplify-graphql-schema-test-library/src/__tests__/index.test.ts
+++ b/packages/amplify-graphql-schema-test-library/src/__tests__/index.test.ts
@@ -1,0 +1,105 @@
+import { AuthTransformer } from '@aws-amplify/graphql-auth-transformer';
+import { DefaultValueTransformer } from '@aws-amplify/graphql-default-value-transformer';
+import { FunctionTransformer } from '@aws-amplify/graphql-function-transformer';
+import { HttpTransformer } from '@aws-amplify/graphql-http-transformer';
+import { IndexTransformer, PrimaryKeyTransformer } from '@aws-amplify/graphql-index-transformer';
+import { ModelTransformer } from '@aws-amplify/graphql-model-transformer';
+import { PredictionsTransformer } from '@aws-amplify/graphql-predictions-transformer';
+import {
+  BelongsToTransformer,
+  HasManyTransformer,
+  HasOneTransformer,
+  ManyToManyTransformer,
+} from '@aws-amplify/graphql-relational-transformer';
+import { SearchableModelTransformer } from '@aws-amplify/graphql-searchable-transformer';
+import { ConflictHandlerType, GraphQLTransform, GraphQLTransformOptions } from '@aws-amplify/graphql-transformer-core';
+import { TransformerPluginProvider } from '@aws-amplify/graphql-transformer-interfaces';
+import { schemas, TransformerPlatform, TransformerSchema, TransformerVersion } from '..';
+type Writeable<T> = { -readonly [P in keyof T]: T[P] };
+const defaultDataStoreConfig = {
+  resolverConfig: {
+    project: {
+      ConflictDetection: 'VERSION',
+      ConflictHandler: ConflictHandlerType.AUTOMERGE,
+    },
+  },
+} as any;
+
+for (const [name, schema] of Object.entries(schemas)) {
+  test(`schema '${name}' passes or fails as expected`, () => {
+    if (!isTransformerVersionSupported(schema, TransformerVersion.v2)) {
+      throw new Error('only v2 schemas are currently supported');
+    }
+
+    if (isPlatformSupported(schema, TransformerPlatform.api)) {
+      expectToPass(name, schema, createV2Transformer());
+    } else {
+      expectToFail(name, schema, createV2Transformer());
+    }
+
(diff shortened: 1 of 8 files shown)


lazpavel (approved):

packages/amplify-graphql-schema-test-library/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,105 @@
...
+  resolverConfig: {
+    project: {
+      ConflictDetection: 'VERSION',
+      ConflictHandler: ConflictHandlerType.AUTOMERGE,
+    },
+  },
+} as any;
+
+test('v2 schemas pass or fail as expected', () => {
+  for (const [name, schema] of Object.entries(schemas)) {
```
alharris-at, line 29: I'm forgetting if this is straightforward in this case (it was a pain in cypress) but could you flip these two lines, so you generate a named test per-schema `${name} v2 schema passes or fails as expected`? May help with readability and quickly identifying which specific cases are failing on a breaking change.

packages/amplify-graphql-schema-test-library/src/index.ts
```diff
@@ -0,0 +1,131 @@
+/* eslint-disable no-bitwise */
+export const enum TransformerVersion {
```
alharris-at, line 2: Discussed offline, but I'd recommend dropping a comment on these enums that we're representing them as bitmaps since you can select multiple target versions or platforms in a request.

packages/amplify-graphql-schema-test-library/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+
+export type TransformerSchema = {
+  description: string;
+  transformerVersion: TransformerVersion;
+  supportedPlatforms: TransformerPlatform;
+  sdl: string;
+};
+
+export const schemas: { [key: string]: TransformerSchema } = {
+  '@model-simple': {
```
alharris-at, line 31: I wonder if this becomes a bit more readable/discoverable if we separate test schemas into separate files, e.g. https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/test-generator/lib/components/componentWithActionSignOut.json and https://github.com/aws-amplify/amplify-codegen-ui/blob/main/packages/test-generator/lib/components/index.ts

alharris-at: More of a nitpick though, that's easy enough to separate later if we deem it necessary.

packages/amplify-graphql-schema-test-library/src/index.ts
```diff
@@ -0,0 +1,131 @@
...
+    sdl: `
+      type Blog @model {
+        id: ID!
+        posts: [Post] @hasMany
+      }
+      type Post @model {
+        id: ID!
+        blog: [Blog] @hasMany
+      }
+    `,
```
alharris-at, line 129: I think adding an default value use-case would be valuable.

```
type Todo @model {
  content: String @default(value: "My new Todo")
}
```

implicit and explicit multi-model has one and has many, so:

```
type Project @model {
  id: ID!
  name: String
  team: Team @hasOne
}

type Team @model {
  id: ID!
  name: String!
}
```

```
type Project @model {
  id: ID!
  name: String
  teamID: ID
  team: Team @hasOne(fields: ["teamID"])
}

type Team @model {
  id: ID!
  name: String!
}
```

```
type Post @model {
  id: ID!
  title: String!
  comments: [Comment] @hasMany
}

type Comment @model {
  id: ID!
  content: String!
}
```

```
type Post @model {
  id: ID!
  title: String!
  comments: [Comment] @hasMany(indexName: "byPost", fields: ["id"])
}

type Comment @model {
  id: ID!
  postID: ID! @index(name: "byPost", sortKeyFields: ["content"])
  content: String!
}
```

As well as adding in the usecases for belongsTo

```
type Project @model {
  id: ID!
  name: String
  team: Team @hasOne
}

type Team @model {
  id: ID!
  name: String!
  project: Project @belongsTo
}
```

```
type Project @model {
  id: ID!
  name: String
  team: Team @hasOne
}

type Team @model {
  id: ID!
  name: String!
  projectID: ID
  project: Project @belongsTo(fields: ["projectID"])
}
```

```
type Post @model {
  id: ID!
  title: String!
  comments: [Comment] @hasMany(indexName: "byPost", fields: ["id"])
}

type Comment @model {
  id: ID!
  postID: ID! @index(name: "byPost", sortKeyFields: ["content"])
  content: String!
  post: Post @belongsTo(fields: ["postID"])
}
```

alharris-at (commented):
Overall LGTM - a few minor comments on how we structure this, and a few additional schemas suggested.

packages/amplify-graphql-schema-test-library/src/__tests__/index.test.ts
cjihrig: Good call. It seems to work nicely with Jest.

PR #9576, opened 2022-01-21
 fix: rest api policy size calculation

attilah:
#### Description of changes

Change REST API policy size calculation logic and numbers. The resource gets collapsed into a single resource with `*` when all the HTTP verbs are allowed for a path.

#### Issue #, if available

N/A

Manual deployments

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/api_5.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/api_5.test.ts
@@ -127,6 +127,7 @@ describe('amplify add api (REST)', () => {
     // add more paths to and test policy slicing
     for (let i = 0; i < 15; i++) {
       await addRestApi(projRoot, {
+        path: `/items${i}`,
         isFirstRestApi: false,
         existingLambda: true,
         restrictAccess: true,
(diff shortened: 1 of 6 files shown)


jhockett (approved):
LGTM

packages/amplify-e2e-tests/src/__tests__/api_5.test.ts
```diff
@@ -123,6 +123,16 @@ describe('amplify add api (REST)', () => {
...
+
+    // add more paths to and test policy slicing
+    for (let i = 0; i < 15; i++) {
+      await addRestApi(projRoot, {
+        isFirstRestApi: false,
+        existingLambda: true,
+        restrictAccess: true,
+        allowGuestUsers: true,
+      });
+    }
```
jhockett, line 135: @yuth I think you need to pass in a path to the e2e helper, since the default is always `/items`

yuth: good catch

cjihrig (approved):

packages/amplify-e2e-tests/src/__tests__/api_5.test.ts
jhockett: Oh, perhaps I misread this. I thought this was adding a new path to an existing API, but this should be fine if they're all distinct resources.

PR #9652, opened 2022-01-30
fix(amplify-storage-simulator): fix schema of events emitted

tonyharrison:
#### Description of changes

The structure and names of events emitted by storage simulator was incorrect.
Properties such as eventName, eventSource etc were not direct descendants of the Record object, they were descendants of an "event" property within the Record.  This does not match the schema of the events delivered to the Lambda functions with an S3 trigger.

Additionally the event names were being constructed using the request method, which is all uppercase and does not match the actual event names.

Ref: [Event Message Structure ](
https://docs.aws.amazon.com/AmazonS3/latest/userguide/notification-content-structure.html)

Also added some error handling around `fs.writeFile` in `handlRequestPut`. Since the mock S3 Server relies on request method and not the operation (ie: PutObject etc) to determine the action to take,  scenarios such as CopyObject being used in the t function would be treated as PutObject requests and cause the mock storage server to crash,without cleaning up. (ie: resolver files for graphql mock would remain).   Logging a warning and not crashing seemed "good enough" for this, any other handling would probably require more refactoring than is practical right now.

*Breaking Change?*
Not really, though it might break a trigger function that is in active development and never been deployed. Any target function that was relying on the incorrect structure would not have worked when deployed unless their was already a logic branch in the function to handle the incorrect structure.

#### Issue #, if available

N/A

Added additional tests to validate schema of emitted events.

All tests in `yarn test` pass.

Tested running amplify mock with linked local, functions still triggered and events have correct structure and eventNames.

#### Checklist

- [x] PR description included
- [X] `yarn test` passes 

### DIFF
--- a/packages/amplify-storage-simulator/src/__tests__/S3server.test.ts
+++ b/packages/amplify-storage-simulator/src/__tests__/S3server.test.ts
@@ -15,6 +15,23 @@ let simulator;
 
 jest.setTimeout(2000000);
 
+const toBeS3Event = (eventName: string, bucket: string, key: string) =>
+  expect.objectContaining({
+    Records: expect.arrayContaining([
+      expect.objectContaining({
+        eventName: eventName,
+        s3: expect.objectContaining({
+          bucket: expect.objectContaining({
+            name: bucket,
+          }),
+          object: expect.objectContaining({
+            key: key,
+          }),
+        }),
+      }),
+    ]),
+  });
+
 beforeAll(async () => {
   AWS.config.update({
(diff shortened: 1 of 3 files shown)


sdstolworthy:
Hi, @tonyharrison,

It looks like these changes are not passing through the pipeline. Because there has not been recent activity on this pull request, we are closing it. Please review the contribution guidelines. Once these changes meet those requirements, feel free to re-open the pull request and tag `@aws-amplify/amplify-cli`, and we will take another look.

Issue #9685, opened 2022-02-03
Incorrect error thrown in amplify-appsync-simulator/lib/resolvers/base-resolver


pkubat:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### Amplify CLI Version

n/a latest

### What operating system are you using?

v14.18.3 (npm v8.3.1)

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

no

### Amplify Categories

Not applicable

### Amplify Commands

Not applicable

### Describe the bug

If there is an error in the response template the simulator returns 'Missing request mapping template'

Also, in the code, it never checks the request template.

Lastly, it should not hide the expectations from velocity, it makes troubleshooting velocity issues impossible.

I will post a PR.

### Expected behavior

Throw an exceptions with the message `Missing response mapping template` and the exception that was thrown. an exception

### Reproduction steps

Have an error in the response template.

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

PR #9687, opened 2022-02-03
fix(amplify-appsync-simulator): base-resolver for response and added…

pkubat:
… better error handling

This is addresses issue #9685.   Please review and applies as you want.  I did test localy but did not run the unit tests. 

#### Description of changes
1. Fix the message returned on exception they bother were _request_, and that leads people to look at the wrong template to resolve a template issue.
2. It never called the _request_ template, update the method to call both response and request
3. Added the message from the underlying exceptions.  It will have the error details from velocity which is much needed to make troubleshooting template syntax issues

Example of the fixed message thrown.  
```
Error: Missing response mapping template Error:Parse error on INLINE_TEMPLATE:2:
Lexical error on line 3. Unrecognized text.
...$err.type, $err.data. $err.info)  #end
-----------------------^
```

#### Issue #, if available
https://github.com/aws-amplify/amplify-cli/issues/9685

#### Checklist

- [ ?] `yarn test` passes

### DIFF
--- a/packages/amplify-appsync-simulator/src/resolvers/base-resolver.ts
+++ b/packages/amplify-appsync-simulator/src/resolvers/base-resolver.ts
@@ -7,13 +7,13 @@ export abstract class AppSyncBaseResolver {
     try {
       this.getResponseMappingTemplate();
     } catch (e) {
-      throw new Error(`Missing request mapping template`);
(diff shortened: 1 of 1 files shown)


lazpavel (approved):
Hi @pkubat, thank you for your contribution. Changes LGTM

PR #9677, opened 2022-02-02
fix: allow init with auth role name overrides

johnpc:
#### Description of changes

This PR enables applying auth role overrides to new environments

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/9643

#### Description of how you validated changes

```
amplify-dev init
amplify-dev override project
amplify-dev push
amplify-dev add env envb
```

Without this change, the result is that `envb` uses an auth role that does not match the override. With this change, `envb` uses an auth role that _does_ match the override.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

Issue #9643, opened 2022-01-29
Overrides not applied when switching environments


sachscode:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

v16.13.0

### Amplify CLI Version

7.6.13

### What operating system are you using?

mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No manual changes made

### Amplify Categories

Not applicable

### Amplify Commands

override

### Describe the bug

Overrides are not getting applied on "amplify env add" even when previous project had the resources overridden.

### Expected behavior

The Customer overrode the properties "permission boundary" and "role names"  for AuthRole and UnAuthRole. However after amplify env add , they observed that the overridden properties were not reflected in the new account.

### Reproduction steps

Customer created an environment in AWS Account 1.
“amplify init”
“amplify override project” 
 Override "permission boundary" and "role names"  for AuthRole and UnAuthRole in Auth resource. 
“amplify push”
“amplify env add” -> switching to AWS Account 2. <--- Overrides aren’t applied on this step.
“amplify push”

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

PR #9688, opened 2022-02-03
fix: allow init with auth role name overrides

johnpc:
#### Description of changes

This PR enables applying auth role overrides to new environments. Replaces https://github.com/aws-amplify/amplify-cli/pull/9677 which could not be updated due to branch protection rules.

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/9643

```
amplify-dev init
amplify-dev override project
amplify-dev push
amplify-dev add env envb
```

Without this change, the result is that `envb` uses an auth role that does not match the override. With this change, `envb` uses an auth role that _does_ match the override.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/overrides/override-root.ts
+++ b/packages/amplify-e2e-tests/overrides/override-root.ts
@@ -1,4 +1,7 @@
+function getRandomInt(max) {
+  return Math.floor(Math.random() * max);
+}
 export function override(props: any): void {
-  props.authRole.roleName = 'mockRole';
+  props.authRole.roleName = `mockRole-${getRandomInt(10000)}`;
   return props;
 }
--- a/packages/amplify-provider-awscloudformation/src/initializer.ts
+++ b/packages/amplify-provider-awscloudformation/src/initializer.ts
@@ -3,6 +3,8 @@ import _ from 'lodash';
 import { transformRootStack } from './override-manager';
 import { rootStackFileName } from './push-resources';
 import { getDefaultTemplateDescription } from './template-description-utils';
+import * as vm from 'vm2';
+import { printer, formatter } from 'amplify-prompts';
 
 const moment = require('moment');
 const path = require('path');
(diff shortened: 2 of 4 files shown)


packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -53,6 +55,40 @@ export async function run(context) {
...
+      unauthRole: {
+        roleName: unauthRoleName,
+      },
+    };
+
+    const noOverrideMsg = 'No override file found. To override, run amplify override auth';
+    try {
+      const backendDir = pathManager.getBackendDirPath();
+      const overrideFilePath = path.join(backendDir, 'awscloudformation', 'build', 'override.js');
+      const overrideCode: string = await fs.readFile(overrideFilePath, 'utf-8').catch(() => {
```
sachscode, line 71: 1. It would be good to move this into a function like "transpileAmplifyOverridesCode()" and parameterize the path parameters .
2. It would be good to create a constant for the 5 sec timeouts.

sachscode (commented):
Over all LGTM. just a few comments.

packages/amplify-provider-awscloudformation/src/initializer.ts
cjihrig, line 71: The mix of vanilla Promises (`.catch()`) and `async/await` here is confusing.

packages/amplify-e2e-tests/overrides/override-root.ts
```diff
@@ -1,4 +1,7 @@
+function getRandomInt(max) {
+  return Math.floor(Math.random() * max);
+}
```
jhockett, line 3: nit: I would prefer to see a uuid used here

packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -53,6 +55,37 @@ export async function run(context) {
...
+          require: {
+            context: 'sandbox',
+            builtin: ['path'],
+            external: true,
+          },
+        });
+        sandboxNode.run(overrideCode).override(configuration);
+      }
+    } catch (e) {
+      printer.debug(`Unable to apply auth role overrides: ${e.message}`);
```
jhockett, line 86: `.debug` hides the error message behind the `--debug` flag. Change this to `.error`, or consider throwing if it makes more sense to fast fail

sachscode (approved):
LGTM

packages/amplify-e2e-tests/overrides/override-root.ts
johnpc: Is there a native uuid function? I’d prefer not to pull a bunch of dependencies into the e2e test app.

johnpc: Is there a native uuid function? I’d prefer not to pull a bunch of dependencies into the e2e test app.

packages/amplify-provider-awscloudformation/src/initializer.ts
johnpc: I think —debug is correct. In the common case we expect to log this debug line.

packages/amplify-e2e-tests/overrides/override-root.ts
jhockett: it's already a dependency for the e2e-test package, but we should export a wrapper function from e2e-core (we actually might already have this). It would probably be cleaner to call something like `getShortUniqueId() => string` instead of what most of the code does today

packages/amplify-provider-awscloudformation/src/initializer.ts
jhockett: That behavior was marked as a bug: https://github.com/aws-amplify/amplify-cli/issues/9695

jhockett (approved):

PR #9689, opened 2022-02-03
fix: fix modify gql schema base case; scope auth notif to v2

danielleadams:
#### Description of changes
- scopes the auth notification to v2 transformer
- fixes the recursive updates to schema files by using a returned base case rather than setting a variable

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/auth-notifications.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/auth-notifications.ts
@@ -71,14 +71,12 @@ async function modifyGraphQLSchema(apiResourceDir: string): Promise<void> {
   if (schemaFileExists) {
     fs.appendFile(schemaFilePath, ' ');
   } else if (schemaDirectoryExists) {
-    await modifyGraphQLSchemaDirectory(schemaDirectoryPath, false);
+    await modifyGraphQLSchemaDirectory(schemaDirectoryPath);
   }
 }
 
-async function modifyGraphQLSchemaDirectory(schemaDirectoryPath: string, modified: boolean): Promise<void> {
+async function modifyGraphQLSchemaDirectory(schemaDirectoryPath: string): Promise<boolean> {
   const files = await fs.readdir(schemaDirectoryPath);
-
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/auth-notifications.ts
```diff
@@ -90,17 +88,20 @@ async function modifyGraphQLSchemaDirectory(schemaDirectoryPath: string, modifie
     const fullPath = path.join(schemaDirectoryPath, fileName);
     const stats = await fs.lstat(fullPath);
 
-    if (stats.isDirectory()) {
-      await modifyGraphQLSchemaDirectory(fullPath, modified);
-    } else if (stats.isFile() && !modified) {
+    if (stats.isDirectory() && (await modifyGraphQLSchemaDirectory(fullPath))) {
```
ammarkarachi, line 91:
```suggestion
    if ((await modifyGraphQLSchemaDirectory(fullPath))) {
       if(stats.isFile()){
        fs.appendFile(fullPath, ' ');
       }
       return true;
    }
```

lazpavel: not sure if this will work, this will attempt to read files from a non directory

lazpavel (approved):
LGTM

packages/amplify-cli/src/extensions/amplify-helpers/auth-notifications.ts
danielleadams: `&&` it will return `false` on the first falsy value

danielleadams: oh wait, ignore my comment. let me update

ammarkarachi: yeah never mind I was thinking about it wrong

ammarkarachi (approved):

PR #9715, opened 2022-02-07
Export AmplifyDependentResourceDefinition interface

dannybpw:
Exported AmplifyDependentResourceDefinition interface


cjihrig:
This is missing some explanation of why the change is being made. Is it because `addCDKResourceDependency()` is exported and it takes `AmplifyDependentResourceDefinition[]` as an argument?

dannybpw:
> This is missing some explanation of why the change is being made. Is it because `addCDKResourceDependency()` is exported and it takes `AmplifyDependentResourceDefinition[]` as an argument?

That's correct, the argument is referencing the type and the use case on my end is creating helper functions to get dependencies / utilize the amplify helpers.

Would prefer to use the interface from the library rather than creating my own.

cjihrig (approved):
Thanks for the contribution.

danielleadams (approved):

PR #9734, opened 2022-02-10
build: add script to check if the versions are already published

yuth:
Added helper scripts to run a sanity check before starting the release. The
script will run learn version and bump the version according to conventional
commit and then check if the bumped versions are already published to NPM
or if there is a git tag for that version due to a failed publishing. This script
needs to be run before starting the release by executing either
`yarn run check-version-conflicts:beta` or `check-version-conflicts:release` by the oncall

#### Checklist

- [x] PR description included


scripts/check-version.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    versionChanges = versionCommand(config);
+  });
+
+  await stashChanges();
+  return versionChanges.updatesVersions;
+};
+
+const stashChanges = () => {
+  execa.sync('git', ['reset', '--hard', 'HEAD'], {
```
jhockett, line 63: I think `resetChanges` would be a better name for this function

jhockett:
```suggestion
const resetChanges = () => {
  execa.sync('git', ['reset', '--hard', 'HEAD'], {
```

scripts/check-version.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const bumpVersions = async (preid?: string, distTag?: string): Promise<Map<string, string>> => {
+  let versionChanges;
+
+  await new Promise((resolve, reject) => {
+    const config: any = {
+      conventionalCommits: true,
+      exact: true,
+      composed: 'publish',
+      lernaVersion: '4.0.0',
```
jhockett, line 38: we might want to read this version from package.json, otherwise we should document that this needs to be updated in order to update lerna

jhockett (approved):

scripts/check-version.ts
yuth: That parameter is not needed. Removed it

danielleadams (approved):
looks good to me. my only nit comment is that the tsconfig.json does not have an extra line at the end of the file. i'm indifferent about that, but it may upset a linter at some point (and I think GitHub "expects" it).

PR #9748, opened 2022-02-11
feat(graphql-auth-transformer): @auth allow more granular access rights for get or list only

lazpavel:
#### Description of changes
- allows use of @auth directive with get or list only access rights

- manual testing
- `yarn test` passes
- additional tests added

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-auth-transformer/src/utils/index.ts
```diff
@@ -56,6 +65,30 @@ export const ensureAuthRuleDefaults = (rules: AuthRule[]) => {
...
+    const operations: (ModelOperation | 'read')[] = rule.operations ?? [];
+
+    if (operations.includes('read') && (operations.includes('list') || operations.includes('get'))) {
+      const offendingOperation = operations.includes('list') ? 'list' : 'get';
+      throw new InvalidDirectiveError(
+        `'${offendingOperation}' operation is specified in addition to 'read'. Either remove 'read' to limit access only to '${offendingOperation}' or only keep 'read' to grant both 'get' and 'list' access.`,
+      );
+    }
+
+    if (operations.includes('read')) {
```
cjihrig, line 81: Can you store the index in a variable. Then you no longer need the `includes()` call.

packages/amplify-graphql-auth-transformer/src/utils/index.ts
```diff
@@ -56,6 +65,30 @@ export const ensureAuthRuleDefaults = (rules: AuthRule[]) => {
   }
 };
 
+export const getAuthDirectiveRules = (authDir: DirectiveWrapper): AuthRule[] => {
+  const rules = authDir.getArguments<{ rules: Array<AuthRule> }>({ rules: [] }).rules;
+
+  for (const rule of rules) {
+    const operations: (ModelOperation | 'read')[] = rule.operations ?? [];
```
cjihrig, line 72: Is `rule.operations` expected to exist on all rules after this function runs? If so, it looks like that is not currently the case.

packages/amplify-graphql-auth-transformer/src/__tests__/__snapshots__/auth-operations.test.ts.snap
```diff
@@ -0,0 +1,204 @@
...
+
+#end
+#if( !$isAuthorized && $util.isNull($ctx.stash.authFilter) )
+$util.unauthorized()
+#end
+$util.toJson({\\"version\\":\\"2018-05-29\\",\\"payload\\":{}})
+## [End] Authorization Steps. **"
+`;
+
+exports[`read get list auth operations 2`] = `undefined`;
```
cjihrig, line 18: Is this correct?

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -416,45 +419,46 @@ Static group authorization should perform as expected.`,
       }
     };
     // default model operations
-    addServiceDirective(ctx.output.getQueryTypeName()!, 'read', modelConfig?.queries?.get);
-    addServiceDirective(ctx.output.getQueryTypeName()!, 'read', modelConfig?.queries?.list);
-    addServiceDirective(ctx.output.getQueryTypeName()!, 'read', modelConfig?.queries?.sync);
+    addServiceDirective(ctx.output.getQueryTypeName()!, 'get', modelConfig?.queries?.get);
+    addServiceDirective(ctx.output.getQueryTypeName()!, 'list', modelConfig?.queries?.list);
+    addServiceDirective(ctx.output.getQueryTypeName()!, 'list', modelConfig?.queries?.sync);
```
cjihrig, line 424: I think this is correct, but just double checking - is `'list'` correct here for sync queries? Same question for some of the subscriptions below.

lazpavel: yes, subscriptions and sync queries are tied to list operation

packages/amplify-graphql-auth-transformer/src/__tests__/__snapshots__/auth-operations.test.ts.snap
lazpavel: updated

packages/amplify-graphql-auth-transformer/src/utils/index.ts
lazpavel: updated

packages/amplify-graphql-auth-transformer/src/utils/index.ts
lazpavel: no, discussed offline

cjihrig (approved):

packages/amplify-graphql-auth-transformer/package.json
```diff
@@ -35,6 +35,7 @@
     "@aws-cdk/aws-dynamodb": "~1.124.0",
     "@aws-cdk/aws-iam": "~1.124.0",
     "@aws-cdk/core": "~1.124.0",
+    "amplify-cli-core": "^2.4.18",
```
jhockett, line 38: monorepo pacakges must use exact types

```suggestion
    "amplify-cli-core": "2.4.18",
```

lazpavel: updated

packages/amplify-cli/src/extensions/amplify-helpers/show-auth-acm.ts
```diff
@@ -27,65 +31,71 @@ export function showACM(sdl: string, nodeName: string) {
...
-    }
+    });
     const truthTable = acm.getAcmPerRole();
 
     if (truthTable.size === 0) {
       printer.warn(`No auth rules have been configured for the "${type.name.value}" model.`);
     }
 
-    for (let [role, acm] of truthTable) {
+    truthTable.forEach((table, role) => {
```
phani-srikar, line 51: Double check if this should be `truthTable.forEach((role, table) =>`

packages/amplify-graphql-auth-transformer/src/__tests__/accesscontrol.test.ts
```diff
@@ -1,7 +1,7 @@
 import { AccessControlMatrix } from '../accesscontrol';
 import { MODEL_OPERATIONS } from '../utils';
 
-test('test access control on object and field', () => {
+test('access control on object and field', () => {
   /*
   given the following schema
   type Student
```
phani-srikar, line 7: can you please also update the schema here with new permissions?

packages/amplify-graphql-auth-transformer/src/utils/index.ts
```diff
@@ -27,6 +33,71 @@ export const splitRoles = (roles: Array<RoleDefinition>): RolesByProvider => ({
...
+/**
+ * returns @auth directive rules
+ */
+export const getAuthDirectiveRules = (authDir: DirectiveWrapper): AuthRule[] => {
+  const splitReadOperation = (rule: AuthRule): void => {
+    const operations: (ModelOperation | 'read')[] = rule.operations ?? [];
+    const indexOfRead = operations.indexOf('read', 0);
+    if (indexOfRead !== -1) {
+      operations.splice(indexOfRead, 1);
+      operations.push('get');
```
phani-srikar, line 45: I see that `read` is being expanded to `get, list, search, sync, listen`. Please confirm this with others if grouping all these ops together under `read` is the right approach.

phani-srikar (changes requested):

packages/amplify-graphql-auth-transformer/src/utils/index.ts
lazpavel: This is today's behavior, changing this will end up changing the behavior on the existing customers applications. 
We did have a design review on this change.

packages/amplify-cli/src/extensions/amplify-helpers/show-auth-acm.ts
lazpavel: no, for/of map iterator is `[key, value]`, while forEach is `(value, key)`

packages/amplify-graphql-auth-transformer/src/__tests__/accesscontrol.test.ts
lazpavel: not sure I follow? the schema that has `read` permission should still work as before without customers to need to make any changes

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-walkthrough.ts
```diff
@@ -116,6 +116,9 @@ const schemaTemplatesV2 = [
   },
 ];
 
+/**
+ *
+ */
```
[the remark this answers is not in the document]
lazpavel: linter introduced this changes on a save, reverted this file as it shouldn't have any changes related to the PR

packages/amplify-graphql-auth-transformer/src/__tests__/accesscontrol.test.ts
phani-srikar: Yeah but in this test, the schema here reflects one constructed in the test and in the test we replace `read` with `get/list etc`. Want to keep the documented schema in parity with code changes

packages/amplify-graphql-auth-transformer/src/utils/index.ts
phani-srikar: I am fine with this if this went through a review.

packages/amplify-graphql-auth-transformer/src/__tests__/accesscontrol.test.ts
lazpavel: updated

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -337,12 +336,24 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
       modelFields.forEach(field => {
         const fieldReadRoles = getReadRolesForField(acm, readRoles, field.name.value);
-        const allowedRoles = fieldReadRoles.filter(r => acm.isAllowed(r, field.name.value, 'read'));
+        const allowedRoles = fieldReadRoles.filter(
+          r => acm.isAllowed(r, field.name.value, 'get')
+                || acm.isAllowed(r, field.name.value, 'list')
+                || acm.isAllowed(r, field.name.value, 'search')
+                || acm.isAllowed(r, field.name.value, 'listen')
+                || acm.isAllowed(r, field.name.value, 'sync'),
+        );
```
sundersc, line 356: should we move this logic into its own function? In that way, this can be reused for `roleDefinitions` assignment as well.

packages/amplify-graphql-auth-transformer/src/graphql-auth-transformer.ts
```diff
@@ -417,45 +428,46 @@ export class AuthTransformer extends TransformerAuthBase implements TransformerA
...
       const config = getSearchableConfig(searchableDirective, def.name.value);
-      addServiceDirective(ctx.output.getQueryTypeName(), 'read', config.queries.search);
+      addServiceDirective(ctx.output.getQueryTypeName(), 'search', config.queries.search);
     }
 
     const subscriptions = modelConfig?.subscriptions;
     if (subscriptions?.level === SubscriptionLevel.on) {
       const subscriptionArguments = acm
-        .getRolesPerOperation('read')
+        .getRolesPerOperation('listen')
```
sundersc, line 453: What if a user have get access on a field but not listen access? Do we set these fields to `null` in mutation response mapping? Because, currently our field level redaction support is not so great.

sundersc (commented):
I have one question, otherwise the changes look good.
How are we handling the fields which user doesn't have permission to `listen` in mutations? Please include a test case for this scenario.

lazpavel:
closing this PR and recreating it in the https://github.com/aws-amplify/amplify-category-api repo

Issue #5860, opened 2020-11-13
How to insert an array of strings on a type with a @key set on that field?


PeteDuncanson:
** Please describe which feature you have a question about? **

I have a type in my schema that accepts an array of strings and a @key on that array to help be search by it (at least that is the idea). However when I try to create that type using a GraphQL mutation I get an error when trying to insert the values for the string array. Am I doing something wrong?

** Provide additional details**

Sample schema:

```
type Area   
  @model
  # If I remove this key everything works as it should I think, but I lose the key!
  @key(
    name: "byPostcode"
    fields: ["postcodes"]
    queryField: "testAreaByPostcode"
  )  {
  name: String
  postcodes: [String!]!
}
```

Mutation run in GraphiQL:

```
mutation AddLeeds($input: CreateAreaInput! ) {
  createArea( input: $input) {
    id
    name
  }
}
```

Query input:

```
{ 
  "input": {
    name: "Leeds",
    postcodes: ["LS1", "LS2", "LS3"]    
  }
}
```

Running the above gives me this error:

```
{
  "data": {
    "createArea": null
  },
  "errors": [
    {
      "message": "Invalid attribute value type",
      "errorType": "DynamoDB:ValidationException",
      "data": null,
      "errorInfo": null,
      "path": [
        "createArea"
      ],
      "locations": [
        {
          "line": 4,
          "column": 3,
          "sourceName": "GraphQL request"
        }
      ]
    }
  ]
}
```

What gives?

nikhname:
@PeteDuncanson We do not support inserting an array of strings on keys in this way. I have marked this as a bug as we should validate this at compile time and provide a better error message.

stwatanabe:
I'm facing same issue.
Im my case, `tagList` is type of `[String!]`
When send a mutate request to mock api server, the message returned from backend.
```GraphQL
mutation MyMutation {
  createVodAsset(input: {title: "", description: "",  tagList: ["aaa","ddd"]}) {
    id
  }
}
```

The result of a request on backend
```JSON
Error while executing Local DynamoDB
{
    "version": "2018-05-29",
    "operation": "PutItem",
    "key": {
        "id": {
            "S": "69cbdff5-f619-4014-a46b-742dbd41cbf3"
        }
    },
    "attributeValues": {
        "title": {
            "S": ""
        },
        "description": {
            "S": ""
        },
        "tagList": {
            "L": [
                {
                    "S": "aaa"
                },
                {
[...]

PR #9749, opened 2022-02-11
fix: disallow lists in @primaryKey and @index fields

cjihrig:
This commit adds validation to @primaryKey and @index to disallow
lists for the fields used by these directives.

Fixes: https://github.com/aws-amplify/amplify-cli/issues/5860


lazpavel (approved):
LGTM

Issue #9666, opened 2022-02-01
When creating any part of the composite sort key for @index '<my index>', you must provide all fields for the key. Missing key: 'createdAt'.


loganpowell:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

14.18.1

### Amplify CLI Version

7.6.14

### What operating system are you using?

Windows 10

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No manual changes made

### Amplify Categories

api

### Amplify Commands

Not applicable

### Describe the bug

When I have a composite index made with two sort keys, specifically `createdAt`, though this should be [automatically handled by a PR made last year](https://github.com/aws-amplify/amplify-cli/issues/4424) by Amplify, when I try to run a mutation, I'm required to add the `createdAt` field to the query.

I've set up a simple example repo that seems to work without having to require the `createdAt`  key, but doesn't work in [my  project](https://github.com/loganpowell/cope/tree/migration), which I'm currently trying to update to the v2 of the transformer.... perhaps that's the problem, IDK :(

I'm sorry I can't seem to reproduce the but using a simpler example.

### Expected behavior

When using a composite sort key with `createdAt`, running a graphql mutation doesn't require it (being autogenerated by the appsync resolvers).

### Reproduction steps

I'm sorry, but I tried to reproduce this with a simple Todo example, but it works there...

### GraphQL schema(s)

<details>

```graphql
# 
# 
#   e88~~8e  888-~88e 888  888 888-~88e-~88e  d88~\ 
#  d888  88b 888  888 888  888 888  888  888 C888   
#  8888__888 888  888 888  888 888  888  888  Y88b  
#  Y888    , 888  888 888  888 888  888  888   888D 
#   "88___/  888  888 "88_-888 888  888  888 \_88P  
# 
# 
enum NodeType {
[...]

PR #9759, opened 2022-02-14
fix: include default values in index arg validation

cjihrig:
Create and update mutations generate a VTL snippet that verifies
that partial sort keys are not provided. However, this snippet
only checks for the arguments in `$ctx.args.input`, which does
not include default or computed values. These values are available
alongside the user inputs in a variable called `$mergedValues`.
This commit updates the argument checking logic to use the
merged values instead of the raw input values.

Fixes: https://github.com/aws-amplify/amplify-cli/issues/9666


lazpavel (approved):
LGTM

sundersc (approved):
LGTM.

PR #9792, opened 2022-02-17
chore: add git secrets husky hook

johnpc:
this commit adds https://github.com/awslabs/git-secrets to our project to prevent accidental checking in of aws secrets.

It does this by adding a husky step to run a pre-commit script that verifies git-secrets is installed properly and scanning the project for secrets.


edwardfoyle (approved):

danielleadams (approved):
This may be a dumb question, but do we need typescript for this? The script is mostly JS. Since it's a pre-commit hook and it will be run repeatedly, we may want the very slight performance boost by just using node/js. (Doesn't need to stop merge, we can always change it later)

PR #9799, opened 2022-02-18
ci: retry failed e2e test in ci 

yuth:
#### Description of changes
E2E tests are flaky due to the resource constraint and cause the pipeline to fail frequently. When the tests fail, all the tests in that suite gets rerun instead of the ones that failed. This causes the retry to run longer and makes probability of already passed test failing possible. With this change, the retry will run only the failed tests when retrying to run them.

This is an example execution where the only the failed tests were re-run
https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/6576/workflows/bb0bd11d-18c7-41ef-8e80-ca607a04ea40/jobs/165144/parallel-runs/0/steps/0-106

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/local_publish_helpers.sh
+++ b/.circleci/local_publish_helpers.sh
@@ -97,11 +97,21 @@ function retry {
     SLEEP_DURATION=5
     FIRST_RUN=true
     n=0
+    FAILED_TEST_REGEX_FILE="./amplify-e2e-reports/amplify-e2e-failed-test.txt"
+    rm -f $FAILED_TEST_REGEX_FILE
     until [ $n -ge $MAX_ATTEMPTS ]
     do
         echo "Attempting $@ with max retries $MAX_ATTEMPTS"
         setAwsAccountCredentials
-        "$@" && break
+        if [ -f  $FAILED_TEST_REGEX_FILE ]; then
+            # read the content of failed tests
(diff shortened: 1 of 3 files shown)


.circleci/local_publish_helpers.sh
```diff
@@ -97,11 +97,21 @@ function retry {
...
-        "$@" && break
+        if [ -f  $FAILED_TEST_REGEX_FILE ]; then
+            # read the content of failed tests
+            faileTests=$(<$FAILED_TEST_REGEX_FILE)
+            cmd="$@ -t \"$faileTests\""
+            eval ${cmd} && break
+        else
+            # is it is not exist then it will be printed
+            "$@" && break
+        fi
```
jhockett, line 114:
```suggestion
        if [ -f  $FAILED_TEST_REGEX_FILE ]; then
            # read the content of failed tests
            failedTests=$(<$FAILED_TEST_REGEX_FILE)
            cmd="$@ -t \"$failedTests\""
            eval ${cmd} && break
        else
            # is it is not exist then it will be printed
            "$@" && break
        fi
```

jhockett (approved):

sundersc (approved):

Issue #8774, opened 2021-11-10
amplify-appsync-simulator - Integer.parseInt does not behave like in AppSync


DavidCaya:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

yarn add amplify-appsync-simulator

### If applicable, what version of Node.js are you using?

v14.17.0

### Amplify CLI Version

amplify-appsync-simulator@1.27.7

### What operating system are you using?

Windows

### Amplify Categories

api

### Amplify Commands

Not applicable

### Describe the bug

Using amplify-appsync-simulator to unit test AppSync resolvers. We are parsing a string as an Int using the strategy defined here: [https://github.com/aws-amplify/amplify-cli/issues/5268](url) 

This works fine in AppSync, but `$Integer.parseInt` will always return `null` using the simulator.

### Expected behavior

A resolver with following code should have its result be set to `5`

> #set($Integer = 1)
> $util.toJson($Integer.parseInt('5'))

### Reproduction steps

[sim-parseint.zip](https://github.com/aws-amplify/amplify-cli/files/7512746/sim-parseint.zip)
1. Unzip in empty folder
2. Go to sim-parseint project
3. yarn
4. yarn test

Expecting test to pass, but getting `null` rather than `5`

### GraphQL schema(s)

<details>

```graphql
# Put schemas below this line


```

</details>

### Log output

<details>

```
# Put your logs below this line


```

</details>

cjihrig:
Confirmed that the following works in AppSync:

```
#set( $Integer = 1 )
#set( $val = $Integer.parseInt("2") )
```

After this executes, `$val` is 2.

PR #9810, opened 2022-02-18
fix: support Java numbers in appsync simulator

cjihrig:
Fixes: https://github.com/aws-amplify/amplify-cli/issues/8774

### DIFF
--- a/packages/amplify-appsync-simulator/src/velocity/value-mapper/integer.ts
+++ b/packages/amplify-appsync-simulator/src/velocity/value-mapper/integer.ts
@@ -1,0 +1,23 @@
+export class JavaInteger {
+  value: number;
+
+  constructor(val: number) {
+    this.value = Math.trunc(val);
+  }
+
+  valueOf(): number {
+    return this.value;
+  }
+
+  parseInt(val: any, radix: any = 10): JavaInteger {
+    return new JavaInteger(Number.parseInt(val, +radix));
+  }
+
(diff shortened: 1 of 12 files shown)


packages/amplify-appsync-simulator/src/velocity/value-mapper/integer.ts
```diff
@@ -0,0 +1,23 @@
...
+  constructor(val: number) {
+    this.value = Math.trunc(val);
+  }
+
+  valueOf(): number {
+    return this.value;
+  }
+
+  parseInt(val: any, radix: any = 10): JavaInteger {
+    return new JavaInteger(Number.parseInt(val, +radix));
```
sachscode, line 13: Why +radix , instead of radix.

sachscode (commented):
LGTM, just one question.

packages/amplify-appsync-simulator/src/velocity/value-mapper/integer.ts
cjihrig: It will ensure that `radix` is converted to a number before passing it to `Number.parseInt()`. If `radix` happens to be a `JavaInteger` object instead of a normal JavaScript value, it will call `valueOf()` to convert it.

sachscode (approved):

sundersc (approved):
LGTM.

PR #9834, opened 2022-02-22
fix(graphql): add new open search instance types to the allowed list

sundersc:
Add OpenSearch instance types to the allowed list.


cjihrig (approved):
LGTM as a first step to get customers unblocked.

Two notes:
- I'm not sure if we should continue investing time making changes for the v1 transformer.
- If I recall correctly, not all instance types are supported in all regions. I think we need to make the list here contingent on the region. No need to make that change in this PR though.

danielleadams (approved):
Agree with @cjihrig about not investing in V1, but 👍🏼 .

sundersc:
> LGTM as a first step to get customers unblocked.
> 
> Two notes:
> 
> * I'm not sure if we should continue investing time making changes for the v1 transformer.
> * If I recall correctly, not all instance types are supported in all regions. I think we need to make the list here contingent on the region. No need to make that change in this PR though.

Agreed. We need to have a map to store the allowed instance types based on the region.
The reason for adding the new instance types to the V1 transformer is to unblock the customer.

michaetto:
Using new types (i.e. *.search) does not work. Most likely because amplify generates AWS::Elasticsearch::Domain in searchable stack instead AWS::OpenSearchService::Domain.
It would be pretty greate to have generated AWS::OpenSearchService::Domain instead, at least under some amplify feature flag in cli.json. We are very close to go to production after hard work and I suppose it could be not easy to change that on production system unless CloudFormation itself allow that migration. Anyway I do elasticsearchVersion = "OpenSearch_1.1" in override.ts, that works with AWS::Elasticsearch::Domain, but only with old instance names, i.e. *.elasticsearch

Issue #9728, opened 2022-02-09
Unable to set ElasticSearchInstanceType in team-provider-info.json


benjamindoe:
**Which Category is your question related to?**
GraphQL

**Amplify CLI Version**
7.6.15

I've tried to follow the guide on the amplify [documentation on how to set the instance type](https://docs.amplify.aws/cli/graphql/search-and-result-aggregations/#set-up-opensearch-for-production-environments). I'm running into the following error:
```
Parameters: [ElasticSearchInstanceType] do not exist in the template
```
My project has been upgraded to the v2 transformer from the v1 transformer. Prior to the upgrade, setting `ElasticSearchInstanceType` worked. I've had a look at the `cloudformation-template.json` file in the `amplify/api/<api-name>/build` folder and discovered that `ElasticSearchInstanceType` is not being set as a parameter for the `SearchableStack`. I've tried making a change to the schema file and I've tried to delete the build folder and rerun `amplify api gql-compile` but these steps do not solve the issue. I have also tried setting `ElasticSearchInstanceType` in `parameters.json` and while it doesn't throw any errors, it doesn't update the instance size.

bcs-gbl:
Im confused about team-provider-info.json and override.ts :

**override.ts:** [doc](https://docs.amplify.aws/cli/graphql/override/#customize-amplify-generated-resources-for-searchable-opensearch-directive)

I can override this values in this config file:
- instanceCount: 1
- instanceType: 'm6g.large.elasticsearch'
- and elasticsearchVersion = 'OpenSearch_1.1'
but, i cannot customize this values by amplify environments

**team-provider-info.json** [doc](https://docs.amplify.aws/cli/graphql/search-and-result-aggregations/#set-up-opensearch-for-production-environments)
I can override this values in this config file:
- "OpenSearchInstanceType", but cannot instancecount, opensearch version (1.1 latest), and others, 
but the environments are separated.

So,
I'd like to override this values but in different environments because We dont want 
to pay a lot for Opensearch nodes and no need high availability in all other environments (dev, stage, and so on).

I didnt find any info or documentation about options in team-provider-info.json like instancecount.

Can anybody help with some doc, or parameters, or any other info?

benjamindoe:
@ammarkarachi Thanks for updating the docs.

I've tried using `OpenSearchInstanceType` in the `team-provider-info.json` but I'm still running into the same error:
```
Parameters: [OpenSearchInstanceType] do not exist in the template
```

[...]

PR #9840, opened 2022-02-23
fix: move searchable parameters to api stack V2

marcvberg:
#### Description of changes
The parameters for OpenSearch should be created in the API stack since they're not easily set in the child stack. Brings us back into alignment with docs.

A related docs PR to switch from elastic to opensearch: https://github.com/aws-amplify/docs/pull/4066

#### Issue #, if available
#9728

Created an application and validated that I could set the searchable instance type

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


cjihrig (approved):

cjihrig:
Is it possible to add a test here? It seems like a unit test would be enough to verify that the parameter is present.

sachscode (approved):

giulioambrogi:
We're still experiencing this issue, see latest comment in  https://github.com/aws-amplify/amplify-cli/issues/9728

PR #9841, opened 2022-02-24
fix: validate the presence of sortKeyFields when index is on the same field as the primary key

### LINKED ISSUE #4138: Using hash key for @connection + Detect secondary index @key on hash key field as error
```graphql
###########################################################
##                  Users
###########################################################
type User
@model(subscriptions: null)
@auth(rules: [
	{ allow: owner, ownerField: "userID", queries: [get], mutations: [create, update] },
	{ allow: private, queries: [get], mutations: null }, # Other logged in users
	{ allow: groups, groups: ["AdminUsers"] }
])
@key(fields: ["userID"]) # Primary key
{
	userID: ID!
	username: String!
	conversations: [UserConvMapping] @connection(keyName: "userID", fields: ["userID"])
	avatar: S3Object
	createdAt: String!
	updatedAt: String!
}

###########################################################
##                  UserConvMapping
###########################################################
type UserConvMapping
@model(subscriptions: null, queries: null)
@auth(rules: [{ allow: owner, ownerField: "userID" }])
@key(fields: ["userID", "convID"]) # Primary key
@key(name: "userID", fields: ["userID"])
@key(name: "convID", fields: ["convID"])
{
	user: User! @connection(fields: ["userID"])
	userID: ID!
	conversation: Conversation! @connection(fields: ["convID"])
	convID: ID!
	createdAt: String!
	updatedAt: String!
}
```

When I deploy this I have this error:
```
CREATE_FAILED UserConvMappingTable Index KeySchema does not have a range key for index: userID
```

When I look at the transformation:
```json
                "KeySchema": [
                    {
                        "AttributeName": "userID",
                        "KeyType": "HASH"
                    },
                    {
                        "AttributeName": "convID",
                        "KeyType": "RANGE"
                    }
                ],
                "LocalSecondaryIndexes": [
                    {
                        "IndexName": "userID",
                        "KeySchema": [
                            {
                                "AttributeName": "userID",
                                "KeyType": "HASH"
                            }
                        ],
                        "Projection": {
                            "ProjectionType": "ALL"
                        }
                    }
                ],
                "GlobalSecondaryIndexes": [
                    {
                        "IndexName": "convID",
                        "KeySchema": [
                            {
                                "AttributeName": "convID",
                                "KeyType": "HASH"
                            }
                        ],
                        "Projection": {
                            "ProjectionType": "ALL"
                        },
                    }
                ]
```

The local secondary index userID shouldn't be there.
Well the error is there:
```graphql
@key(fields: ["userID", "convID"]) # Primary key
# Yes of course I don't need this index because this is the hash key
@key(name: "userID", fields: ["userID"])
@key(name: "convID", fields: ["convID"])
```

Let's deploy
```
Key userID does not exist for model UserConvMapping
```

What must be done:
- Adding a check that the user don't create a index on a attribute that is also the hash key
- Allow `@connection` on the hash key (like we can query with DynamoDB on the hash key)

flozender:
#### Description of changes
Check if the LSI field is also a primary key and then make sure it has a `sortKeyFields` attribute. Appropriate unit test added.

#### Issue #, if available
Closes #4138

`yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


flozender:
Looks like the test `owner-auth.test.ts` in `amplify-graphql-auth-transformer` has a field with an `index` on a `primaryKey` field with no `sortKeyFields`.

https://github.com/aws-amplify/amplify-cli/blob/fcf9100c778e98e16f0ed695112f2a7251aca06a/packages/amplify-graphql-auth-transformer/src/__tests__/owner-auth.test.ts#L241

What would you suggest I do here to fix this? Thanks!

cjihrig (changes requested):
@flozender you'll need to set the `secondarykeyasgsi` feature flag to `false` in your test, otherwise the Transformer will create a GSI by default. The validation logic will also need to check for that flag.

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
```diff
@@ -88,6 +88,7 @@ export class IndexTransformer extends TransformerPluginBase {
 
 function validate(config: IndexDirectiveConfiguration, ctx: TransformerContextProvider): void {
   const { name, object, field, sortKeyFields } = config;
+  const defaultGSI = ctx.featureFlags.getBoolean('secondaryKeyAsGSI', false);
```
cjihrig, line 91: I think this should have a default value of `true`.

```suggestion
  const defaultGSI = ctx.featureFlags.getBoolean('secondaryKeyAsGSI', true);
```

packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-index-transformer.test.ts
```diff
@@ -52,6 +52,35 @@ test('throws if an invalid LSI is created', () => {
...
+    }`;
+
+  const transformer = new GraphQLTransform({
+    transformers: [new ModelTransformer(), new PrimaryKeyTransformer(), new IndexTransformer()],
+    featureFlags: {
+      getBoolean: jest.fn().mockImplementation((name, defaultValue) => {
+        if (name === 'secondaryKeyAsGSI') {
+          return false;
+        }
+        return;
```
cjihrig, line 69:
```suggestion
        return defaultValue;
```

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
```diff
@@ -107,6 +108,19 @@ function validate(config: IndexDirectiveConfiguration, ctx: TransformerContextPr
     fieldMap.set(objectField.name.value, objectField);
 
     for (const peerDirective of objectField.directives!) {
+      if (
+        !defaultGSI &&
+        peerDirective.name.value === directiveName &&
+        objectField == config.primaryKeyField &&
+        !peerDirective.arguments!.some((arg: any) => arg.name.value === 'sortKeyFields')
```
cjihrig, line 115: This validation has some overlap with the validation on lines 131-134. Is it possible to do some consolidation.

cjihrig (commented):
A few little comments, but this is looking good. Thanks.

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
flozender: added `hasSortFields` to move the check into its own variable.

flozender:
Thanks, I made the changes. Please let me know if something else seems off!

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
```diff
@@ -107,17 +108,24 @@ function validate(config: IndexDirectiveConfiguration, ctx: TransformerContextPr
     fieldMap.set(objectField.name.value, objectField);
 
     for (const peerDirective of objectField.directives!) {
+      const hasSortFields = peerDirective.arguments!.some((arg: any) => arg.name.value === 'sortKeyFields');
+
+      if (!defaultGSI && !hasSortFields && peerDirective.name.value === directiveName && objectField == config.primaryKeyField) {
```
cjihrig, line 113: Does it work to move this validation down? The way this is currently written, it's possible that `config.primaryKeyField` won't be set yet. I'm also not sure that we need the `peerDirective.name.value === directiveName` part of this since we already know we are validating an `@index` directive.

flozender: From what I understand, the `@index` shouldn't exist without a `sortKeyFields` when the `@index` is on the same field as a `@primaryKey`. So the only way `config.primaryKeyField` could be undefined would be if we had something like `foo: @index @primaryKey` would that be possible? Unfortunately, I cannot test it out myself since my AWS account is suspended right now.

cjihrig: > `foo: @index @primaryKey` would that be possible?

Yes, I believe that would be possible.

> Unfortunately, I cannot test it out myself since my AWS account is suspended right now.

Since this only deals with input validation, you shouldn't need to deploy/test this in AWS.

flozender: Thanks for the input, I made the changes and pushed it over again.

flozender: Sorry, looks like moving it down fails quite a few tests. Do you think I should instead run a second `objectField.directives` loop for the validation after the current one so we know `config.primaryKeyField` will have to exist then?

packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-index-transformer.test.ts
```diff
@@ -52,6 +52,47 @@ test('throws if an invalid LSI is created', () => {
...
 });
 
+test('throws if an LSI is missing sort fields', () => {
+  const schema = `
+    type Test @model {
+      id: ID! @primaryKey(sortKeyFields: ["foo"]) @index(name: "index1")
+      foo: ID!
+    }`;
+
+  const schemaInverted = `
```
flozender, line 62: added another test schema for the discussed case

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
```diff
@@ -112,12 +113,10 @@ function validate(config: IndexDirectiveConfiguration, ctx: TransformerContextPr
       }
 
       if (peerDirective.name.value === 'primaryKey') {
+        const hasSortFields = peerDirective.arguments!.some((arg: any) => arg.name.value === 'sortKeyFields');
```
sundersc, line 116: Does it accommodate if sortKeyFields explicitly set as empty `sortKeyFields: []`?

flozender: Yes, it does.

sundersc: This schema failed to compile.
https://github.com/aws-amplify/amplify-cli/blob/abd8c67f7f4b76ece729f34447daf530179e1a1a/packages/amplify-util-mock/src/__e2e_v2__/model-relational-with-key-transformer.e2e.test.ts#L61-L66

sundersc:
Thanks for your contribution. The changes look good, except for the failing test.

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
flozender: Thank you for your response, I was able to figure out what was going wrong. The tests should pass now.

sundersc (approved):
LGTM.

lazpavel (approved):
LGTM

phani-srikar (changes requested):
One unit test to cover a case. Otherwise LGTM

flozender:
Thanks for your comment @phani-srikar, I found that the code wasn't checking if `sortKeyFields` was empty if it was in the `index` directive. I've fixed that and added the test!

packages/amplify-graphql-index-transformer/src/graphql-index-transformer.ts
```diff
@@ -24,6 +24,9 @@ const directiveDefinition = `
   directive @${directiveName}(name: String!, sortKeyFields: [String], queryField: String) repeatable on FIELD_DEFINITION
 `;
 
+/**
+ *
+ */
```
flozender, line 29: `eslint` force added this for `JSDocs`

packages/amplify-graphql-index-transformer/src/__tests__/amplify-graphql-index-transformer.test.ts
```diff
@@ -43,14 +43,83 @@ test('throws if an invalid LSI is created', () => {
       id: ID! @primaryKey @index(name: "index1")
     }`;
 
+  const schemaEmptySortKeyFields = `
```
flozender, line 46: I found that this test case was missed for when using the `sortKeyFields` directive on `primaryKey`, so I added that test as well.

phani-srikar (approved):
Thanks for the changes. LGTM!

PR #9843, opened 2022-02-24
fix: extend container test timeout

sachscode:
#### Description of changes
Container tests for retrieving  api secrets was timing out. Extending the API timeout in Express allows the test to pass. This change increases timeout to 500 secs from 120 secs ( default )

`npm run e2e src/__tests__/containers-api-secrets.test.ts`
#### Checklist

- [*] PR description included
- [*] `yarn test` passes


sundersc (changes requested):
As discussed, container tests are failing due to the problem with flask dependencies. This change may not be required.
https://serverfault.com/questions/1094062/error-from-itsdangerous-import-json-as-json-importerror-cannot-import-name-j

PR #9899, opened 2022-03-02
feat(geo): add headless command for map remove

AaronZyLee:
#### Description of changes

Add headless command for removing map in geo category, including

- Interface and schema definition for headless payload
- Headless entry point for remove in geo category
- Unit and e2e tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-geo/src/provider-controllers/map.ts
+++ b/packages/amplify-category-geo/src/provider-controllers/map.ts
@@ -8,8 +8,9 @@ import { printNextStepsSuccessMessage, setProviderContext } from './index';
 import { ServiceName } from '../service-utils/constants';
 import { printer } from 'amplify-prompts';
 import { getMapStyleComponents } from '../service-utils/mapParams';
-import { GeoServiceConfiguration, GeoServiceModification } from 'amplify-headless-interface';
+import { GeoServiceConfiguration, GeoServiceModification, GeoServiceRemoval } from 'amplify-headless-interface';
 import { merge } from '../service-utils/resourceUtils';
+import { checkGeoResourceExists, updateDefaultResource } from '../service-utils/resourceUtils';
 
 export const addMapResource = async (
   context: $TSContext
@@ -38,33 +39,12 @@ export const updateMapResource = async (
 export const removeMapResource = async (
   context: any
 ): Promise<string | undefined> => {
-  const { amplify } = context;
-  const resourceToRemove = await removeWalkthrough(context, ServiceName.Map);
-  if (!resourceToRemove) return;
+  const mapToRemove = await removeWalkthrough(context, ServiceName.Map);
+  if (!mapToRemove) return;
 
-  const resourceParameters = await getCurrentMapParameters(resourceToRemove);
+  await removeMapResourceWithParams(context, mapToRemove);
 
-  try {
-    await amplify.removeResource(context, category, resourceToRemove)
-    .then(async (resource: { service: string; resourceName: string }) => {
-      if (resource?.service === ServiceName.Map && resourceParameters.isDefault) {
-        // choose another default if removing a default map
-        await updateDefaultMapWalkthrough(context, resource.resourceName);
-      }
-    });
-  } catch (err: $TSAny) {
-    if (err.stack) {
-      printer.error(err.stack);
-      printer.error(err.message);
-      printer.error(`An error occurred when removing the geo resource ${resourceToRemove}`);
-    }
-
-    context.usageData.emitError(err);
-    process.exitCode = 1;
-  }
-
-  printNextStepsSuccessMessage(context);
-  return resourceToRemove;
+  return mapToRemove
 };
 
 export const addMapResourceHeadless = async (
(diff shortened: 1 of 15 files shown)


packages/amplify-headless-interface/schemas/geo/1/RemoveGeoRequest.schema.json
```diff
@@ -0,0 +1,69 @@
...
+                    "type": "string"
+                },
+                "newDefaultResourceName": {
+                    "description": "Optional param. The name of new default resource when the removed one is default",
+                    "type": "string"
+                }
+            },
+            "required": [
+                "name",
+                "serviceName"
```
phani-srikar, line 48: Is the `newDefaultResourceName` also required here?

packages/amplify-e2e-tests/src/__tests__/geo-headless.test.ts
```diff
@@ -40,21 +40,34 @@ describe('Geo headless tests', () => {
...
       await amplifyPushAuth(projRoot);
-      const newMeta = getProjectMeta(projRoot);
-      expect(newMeta.geo[mapId].accessType).toBe('AuthorizedAndGuestUsers');
+      meta = getProjectMeta(projRoot);
+      expect(meta.geo[mapId].accessType).toBe('AuthorizedAndGuestUsers');
+      //remove map
+      await removeHeadlessGeo(projRoot, removeGeoRequest);
+      await amplifyPushAuth(projRoot);
+      meta = getProjectMeta(projRoot);
+      expect(meta.geo[mapId]).toBeUndefined();
```
phani-srikar, line 70: Can you try to add 2 Maps and remove the default map to check if the second will be set to default? That would give completeness to this test. Thanks.

packages/amplify-e2e-tests/src/__tests__/geo-headless.test.ts
```diff
@@ -40,21 +40,34 @@ describe('Geo headless tests', () => {
           setAsDefault: true
         }
       };
+      const removeGeoRequest: RemoveGeoRequest = {
```
phani-srikar, line 43: nit: change the test description to `add/update/remove` geo map ...

packages/amplify-category-geo/src/provider-controllers/map.ts
```diff
@@ -120,3 +104,57 @@ export const updateMapResourceWithParams = async (
...
+        if (isHeadlessCommand) {
+          const remainingMaps = ((await context.amplify.getResourceStatus()).allResources as any[])
+            .filter(resource => resource.service === ServiceName.Map)
+            .map(resource => resource.resourceName);
+          if (remainingMaps.length > 0) {
+            // directly update the new default map when provided
+            if (newDefaultMap && remainingMaps.includes(newDefaultMap)) {
+              await updateDefaultResource(context, ServiceName.Map, newDefaultMap);
+            }
+            // otherwise select first available map as default
```
phani-srikar, line 131: I'm a little skeptical with going this route as this assumes that the user wants to set the "first" available map as default which might not be true. I think it is better to explicitly require the new default map (unless there is only one map left after removing) to be specified by the user.

packages/amplify-category-geo/src/provider-controllers/map.ts
```diff
@@ -120,3 +104,57 @@ export const updateMapResourceWithParams = async (
...
+  const { amplify } = context;
+  const mapParams = await getCurrentMapParameters(mapToRemove);
+  try {
+    await amplify.removeResource(context, category, mapToRemove, { headless: isHeadlessCommand })
+    .then(async (resource: { service: string; resourceName: string } | undefined) => {
+      // choose another default if removing a default map
+      if (resource?.service === ServiceName.Map && mapParams.isDefault) {
+        // select new default map in headless mode
+        if (isHeadlessCommand) {
+          const remainingMaps = ((await context.amplify.getResourceStatus()).allResources as any[])
```
phani-srikar, line 123: Don't we need to filter the current map being removed also to get the remaining maps or at this point the current map info is already deleted from `amplify-meta` ?

phani-srikar (changes requested):

packages/amplify-headless-interface/schemas/geo/1/RemoveGeoRequest.schema.json
AaronZyLee: It is optional.

packages/amplify-category-geo/src/provider-controllers/map.ts
AaronZyLee: The initial thought for optional `newDefaultMap` is that this variable is only useful when a user has more than 2 maps. If you have 2 or 1 map, you do not need to pass this parameter because the next default map is determined. When you have 3 or more maps, there are following scenarios. 

1. If I remove a non-default map, I don't need the optional field in headless payload.
2. If I remove the default map with new default map defined and existing, I can assign a new remaining one as the default map, where the optional field works
3. If I remove the default map with new default map defined but not existing, the first available map will be chosen as default
4. If I remove the default map without defining next default map, the first available map will be chosen as default

Right now the code can handle all the four scenarios. Typically the fourth one is applied when a user have multiple maps and he just wants to remove them all, in which case the new default map does not matter in each remove headless command. That's why the default map will be automatically selected.

packages/amplify-category-geo/src/provider-controllers/map.ts
AaronZyLee: The current map is already deleted from meta

Issue #7848, opened 2021-08-01
MFA token code message is hidden


g-grass:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

npm (6.14.13)

### If applicable, what version of Node.js are you using?

v14.17.3

### Amplify CLI Version

5.2.1

### What operating system are you using?

Mac (Mojave 10.14.6)

### Amplify Categories

Not applicable

### Amplify Commands

push

### Describe the bug

- I'm sorry if this is a misunderstanding. This happened when I created a new project, referring to the [official tutorial](https://docs.amplify.aws/start/getting-started/setup/q/integration/angular#importing-the-amplify-angular-ui-module).
- When I ran the `amplify push` command, it seemed to freeze with the following message displayed.
```
It requires MFA authentication. The MFA device is
  arn:aws:iam::012345678910:mfa/sample-mfa
⠙ Fetching updates to backend environment: dev from the cloud.
```

- However, when I pressed the Enter key in this state, the following error message was output:
```
  arn:aws:iam::012345678910:mfa/sample-mfa
⠙ Fetching updates to backend environment: dev from the cloud.
>> Must have length equal to 6
```
- This was the first time I noticed that I was waiting for a MFA token code to be entered.

### Expected behavior

- I hope it looks like this:
```
It requires MFA authentication. The MFA device is
  arn:aws:iam::012345678910:mfa/sample-mfa
? Enter the MFA token code:
```

### Reproduction steps

- Is is basically in line with the official tutorial.
1.  amplify init

<details>

```bash
$ amplify init
Note: It is recommended to run this command from the root of your app directory
? Enter a name for the project debugapp
The following configuration will be applied:

Project information
| Name: debugapp
| Environment: dev
| Default editor: Visual Studio Code
| App type: javascript
| Javascript framework: angular
| Source Directory Path: src
| Distribution Directory Path: dist/amplify-app
| Build Command: npm run-script build
| Start Command: ng serve

[...]

PR #9954, opened 2022-03-10
fix: show mfa prompt

johnpc:
#### Description of changes

This PR fixes an issue originally resolved by https://github.com/aws-amplify/amplify-cli/pull/7866, which was reverted due to e2e failures. 

The issue is that the MFA token prompt was swallowed by the ora spinner and is not visible to the developer.

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/7848

Manually set up mfa and ran amplify push. See [before](https://user-images.githubusercontent.com/4824042/128068323-40ac6cf4-0eb6-4c81-9f4e-bda0ed007ef9.mov) and [after](https://user-images.githubusercontent.com/4824042/128068364-6a8ceebb-4a6d-4d25-8f11-18eafd17643e.mov) screen recordings.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
```diff
@@ -93,15 +95,19 @@ export class SSMClientWrapper {
...
     await this.ssmClient.deleteParameters({ Names: secretNames }).promise();
   };
 }
 
-const getSSMClient = async (context: $TSContext) => {
+const getSSMClient = async (context: $TSContext): Promise<aws.SSM> => {
   const spinner = ora('Initializing SSM Client');
+  if (context.exeInfo) {
+    context.exeInfo.spinner = spinner;
+  }
```
ammarkarachi, line 109: I have moved the spinner to amplify-cli-core and you can use that instead of passing it through the context

ammarkarachi (changes requested):

packages/amplify-provider-awscloudformation/src/system-config-manager.ts
```diff
@@ -154,9 +163,13 @@ async function getRoleCredentials(context: $TSContext, profileName: string, prof
   }
 
   return roleCredentials;
-}
+};
 
-async function getMfaTokenCode() {
+const getMfaTokenCode = async (context): Promise<string> => {
```
danielleadams, line 168: I think `context` needs a type.
```suggestion
const getMfaTokenCode = async (context: $TSContext): Promise<string> => {
```

packages/amplify-provider-awscloudformation/src/system-config-manager.ts
```diff
@@ -154,9 +163,13 @@ async function getRoleCredentials(context: $TSContext, profileName: string, prof
...
+  let spinner;
+  if (context.exeInfo && context.exeInfo.spinner && context.exeInfo.spinner.isSpinning) {
```
danielleadams, line 170:
```suggestion
  if (context.exeInfo?.spinner?.isSpinning) {
```

danielleadams (changes requested):

packages/amplify-provider-awscloudformation/src/system-config-manager.ts
danielleadams: Based on what @ammarkarachi said, my comments may be obsolete

packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
johnpc: I'm not sure whether I can use this, since `amplify-cli-core` creates a new spinner every time but `context.exeInfo.spinner` is a singleton where I can check if it's already spinning or not.

danielleadams (approved):

packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
ammarkarachi: We want to reduce the dependency on the context object and the imported instance from the core is a singleton

johnpc: I might be misunderstanding but it doesn't look like a singleton here: https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-cli-core/src/spinner/index.ts#L3

ammarkarachi (approved):

PR #9959, opened 2022-03-10
fix(test): check aws-exports on sandbox pull

sundersc:
Update sandbox pull e2e test to validate aws-exports.js, models/index.js and models/schema.js files.


jhockett (approved):

marcvberg (approved):
LGTM

PR #9978, opened 2022-03-14
chore: explicitly declare some dependencies we're pulling implicitly today

alharris-at:
#### Description of changes
Currently, some of the packages in the CLI repo have expectations around dependencies existing which they get transitively through lerna (i.e. I don't declare a dependency on `@types/lodash` but it exists in the root because at least 2 deps in the repo do depend on it). This makes a few of those explicit. These were identified while working through a workstream to separate graphql transformers into a new repo (which will be tracked separately).

#### Issue #, if available
N/A

Built and ran `yarn test` on the package.

After building, there were no changes to `yarn.lock` which is expected, since there should be no new or changed package version associated with this change, just ensuring that if we were to pull these repos apart, we maintain the right dep tree.

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,6 +88,7 @@
     "@commitlint/config-lerna-scopes": "^15.0.0",
     "@commitlint/cz-commitlint": "^15.0.0",
     "@commitlint/prompt": "^15.0.0",
+    "@types/glob": "^7.1.1",
     "@types/jest": "^26.0.20",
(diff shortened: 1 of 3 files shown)


jhockett (approved):

marcvberg (approved):
LGTM

PR #9980, opened 2022-03-14
fix(graphql): avoid duplicate function when overriding resolvers

### LINKED ISSUE #9650: overrid an auto generated pipeline resolver create another function
>**Note**: If your question is regarding the AWS Amplify Console service, please log it in the 
[AWS Amplify Console repository](https://github.com/aws-amplify/amplify-console/issues)

I overriden an auto generated pipline resolver (just the res.vtl file) using this gide https://docs.amplify.aws/cli/graphql/custom-business-logic/#override-amplify-generated-resolvers  it overrid it just fine, but  it also created another function with the same res.vtl and $util.toJson({}) for the req.vtl

**Amplify CLI Version** 
7.6.13
You can use `amplify -v` to check the amplify cli version on your system

**What AWS Services are you utilizing?**

**Provide additional details e.g. code snippets**

sundersc:
#### Description of changes

Avoid creating a duplicate function when overriding a generated resolver.

#### Issue #, if available

Fixes #9650 

- E2E test
- Manual test

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -162,14 +162,59 @@ export class TransformerResolver implements TransformerResolverProvider {
       slotEntry = [];
     }
 
-    slotEntry.push({
-      requestMappingTemplate,
-      responseMappingTemplate,
-      dataSource,
-    });
+    if (!this.updateSlotIfExists(slotName, requestMappingTemplate, responseMappingTemplate)) {
+      slotEntry.push({
```
alharris-at, line 166: I think we can improve the understandability of this snippet here. Having `updateSlotIfExists` return control flow signals seems like one of those things we'll confuse or mix up later on. I'd prefer if we either 1) treat this method as an upsert taking the 4 params we're using in both places, or 2) have control flow helper that determines if slotExists, then a separate push/update flow that follows.

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -162,14 +162,59 @@ export class TransformerResolver implements TransformerResolverProvider {
...
+    }
     this.slotMap.set(slotName, slotEntry);
   };
 
+  updateSlotIfExists = (
+    slotName: string,
+    requestMappingTemplate?: MappingTemplateProvider,
+    responseMappingTemplate?: MappingTemplateProvider,
+  ): boolean => {
+    const slotEntires = this.slotMap.get(slotName)!;
```
alharris-at, line 180: typo on Entires

alharris-at: You also don't need the bang here, since we're null checking down below, right?

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -162,14 +162,59 @@ export class TransformerResolver implements TransformerResolverProvider {
...
+    let slotIndex = 1;
+
+    if (!slotEntires) {
+      return false;
+    }
+
+    for (const slotItem of slotEntires) {
```
alharris-at, line 187: Rather than needing a mutable `slotIndex` can we just call `slotEntries.forEach((slotEntry, slotIndex) => { /** your logic below **/ };`

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -162,14 +162,59 @@ export class TransformerResolver implements TransformerResolverProvider {
...
+      const [slotItemRequestMappingTemplate, slotItemResponseMappingTemplate] = [
+        (slotItem.requestMappingTemplate as any)?.name ?? '',
```
alharris-at, line 189: Is there a specific type we can use here instead of `any`?

packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
```diff
@@ -3531,139 +3531,6 @@ $util.toJson({})
 #set( $filterArgsMap = {} )
 ## [End] Set map initialization for @key **
 ## [Start] Set query expression for @key **
-$util.qr($QueryMap.put('status+createdAt' , 'ByStatus'))
```
alharris-at, line 3534: I'm not really able to tell from this test on it's own, but to clarify, what's being kept here is the customer overridden logic only, not the one we'd write?

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
```diff
@@ -162,14 +162,59 @@ export class TransformerResolver implements TransformerResolverProvider {
...
+        .map(name => name.replace('{slotName}', slotName).replace('{slotIndex}', slotIndex));
+
+      // If both request and response mapping templates are inline, return false
+      if (slotItemRequestMappingTemplate === '' && slotItemResponseMappingTemplate === '') {
+        return false;
+      }
+
+      // If name matches, then it is an overridden resolver
+      if (
+        slotItemRequestMappingTemplate === ((requestMappingTemplate as any)?.name ?? '')
```
lazpavel, line 201: nit: can you use $TSAny if there isn't a better alternative?

lazpavel (approved):
LGTM

packages/amplify-graphql-index-transformer/src/__tests__/__snapshots__/amplify-graphql-index-transformer.test.ts.snap
sundersc: That's right. The customer overridden logic will replace the one we create.

packages/amplify-e2e-tests/src/__tests__/resolvers.test.ts
```diff
@@ -47,6 +47,30 @@ describe('user created resolvers', () => {
...
+      await generateModels(projectDir);
+      await amplifyPushUpdate(projectDir);
+
+      const todoJsonPath = join(projectDir, 'amplify', 'backend', 'api', apiName, 'build', 'stacks', 'Todo.json');
+      const todoJson = JSON.parse(fs.readFileSync(todoJsonPath).toString());
+
+      expect(fs.readFileSync(generatedResolverPath).toString()).toEqual(slot);
+      expect(todoJson.Resources.GetTodoResolver.Properties.PipelineConfig.Functions).toHaveLength(3);
+      // The function count should be 3 even after overriding the auth resolver
+      expect(todoJson.Resources.ListTodoResolver.Properties.PipelineConfig.Functions).toHaveLength(3);
```
phani-srikar, line 72: If it's not too much additional work, can we add a check that the contents of the custom resolver slot are preserved and not overwritten? Thanks

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
phani-srikar, line 201: In the case the user only wants to override the request mapping template but not the response VTL, will this logic work?

phani-srikar (approved):

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
sundersc: Good catch. Refactored into 3 different functions - findSlot, updateSlot and slotExists.

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
sundersc: Updated.

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
sundersc: This block requires early termination if match is found. The only way to early exit foreach is using exception, that would look ugly. I have also considered the usage of every() and some(), however `for .. of ...` seems to be the best fit here. I'm open to suggestions if there is a better way.

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
sundersc: Unfortunately no. We could alter the class structure we have for storing templates but that would be outside the scope of the change intended here and requires quite a bit of refactoring.

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
sundersc: $TSAny is defined on `amplify-cli-core`. Currently, amplify-graphql-transformer-core doesn't have a dependency on `amplify-cli-core`. We can consider adding something similar to $TSAny type within data package once we split the repository.

sundersc: Yes. For example, if response VTL is not overridden, `(responseMappingTemplate as any)?.name` would be `undefined`. In that case, the slot's response VTL is preserved.

```js
slot.responseMappingTemplate = (responseMappingTemplate as any)?.name
        ? responseMappingTemplate
        : slot.responseMappingTemplate;
```

packages/amplify-e2e-tests/src/__tests__/resolvers.test.ts
sundersc: Updated the e2e test to check the overridden resolver folder and make sure that the file exists and the contents match the custom resolver content.

packages/amplify-graphql-transformer-core/src/transformer-context/resolver.ts
alharris-at: Ahh, I see. I'm fine w/ this as-is in that case.

alharris-at (commented):
Generally LGTM - one n.p. on the `slotExists` method.

alharris-at (approved):

PR #9984, opened 2022-03-15
fix: use lib paths instead of src to reference submodule imports across data-cli split

alharris-at:
#### Description of changes
While prepping to split the first round of data and cli packages, I realized we had a few places where we were importing locally from src, rather than from `lib` of the distributed package. This addresses the two issues I spotted. I was hoping to get a generalized solution and check in place using something like tslint/no-submodule-imports, but it looks like we aren't onboarded w/ tslint yet, and I couldn't get the eslint package no-internal-modules working quite as expected, so ended up just addressing these two.

#### Issue #, if available
N/A

Built and ran unit tests.

Current path imports (fails unless local)
* https://unpkg.com/@aws-amplify/graphql-transformer-core@0.16.2/src/transformation/types.d.ts
* https://unpkg.com/graphql-transformer-core@7.4.9/src/TransformerContext.d.ts

Updated path imports
* https://unpkg.com/@aws-amplify/graphql-transformer-core@0.16.2/lib/transformation/types.d.ts
* https://unpkg.com/graphql-transformer-core@7.4.9/lib/TransformerContext.d.ts

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/graphql-transformer/transform-graphql-schema.ts
+++ b/packages/amplify-provider-awscloudformation/src/graphql-transformer/transform-graphql-schema.ts
@@ -21,7 +21,7 @@ import {
 } from '@aws-amplify/graphql-transformer-core';
 import { Template } from '@aws-amplify/graphql-transformer-core/lib/config/project-config';
 import { MapsToTransformer } from '@aws-amplify/graphql-maps-to-transformer';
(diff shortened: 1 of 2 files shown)


jhockett (approved):
Changes LGTM, but I would prefer to see these types exported from `@aws-amplify/graphql-transformer-core` and `graphql-transformer-core` respectively unless there's a reason not to.

PR #10030, opened 2022-03-21
fix(graphql-relational-transformer): belongs to when linked to hasMany with primary/sort key fields

lazpavel:
#### Description of changes
- adds has many sort key to the belongs to related type for proper lookup
- lint related changes

- [x] `yarn test`
- [x] `yarn cloud-e2e`
- [x] manual testing
- [x] unit test added

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


alharris-at:
What part of this PR are the actual changes? I'm scanning through and it looks like the majority is lint-related, I want to make sure I'm digging into the right part of the change.

packages/amplify-graphql-relational-transformer/src/schema.ts
```diff
@@ -120,179 +137,235 @@ export function ensureHasOneConnectionField(config: HasOneDirectiveConfiguration
...
+  config.connectionFields.push(connectionFieldName, ...sortKeyFields.map(it => it.name.value));
+
   const connectionAttributeName = getConnectionAttributeName(object.name.value, field.name.value);
 
   const relatedTypeObject = ctx.output.getType(relatedType.name.value) as ObjectTypeDefinitionNode;
   if (relatedTypeObject) {
-    ctx.output.putType(updateTypeWithConnectionField(relatedTypeObject, connectionAttributeName, isNonNullType(field.type)));
+    const updatedFields = [...relatedTypeObject.fields!];
+    updatedFields.push(...getTypeFieldsWithConnectionField(updatedFields, connectionAttributeName, isNonNullType(field.type)));
+    sortKeyFields.forEach(it => {
```
lazpavel, line 253: @alharris-at this is one the actual change, adding the sortKeyFields from the hasMany object to the belongsTo object for proper lookup

phani-srikar: If we add the sortKeyFields from the hasMany object w/o any renaming, there might be duplicates if the belongsTo model has same field names as sortKeyFields right? I believe that is the reason why we generate a `{hasManyModelName}{belongsToModelName}Id` field versus using the direct primary key field name from the hasMany model.

lazpavel: That is right, will update to use same naming structure for sortKeyFields as well

packages/amplify-provider-awscloudformation/src/graphql-transformer/transform-graphql-schema.ts
```diff
@@ -73,112 +76,114 @@ function warnOnAuth(map, docLink) {
...
+    identityPoolId: options.identityPoolId,
+  });
+  const transformerList: TransformerPluginProvider[] = [
+    modelTransformer,
+    new FunctionTransformer(),
+    new HttpTransformer(),
+    new PredictionsTransformer(options?.storageConfig),
+    new PrimaryKeyTransformer(),
+    indexTransformer,
+    new HasManyTransformer(),
```
phani-srikar, line 100: Is it intended to change the order of execution of the transformers?

phani-srikar (commented):
The changes look good. I would wait for other approvals. The lint changes make it hard to keep track of actual changes.

packages/amplify-graphql-relational-transformer/src/schema.ts
```diff
@@ -120,179 +137,243 @@ export function ensureHasOneConnectionField(config: HasOneDirectiveConfiguration
...
+  });
+
+  config.connectionFields.push(connectionFieldName, ...sortKeyFields.map(it => it.name.value));
+
   const connectionAttributeName = getConnectionAttributeName(object.name.value, field.name.value);
 
   const relatedTypeObject = ctx.output.getType(relatedType.name.value) as ObjectTypeDefinitionNode;
   if (relatedTypeObject) {
-    ctx.output.putType(updateTypeWithConnectionField(relatedTypeObject, connectionAttributeName, isNonNullType(field.type)));
+    const updatedFields = [...relatedTypeObject.fields!];
```
alharris-at, line 251: Why do we need the bang here? Should we be defaulting to an empty list, or have a type that defines a required list?

lazpavel: ObjectTypeDefinitionNode has fields optional, but since this is a related type it has to be an @model so it will always have fields or this can be rewritten to `relatedTypeObject.fields ?? []`

lazpavel:
closed in favor of https://github.com/aws-amplify/amplify-category-api/pull/462

PR #10073, opened 2022-03-25
fix(graphql-model-transformer): fix create mutation when index field is null

lazpavel:
#### Description of changes
- GSI doesnt allow null for the partition key. The solution is to not include that attribute when it is the partition key of a GSI and its null which will ensure the item is never propagated to the GSI which will be the correct behavior and save storage and write cost

- manual testing
- e2e test added

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


alharris-at (approved):
Overall lgtm, but I had to go through and do a bit of reading on the VTL template and how the utils/ctx work in general today, so you probably also want to grab a shipit from somebody more authoritative :D

packages/amplify-graphql-model-transformer/package.json
```diff
@@ -44,6 +44,7 @@
         "@aws-cdk/custom-resources": "~1.124.0",
         "@aws-cdk/cx-api": "~1.124.0",
         "@aws-cdk/region-info": "~1.124.0",
+        "amplify-cli-core": "^2.5.0",
```
sundersc, line 47: IMO we shouldn't introduce dependency on CLI core for GraphQL transformers.

packages/graphql-transformers-e2e-tests/package.json
```diff
@@ -22,6 +22,7 @@
     "build-tests": "yarn tsc --build tsconfig.tests.json"
   },
   "dependencies": {
+    "amplify-cli-core": "^2.5.0",
```
sundersc, line 25: Same comment as above. We should avoid this dependency.

sundersc (commented):
LGTM except for the CLI core dependency.

packages/graphql-transformers-e2e-tests/package.json
lazpavel: updated

packages/amplify-graphql-model-transformer/package.json
lazpavel: updated

sundersc (approved):
LGTM.

PR #10080, opened 2022-03-27
fix(amplify-util-mock): Change exposed attr to reference correct attr

loganhersh:
#### Description of changes
Changes the field name provided by the FunctionArn exposed attribute for processed appsync function resources so that it references the correct property.

Previously, the exposed attribute pointed to the 'Ref' field which did not exist. It should point to the 'ref' field. This was causing 'amplify mock api' to fail if the resource's FunctionArn was referenced later on in the template.

I've also updated a related test file to fix a broken test case and lint errors.

#### Issue #, if available
No existing issue.

Added a unit test to verify that the property referenced by FunctionArn is defined in the resource.
I've also made this change in my local development environment where I came upon this issue.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


jhockett (approved):
LGTM, thanks for the contribution @loganhersh!

johnpc (approved):
lgtm

PR #10165, opened 2022-04-08
chore: add dependencies necessary for aws-cli and cypress

johnpc:
our pipeline now requires `awscli` and `cypress`, so our docker image needs to be updated to include all required dependencies so we don't have to install them at run time.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,24 @@ RUN sudo npm install -g npm@7
 RUN npm -v
 WORKDIR /tmp
 RUN sudo apt-get update && sudo apt-get install -y \
-  xdg-utils
+  xdg-utils \
+  libatk-bridge2.0-0 \
+  libgtk-3.0 \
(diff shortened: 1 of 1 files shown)


jhockett (approved):

Dockerfile
```diff
@@ -3,7 +3,24 @@ RUN sudo npm install -g npm@7
...
+  zip \
+  lsof \
+  jq \
+  groff \
+  python \
+  python-pip \
+  libpython-dev \
+  default-jdk
+
+RUN sudo pip install awscli
```
sachscode, line 23: Do you need to install cypress as well ?

sachscode (commented):
one question

sachscode (approved):

Dockerfile
johnpc: no, it gets installed via npm inside the uibuilder test

Issue #9362, opened 2021-12-21
 Custom Transformer v2 - Error: Template asset can be used only with TransformerStackSynthesizer


mrsimply:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.

### How did you install the Amplify CLI?

curl

### If applicable, what version of Node.js are you using?

v17.0.1

### Amplify CLI Version

7.6.4

### What operating system are you using?

Mac

### Amplify Categories

api

### Amplify Commands

Not applicable

### Describe the bug

When **adding a slot** to an existing resolver, the following error is thrown: `Template asset can be used only with TransformerStackSynthesizer`. It refers to `amplify-graphql-transformer-core/src/cdk-compat/file-asset.ts`

### Expected behavior

I expect additional VTL files to be automatically generated when adding a slot. In my example `Mutation.createTodo.auth.2.req.vtl`

The Transformer v2 is very sparsely documented - any fix, help or workaround is appreciated 😥

### Reproduction steps

1. Init amplify
2. Create GraphQL API with Basic Cognito Auth and the Demo "Todo" GraphQL Schema
3. Create a Custom Transformer (code below), configure `transform.conf.json`, add @tenant directive to GraphQL Schema
3. Run `amplify push`

Custom Transformer code:
```ts
import {
  MappingTemplate,
  TransformerPluginBase,
} from '@aws-amplify/graphql-transformer-core';
import {
  TransformerContextProvider,
  TransformerResolverProvider,
  TransformerSchemaVisitStepContextProvider,
} from '@aws-amplify/graphql-transformer-interfaces';
import { DirectiveNode, ObjectTypeDefinitionNode } from 'graphql';
import { printBlock, raw } from 'graphql-mapping-template';

const directiveName = 'tenant';
const directiveDefinition = `
  directive @${directiveName} on OBJECT
`;

export class TenantTransformer extends TransformerPluginBase {
  private typesWithModelDirective: Set<string> = new Set();

  constructor() {
[...]

PR #10188, opened 2022-04-11
fix: improve instanceof checks to support custom transformers

johnpc:
This commit updates two `instanceof` checks to instead check
the constructor name. This is done to better support custom
transformers. Because custom transformers `import`/`require`
potentially different copies of the transformer modules,
the `instanceof` checks do not work. This required v2 custom
transformers to rely on `NODE_PATH` hacks and other
workarounds.

Fixes: https://github.com/aws-amplify/amplify-cli/issues/9362

Duplicates https://github.com/aws-amplify/amplify-cli/pull/9970 because @cjihrig's fork of amplify-cli no longer exists so the change cannot be rebased.

### DIFF
--- a/packages/amplify-graphql-transformer-core/src/cdk-compat/file-asset.ts
+++ b/packages/amplify-graphql-transformer-core/src/cdk-compat/file-asset.ts
@@ -20,8 +20,11 @@ export class FileAsset extends cdk.Construct implements cdk.IAsset {
     const rootStack = findRootStack(scope);
     const sythesizer = rootStack.synthesizer;
 
-    if (sythesizer instanceof TransformerStackSythesizer) {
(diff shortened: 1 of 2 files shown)


jhockett (approved):

cjihrig:
https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit :smile:

johnpc:
The author doesn't matter. I could not clone down the change to rebase it because the original branch is a 404 page: https://github.com/cjihrig/amplify-cli/tree/instancenotof

Also, I could not use GitHub's "resolve conflicts" feature - it caused GitHub to get stuck in an error loop.

johnpc:
Oh unless you are asking me to do this to properly credit you for writing this code :P. One sec

akshbhu (approved):

PR #10218, opened 2022-04-15
ci: reduce integration-test pipeline step flakiness

johnpc:
#### Description of changes

This change moves the installation of cypress to _before_ the react server has started. This is necessary because there is a race condition where the react development server is starting at the same time that yarn is mucking with `node_modules` contents. This can cause the react server to not start and the pipeline step to fail and require a retry.

#### Issue #, if available

n/a

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


sachscode (approved):

phani-srikar (approved):
verified that this fixes the issue we saw in Data repo test pipeline. Looks good.

PR #10228, opened 2022-04-15
chore(amplify-graphiql-explorer): upgrade react-scripts

johnpc:
This PR updates react-scripts in amplify-graphiql-explorer.

This required:

1. running `yarn policies set-version 1.18` to allow appropriate lerna dependency behavior
2. upgrading react-scripts to ^5.0.0
3. ejecting from react-scripts to add polyfills to the webpack config necessary for the `jsonwebtoken` dependency
4. adding a script to fixup the `semantic-ui` dependency at build time to work properly with webpack5
5. using `"skipLibCheck": true` in our tsconfig setup to prevent builds from failing due to the new dependencies
6. upgrading windows e2e test harness to use node 14, since new dependencies don't support node 12.


packages/amplify-graphiql-explorer/scripts/patch.js
```diff
@@ -0,0 +1,74 @@
+const fs = require("fs");
+const path = require("path");
+const nodeModulesPath = path.resolve(`${process.cwd()}/../..`, "node_modules");
```
johnpc, line 3: Builds were failing because of an incompatibility with `semantic-ui` and `webpack5`. A patch package was created to bridge the issue (https://github.com/Semantic-Org/Semantic-UI/issues/7073#issuecomment-1016579441). 

This script is a lightly modified version of [that patch package](https://github.com/layershifter/semantic-ui-tools/blob/master/packages/css-patch/src/index.js#L6) modified to work with our lerna setup.

packages/amplify-graphiql-explorer/config/env.js
```diff
@@ -0,0 +1,104 @@
+'use strict';
```
johnpc, line 1: Almost all the changes in this PR come from running CRA's eject script, which is necessary to add the polyfills needed in this package.

packages/amplify-graphiql-explorer/config/webpack.config.js
```diff
@@ -0,0 +1,762 @@
...
+    },
+    resolve: {
+      // This allows you to set a fallback for where webpack should look for modules.
+      // We placed these paths second because we want `node_modules` to "win"
+      // if there are any conflicts. This matches Node resolution mechanism.
+      // https://github.com/facebook/create-react-app/issues/253
+      modules: ['node_modules', paths.appNodeModules].concat(
+        modules.additionalModulePaths || []
+      ),
+      fallback:  {
```
johnpc, line 301: These fallbacks I added to this generated file to support `stream`, `util`, `crypto` and `buffer` required by the `jsonwebtoken` dependency in this package.

tsconfig.base.json
```diff
@@ -22,6 +22,7 @@
 
     /* Strict Type-Checking Options */
     "strict": true /* Enable all strict type-checking options. */,
+    "skipLibCheck": true,
```
johnpc, line 25: typescript was having some problems digesting the fact that `@types/eslint` and `@types/eslint-scope` are both in the dependency tree of this package.

.yarnrc
```diff
@@ -0,0 +1,5 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+yarn-path ".yarn/releases/yarn-1.18.0.cjs"
```
johnpc, line 5: this silenced an error I was seeing when adding dependencies to this package

```
error An unexpected error occurred: "expected workspace package to exist for \"@babel/template\"".
```

Resolved by running `yarn policies set-version 1.18` which generated these files. See https://github.com/yarnpkg/yarn/issues/7807 for more

lazpavel (approved):

ammarkarachi (approved):

sachscode (approved):
LGTM - changes are all localized to graphiql-explorer
Since this is GraphiQL related, it would be good to test this manually after clearing all local state. 
( create-app + use graphiql + push + pull with amplify-tagged-release/dev + graphiql ).
we must test the same with datastore.

PR #10239, opened 2022-04-18
fix: remove prompt from rds headless pull

johnpc:
#### Description of changes

Currently, there is a bug for applications using RDS-backed apis with Studio / headless CI environments:

```
amplify init
amplify add api
amplify api add-graphql-datasource # using instructions from https://docs.amplify.aws/cli-legacy/graphql-transformer/relational/
amplify push 
amplify pull # prompts/fails if headless args are set 
```

The root cause of the issue is the RDS metadata is stored in team-provider-info.json, which is not available on `amplify pull` from an empty workspace.

This PR implements a bit of a ~~hack~~ creative solution that reads the necessary values from pulled cloud formation templates when available instead of prompting.

This can be removed when team-provider-info is moved to a service.

#### Issue #, if available

n/a

- manually created rds app and pulled it headlessly. failed without the change, succeeded with the change.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
+++ b/packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
@@ -1,9 +1,11 @@
-import { $TSContext, $TSObject, exitOnNextTick, ResourceCredentialsNotFoundError, ResourceDoesNotExistError } from 'amplify-cli-core';
+import { $TSContext, $TSObject, exitOnNextTick, ResourceCredentialsNotFoundError, ResourceDoesNotExistError, pathManager } from 'amplify-cli-core';
 import { printer, prompter } from 'amplify-prompts';
 import chalk from 'chalk';
 import { DataApiParams } from 'graphql-relational-schema-transformer';
 import ora from 'ora';
+import fs from 'fs-extra';
 
+const cfnRootStackFileName = 'root-cloudformation-stack.json';
 const spinner = ora('');
 const category = 'api';
 const providerName = 'awscloudformation';
@@ -42,8 +44,14 @@ export async function serviceWalkthrough(context: $TSContext, datasourceMetadata
 
   const { inputs, availableRegions } = datasourceMetadata;
 
+  // FIXME: We should NOT be treating CloudFormation templates as inputs to prompts! This a temporary exception while we move team-provider-info to a service.
+  const cfnJson = fs.readJSONSync(`${pathManager.getCurrentCloudRootStackDirPath(pathManager.findProjectRoot())}/${cfnRootStackFileName}`);
+  const cfnJsonParameters = cfnJson?.Resources[`api${appSyncApi}`]?.Properties?.Parameters || {};
+  let selectedRegion = cfnJsonParameters.rdsRegion;
   // Region Question
-  const selectedRegion = await promptWalkthroughQuestion(inputs, 0, availableRegions);
+  if (!selectedRegion) {
+    selectedRegion = await promptWalkthroughQuestion(inputs, 0, availableRegions);
+  }
 
   const AWS = await getAwsClient(context, 'list');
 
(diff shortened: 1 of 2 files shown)


yppartyk (approved):

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
```diff
@@ -42,8 +44,14 @@ export async function serviceWalkthrough(context: $TSContext, datasourceMetadata
 
   const { inputs, availableRegions } = datasourceMetadata;
 
+  // FIXME: We should NOT be treating CloudFormation templates as inputs to prompts! This a temporary exception while we move team-provider-info to a service.
```
sachscode, line 47: could we please move this part into a function. like 
.. application state ..
const execMode = context.getIsModeInteractive() // check mode is -y or --headless
const template = get<name>CFNTemplate() 

.. business logic...
const selectedRegion = getSelectedRegion( template, execMode). 

Imo it's safer to not rely on cloudformation input to infer execution state. It would be great if we can get this state from  a single signal.

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
```diff
@@ -70,6 +88,18 @@ export async function serviceWalkthrough(context: $TSContext, datasourceMetadata
   };
 }
 
+async function getClusterResourceIdFromArn(arn: string|undefined, AWS) {
```
sachscode, line 91: please change this to get`RDS`ClusterResourceID .

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
```diff
@@ -70,6 +88,18 @@ export async function serviceWalkthrough(context: $TSContext, datasourceMetadata
...
+  if (!arn) {
+    return;
```
sachscode, line 93: Is there a case when the arn is empty? if so, could you pl add a comment here. If not do we need to throw an error ?

sachscode (commented):
LGTM, just a few comments

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
johnpc: We're not inferring the execution state here, just bypassing the prompt if the necessary data is already available. The code here is not aware of execution state.

johnpc: the prompt itself will fail in headless mode if the value can't be derived from available info (or prompt if mode is not headless). this is handled by the prompter library.

sachscode: I see, so the code is agnostic to the execution mode and the prompter returns the error-message for API failures of non-interactive calls . 
This is okay for now and we can make the change as part of the bigger CLI refactor: 

With the current implementation we may face problems when we scale the application beyond the current implementation to say add more states of execution or different input modes. 
Imo, The Prompter library should be state agnostic.  However business-logic is better state-aware.
`Client=>HeadlessAPI-Call=> ValidateRequestAgainstState  => WalkthroughInHeadlessMode => Prompter`
                                                     (api failure is handled here)
- what we have is
`Client=>HeadlessAPI-Call=> WalkthroughModeAgnostic => Prompter=>Validate

sachscode (approved):

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
johnpc: Agreed - the good news is we do have plan to resolve this with @edwardfoyle's work on extracting this metadata into a service. Once that is complete we can remove this (but it may be a few months, and customers are in a broken state today).

akshbhu (commented):
LGTM

I have one question . Since we have fetching cfn parameter values, Are these values not present `parameters.json` file ?

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
```diff
@@ -42,8 +44,14 @@ export async function serviceWalkthrough(context: $TSContext, datasourceMetadata
 
   const { inputs, availableRegions } = datasourceMetadata;
 
+  // FIXME: We should NOT be treating CloudFormation templates as inputs to prompts! This a temporary exception while we move team-provider-info to a service.
+  const cfnJson = fs.readJSONSync(`${pathManager.getCurrentCloudRootStackDirPath(pathManager.findProjectRoot())}/${cfnRootStackFileName}`);
+  const cfnJsonParameters = cfnJson?.Resources[`api${appSyncApi}`]?.Properties?.Parameters || {};
+  let selectedRegion = cfnJsonParameters.rdsRegion;
```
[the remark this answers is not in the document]
johnpc: is this necessary given it's using `|| {}` on the previous line?

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
```diff
@@ -70,6 +88,19 @@ export async function serviceWalkthrough(context: $TSContext, datasourceMetadata
...
 }
 
+async function getRdsClusterResourceIdFromArn(arn: string|undefined, AWS) {
+  // If the arn was not already existing in cloudformation template, return undefined to prompt for input.
+  if (!arn) {
+    return;
+  }
+
+  const RDS = new AWS.RDS();
+  const describeDBClustersResult = await RDS.describeDBClusters().promise();
```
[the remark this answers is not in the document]
johnpc: No, we already have existing calls to `RDS.describeDBClusters` in this file

packages/amplify-category-api/src/provider-utils/awscloudformation/service-walkthroughs/appSync-rds-walkthrough.ts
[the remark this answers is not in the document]
akshbhu: yeah I didn't catch that

akshbhu (approved):

PR #10251, opened 2022-04-19
chore: separate graphql-transformer and resource-mangers for future refactor

alharris-at:
#### Description of changes
Create a new directory called `graphql-resource-manager` within the `amplify-provider-awscloudformation` to tease out some of the lower-level deployment level concerns that are tightly coupled with our current provider, and the directive transformer logic that lives in the `graphql-transformer` directory. This creates a 1-way dependency where graphql-transformer will still have some dependencies on the provider until we've refactored the multi-step deployments a bit, but doesn't create a circular dependency back from the provider into the `graphql-transformer` lib. This should free us up to hoist the `compileSchema` and `getTransformerDirectives` APIs out of provider and into the API category as a next step. This PR is separated from that primarily for readability.

There are no functional changes in this PR, mostly files were just moved wholesale over to the new director (including tests) with the exception of the `graphql-transformer/utils.ts` file, which was separated on a per-utility-fn basis since some of the utilities and types applied to the resource management process, and some were related to the graphql-transformer itself. There was one utility that was used in `push-resources.ts` which was just inlined there instead of living in a separate file.

#### Issue #, if available
N/A

`yarn clean && yarn setup-dev && yarn test`

#### Checklist
- [X] PR description included
- [X] `yarn test` passes


packages/amplify-provider-awscloudformation/src/graphql-transformer/utils.ts
```diff
@@ -4,37 +4,21 @@ import { TransformerProjectConfig, DeploymentResources } from '@aws-amplify/grap
...
-
-export type DiffChanges<T> = Array<Diff<DiffableProject, DiffableProject>>;
-
-export interface GQLDiff {
-  diff: DiffChanges<DiffableProject>;
-  next: DiffableProject;
-  current: DiffableProject;
-}
 
 export const getIdentityPoolId = async (ctx: $TSContext): Promise<string | undefined> => {
```
sundersc, line 23: I don't have complete context on this. Should we move `getIdentityPoolId` and `writeDeploymentToDisk` to resource-manager folder?

alharris-at: These utilities are both used downstream of `transformGraphQLSchema`, which I'm planning to port over to our library. We're `transformGraphQLSchema` is pretty tightly coupled to the provider in general, but I think that the refactor effort for our GQL compilation should address this a bit better. We'll still be somewhat tightly coupled, but I'd bias towards having that coupling live in our domain for the time being than live in the provider.

marcvberg (approved):

sundersc (approved):
LGTM.

packages/amplify-provider-awscloudformation/src/graphql-transformer/utils.ts
alharris-at: Chatted offline, but we're planning to leave as-if for this commit, since this is actively being refactored. The utilities are basically remaining in the package which they have the most in common with, and we should be addressing this in a separate commit.

PR #10267, opened 2022-04-22
chore: use jest 26

johnpc:
This PR pins our jest version to 26, since there appears to be an issue in jest 27 that causes this error in our unit tests:

```
 ReferenceError: setImmediate is not defined
```


sachscode (approved):
LGTM

ammarkarachi (approved):

PR #10336, opened 2022-05-02
chore: initially tease transformer-factory out from transformGraphQLSchema

alharris-at:
#### Description of changes
In preparation to move some of the key pieces of API-related logic out of the provider package, and into the API I've been working to simplify the surface area here a bit, and consolidate down some of the places we have differences in the v1 vs. v2 code paths. In practices this means the following things have changed.

* Provide a single entry point for the following APIs, which will do the right thing based on whether or not you're operating in a v1 vs. v2 app.
  * `getTransformerVersion`
  * `getDirectiveDefinitions`
* Not yet consolidated into a consistent API
  * `getTransformerFactory` - this turned out to be a bit more involved, since the APIs are slightly different, so pulling that into a separate PR.
* Create utility function files which contain reused helpers across v1/v2 scopes, either grouped by whether they relate more strongly to the api category (i.e. internal representation of a project) or the provider (i.e. state of project AWS resources).

For this set of changes I'm relying on existing unit and e2e tests to protect against regressions, and there have been no test changes as part of this (other than updating import paths).

#### Issue #, if available
N/A

I've run `yarn setup-dev && yarn test` locally to verify unit tests and builds look good, I've also manually run the e2e test suite against this change https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/8263/workflows/5f26aca4-a443-469a-a5f0-870abe3b785c

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-provider-awscloudformation/src/graphql-transformer-factory/transformer-version.ts
```diff
@@ -0,0 +1,55 @@
...
+  $TSContext,
+  pathManager,
+  stateManager,
+} from 'amplify-cli-core';
+import { FeatureFlags } from 'amplify-cli-core';
+
+/**
+ * Shorthand for Feature flag retrieval.
+ */
+const useExperimentalPipelinedTransformerFF = () => FeatureFlags.getBoolean('graphQLTransformer.useExperimentalPipelinedTransformer')
```
marcvberg, line 11: Could this potentially be an issue? Since this doesn't happen at function call time the loading order of packages could mean the `FeatureFlags` class not being initialized, or if we change a feature flag value during the execution (such as v1 -> v2 migration) these values would be out of date

marcvberg: oh nevermind I missed that these are written as a function rather than just resolved values

marcvberg (approved):
LGTM

danielleadams (approved):

PR #10569, opened 2022-06-12
fix: golang function build for goversion above 1.17

sky4git:
#### Description of changes

Golang version above 1.17 has deprecated `go get` command and instead suggested to use `go install`
https://go.dev/doc/go-get-install-deprecation
To fix the issue in the local environment while working with `amplify build function GoFunction -y`, we need to change the command to fetch and install the `github.com/aws/aws-lambda-go/cmd/build-lambda-zip`.

Changes are made to work with golang version above > 1.17 and below it.

- [X] PR description included
- [X] `yarn test` passes


packages/amplify-go-function-runtime-provider/src/runtime.ts
```diff
@@ -161,8 +177,14 @@ export const packageResource = async (request: PackageRequest, context: any): Pr
 };
 
 const winZip = async (src: string, dest: string, print: any) => {
-  // get lambda zip tool
-  await execa(executableName, ['get', '-u', 'github.com/aws/aws-lambda-go/cmd/build-lambda-zip']);
+  // get lambda zip tool with the fix of https://go.dev/doc/go-get-install-deprecation
+  try {
+  // for golang version above > 1.17
+    await execa(executableName, ['install', '-u', 'github.com/aws/aws-lambda-go/cmd/build-lambda-zip']);
```
josefaidt, line 183:
```suggestion
    await execa(executableName, ['install', 'github.com/aws/aws-lambda-go/cmd/build-lambda-zip']);
```

Hey :wave: just a minor tweak, `-u` is not a valid option for `install`

sachscode (commented):
Thanks a lot @sky4git for your contribution. Mostly LGTM. 
Could you please address @josefaidt 's comment above so that we could accept in our next release.

sachscode (approved):
LGTM

lazpavel (approved):

lazpavel:
Hi @sky4git, changes LGTM, thank you for your contribution!

packages/amplify-go-function-runtime-provider/src/runtime.ts
```diff
@@ -63,6 +70,9 @@ const isBuildStale = (resourceDir: string, lastBuildTimeStamp: Date, outDir: str
   return !!fileUpdatedAfterLastBuild;
 };
 
+/**
+ *
+ */
```
danielleadams, line 75: Sorry to be nitpicky because I know these were auto generated, but can you remove these empty function definitions? You should be able to make these changes without running the lint filter with the `--no-verify` flag.

danielleadams (changes requested):

packages/amplify-go-function-runtime-provider/src/runtime.ts
```diff
@@ -149,6 +162,9 @@ export const checkDependencies = async (_runtimeValue: string): Promise<CheckDep
   };
 };
 
+/**
+ *
+ */
```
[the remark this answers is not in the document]
ammarkarachi: Removing autogenerated comments

packages/amplify-go-function-runtime-provider/src/runtime.ts
```diff
@@ -12,15 +12,22 @@ import archiver from 'archiver';
...
 
 const executableName = 'go';
 const minimumVersion = <SemVer>coerce('1.0');
 const maximumVersion = <SemVer>coerce('2.0');
 
 let executablePath: string | null;
 
+/**
+ *
+ */
```
ammarkarachi, line 30: Removing auto generated method comment
```suggestion
```

ammarkarachi (approved):

packages/amplify-go-function-runtime-provider/src/runtime.ts
sky4git: Changes are done.

packages/amplify-go-function-runtime-provider/src/runtime.ts
```diff
@@ -161,8 +165,14 @@ export const packageResource = async (request: PackageRequest, context: any): Pr
...
-  // get lambda zip tool
-  await execa(executableName, ['get', '-u', 'github.com/aws/aws-lambda-go/cmd/build-lambda-zip']);
+  // get lambda zip tool with the fix of https://go.dev/doc/go-get-install-deprecation
+  try {
+  // for golang version above > 1.17
+    await execa(executableName, ['install', 'github.com/aws/aws-lambda-go/cmd/build-lambda-zip']);
+  } catch (error: unknown) {
+  // for golang version below < 1.17
+    await execa(executableName, ['get', '-u', 'github.com/aws/aws-lambda-go/cmd/build-lambda-zip']);
+  }
```
danielleadams, line 175: Is there any way to get the go version so that only 1 of these is run? It looks like [this StackOverflow post](https://stackoverflow.com/questions/59820549/how-to-extract-only-version-from-go-version-command-using-sed-or-other-bash-co) has directions for piping out just the number. Then you can use an if/else statement instead of a try/catch.

sky4git: @danielleadams Please check it again.

danielleadams (approved):

danielleadams:
@sky4git thanks again for the work here - do you mind rebasing this? Some package names changed and it's breaking the tests.

sky4git:
> @sky4git thanks again for the work here - do you mind rebasing this? Some package names changed and it's breaking the tests.

Please check it again.

PR #10597, opened 2022-06-14
Feature: add disabled flag for deep merging directive defaults

marcvberg:
#### Description of changes
This adds a new feature flag, not yet defaulting to true for new projects. The purpose is to deep merge directive defaults, so that if the default is:

```json
{
    "foo": "bar",
    "obj": {
        "val": true,
        "this": "that"
    }
}
```
and a customer provides
```json
{
    "obj": {
        "val": false
    }
}
```

the result will be
```json
{
    "foo": "bar",
    "obj": {
        "val": false,
        "this": "that"
    }
}
```

instead of the current result, which is

```json
{
    "foo": "bar",
    "obj": {
        "val": false
    }
}
```
since only top level values are currently being merged for config defaults.

Documentation PR will be coming, I'm waiting until the feature is actually out. This enables work on the API category

#### Issue #, if available
Internally tracked

It's a feature flag, no way to validate changes

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli-core/src/feature-flags/featureFlags.ts
+++ b/packages/amplify-cli-core/src/feature-flags/featureFlags.ts
@@ -643,6 +643,12 @@ export class FeatureFlags {
         defaultValueForExistingProjects: false,
         defaultValueForNewProjects: false,
       },
+      {
+        name: 'shouldDeepMergeDirectiveConfigDefaults',
+        type: 'boolean',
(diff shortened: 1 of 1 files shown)


lazpavel (approved):
LGTM

sachscode (approved):

PR #10603, opened 2022-06-14
Revert "chore: add FF for primary key connection field attributes"

pavellazar:
Reverts aws-amplify/amplify-cli#10556


sachscode (approved):
LGTM

jhockett (approved):

danielleadams:
Why are we reverting this if it doesn't touch any code paths?

sachscode:
> Why are we reverting this if it doesn't touch any code paths?
@danielleadams , the consensus between @lazpavel  and Eric Straub was to not have any PRs along side a sev2 fix if it doesn't directly impact the sev2 problem space and the other thought was that if a feature flag is enabled and if for some reason studio has to rollback the version, it might result in projects which have the feature flag in the persistent state but doesn't have have the logic to handle.

PR #10850, opened 2022-08-05
fix: detect table requiring replacement

marcvberg:
#### Description of changes
Add in a condition to detect destructive updates when a sort key is added to a primary key, currently there is no warning and deploys just fail, using the `--allow-destructive-graphql-schema-updates` flag doesn't work

This is tied to the API category PR which adds in the sanity check and warning https://github.com/aws-amplify/amplify-category-api/pull/714

This is the same condition that I used in the API category, manually tested to verify it picks up the right changes

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


edwardfoyle (approved):

danielleadams (approved):

marcvberg:
For the time being, do not merge, this will go once the API category change is ready to go into release as well

jhockett (approved):

Issue #3655, opened 2020-03-12
Amplify mock subscription error: "Connection failed: An invalid or illegal string was specified"


kldeb:
**Describe the bug**
The same code that works when subscribing to an AppSync endpoint doesn't work when using `amplify mock`.

**Amplify CLI Version**
4.13.1

**To Reproduce**
1. Create react app
2. run amplify init
3. add graphql api with api key (for test i've tried with api key and cognito user pool)
4. npm install amplify modules
5. deploy to server
6. copy paste code below 
7. run aginst server and see connection
8. run amplify mock and see error

Code: 

```js
import React, { useEffect } from 'react';
import Amplify from '@aws-amplify/core';
import API from '@aws-amplify/api';
import PubSub from '@aws-amplify/pubsub';
import config from './aws-exports';
import { onCreateMyType } from './graphql/subscriptions';

Amplify.configure(config);
PubSub.configure(config);

// Amplify.Logger.LOG_LEVEL = 'DEBUG';

function App() {
  useEffect(() => {
    let subscription;

    const run = () => {
      subscription = API.graphql({
        query: onCreateMyType,
      }).subscribe({
        next: response => console.log('response', response),
        error: response => console.log('error response', response),
      });
    };

    run();

    return () => subscription.unsubscribe();
  }, []);

  return <div>Pub Sub Test</div>;
}

export default App;

```

**Expected behavior**
Be able to test subscriptions locally 

**Desktop (please complete the following information):**
 - OS: macOS
 - Node Version: v12.10.0

yuth:
@kldeb could you share the schema and the `onCreateMyType` subscription statements as well

kldeb:
sure thing @yuth 

```gql
type MyType @model {
  id: ID!
  title: String!
  content: String!
  price: Int
  rating: Float
}
```

```js
export const onCreateMyType = /* GraphQL */ `
  subscription OnCreateMyType {
    onCreateMyType {
      id
      title
      content
      price
      rating
    }
  }
`;
```

yuth:
@kldeb what version of Amplify JS package are you using. Specifically `import API from 'amplify/api` and `@aws-amplify/pubsub`.

The mock does not support Websocket based subscription yet. We should have mock support for websocket based subscription in a few weeks.

ranjan-purbey:
[...]

PR #10870, opened 2022-08-10
fix(amplify-appsync-simulator): mock websocket subscriptions

phani-srikar:
#### Description of changes

The Amplify `API` client libraries [example JS](https://github.com/aws-amplify/amplify-js/pull/9171/files#diff-5f5d479eb66a89cef0dd03d15fe5da529c8b3221f774c264c240bc6209cc65ad) use web-sockets to connect to the mock subscription server. They treat the mock appsync endpoint as custom domain and hence, try to connect to the path `graphql/realtime`, as mentioned in Appsync docs [here](https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html#handshake-details-to-establish-the-websocket-connection). The mock appsync endpoint that clients try to connect to is `ws://localhost:20002/graphql/realtime`.

The current mock implementation however, starts the web-socket at `graphql` path instead of `graphql/realtime` and also uses port `20003`. This prevents the Amplify customers from testing their subscriptions via Amplify client libraries and the mock plugin. Also note that the mock GraphiQL interface does not support running subscriptions directly from the dashboard. 

With this change, we update the mock web-socket server to run on the endpoint that the Amplify clients expect from a custom appsync domain. We also remove support for legacy MQTT subscription server which runs on and hence blocked the port `20002`.

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/122

Tested using React JS sample app that we are able to subscribe and get mutation events.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-appsync-simulator/src/__tests__/server/subscription/websocket-server/server.test.ts
```diff
@@ -52,9 +52,10 @@ describe('WebsocketSubscriptionServer', () => {
   const onSubscribeHandler = jest.fn();
   const onConnectHandler = jest.fn();
   let serverPort = 20005;
-  const SUBSCRIPTION_PATH = '/graphql';
+  const SUBSCRIPTION_PATH = '/graphql/realtime';
```
stocaaro, line 55: I got my brain wrapped around this for a bit. So, yes, since the localhost url won't match [the pattern](https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts#L69-L70) of an aws graphql endpoint, it falls back the a url pattern that has [realtime concatinated](https://github.com/aws-amplify/amplify-js/blob/main/packages/pubsub/src/Providers/AWSAppSyncRealTimeProvider/index.ts#L617-L620) onto it.

This seems like an artifact of the websocket cutover, supporting the beta endpoints along with release endpoints, given the special case logic I see around this. Is this the right pattern to maintain forward? Maybe it's not worth a change to all platforms for a better testing experience, but happy to have that question asked.

This behavior is released today, do customers have a workaround to enable them to use this?

packages/amplify-appsync-simulator/package.json
```diff
@@ -49,24 +49,17 @@
     "moment": "^2.24.0",
     "moment-jdateformatparser": "^1.2.1",
     "moment-timezone": "0.5.34",
-    "mqtt-connection": "4.0.0",
```
stocaaro, line 52: In terms of removing MQTT support, does this mean we are no longer supporting local testing for IoT?  Do customers use this behavior? If so, what is our guidance for these customers with this change?

stocaaro (commented):
👋 Aaron from the JS team. This caught my interest. I've spent a lot of time down in the JS code related to this. Leaving a couple questions inline, but otherwise this change does what it proposes and I have no concerns about the semantics of the change.

packages/amplify-appsync-simulator/package.json
phani-srikar: I don't think the usage of MQTT here is to support IoT use-cases. From the data that I could find, the older implementations in Amplify Client Libraries allowed MQTT based AppSync subscriptions before shifting to use real-time websockets. Refer to [this comment](https://github.com/aws-amplify/amplify-cli/issues/3655#issuecomment-599662423) and [this one](https://github.com/aws-amplify/amplify-cli/issues/3008#issuecomment-599169057). I've tried to keep both of these approaches, but ran into issues because both processes need to use the same port. I believe that is the reason @yuth removed it too in his [draft PR](https://github.com/aws-amplify/amplify-cli/pull/9709).

packages/amplify-appsync-simulator/src/__tests__/server/subscription/websocket-server/server.test.ts
phani-srikar: The handling of the API endpoints with custom domain (which is the case here with mock endpoint) is defined in this [Appsync doc](https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html#handshake-details-to-establish-the-websocket-connection). I've verified that latest versions of Amplify Client libraries [JS](https://github.com/aws-amplify/amplify-js/pull/9171/files), [iOS](https://github.com/aws-amplify/aws-appsync-realtime-client-ios/pull/53/files#diff-af83dacdfaccd65d06c9374e7782b95ab7665aa32df46423b828b737ec9517c8) and [Android](https://github.com/aws-amplify/amplify-android/blob/0af580310b604fd297d25be77de668045594eecb/aws-api/src/main/java/com/amplifyframework/api/aws/SubscriptionEndpoint.java#L325) all respect this pattern and concatenate `realtime` to the path for custom domains. So I believe this is the right way going forward. The current status is that mock subscriptions do not work in any of the platforms unless the developer defines proxies in his app as described here. With this PR, we will fix that so that [such workaround is not needed](https://github.com/aws-amplify/amplify-category-api/issues/122#issuecomment-1129353449).

packages/amplify-util-mock/src/__e2e__/subscriptions-with-auth.e2e.test.ts
```diff
@@ -247,41 +247,6 @@ test('Test that only authorized members are allowed to view subscriptions', asyn
   return subscriptionPromise;
 });
 
-test('Test that a user not in the group is not allowed to view the subscription', async () => {
```
phani-srikar, line 250: This test should not throw but there should not be any event received in the subscription. Since the async callback never gets invoked, the promise remains unresolved causing the test to timeout and fail ungracefully. Verified this behavior from console.

packages/amplify-appsync-simulator/package.json
stocaaro: Yep. The first iteration of AppSync was MQTT based, fully migrated in January, we're deprecating the MQTT AppSync provider with the next major version bump. MQTT is also offered as a PubSub provider for non-amazon MQTT servers and then is the protocol for IoT integrations on the PubSub category.

packages/amplify-appsync-simulator/package.json
```diff
@@ -26,7 +26,7 @@
     "clean": "rimraf ./lib tsconfig.tsbuildinfo tsconfig.tests.tsbuildinfo node_modules",
     "watch": "tsc -w",
     "start": "node ./lib/index.js",
-    "test": "jest"
+    "test": "jest --forceExit"
```
lazpavel, line 29: why is this needed, seems to be more of a [escape hatch](https://jestjs.io/docs/cli#--forceexit)

packages/amplify-appsync-simulator/src/__tests__/server/subscription/websocket-server/server.test.ts
```diff
@@ -76,9 +77,20 @@ describe('WebsocketSubscriptionServer', () => {
...
+    server?.stop();
+    httpServer?.close();
+  });
+
+  beforeAll(done => {
+    done();
+  });
+
+  afterAll(done => {
+    server?.stop();
```
lazpavel, line 91: aren't this 2 lines executed in the afterEach section?

lazpavel (commented):
added a few comments but overall LGTM

packages/amplify-appsync-simulator/src/server/index.ts
```diff
@@ -13,22 +12,19 @@ const MAX_PORT = 9999;
...
   private _realTimeSubscriptionServer: AppSyncSimulatorSubscriptionServer;
   private _url: string;
 
   constructor(private config: AppSyncSimulatorServerConfig, private simulatorContext: AmplifyAppSyncSimulator) {
-    this._subscriptionServer = new SubscriptionServer(config, simulatorContext);
-    this._operationServer = new OperationServer(config, simulatorContext, this._subscriptionServer);
+    this._operationServer = new OperationServer(config, simulatorContext);
     this._httpServer = createServer(this._operationServer.app);
-    this._realTimeSubscriptionServer = new AppSyncSimulatorSubscriptionServer(simulatorContext, this._httpServer, '/graphql');
+    this._realTimeSubscriptionServer = new AppSyncSimulatorSubscriptionServer(simulatorContext, this._httpServer, '/graphql/realtime');
```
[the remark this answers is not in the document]
phani-srikar: re-using the same constant. Thanks

packages/amplify-appsync-simulator/src/__tests__/server/subscription/websocket-server/server.test.ts
phani-srikar: remove it here

packages/amplify-appsync-simulator/src/__tests__/server/subscription/websocket-server/server.test.ts
```diff
@@ -76,14 +77,24 @@ describe('WebsocketSubscriptionServer', () => {
...
+    httpServer?.close();
+  });
+
+  beforeAll(done => {
+    done();
+  });
+
+  afterAll(done => {
+    httpServer?.close();
```
lazpavel, line 91: same for httpServer, we already close it in the afterEach

packages/amplify-appsync-simulator/package.json
phani-srikar: Actually the unit tests terminate graciously on my machine but in CI they pass but for some reason fail to exit. I have tried to use `jest --detectOpenHandles` to see what's going on but even that doesn't give me any info. So, had to add this flag like in the mock plugin tests.

lazpavel (approved):

packages/amplify-appsync-simulator/tsconfig.json
```diff
@@ -5,9 +5,21 @@
...
     "declaration": true,
     "skipLibCheck": true,
   },
-  "exclude": ["node_modules", "lib", "**/__mocks__/**/*", "**/__tests__/**/*"]
-}
+  "exclude": [
+    "node_modules",
+    "lib",
+    "**/__mocks__/**/*",
+    "**/__tests__/**/*"
```
sundersc, line 23: Don't we have to compile the ts files moved inside `__tests__/server` folder?

sundersc (commented):
Few nits, LGTM otherwise.

danielleadams:
@phani-srikar looks like this needs a rebase

packages/amplify-appsync-simulator/src/__tests__/server/subscription/websocket-server/server.test.ts
```diff
@@ -4,6 +4,7 @@ import WS from 'ws';
 import getPort from 'get-port';
 import { PubSub } from 'graphql-subscriptions';
 import { parse } from 'graphql';
+import { REALTIME_SUBSCRIPTION_PATH } from '../../../../server/subscription/websocket-server/server';
```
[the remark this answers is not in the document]
phani-srikar: This package always had test files written in Typescript inside the `src/__tests__` folder.

packages/amplify-appsync-simulator/tsconfig.json
phani-srikar: This is an interesting question about how the TS files in __tests__ are compiled. Doing a bit of investigation, `jest` actually takes care of `TS->JS` transformation for the test files it discovers. We configure how to do this using the below config which you can find in [package.json]()
```
"transform": {
      "^.+\\.tsx?$": "ts-jest"
    }
```
This basically means for all the TS files, use `ts-jest`(built-in) to transform them to JS before executing the test.
So, each time you run the test, it's also inherently compiled by `jest`. We use this for several packages already. 

Resources: 
- https://jestjs.io/docs/configuration#transform-objectstring-pathtotransformer--pathtotransformer-object
- https://jestjs.io/docs/code-transformation

sundersc (approved):

PR #10874, opened 2022-08-10
fix: stop the timer correctly on init env errors

jhockett:
#### Description of changes
The initialize-env timer code has a bug when an error is thrown. Due to process.exit() in the catch block, the finally block would never execute and the timer would never be properly stopped in this case.

Created a demo to see what would happen.

```js
function main() {
  try {
    console.log('try block')
    throw '';
  } catch {
    console.log('catch block')
    process.exit(1);
  } finally {
    console.log('finally block')
  }
}

main()
```

Output
```
try block
catch block
```

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/amplify-cli/src/initialize-env.ts
```diff
@@ -76,10 +76,10 @@ export const initializeEnv = async (
     } catch (e) {
       printer.error(`Could not initialize '${currentEnv}': ${e.message}`);
       context.usageData.emitError(e);
-      process.exit(1);
-    } finally {
       context.usageData.stopCodePathTimer(ManuallyTimedCodePath.INIT_ENV_PLATFORM);
+      exitOnNextTick(1);
```
danielleadams, line 80: What is `exitOnNextTick` for?

jhockett: It calls process.exit(), but first closes the logger:
https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-cli-core/src/exitOnNextTick.ts

pavellazar (dismissed):

edwardfoyle (dismissed):

jhockett:
No longer relevant.

PR #10915, opened 2022-08-19
fix: add FF input to show ACM correctly

marcvberg:
#### Description of changes
Due to fun typescript issues we can't access the feature flag from the transformer core package, so it needs to be given as input from outside. This adds the necessary input for showing ACM rules. This should not be merged until the API category version is bumped that includes this PR: https://github.com/aws-amplify/amplify-category-api/pull/747

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/show-auth-acm.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/show-auth-acm.ts
@@ -10,6 +10,7 @@ import {
 import { parse, ObjectTypeDefinitionNode, DirectiveNode, FieldDefinitionNode } from 'graphql';
 import { printer } from 'amplify-prompts';
 import { DirectiveWrapper } from '@aws-amplify/graphql-transformer-core';
+import { FeatureFlags } from 'amplify-cli-core';
 
 export function showACM(sdl: string, nodeName: string) {
   const schema = parse(sdl);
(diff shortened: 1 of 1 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/show-auth-acm.ts
```diff
@@ -24,7 +25,9 @@ export function showACM(sdl: string, nodeName: string) {
     const acm = new AccessControlMatrix({ name: type.name.value, operations: MODEL_OPERATIONS, resources: fields });
     const parentAuthDirective = type.directives?.find(dir => dir.name.value === 'auth');
     if (parentAuthDirective) {
-      const authRules: AuthRule[] = getAuthDirectiveRules(new DirectiveWrapper(parentAuthDirective));
+      const authRules: AuthRule[] = getAuthDirectiveRules(new DirectiveWrapper(parentAuthDirective), false, {
```
phani-srikar, line 28: There is one more place where we have the same method call [here](https://github.com/aws-amplify/amplify-cli/pull/10915/files#diff-86aacfb3d4a9e3ac7d29ce386169b1c55f7eea125966104a92d81ef0318decdfR39)

packages/amplify-cli/src/extensions/amplify-helpers/show-auth-acm.ts
```diff
@@ -24,7 +25,9 @@ export function showACM(sdl: string, nodeName: string) {
...
+        deepMergeArguments: FeatureFlags.getBoolean('shouldDeepMergeDirectiveConfigDefaults'),
```
phani-srikar, line 29: shouldn't this be `graphqlTransformer.<something>`?

phani-srikar (changes requested):

phani-srikar:
These changes should be in the same version bump PR and should not be merged independently. Closing this to avoid confusion.

PR #10954, opened 2022-08-31
fix: raise incorrect post event data

fossamagna:
#### Description of changes

Using an `AmplifyPostPushEventData` instance when raised a `PostInit` event, and an `AmplifyPostPushEventData` instance when raised a `InitEvent` event. These use the wrong event data for each other.
Fixed to use the correct event data.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
     "coverage": "codecov || exit 0",
     "release-rc": "./scripts/release-rc.sh",
     "promote-rc": "./scripts/promote-rc.sh",
-    "finish-release": "./scripts/finish-release.sh"
+    "finish-release": "./scripts/finish-release.sh",
+    "update-data-packages": "./scripts/update-data-dependencies.sh && yarn"
   },
   "bugs": {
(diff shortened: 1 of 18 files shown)


danielleadams:
Hey @fossamagna - Thanks for the contribution! I see the function names were changed to match, but what is the context here? Was it breaking anything and was this intentional? Are there any tests we can add to ensure this works? Going to run the CLI test suite now.

fossamagna:
@danielleadams Thank you for your review. 
I think that this change is not breaking, because `AmplifyPostPushEventData` and `AmplifyPostPushEventData ` is not same class, but both class is extended same parent class and both class no have property. 
I will add tests to ensure this work.

fossamagna:
@danielleadams I added test for `execution-manager.ts`. It looks like there were nothing failures in the CI, Is there anything blocking to merging this PR?

jhockett (approved):

hackmajoris:
Not sure if this issue is related, but I'll leave it here: https://github.com/aws-amplify/amplify-cli/issues/11021

PR #11017, opened 2022-09-16
chore: converted analytics category from JS to TS

lazpavel:


packages/amplify-category-analytics/src/commands/analytics.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  printer.info('');
+  return context;
+};
+
+module.exports = {
+  name: featureName,
+  run: analyticsRun,
+  analyticsPush,
+};
```
jhockett, line 57: nit: remove this in favor of `export`, here and everywhere else

packages/amplify-category-analytics/package.json
```diff
@@ -21,6 +21,7 @@
   },
   "dependencies": {
     "amplify-cli-core": "3.0.0",
+    "amplify-prompts": "2.3.0",
```
sobolk, line 24: why is new dependency part of this refactoring?

packages/amplify-category-analytics/src/lib/constants.ts
```diff
@@ -0,0 +1,11 @@
+export const CategoryName = 'analytics';
```
jhockett, line 1: I'm not sure if it makes sense to change in this PR, but we don't want `src` files in a `lib` folder

packages/amplify-category-analytics/tsconfig.json
```diff
@@ -19,6 +20,8 @@
     "src/__tests__"
   ],
   "references": [
-    { "path": "../amplify-cli-core" }
+    {
+      "path": "../amplify-cli-core"
```
jhockett, line 24: We should add `amplify-prompts` here

packages/amplify-category-analytics/src/index.ts
```diff
@@ -40,18 +45,24 @@ async function console(context) {
...
-  const resourceAttributes = [];
+  const permissionPolicies: $TSAny[] = [];
+  const resourceAttributes: $TSAny[] = [];
 
   Object.keys(resourceOpsMapping).forEach(resourceName => {
     try {
       const providerName = amplifyMeta[category][resourceName].providerPlugin;
       if (providerName) {
+        // eslint-disable-next-line import/no-dynamic-require, global-require, @typescript-eslint/no-var-requires
         const providerController = require(`./provider-utils/${providerName}/index`);
```
sobolk, line 66: here (and other places like this) is this by design? I.e. should we fix these linter warnings in the feature or not?

Perhaps we should have "ProviderProvider/ProviderFactory/ProviderLoader" and centralize this logic?

This is something that can be explored afterwards since this seems to be existing pattern.

sobolk (commented):
Could you please run e2e tests for this change?

packages/amplify-category-analytics/package.json
lazpavel: In the refactor I replaced `context.print` with `printer` from `amplify-prompts` as the `context.print` is deprecated. It is a small change that I figured I can do as part of this refactor, let me know if you feel this shouldn't be here and I can keep the context.print instead

packages/amplify-category-analytics/src/index.ts
lazpavel: Like the idea of either have a design where dynamic import is not needed or use centralized factory of some kind if the first isn't possible, but will leave it for now as might increase the scope of this particular move from JS to TS

packages/amplify-category-analytics/src/lib/constants.ts
lazpavel: I feel that this is a small enough change that we can take as part of this refactor

packages/amplify-category-analytics/package.json
sobolk: sounds good.

jhockett (dismissed):

packages/amplify-category-analytics/src/commands/analytics.ts
```diff
@@ -0,0 +1,56 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable global-require */
+/* eslint-disable import/no-dynamic-require */
+import { $TSAny, $TSContext } from 'amplify-cli-core';
+import { printer } from 'amplify-prompts';
+import { run as analyticsPush } from './analytics/push';
```
jhockett, line 6: nit: no need for this import

```suggestion
```

jhockett (approved):

sobolk (approved):

PR #11024, opened 2022-09-19
feat: append whitespace to cli-inputs.json for updated build dir

danielleadams:
#### Description of changes
This inserts a whitespace in the `cli-inputs.json` file on `amplify push` when `AttributesRequireVerificationBeforeUpdate` is detected in the local build directory and not in the cloud-backend build directory.

The build directory is not usually tracked for local <--> cloud diffs, so updating the cli-inputs.json file (a tracked file) will allow a push without `--force` tag. When the whitespace is inserted, the `push` script detects a change in the and will push the entire updated build directory.

I performed the following test:
- install amplify v9.2.1
- run `amplify init` on project
- run `amplify add auth` on project
- run `amplify push` on project
- run `amplify-dev` pull with the locally build bin
- run `amplify-dev push`

Result: the CLI detects a change and will update the auth category for the user.

Also added unit tests to ensure the code that appends white space is working as expected.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/push.ts
+++ b/packages/amplify-cli/src/commands/push.ts
@@ -15,6 +15,7 @@ import {
 import {
   getProviderPlugins,
 } from '../extensions/amplify-helpers/get-provider-plugins';
+import { updateCognitoTrackedFiles } from '../extensions/amplify-helpers/update-tracked-files';
 
 /**
  * Download and unzip deployment bucket contents to #current-cloud-backend so amplify status shows correct state
@@ -53,6 +54,13 @@ const syncCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
 };
 
 /**
+ * Updates tracked files for auto updates in the build directory that will not be detected for 'amplify push'
+ */
+const updateTrackedFiles = async (): Promise<void> => {
+  await updateCognitoTrackedFiles();
+};
+
+/**
  * Runs push command
  */
 export const run = async (context: $TSContext): Promise<$TSAny> => {
@@ -64,5 +72,6 @@ export const run = async (context: $TSContext): Promise<$TSAny> => {
     context.exeInfo.forcePush = true;
   }
   await syncCurrentCloudBackend(context);
+  await updateTrackedFiles();
   return context.amplify.pushResources(context);
 };
--- a/packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
@@ -1,0 +1,80 @@
+import * as path from 'path';
+import fs from 'fs-extra';
+import {
+  $TSAny, JSONUtilities, pathManager, stateManager,
+} from 'amplify-cli-core';
+
+const {
+  readJson,
+} = JSONUtilities;
+
+/**
+ * Updates Cognito files that are tracked so that the diff is detected for an `amplify push`
+ */
+export const updateCognitoTrackedFiles = async (): Promise<void> => {
+  const currentCloudBackendDir = pathManager.getCurrentCloudBackendDirPath();
+  const localBackendDir = pathManager.getBackendDirPath();
+  const amplifyMeta = stateManager.getMeta();
+  const cognitoResource = stateManager.getResourceFromMeta(amplifyMeta, 'auth', 'Cognito', undefined, false);
+
+  if (!fs.existsSync(currentCloudBackendDir) || !cognitoResource) {
+    return;
+  }
+
+  const { resourceName } = cognitoResource;
+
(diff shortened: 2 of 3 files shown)


packages/amplify-cli/src/commands/push.ts
```diff
@@ -52,6 +53,13 @@ const syncCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
   }
 };
 
+/**
+ * Updates tracked files for auto updates in the build directory that will not be detected for 'amplify push'
+ */
+const updateTrackedFiles = async (): Promise<void> => {
```
lazpavel, line 59: is this wrapper method needed? can we just use `await updateCognitoTrackedFiles();` directly?

danielleadams: Yes, I figured this could be updated to add each category since they all probably have different ways of changing "tracked" files.

lazpavel (approved):

packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
```diff
@@ -0,0 +1,80 @@
...
+  currentCloudBackendDir: string,
+  localBackendDir: string,
+  resourceName: string,
+): Promise<boolean> => detectCognitoAttributesRequireVerificationBeforeUpdateDiff(
+  currentCloudBackendDir,
+  localBackendDir,
+  resourceName,
+);
+
+const detectCognitoAttributesRequireVerificationBeforeUpdateDiff = async (
```
awsluja, line 41: It seems you have two things going on here; one very generic thing "checking if a local update is not present in the cloud" and one very specific thing "checking if a cognito setting is a specific value that differs locally vs in the cloud" . I'm guessing the original issue is more generic "any setting locally should also reflect in the cloud & trigger an update", otherwise it seems like we are patching a bug with that setting specifically. Can you explain why we only check this specific setting; and why we don't provide a more high level diff check

packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
```diff
@@ -0,0 +1,80 @@
...
+    && backendUserAttrUpdateSettings?.AttributesRequireVerificationBeforeUpdate[0] === 'email';
+
+  return updateNotInCloudBackend && updateInLocalBackend;
+};
+
+type UserAttributeUpdateSettings = {
+  AttributesRequireVerificationBeforeUpdate: string[]
+}
+
+const readCfnTemplateUserAttributeSettings = async (
```
awsluja, line 60: the utility of this function is very localized to the **detectCognitoAttributesRequireVerificationBeforeUpdateDiff** function, so it may be worth declaring it as an arrow function within that; it won't see re-use outside of there. Also, resourceName is not changing, just the backendDir path; so you can exclude it from params in your new arrow function as the param will be in the same closure

packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
```diff
@@ -0,0 +1,80 @@
...
+  const cognitoResource = stateManager.getResourceFromMeta(amplifyMeta, 'auth', 'Cognito', undefined, false);
+
+  if (!fs.existsSync(currentCloudBackendDir) || !cognitoResource) {
+    return;
+  }
+
+  const { resourceName } = cognitoResource;
+
+  if (await detectCognitoDiff(currentCloudBackendDir, localBackendDir, resourceName)) {
+    await addExtraLineToCliInputsJson(localBackendDir, resourceName);
```
awsluja, line 27: if the goal is to trigger change detection on the CLIInputsJSON file, it may help to rename this "triggerChangeDetectionOnCLIInputsJSON"; then leave implementation details out. Implementations change, but intention is more resilient

awsluja (changes requested):
I am curious if this is pointing to a high level issue in our change detection for local vs cloud build directories; I am curious as to the source of the problem but have some other notes as well

packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
awsluja: resolving this as we only care about this single property; and there are other issues with diff tracking

packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
awsluja: resolving as style suggestion

awsluja (approved):
Comments resolved as they are simple style improvements that can be done later; looks good

PR #11089, opened 2022-09-30
feat: add Searchable mocking feature

phani-srikar:
#### Description of changes

This PR adds support for mocking the behavior of @searchable GraphQL directive by using a local instance of OpenSearch so that the developers can locally index the records and run search queries on the @model GraphQL types.

**Summary of Changes:**
- Add a package to provide the ability to download, start, stop a local OpenSearch emulator.
- Store all mocking artifacts for API mocking in `mock-api-resources` local directory. These include opensearch data directory, searchable python lambda trigger artifacts.
- Download the opensearch binary into the Amplify lib folder `~/.amplify/lib`
- Retain the same DX for windows platform where this feature is not supported. The current DX includes throwing a warning `@searchable mocking is not supported. Search queries will not work as expected.` and returning empty list as data for search queries.
- Searchable artifacts are downloaded and used only when the schema has a searchable model.
- Searchable lambda trigger is automatically setup to store the records in local OpenSearch indices when a GraphQL mutation is performed.
- `Amazon_Elasticsearch` data source is now supported for search queries where in the search query is sent to local OpenSearch instance to retrieve search results from local indices.

- Added unit tests for OpenSearch simulator and mock plugin changes.
- Added e2e tests that cover all search use-cases mentioned in the [docs](https://docs.amplify.aws/cli/graphql/search-and-result-aggregations/).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-appsync-simulator/src/data-loader/opensearch/index.ts
```diff
@@ -1,8 +1,21 @@
...
 
 export class OpenSearchDataLoader implements AmplifyAppSyncSimulatorDataLoader {
-  load(request): any {
-    console.error('@searchable mocking is not supported.');
-    return null;
+  constructor(private _config) {}
+
+  async load(payload, extraData) {
+    try {
+      console.log(payload);
```
phani-srikar, line 8: Todo: Remove this log.

packages/amplify-opensearch-simulator/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "keywords": [
+        "graphql",
+        "appsync",
+        "aws"
+    ],
+    "scripts": {
+        "test": "jest"
+    },
+    "dependencies": {
+        "amplify-cli-core": "2.10.0",
```
phani-srikar, line 22: todo: update this to latest cli-core version

packages/amplify-appsync-simulator/src/velocity/util/opensearch-utils.ts
```diff
@@ -0,0 +1,631 @@
...
+     *      The value in the expression.
+     * @return
+     *      The converted Elasticsearch expression. Returns null if fieldName is invalid.
+     */
+    public toEqExpression(fieldName: string, value: any): any {
+        if (!fieldName) {
+            return null;
+        }
+
+        const updatedFieldName: string = ((typeof value) === "string") ? (fieldName + ".keyword") : fieldName;
```
sundersc, line 40: What if customer is passing a string for integer field like "10"?

packages/amplify-util-mock/resources/mock-searchable-lambda-trigger/src/index.py
```diff
@@ -0,0 +1,256 @@
+import base64
```
sundersc, line 1: Is this the copy of open search streaming lambda?

sundersc (dismissed):
LGTM. Left a couple of questions.

Also, consider fixing the lgtm-com bot alerts.

packages/amplify-appsync-simulator/src/velocity/util/opensearch-utils.ts
```diff
@@ -0,0 +1,631 @@
+/**
```
lazpavel, line 1: should the class be named OpensearchUtils to match the file?

packages/amplify-appsync-simulator/src/velocity/util/opensearch-utils.ts
```diff
@@ -0,0 +1,631 @@
...
+                [ElasticsearchUtils.MUST_NOT]: expression
+            }
+        };
+
+        return andExpression;
+    }
+
+}
+
+class ElasticsearchHelper {
```
lazpavel, line 440: should this be in it's own file?

packages/amplify-opensearch-simulator/__test__/index.test.js
```diff
@@ -0,0 +1,135 @@
...
+        })
+        .on("error", (err) => {
+          reject(err);
+        });
+    });
+  };
+
+  if (process.platform.startsWith('win')) {
+    it('should fail to launch on windows OS', async () => {
+        try {
```
lazpavel, line 64: nit: inconsistent identation

packages/amplify-opensearch-simulator/__test__/index.test.js
```diff
@@ -0,0 +1,135 @@
...
+            await openSearchEmulator.launch(mockSearchableResourcePath);
+            fail('launching simulator is expected to throw but did not');
+        } catch (error) {
+            expect(error.message).toEqual('Cannot launch OpenSearch Simulator on windows OS');
+        }
+    });
+  }
```
lazpavel, line 71: nit: `else` should be on the previous line

packages/amplify-opensearch-simulator/package.json
```diff
@@ -0,0 +1,44 @@
+{
```
lazpavel, line 1: is this a new package? is there a reason this is not written in typescript?

packages/amplify-util-mock/src/utils/opensearch.ts
```diff
@@ -0,0 +1,52 @@
+import { $TSObject } from "amplify-cli-core"
```
lazpavel, line 1: inconsistend identation with the rest of the project

packages/amplify-util-mock/src/utils/opensearch.ts
```diff
@@ -0,0 +1,52 @@
...
+            value: number
+        }
+    },
+    aggregations: $TSObject,
+
+}
+
+export const querySearchable = async (endpoint: string, searchConfig: OpensearchQueryConfig): Promise<OpensearchQueryResult> => {
+    if (!endpoint) {
+        throw new Error('The local opensearch endpoint is not found');
```
lazpavel, line 31: use AmplifyError

packages/amplify-util-mock/src/utils/opensearch.ts
```diff
@@ -0,0 +1,52 @@
...
+    const url = endpoint.replace(/\/+$/, '') + searchConfig.path;
+
+    const result = await fetch(url, {
+        method: 'POST',
+        body: JSON.stringify(searchConfig.params.body),
+        headers: {
+            'Content-type': 'application/json',
+        }
+    });
+    const resultJSON = await result.json();
```
lazpavel, line 50: `return result.json();`

packages/amplify-util-mock/src/api/lambda-invoke.ts
```diff
@@ -1,39 +1,87 @@
...
-        printer.success('Result:');
-        printer.info(stringResult);
-      } catch (err) {
-        printer.error(`${functionName} failed with the following error:`);
-        printer.info(err);
-      } finally {
-        printer.info('Finished execution.');
-      }
+    if (!lambdaConfig?.handler) {
+      throw new Error(`Could not parse handler for ${functionName} from cloudformation file`);
```
lazpavel, line 23: will stop commenting related to new Error, but we should use `new AmplifyError` or `new AmplifyFault` everywhere

pavellazar (changes requested):

packages/amplify-appsync-simulator/src/velocity/util/transform/index.ts
```diff
@@ -15,4 +19,21 @@ export const transformUtils = {
...
+    if (!filter) {
+      return null;
+    }
+
+    try {
+      const queryDSL: $TSObject = elasticsearchHelper.getQueryDSL(filter.toJSON());
+      return JSON.stringify(queryDSL);
+    }
+    catch (err) {
+        console.error("Error when constructing the Elasticsearch Query DSL using the model transform utils. {}");
```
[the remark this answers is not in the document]
phani-srikar: This shouldn't throw especially for Windows platform where this feature is not supported yet. Using `printer` now.

packages/amplify-appsync-simulator/src/velocity/util/opensearch-utils.ts
phani-srikar: Well AppSync still uses the "ElasticSearch" terminology for it's [utility methods](https://docs.aws.amazon.com/appsync/latest/devguide/transformation-helpers-in-utils-transform.html) that we're mocking here, so can I keep it as is to avoid more confusion?

packages/amplify-appsync-simulator/src/velocity/util/opensearch-utils.ts
phani-srikar: sorry missed this one. Moved it. Thanks

packages/amplify-opensearch-simulator/package.json
phani-srikar: Yes it is a new package but a bit similar to the existing dynamodb simulator package. I wanted to keep the logic to start/stop the emulator in available ports, same as we have in other place in JS, since I was running into errors re-writing it in TS.

packages/amplify-util-mock/resources/mock-searchable-lambda-trigger/src/index.py
phani-srikar: It is almost a copy. The changes that I did were:
1. Remove logic to make valid signed AWS SDK calls to interact with ElasticSearch cluster.
2. Remove the extra dependencies for AWS SDKs.
3. Pass the local cluster info as environment variables to call the local OpenSearch instance.

packages/amplify-appsync-simulator/src/velocity/util/opensearch-utils.ts
phani-srikar: It only depends on the type of the field and not on value passed. All `String` type fields are mapped to have `.keyword` postfixes in Elasticsearch to make some operations like "word searching" optimal.

packages/amplify-util-mock/src/api/lambda-invoke.ts
phani-srikar: Migrated all the higher level errors to use `AmplifyFault`.

pavellazar (dismissed):

sundersc (dismissed):
LGTM. Nice work.

pavellazar (dismissed):

sundersc (dismissed):

packages/amplify-opensearch-simulator/index.js
```diff
@@ -0,0 +1,293 @@
...
+    waiter && waiter.cancel();
+    if (typeof prematureExit === 'function') {
+      proc.removeListener('exit', prematureExit);
+    }
+  }
+
+  return new OpenSearchEmulator(proc, opts);
+}
+
+const ensureOpenSearchLocalExists = async (pathToOpenSearchLocal) => {
```
edwardfoyle, line 228: not sure exactly where `pathToOpenSearchLocal` gets set, but we should use `pathManager.getAmplifyPackageLibDirPath('amplify-opensearch-simulator')`. This will return the path `~/.amplify/lib/amplify-opensearch-simulator`. The `~/.amplify/lib` folder is where all package assets go

packages/amplify-opensearch-simulator/index.js
```diff
@@ -0,0 +1,293 @@
...
+  /*
+   This is a fairly complex set of logic, similar to the DynamoDB emulator, 
+   to retry starting the emulator if it fails to start. We need this logic due
+   to possible race conditions between when we find an open
+   port and bind to it. This situation is particularly common
+   in jest where each test file is running in it's own process
+   each competing for the open port.
+  */
+  try {
+    waiter = wait(opts.startTimeout);
```
edwardfoyle, line 152: can we break this into a few functions (maybe one for each promise that is racing) so it's easier to read?

packages/amplify-opensearch-simulator/index.js
```diff
@@ -0,0 +1,293 @@
+const { fromEvent } = require('promise-toolbox');
+const waitPort = require('wait-port');
+const detectPort = require('detect-port');
+const log = require('logdown')('opensearch-emulator');
```
edwardfoyle, line 4: where are these logs going?

packages/amplify-appsync-simulator/src/velocity/util/elasticsearch-helper.ts
```diff
@@ -0,0 +1,194 @@
...
+   * This FilterInput object is defined by the appsync-model-transform code,
+   *  so this method is opinionated to follow this pattern.
+   *
+   * @param filterInput
+   *    The FilterInput object.
+   * @return
+   *    The Elasticsearch Query DSL
+   * 
+   */
+  public getQueryDSL(filterInput: any): any {
```
jhockett, line 68: Is there a stricter type that can be used instead of `any`? New code should not be using `any` unless absolutely necessary. Here and everywhere else.

edwardfoyle (changes requested):
The 2 blockers I see are adding js files instead of ts and making sure that the searchable mock binary is placed at `pathManager.getAmplifyPackageLibDirPath('amplify-searchable-simulator')` (maybe it already is but I didn't see it)

packages/amplify-opensearch-simulator/package.json
```diff
@@ -0,0 +1,44 @@
...
+        "graphql",
+        "appsync",
+        "aws"
+    ],
+    "scripts": {
+        "test": "jest"
+    },
+    "dependencies": {
+        "amplify-cli-core": "3.3.0",
+        "aws-sdk": "^2.1113.0",
```
jhockett, line 23: The sdk version here must match the rest of the CLI.

```suggestion
        "aws-sdk": "^2.1233.0",
```

packages/amplify-util-mock/src/api/api.ts
```diff
@@ -92,6 +110,17 @@ export class APITest {
...
 
+    try {
+      if (this.opensearchEmulator) {
+        await this.opensearchEmulator.terminate();
+        this.opensearchEmulator = null;
+        this.opensearchURL = null;
+      }
+    } catch (e) {
+      // failed to stop opensearch emulator
+      context.print.error(`Failed to stop OpenSearch Local Server ${e.message}`);
```
jhockett, line 121: What action should someone take if this happens?

Also, please use `printer.error` from `amplify-prompts`

packages/amplify-util-mock/src/api/api.ts
```diff
@@ -307,6 +391,71 @@ export class APITest {
...
+            headers: {
+                'Content-type': 'application/json',
+            }
+        });
+        const status = await result.json();
+        if (!(status?.acknowledged)) {
+          throw new AmplifyFault('MockProcessFault', {
+            message: errMessage,
+            link: AMPLIFY_SUPPORT_DOCS.CLI_GRAPHQL_TROUBLESHOOTING.url
+          });
```
jhockett, line 434: What's the purpose of initializing this exception? The outer catch doesn't use it

packages/amplify-opensearch-simulator/package.json
```diff
@@ -0,0 +1,44 @@
...
+        "detect-port": "^1.3.0",
+        "execa": "^5.1.1",
+        "fs-extra": "^8.1.0",
+        "logdown": "^3.3.0",
+        "openpgp": "^4.10.10",
```
jhockett, line 28: This package is nearly 22mb in size unpacked according to [npm](https://www.npmjs.com/package/openpgp). Is it absolutely necessary to add this?

packages/amplify-util-mock/resources/mock-searchable-lambda-trigger/src/index.py
jhockett: Please add this context in a comment.

packages/amplify-util-mock/src/utils/index.ts
```diff
@@ -45,3 +45,7 @@ function isUnsupportedJavaVersion(stderr: string | null): boolean {
 }
 
 export const _isUnsupportedJavaVersion: (stderr: string | null) => boolean = isUnsupportedJavaVersion;
+
+export const isWindowsPlatform = () => {
+  return process?.platform?.startsWith('win');
+}
```
jhockett, line 51: I think this check exists elsewhere, can you check that this isn't a duplicate function?

If you need to keep it, please add the missing return type

packages/amplify-util-mock/src/utils/lambda/find-lambda-triggers.ts
```diff
@@ -29,16 +46,47 @@ type LambdaTriggersMap = { [index: string] : LambdaTriggerNames};
...
         else {
-          lambdaTriggersMap[attachedTable] = [resourceName];
+          lambdaTriggersMap[attachedTable] = [{ name: resourceName }];
         }
       });
     })
     return lambdaTriggersMap;
 }
 
+export const findSearchableLambdaTriggers = async (context: $TSContext , tables: string[], opensearchEndpoint?: URL): Promise<{ [index: string] : LambdaTrigger }> => {
```
jhockett, line 59:
```suggestion
export const findSearchableLambdaTriggers = async (context: $TSContext, tables: string[], opensearchEndpoint?: URL): Promise<{ [index: string] : LambdaTrigger }> => {
```

packages/amplify-util-mock/src/utils/opensearch.ts
```diff
@@ -0,0 +1,57 @@
+import { $TSObject, AmplifyFault, AMPLIFY_SUPPORT_DOCS } from "amplify-cli-core"
```
jhockett, line 1:
```suggestion
import { $TSObject, AmplifyFault, AMPLIFY_SUPPORT_DOCS } from 'amplify-cli-core';
```

jhockett (changes requested):

packages/amplify-opensearch-simulator/index.js
phani-srikar: These are printed out to the console. However, I removed the `logdown` dependency all together in favor of `amplify-prompts`.

packages/amplify-util-mock/src/api/api.ts
```diff
@@ -307,6 +391,71 @@ export class APITest {
...
+        }
+      }
+      catch(error) {
+        throw new AmplifyFault('MockProcessFault', {
```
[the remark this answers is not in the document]
phani-srikar: using the `details` property of `AmplifyFault` to pass on the error message. Thanks.

packages/amplify-util-mock/src/api/api.ts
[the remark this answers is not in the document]
phani-srikar: updated this throughout the file.

packages/amplify-util-mock/src/api/api.ts
phani-srikar: As Ed suggested above, for the Exceptions thrown from inside a `catch`, I've used the `details` part of the `AmplifyFault` to pass this on.

packages/amplify-opensearch-simulator/package.json
phani-srikar: The [recommended way](https://opensearch.org/verify-signatures.html) to verify the authenticity of downloaded build file is using the public PGP key that is provided. This is the only reliable helper package for us to take advantage of.

packages/amplify-util-mock/src/__e2e_v2__/searchable-transformer.e2e.test.ts
```diff
@@ -61,19 +124,393 @@ describe('@searchable transformer', () => {
   /**
    * Test queries below
    */
-  test('@searchable allows the mock server to run', async () => {
+
+  if (isWindowsPlatform()) {
```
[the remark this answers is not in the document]
phani-srikar: I have spent some time to mock the CLI core appropriately but am running into issues. Since this logic is localized to this test, can I take it up later? We're planning to support windows too in the future, so this will go away anyways.

packages/amplify-opensearch-simulator/index.js
phani-srikar: There is a slight difference in packaging for the opensearch simulator compared to dynamodb simulator. Here, given the following conditions, we decided to download the binary locally to `amplify/mock-api-resources` directory similar to `mock-data`, we have today.
- The binary size (~220MB) is a lot to be packaged along with the CLI.
- Majority of the customers do not use `@searchable` and hence, does not make sense to install the binary during CLI's installation but only when they have searchable models, determined by the mock process. 
- There are optimizations in place for example, we do not re-download the binary if one is already present in mock directory.

packages/amplify-appsync-simulator/src/velocity/util/elasticsearch-helper.ts
phani-srikar: Determining the typings for this Appsync util methods is hard since it depends on user inputs to the search query.

packages/amplify-opensearch-simulator/index.js
```diff
@@ -0,0 +1,293 @@
+const { fromEvent } = require('promise-toolbox');
```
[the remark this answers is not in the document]
phani-srikar: Some of the retry logic in there was re-used from dynamodb simulator (JS) and I was hoping to add a separate more generic emulator package in TS at a later point.
I've converted this file to TS. Will do more cleanup later.

packages/amplify-opensearch-simulator/index.js
edwardfoyle: Yeah we definitely don't want to include this binary with the CLI, but I think we should install it to the `~/.amplify/lib` directory. This is where all other executable assets that the CLi uses are installed to. This also allows different projects to share the same searchable mock binary

packages/amplify-appsync-simulator/src/data-loader/opensearch/index.ts
```diff
@@ -11,11 +12,10 @@ export class OpenSearchDataLoader implements AmplifyAppSyncSimulatorDataLoader {
...
     } catch (e) {
-      console.log('Opensearch Data source failed with the following error');
-      console.error(e);
-      throw new AmplifyError('ConfigurationError', {
+      printer.info('Opensearch Data source failed with the following error:' + e?.message);
+      throw new AmplifyFault('MockProcessFault', {
         message: 'Failed to load data from Opensearch data source',
-        resolution: 'Restart the mock process'
+        link: AMPLIFY_SUPPORT_DOCS.CLI_GRAPHQL_TROUBLESHOOTING.url
       });
```
edwardfoyle, line 19:
```suggestion
      }, e);
```

packages/amplify-cli-core/src/utils/windows-utils.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Check if running on Windows OS
+ * @returns true if the CLI is used from Windows machine, false otherwise
+ */
+export const isWindowsPlatform = (): boolean => {
+  return process?.platform?.startsWith('win');
+};
```
edwardfoyle, line 7: nit this can just be a const value:
```suggestion
export const isWindowsPlatform = process?.platform?.startsWith('win');
```

packages/amplify-util-mock/src/api/api.ts
```diff
@@ -118,7 +119,7 @@ export class APITest {
       }
     } catch (e) {
       // failed to stop opensearch emulator
-      context.print.error(`Failed to stop OpenSearch Local Server ${e.message}`);
+      printer.error(`Failed to stop OpenSearch Local Server ${e.message}. Please kill the mock process and restart it.`);
```
edwardfoyle, line 122:
```suggestion
      printer.error(`Failed to stop OpenSearch Local Server ${e.message}. Kill the mock process and restart it.`);
```
Additionally, is there any way we could get the PID of the process so we could print `Kill the process using "kill -9 <PID>"`

packages/amplify-util-mock/src/api/api.ts
```diff
@@ -409,38 +410,40 @@ export class APITest {
...
-          message: errMessage,
+          message: 'The action to delete all items in an index is not acknowledged by the Opensearch server.',
           link: AMPLIFY_SUPPORT_DOCS.CLI_GRAPHQL_TROUBLESHOOTING.url
         });
       }
     }
+    catch(error) {
+      throw new AmplifyFault('MockProcessFault', {
+        message: errMessage,
+        details: error?.message || '',
```
edwardfoyle, line 443: this shouldn't be necessary if passing the error as the 3rd arg to the AmplifyFault constructor

edwardfoyle (changes requested):

packages/amplify-opensearch-simulator/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,169 @@
...
+      const fullPathToOpenSearchBinary = await openSearchEmulator.getPathToOpenSearchBinary(mockSearchableResourcePath);
+      expect(fullPathToOpenSearchBinary).toEqual(join(mockSearchableResourcePath, 'opensearchLib', 'bin', 'opensearch'));
+    });
+    
+    it('skips downloading another opensearch binary when one is locally available', async () => {
+      const openSearchExists = await openSearchEmulator.openSearchLocalExists(mockSearchableResourcePath);
+      // returns false when there is no local binary
+      expect(openSearchExists).toEqual(false);
+
+      const nodefetch = require('node-fetch');
```
jhockett, line 110: Why dynamically require node-fetch if mocking it on the next line?

packages/amplify-opensearch-simulator/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "openpgp": "^4.10.10",
+    "promise-toolbox": "^0.20.0",
+    "tar": "^6.1.11",
+    "wait-port": "^0.2.7",
+    "which": "^2.0.2"
+  },
+  "devDependencies": {
+    "@types/node": "^12.12.6",
+    "@types/openpgp": "^4.4.18",
+    "hash.js": "^1.1.7",
```
jhockett, line 42: If using `opengpg`, `hash.js` is probably not needed. I cannot find any usage of it in this PR.

packages/amplify-opensearch-simulator/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "uuid": "^8.3.2"
```
jhockett, line 43: I don't see uses of `uuid`, remove if not needed.

packages/amplify-util-mock/src/utils/opensearch.ts
```diff
@@ -0,0 +1,57 @@
...
+  const url = endpoint.replace(/\/+$/, '') + searchConfig.path;
+
+  const result = await fetch(url, {
+    method: 'POST',
+    body: JSON.stringify(searchConfig.params.body),
+    headers: {
+      'Content-type': 'application/json',
+    }
+  });
+  return await result.json();
```
jhockett, line 56:
```suggestion
  return result.json();
```

packages/amplify-opensearch-simulator/package.json
```diff
@@ -0,0 +1,49 @@
...
+    "execa": "^5.1.1",
+    "fs-extra": "^8.1.0",
+    "get-port": "^5.1.1",
+    "gunzip-maybe": "^1.4.2",
+    "node-fetch": "^2.6.7",
+    "openpgp": "^4.10.10",
+    "promise-toolbox": "^0.20.0",
+    "tar": "^6.1.11",
+    "wait-port": "^0.2.7",
+    "which": "^2.0.2"
```
jhockett, line 37: I don't see uses of `which`, remove if not needed.

packages/amplify-opensearch-simulator/index.js
phani-srikar: I am afraid that'll have unintended consequences if we share it across projects. The default config for local opensearch cluster used is `single-node` and I've run into some issues with spinning multiple clusters from the same binary. The name of the indices that are created might also clash if 2 projects use the same API schema. Also, the contents of `~/.amplify/lib` from what I see are mostly build-time assets that are included in `resources` directory of various packages.

packages/amplify-opensearch-simulator/package.json
phani-srikar: we do use it in the unit test and this is just a dev dependency.

packages/amplify-opensearch-simulator/src/__tests__/index.test.ts
phani-srikar: I want to check the number of calls to this method which requires me to mock it. I however, only want to mock it locally for this test, hence the dynamic import.

packages/amplify-opensearch-simulator/index.js
edwardfoyle: could the executable be stored in `~/.amplify/lib` but then store the index data in the project directory? This is what the ddb simulator does. The jar file lives in `~/.amplify/lib` but the table data is in the project

phani-srikar: Yes, I've made these changes. Thanks

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -347,3 +353,16 @@ export const getPathToOpenSearchBinary = async (pathToOpenSearchLocal?: string):
   }
   return join('opensearchLib', 'bin', 'opensearch');
 };
+
+export const getPackageAssetPaths: GetPackageAssetPaths = async () => [relativePathToOpensearchLocal];
+
+export const getOpensearchLocalDirectory = () => {
+  let opensearchLocalDir = join(pathManager.getAmplifyLibRoot(), packageName);
+  try {
+    opensearchLocalDir = pathManager.getAmplifyPackageLibDirPath(packageName);
```
edwardfoyle, line 362: if `getAmplifyPackageLibDirPath` it should just fail. It means something else has gone wrong with the lib setup

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -347,3 +353,16 @@ export const getPathToOpenSearchBinary = async (pathToOpenSearchLocal?: string):
   }
   return join('opensearchLib', 'bin', 'opensearch');
 };
+
+export const getPackageAssetPaths: GetPackageAssetPaths = async () => [relativePathToOpensearchLocal];
```
edwardfoyle, line 357: What files are being copied on CLI install?

packages/amplify-opensearch-simulator/src/__tests__/index.test.ts
```diff
@@ -0,0 +1,176 @@
...
+const execaMock = execa as jest.MockedFunction<typeof execa>;
+(execaMock as any).mockImplementation(async () => ({ stdout: 'mock-process-output' }));
+
+describe('emulator operations', () => {
+  const getMockSearchableFolder = (): string => {
+    let pathToSearchableMockResources = join(__dirname, '..', '..', 'resources', );
+    while (true) {
+      pathToSearchableMockResources = join('/tmp', `amplify-cli-opensearch-emulator-${v4()}`, 'mock-api-resources', 'searchable');
+      if (!fs.existsSync(pathToSearchableMockResources)) break;
+    }
```
jhockett, line 20:
```suggestion
    do {
      pathToSearchableMockResources = join('/tmp', `amplify-cli-opensearch-emulator-${v4()}`, 'mock-api-resources', 'searchable');
    } while (fs.existsSync(pathToSearchableMockResources));
```

packages/amplify-opensearch-simulator/src/__tests__/index.test.ts
jhockett:
```suggestion
      const nodefetch = await import('node-fetch');
```

packages/amplify-opensearch-simulator/src/index.ts
phani-srikar: No files are copied, just the directory structure is created as `~/.amplify/lib/amplify-opensearch-simulator/opensearch/1.3.0`.

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -0,0 +1,368 @@
...
+
+  public get port(): number|undefined {
+    return this.opts?.port;
+  }
+
+  public get url(): string {
+    return `http://localhost:${this.port}/`;
+  }
+
+  public terminate() {
```
jhockett, line 63: As best I can tell, this has a return type of the following:

```ts
  public terminate(): number | undefined | Promise<unknown> {
```

Please define the return types on all functions. If the above return type is accurate, a refactor is in order.

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -0,0 +1,368 @@
...
+    // already exited
+    if (this.proc?.exitCode != null) {
```
jhockett, line 65: `!==`

packages/amplify-opensearch-simulator/src/__tests__/index.test.ts
jhockett: Please use camel-case for variable names. The name in this case should be nodeFetch.

jhockett (commented):
I am seeing a lot of inconsistencies that would normally be caught by eslint. Please run the linter on all new files.

packages/amplify-opensearch-simulator/src/index.ts
phani-srikar: I've tried to disambiguate types wherever possible. There are still couple of functions like this one which need a deep dive and refactor. Like I noted above, can we do this when we abstract out common components of DynamoDB and opensearch emulators?

packages/amplify-util-mock/src/api/lambda-trigger-handler.ts
```diff
@@ -15,28 +16,42 @@ import { printer } from 'amplify-prompts';
...
-        throw new Error(`Unable to mock ${lambdaTriggerName}. ${mockable.reason}`);
+    if (lambdaTrigger?.name) {
+        // Lambda functions with layers are not mockable
+        const mockable = isMockable(context, lambdaTrigger?.name);
+        if (!mockable.isMockable) {
+            throw new AmplifyFault('MockProcessFault', {
+              message: `Unable to mock ${lambdaTrigger?.name}. ${mockable.reason}`,
+              link: AMPLIFY_SUPPORT_DOCS.CLI_GRAPHQL_TROUBLESHOOTING.url
+            });
+        }
```
jhockett, line 49: These are all errors not faults.

packages/amplify-util-mock/src/utils/opensearch.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  try {
+    searchConfig = searchConfig as OpensearchQueryConfig;
+  }
+  catch(e) {
+    throw new AmplifyFault('MockProcessFault', {
+      message: 'Given search query configuration is not valid',
+      link: AMPLIFY_SUPPORT_DOCS.CLI_GRAPHQL_TROUBLESHOOTING.url
+    });
+  }
```
jhockett, line 45: These are also errors

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -0,0 +1,362 @@
...
+};
+
+export const launch = async (
+  pathToOpenSearchData: string,
+  givenOptions: OpenSearchEmulatorOptions = {}, 
+  retry:number = 0, 
+  startTime: number = Date.now()
+): Promise<OpenSearchEmulator> => { 
+  if (isWindowsPlatform) {
+    throw new AmplifyFault('MockProcessFault', {
```
jhockett, line 120: This is an error

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -0,0 +1,362 @@
...
+    });
+  }
+
+  let { port } = { ...defaultOptions, ...givenOptions };
+  if (!port) {
+    port = await detectPort(basePort);
+  } else {
+    const freePort = await detectPort(port);
+    if (freePort !== port) {
+      throw new AmplifyFault('MockProcessFault', {
```
jhockett, line 149: I'm a little on the fence, but I think this is an error too

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -0,0 +1,362 @@
+import waitPort from 'wait-port';
+import detectPort from 'detect-port';
+import execa from 'execa';
+import { ensureDir, writeFileSync, existsSync } from 'fs-extra';
+import gunzip from 'gunzip-maybe';
+import nodefetch from 'node-fetch';
```
jhockett, line 6:
```ts
import nodeFetch from 'node-fetch';
```

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -60,10 +60,10 @@ export class OpenSearchEmulator {
     return `http://localhost:${this.port}/`;
   }
 
-  public terminate() {
+  public terminate(): Promise<unknown> {
```
edwardfoyle, line 63: is this return value used anywhere? If not, the return type should be `Promise<void>`

edwardfoyle (dismissed):
Left one last nit

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -117,7 +117,7 @@ export const launch = async (
   startTime: number = Date.now()
 ): Promise<OpenSearchEmulator> => { 
   if (isWindowsPlatform) {
-    throw new AmplifyFault('MockProcessFault', {
+    throw new AmplifyError('MockProcessError', {
```
jhockett, line 120:
```suggestion
    throw new AmplifyError('OpensearchMockUnsupportedPlatformError', {
```

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -146,7 +146,7 @@ export const launch = async (
   } else {
     const freePort = await detectPort(port);
     if (freePort !== port) {
-      throw new AmplifyFault('MockProcessFault', {
+      throw new AmplifyError('MockProcessError', {
```
jhockett, line 149:
```suggestion
      throw new AmplifyError('OpensearchMockPortUnavailableError', {
```

packages/amplify-util-mock/src/api/lambda-trigger-handler.ts
```diff
@@ -15,28 +16,42 @@ import { printer } from 'amplify-prompts';
 export const ddbLambdaTriggerHandler = async (
     context: $TSContext,
     streamArn?: string,
-    lambdaTriggerName?: string,
+    lambdaTrigger?: LambdaTrigger,
     localDynamoDBEndpoint?: Endpoint
 ): Promise<void> => {
-    if (!lambdaTriggerName) {
-        throw new Error('Name of the lambda trigger function must be specified');
+    if (!lambdaTrigger || (!lambdaTrigger?.name && !lambdaTrigger?.config)) {
```
jhockett, line 22: @phani-srikar I'm very sorry, I misread this code. It was correct earlier when these were Faults.

packages/amplify-util-mock/src/utils/opensearch.ts
```diff
@@ -0,0 +1,57 @@
...
+
+  try {
+    searchConfig = searchConfig as OpensearchQueryConfig;
+  }
+  catch(e) {
+    throw new AmplifyError('MockProcessError', {
+      message: 'Given search query configuration is not valid',
+      link: AMPLIFY_SUPPORT_DOCS.CLI_GRAPHQL_TROUBLESHOOTING.url
+    });
```
jhockett, line 44: After diving into the code I think these are Faults, sorry about the mistakes @phani-srikar

packages/amplify-opensearch-simulator/src/index.ts
```diff
@@ -0,0 +1,362 @@
...
+  printer.info('Spawning OpenSearch Simulator with options: ' + JSON.stringify({ args, cwd: getOpensearchLocalDirectory() }));
+  const openSearchBinPath = await getPathToOpenSearchBinary();
+
+  const proc = execa(openSearchBinPath, args, {
+    cwd: getOpensearchLocalDirectory(),
+  });
+
+  const emulator = await startOpensearchEmulator(opts, proc, port, startTime, givenOptions, getOpensearchLocalDirectory(), retry);
+  if (!emulator) {
+    throw new AmplifyFault('MockProcessFault', {
```
jhockett, line 167:
```suggestion
    throw new AmplifyError('OpensearchMockProcessFault', {
```

jhockett (approved):

sundersc (approved):

edwardfoyle (approved):

pavellazar (approved):

PR #11095, opened 2022-10-03
chore: error handling migration for delete command

awsluja:
#### Description of changes

Migrate the 'delete' flow to use the new error handling.

Run 'delete' command on an existing amplify project
Run delete-project-test.ts

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli-core/src/feature-flags/featureFlags.ts
```diff
@@ -158,7 +159,11 @@ export class FeatureFlags {
     FeatureFlags.ensureInitialized();
 
     if (!envNames) {
-      throw new Error("'envNames' argument is required");
+      // this is an internal issue, we either couldn't load the environment names
+      // or their configuration is invalid, further troubleshooting is needed
+      throw amplifyFaultWithTroubleshootingLink('ConfigurationFault', {
+        message: 'Environment names could not be loaded or were not provided.',
+      });
```
sobolk, line 166: do you know why do we have `amplifyFaultWithTroubleshootingLink` factory method instead of defaulting the `link` in `AmplifyError` ctor ?
The naming of `amplifyFaultWithTroubleshootingLink` is quite unfortunate. If `a` was capital letter it would look like subtype of `AmplifyError`. There's a verb missing there.

packages/amplify-cli/src/__tests__/extensions/amplify-helpers/delete-project.test.ts
```diff
@@ -1,3 +1,6 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+/* eslint-disable @typescript-eslint/no-empty-function */
```
sobolk, line 2: how commonly do we disable those linter rules (and the other one below) in tests? If it's a pattern perhaps tests could use different set of rules.

awsluja: very often, I'm not sure why yet, but tests have a unique import resolution strategy that I don't understand

packages/amplify-cli-core/src/feature-flags/featureFlags.ts
awsluja: I have proposed a slight refactor of our error handling classes and utilities (we can use decorator or factory); but it is a second priority to this at the moment. We are concluding work with the error handling migration, so it became a lower priority (we have already completed the work for the top 6 commands, which was the original goal); I will refactor if I have extra time this sprint

awsluja: here is the task for follow up https://app.asana.com/0/1203051621673505/1203051593370893/f

sobolk: Yup. It's a legacy from POV of this PR. However, if we spread coverage the follow up might be big. so good to think about doing that with next error handler migration piece.

sobolk (dismissed):

lazpavel (dismissed):

packages/amplify-cli-core/src/errors/amplify-exception.ts
```diff
@@ -123,6 +123,7 @@ export type AmplifyFaultType =
   | 'AmplifyBackupFault'
   | 'BackendPullFault'
   | 'BackendDeleteFault'
+  | 'ConfigurationFault'
```
sobolk, line 126: possible follow up. Should `Fault` and `Error` be separate dimension? Looks like we're mixing something I'd call area or component with severity.

sobolk (dismissed):

packages/amplify-cli-core/src/feature-flags/featureFlags.ts
sobolk: Btw I noticed on the other PR that amplifyFaultWithTroubleshootingLink is deprecated.

lazpavel (dismissed):

packages/amplify-cli/src/commands/delete.ts
```diff
@@ -1,11 +1,15 @@
...
+export const run = async (context: $TSContext) : Promise<void> => {
   if (Array.isArray(context.parameters.array) && context.parameters.array.length > 0) {
-    context.print.error('"delete" command does not expect additional arguments.');
-    context.print.error('Perhaps you meant to use the "remove" command instead of "delete"?');
-    context.usageData.emitError(new UnknownArgumentError());
-    exitOnNextTick(1);
+    throw new AmplifyError('CommandNotSupportedError', {
+      message: 'The "delete" command does not expect additional arguments.',
+      details: 'Perhaps you meant to use the "remove" command instead of "delete"?',
+      resolution: 'Try the "delete" command again without any additional arguments.',
```
jhockett, line 11: Maybe we can be clear on the command's intent in the resolution when additional args are present:

```suggestion
      resolution: 'If you intend to delete this project and all backend resources, try the "delete" command again without any additional arguments.',
```

lazpavel (approved):

lazpavel (approved):

PR #11118, opened 2022-10-05
fix: regression caused by new error thrown

awsluja:
#### Description of changes
Previously, we were ignoring errors that had the type of "NotFoundException" when an S3 bucket did not exist and a user tried to delete it. If this is the case, we would exit without throwing an error. 
The problem is that the "NotFoundException" was changed to "BucketNotFoundError" of AmplifyError type.
The remove-env-from-cloud.ts file was not updated to match, so it is throwing errors when S3 bucket does not exist/is already deleted.

Passing unit tests:
image

E2e run pending

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/extensions/amplify-helpers/remove-env-from-cloud.ts
```diff
@@ -31,12 +31,12 @@ export const removeEnvFromCloud = async (context, envName, deleteS3): Promise<vo
   try {
     await Promise.all(providerPromises);
     await raiseInternalOnlyPostEnvRemoveEvent(context, envName);
-  } catch (e) {
-    if (e.code !== 'NotFoundException') {
+  } catch (ex) {
+    if (!('name' in ex && ex.name === 'BucketNotFoundError')) {
```
sobolk, line 35: nit: would `ex?.name` simplify this ?

packages/amplify-cli/src/__tests__/extensions/amplify-helpers/remove-env-from-cloud.test.ts
```diff
@@ -68,8 +69,10 @@ describe('remove-env-from-cloud', () => {
   });
 
   it('does not throw not found error when deleteEnv promise rejected', async () => {
-    const e: any = new Error('deleteEnv error');
-    e.code = 'NotFoundException';
+    const e: any = new AmplifyError('BucketNotFoundError', {
+      message: 'Project deployment bucket has not been created yet.',
+      resolution: 'Use amplify init to initialize the project.',
+    });
```
sobolk, line 75: is it safe to change error code?

lazpavel (approved):

packages/amplify-cli/src/__tests__/extensions/amplify-helpers/remove-env-from-cloud.test.ts
awsluja: I just updated it so its' more clear what we're testing. In this case, we want to make sure that we aren't throwing exceptions if its just a "BucketNotFoundError" when deleting

jhockett (approved):

akshbhu (approved):

packages/amplify-cli/src/extensions/amplify-helpers/remove-env-from-cloud.ts
```diff
@@ -31,12 +31,12 @@ export const removeEnvFromCloud = async (context, envName, deleteS3): Promise<vo
   try {
     await Promise.all(providerPromises);
     await raiseInternalOnlyPostEnvRemoveEvent(context, envName);
-  } catch (e) {
-    if (e.code !== 'NotFoundException') {
+  } catch (ex) {
+    if (!(ex?.name === 'BucketNotFoundError')) {
```
lazpavel, line 35: just a nit, it might not be the case but if the `ex` is not instance of `AmplifyException`, it will throw an error as it will try to access a property that doesn't exist, if we keep this version can we use `(ex?.name !== 'BucketNotFoundError')` to remove some braces

Issue #1525, opened 2019-05-27
Remove aws-sdk from functions as default library


slaat:
**Is your feature request related to a problem? Please describe.**
The total size of a lambda function (dist file) is at about 8MB. Most of it is because of the aws-sdk library that is being included. Lambda Functions have this library preinstalled which means that this is not necessary. 

**Describe the solution you'd like**
Remove the aws-sdk as a default from the package.json dependencies

**Additional context**
Dist file size before: 8mb
Dist file size after: 700kb

Function folder size before: 45mb
Function folder size after: 3mb

The total size can be reduced by over 90%. My overall project size (a project with 17 functions) has reduced by ~800mb, which makes a huge difference.

kstro21:
Ideally, `aws-sdk` is added in **package.json** as a **devDependencies**, later, in the process of _zipping_, only **production** dependencies are added, by using `npm install --production`.

slaat:
That would still add about 40mb of data to the node_modules of each lambda function...

PR #11128, opened 2022-10-06
chore: add aws-sdk v2 to devDependencies for function mock

josefaidt:
#### Description of changes

- adds `aws-sdk` v2 to Node.js functions' `package.json` as a devDependency. Enables local mock with AWS SDK

#### Issue #, if available

related https://github.com/aws-amplify/amplify-cli/issues/1525
related https://github.com/aws-amplify/amplify-cli/issues/12428

#### Checklist

- [x] PR description included


danielleadams:
@josefaidt it looks like you linked this to a closed issue. Is this still an ongoing issue?

josefaidt:
Hey @danielleadams it's indirectly related to that issue. When folks want to use `aws-sdk` they will need to add this manually as a devDependency for `mock`, where it is provided OOTB in the Lambda environment. This change will ensure the dependency is auto-installed on `mock` but should not conflict with the provided SDK

danielleadams:
@josefaidt it looks like a lot of tests are failing when I reran this 3 days ok, so you may need to merge dev so that we can rerun tests.

josefaidt:
@danielleadams done

josefaidt:
fwiw this may cause issues once we default to creating Node 18 Lambda's since the updated runtime provides AWS SDK v3 (ref https://github.com/aws-amplify/amplify-cli/issues/11444)

packages/amplify-nodejs-function-template-provider/resources/lambda/crud/package.json.ejs
```diff
@@ -11,6 +11,6 @@
   },
   "devDependencies": {
     "@types/aws-lambda": "^8.10.92",
-    "aws-sdk": "^2.1113.0"
+    "aws-sdk": "^2.1230.0"
```
evcodes, line 14: I think that this is no-op because the carat symbol would already resolve ^2.1113.0 to the same version as ^2.1230.0

evcodes (commented):
small note

packages/amplify-nodejs-function-template-provider/resources/lambda/crud/package.json.ejs
jhockett: While true in a semantic versioning sense, in practice it's not necessarily true if dependencies are locked. We should standardize to a single version so that only a single version of the `aws-sdk` is needed.

danielleadams (approved):

sobolk:
Closing this PR.

1. We migrated labmda templates to use Node 18.x runtime. That works with SDK V3.
2. Templates that require SDK deps got them added to dev deps in https://github.com/aws-amplify/amplify-cli/pull/12394 .

PR #11287, opened 2022-10-29
fix: stop timer on early push resources exit

pavellazar:
#### Description of changes
- prevents leaving the usage timer started on early exit from push resources
note: issue was discovered while testing notifications migration which triggers push for auth resource followed up by a push for analytics, if auth was previously pushed, first push will exit early leaving the timer started and followup push for analytics will throw an exception attempting to start a started timer

- manual testing
- unit test added

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
@@ -78,8 +78,8 @@ export const pushResources = async (
 
   // no changes detected
   if (!hasChanges && !context.exeInfo.forcePush && !rebuild) {
+    context.usageData.stopCodePathTimer(ManuallyTimedCodePath.PUSH_TRANSFORM);
     printer.info('\nNo changes detected');
-
     return false;
   }
 
--- a/packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
@@ -78,8 +78,8 @@ export const pushResources = async (
 
   // no changes detected
   if (!hasChanges && !context.exeInfo.forcePush && !rebuild) {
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
```diff
@@ -78,8 +78,8 @@ export const pushResources = async (
 
   // no changes detected
   if (!hasChanges && !context.exeInfo.forcePush && !rebuild) {
+    context.usageData.stopCodePathTimer(ManuallyTimedCodePath.PUSH_TRANSFORM);
     printer.info('\nNo changes detected');
-
```
danielleadams, line 82: Should this go after the print? Not sure if it really makes a difference in the metric, but it might be worth factoring in.

packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
```diff
@@ -78,8 +78,8 @@ export const pushResources = async (
 
   // no changes detected
   if (!hasChanges && !context.exeInfo.forcePush && !rebuild) {
+    context.usageData.stopCodePathTimer(ManuallyTimedCodePath.PUSH_TRANSFORM);
```
sobolk, line 81: Could this be solved by `try { } finally {}` ? Or some AOP technique? 
This will come back if somebody adds another exit point in this function.

lazpavel: I didn't added a try finally because the stop is actually called here, looks that the push timer is split in 2 different methods, one is starting it, and the second one is stopping: https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-provider-awscloudformation/src/push-resources.ts#L276
cc: @edwardfoyle

danielleadams (approved):

sobolk (approved):

Issue #11252, opened 2022-10-25
Use Node.js runtime v16.x by default


fossamagna:
### Is this feature request related to a new or existing Amplify category?

function

### Describe the feature you'd like to request

I'd like to use Node.js v16.x by default in function category.

### Describe the solution you'd like

Node.js Lambda functions and triggers will use 16.x by default.

### Describe alternatives you've considered

Update manually CloudFormation template.
See: https://docs.amplify.aws/guides/functions/configuring-lambda/q/platform/js/#updating-the-runtime

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

PR #11317, opened 2022-11-03
chore: update node.js version handling for CLI

fossamagna:
#### Description of changes

- Lambda resource handlers will use 16.x by default
- Lambda functions and triggers will use 16.x by default

#### Issue #, if available

fix: https://github.com/aws-amplify/amplify-cli/issues/11252

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-analytics/src/provider-utils/awscloudformation/cloudformation-templates/pinpoint-cloudformation-template.json
```diff
@@ -121,13 +119,9 @@
             {
               "Effect": "Allow",
               "Principal": {
-                "Service": [
-                  "lambda.amazonaws.com"
-                ]
+                "Service": ["lambda.amazonaws.com"]
```
danielleadams, line 122: There seem to be a lot of unrelated formatting changes. Do you have a linter that is turned on? Is there any chance that you can separate the linter changes from the upgrade changes? This makes the pull request more difficult to review with all the additional changes.

fossamagna: @danielleadams Thanks for the suggestion. I have turned off the linter. I have committed and pushed only changes that update the nodejs version.

danielleadams: great - thank you.

danielleadams:
@fossamagna I started an integration test run for CLI. Not sure why its not updating the GitHub PR, but I will keep an eye on it to see that it passes.

danielleadams:
@fossamagna running your changes through the test suite.

GeorgeBellTMH:
It would be good if before pushing we checked to ensure that both the current and updated node versions are supported by AWS. If the current one isn't supported we may not be able to roll back, if the future one isn't then it shouldn't push. In both cases there should be an override with some warning about how to manually fix if a rollback gets stuck (ie. you have to go in and edit the lambda directly).

danielleadams:
@fossamagna The tests seem to have stalled at the very end, so I'm going to restart them, but I've marked this as semver-major, so they will go out in v11 (don't have a date for that yet).

danielleadams:
> It would be good if before pushing we checked to ensure that both the current and updated node versions are supported by AWS. 

@GeorgeBellTMH Both Node.js 14 and 16 are supported by AWS Lambda: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html.

GeorgeBellTMH:
They won't be forever...

sobolk:
@GeorgeBellTMH 
since this change warrants major version bump. could you please target this PR to `cdkv2` branch where we're staging V11?

sobolk:
@fossamagna 
since this change warrants major version bump. could you please target this PR to `cdkv2` branch where we're staging V11?

fossamagna:
@danielleadams @sobolk Thanks, I changed PR target to `cdkv2` branch. I look forward to V11.

sobolk (approved):

PR #11343, opened 2022-11-07
fix: generate default theme if user has no themes

letsbelopez:
#### Description of changes

Use the default theme from amplify ui in the users project if they don't own their own themes.

Upgrade codegen-ui in util-uibuilder package

#### Issue #, if available

https://app.asana.com/0/1202185248783861/1203287913178604/f

Unit test

In a new account without a theme, I went through the getting started instructions for Form builder. And the default theme was used and generated instead.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/uibuilder.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/uibuilder.test.ts
@@ -75,7 +75,9 @@ describe('amplify pull with uibuilder', () => {
     expect(fileList).toContain('MyIcon.d.ts');
     expect(fileList).toContain('MyIcon.jsx');
     expect(fileList).toContain('index.js');
-    expect(fileList).toHaveLength(5);
+    expect(fileList).toContain('studioTheme.js');
+    expect(fileList).toContain('studioTheme.js.d.ts');
+    expect(fileList).toHaveLength(7);
 
     spawnSync(
       getNpmPath(),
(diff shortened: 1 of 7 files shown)


packages/amplify-util-uibuilder/src/commands/generateComponents.ts
```diff
@@ -42,7 +43,7 @@ export const run = async (context: $TSContext): Promise<void> => {
 
     const generatedResults = {
       component: generateUiBuilderComponents(context, componentSchemas.entities, dataSchema),
-      theme: generateUiBuilderThemes(context, themeSchemas.entities),
+      theme: themeSchemas.entities.length > 0 ? generateUiBuilderThemes(context, themeSchemas.entities) : [generateDefaultTheme(context)],
```
jhockett, line 46: `commands/generateComponents.ts` doesn't seem like the best place for this logic to live.

I would rather see `generateUiBuilderThemes` fall back to the default theme if no `themeSchemas.entities` are present. That's one less exported method to reason about.

jhockett (changes requested):

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
letsbelopez: @jhockett I made the update

jhockett (dismissed):

jhockett (approved):

akshbhu (approved):

PR #11346, opened 2022-11-07
chore: add api extractor

sobolk:
#### Description of changes

This PR:
- brings `api-extractor` into toolchain.
- adds `API.md` files that extract public contract
- adds Circle Ci check that fails if API.md files has not be regenerated.

This accomplishes two goals:
- we know what our public surface is
- there's a gate in PR process that forces API.md regeneration, thus giving reviewers and author a chance to review public are changes in separation. (I.e. asses what version number component they should bump.
- we know which 3rd party types we expose (I.e. which CDK constructs are part of our public surface).

Out of scope.
Automatic update of API.md files.
The scan and generation of API.md files takes ~1-2 minutes. Presumably this should happen rarely and doesn't block dev-test feedback loop.
Possible solutions that could work in the future:
- git hooks and re-generation of delta. I.e. detect which packages should be updated and run extractor only there. (still heavy and still requires circle ci check.
- have circle CI commit delta directly to PR if change is detected. (inspired by https://github.com/aws-amplify/amplify-cli-export-construct/blob/9e1fa94157ad4b65b2eaa2b9b1f29426374bf381/.github/workflows/build.yml#L34-L40 )

#### Checklist

- [x ] `yarn test` passes
- [ x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)


scripts/extract-api.ts
```diff
@@ -0,0 +1,40 @@
+import * as fs from 'fs-extra';
```
awsluja, line 1: should we just add this as a pre-push hook?
We do something like this for split-e2e-tests, which generates our circleci template & verifies correctness

sobolk: @awsluja I have explained that in PR description.

awsluja: we may also be able to leverage lerna/nx so that this script runs only for packages that have changed

nx.json
```diff
@@ -10,31 +10,44 @@
...
-    "package": [
-      {
-        "target": "package",
-        "projects": "dependencies"
-      }
+  "namedInputs": {
+    "default": [
+      "{projectRoot}/**/*",
+      "!{projectRoot}/**/*.md"
     ]
```
sobolk, line 23: image

https://nx.dev/concepts/task-pipeline-configuration#define-task-dependencies-(aka-task-pipelines)

nx.json
```diff
@@ -10,31 +10,44 @@
...
+      "dependsOn": [ "build" ]
+    },
+    "extract-api": {
+      "dependsOn": [ "build" ],
+      "inputs": [
+        "{projectRoot}/src/**/*",
+        "!{projectRoot}/src/__tests__/**/*"
+      ],
+      "outputs": [ "{projectRoot}/API.md" ]
+    }
```
sobolk, line 49: see https://nx.dev/more-concepts/customizing-inputs and https://nx.dev/reference/project-configuration for reference.

this also replaces older syntax.

scripts/extract-api.ts
sobolk: @awsluja I added verification on push and harnessed nx to make it more preformant.

jhockett (commented):
PR LGTM, but my one concern is that extracting an API from private packages like `amplify-e2e-core` is probably not needed. Removing this will should help with performance a bit.

akshbhu (commented):
PR looks good. 
One thing I was thinking is do we want to expose all details of all packages explicitly.
One other way  is to have snapshots  testing to keep it more internal and  to know the diffs and what are we exposing extra

jhockett (approved):

sobolk:
> PR looks good. One thing I was thinking is do we want to expose all details of all packages explicitly. One other way is to have snapshots testing to keep it more internal and to know the diffs and what are we exposing extra

This tooling is intended to help reviewers assess public API changes. This is a report, neither test nor documentation.

jhockett (approved):

akshbhu (approved):

PR #11349, opened 2022-11-08
chore(api): update api dependencies to 4.0.9

sundersc:
- Update API dependencies to use the version 4.0.9.
- Clean up CDK dependencies used by CLI and lock to 1.172.0.

PR #11354, opened 2022-11-09
test(api): update codegen snapshot to match the latest codegen version

sundersc:
Update studio modelgen snapshot to match the latest codegen version.


sundersc:
Added this change to update API dependencies PR #11349.

PR #11366, opened 2022-11-09
fix(cli): fixed list permissions and if statement condition

goldbez:
#### Description of changes

The binaries upload script fails to list the contents of buckets in s3://aws-amplify-cli-do-not-delete due to a permissions issue. It also fails to exit when this permissions issue occurs. 

The permissions issue was caused by the omission of the —profile=s3-uploader flag on line 34. The failure to exit was caused by the AWS CLI outputting noting to stdout in the case of an error. This empty output was piped to egrep and wc. Since its input was empty, wc ended up counting zero lines, preventing the IF-statement from being entered.

To fix this, I added the necessary flag and added a pipefail so that the positive exit code of the AWS CLI would be persisted through the pipe. I then save this exit code and the output of the pipe. If the exit code is positive, it means there was a permissions error with the list operation. If the pipe output is positive, it means there are files that cannot be overwrite. In either case, the script exits due to the new IF-statements I added.

#### Issue #, if available
Screen Shot 2022-11-09 at 3 34 56 PM

Tested all combinations of permissions/file errors in local_publish_helpers_test.sh.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.circleci/local_publish_helpers.sh
```diff
@@ -31,7 +31,15 @@ function uploadPkgCli {
         aws --profile=s3-uploader s3 cp amplify-pkg-linux-arm64.tgz s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-arm64-$(echo $hash).tgz
         aws --profile=s3-uploader s3 cp amplify-pkg-linux-x64.tgz s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-x64-$(echo $hash).tgz
 
-        if [ "0" -ne "$(aws s3 ls s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-x64 | egrep -v "amplify-pkg-linux-x64-.*" | wc -l)" ]; then
+        INCORRECT_FILES="$(set -o pipefail && aws --profile=s3-uploader s3 ls s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-x64 | ( egrep -v "amplify-pkg-linux-x64-.*" || true ) | wc -l | xargs)"
```
edwardfoyle, line 34: what is the `xargs` on the end doing here?

.circleci/local_publish_helpers.sh
```diff
@@ -31,7 +31,15 @@ function uploadPkgCli {
...
+        INCORRECT_PERMISSIONS=$?
+
+        if [ INCORRECT_PERMISSIONS -ne "0" ]; then
```
edwardfoyle, line 37: I see below you tested with double bracketed conditions but here they are single brackets. My bash knowledge isn't good enough to know if it matters here but I know there are subtle differences between the two. Just making sure this was intentional

.circleci/local_publish_helpers.sh
goldbez: Screen Shot 2022-11-09 at 2 45 29 PM

.circleci/local_publish_helpers_test.sh
```diff
@@ -0,0 +1,132 @@
...
+FILE_NAME="test"
+BAD_FILE_NAME="test2"
+
+touch $FILE_NAME
+touch $BAD_FILE_NAME
+
+
+#####
+
+echo "Testing with correct permissions and correct files"
```
sobolk, line 20: how do we run this test as part of PR validation ?

.circleci/local_publish_helpers.sh
```diff
@@ -31,7 +31,15 @@ function uploadPkgCli {
...
+            echo "Insufficient permissions to list s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-x64"
+            exit 1
+        fi
+
+        if [ INCORRECT_FILES -ne "0" ]; then
```
sobolk, line 42: does "incorrect" mean "already existing" in this context ?

.circleci/local_publish_helpers.sh
```diff
@@ -31,7 +31,15 @@ function uploadPkgCli {
         aws --profile=s3-uploader s3 cp amplify-pkg-linux-arm64.tgz s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-arm64-$(echo $hash).tgz
         aws --profile=s3-uploader s3 cp amplify-pkg-linux-x64.tgz s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-x64-$(echo $hash).tgz
 
-        if [ "0" -ne "$(aws s3 ls s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-x64 | egrep -v "amplify-pkg-linux-x64-.*" | wc -l)" ]; then
+        INCORRECT_FILES="$(set -o pipefail && aws --profile=s3-uploader s3 ls s3://aws-amplify-cli-do-not-delete/$(echo $version)/amplify-pkg-linux-x64 | ( egrep -v "amplify-pkg-linux-x64-.*" || true ) | wc -l | xargs)"
+        INCORRECT_PERMISSIONS=$?
```
sobolk, line 35: what does this do?

.circleci/local_publish_helpers.sh
goldbez: Yes

.circleci/local_publish_helpers.sh
sobolk: I'm refering to `INCORRECT_PERMISSIONS=$?`

.circleci/local_publish_helpers.sh
sobolk: Please give it more descriptive name then.

.circleci/local_publish_helpers.sh
goldbez: Line 34 sets pipefail so that the rightmost non-zero exit code in the pipe is returned, calls ls on the S3 bucket, pipes the output to egrep to select all files that do not begin with "amplify-pkg-linux-x64-.*" (the egrep command is ORed with true so that it is forced to have a zero exit code because egrep exits with code 1 if one or more line is returned), pipes the output to wc to see if any such files exist, and then pipes that to xargs to remove the whitespace that wc adds. This output is saved to the variable INCORRECT_FILES. If this is non-zero, it means that files that shouldn't be overwritten exist in the bucket.

Line 35 saves the exit code of the whole pipe to the variable INCORRECT_PERMISSIONS. If the AWS CLI ls call fails because of a permissions error, it will exit with a non-zero exit code.

goldbez: `INCORRECT_PERMISSIONS=$?` saves the exit code of the whole pipe to the variable INCORRECT_PERMISSIONS. $? is a special variable that holds the value of the exit code of the last command.

.circleci/local_publish_helpers.sh
goldbez: I just added another commit with this change.

.circleci/local_publish_helpers_test.sh
goldbez: The team decided to remove this test file and to later figure out a way to convert the bash scripts to typescript so that they can be tested better. This file has now been removed from this PR.

sobolk (approved):

edwardfoyle (approved):

PR #11374, opened 2022-11-10
fix: should not print auth warn when adding auth to new project

lazpavel:
#### Description of changes
- prevents showing auth warning when adding auth to a new project

- manual and unit tests added

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-provider-awscloudformation/src/__tests__/utils/auto-updates.test.ts
```diff
@@ -74,9 +79,34 @@ describe('showBuildDirChangesMessage', () => {
...
+      fsMock.existsSync.mockReturnValue(false);
+    });
+
+    it('does not call warn', async () => {
+      await showBuildDirChangesMessage();
+      expect(printerMock.warn).not.toBeCalled();
+    });
+  });
+
+  describe('project does not have auth 2', () => {
```
danielleadams, line 101: Is this named right `existsSync` returns `true` means that auth is in backend cloud right?

danielleadams (changes requested):
I'm a little confused about the test naming, but other than that looks good.

packages/amplify-provider-awscloudformation/src/__tests__/utils/auto-updates.test.ts
lazpavel: this is the behavior that was previously tested, folder exists but auth not in the cloud, will update test names

packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
```diff
@@ -12,37 +12,31 @@ const {
  * Updates Cognito files that are tracked so that the diff is detected for an `amplify push`
  */
 export const updateCognitoTrackedFiles = async (): Promise<void> => {
+  if (await detectCognitoAttributesRequireVerificationBeforeUpdateDiff()) {
+    const { resourceName } = stateManager.getResourceFromMeta(stateManager.getMeta(), 'auth', 'Cognito', undefined, false)!;
+    await addExtraLineToCliInputsJson(pathManager.getBackendDirPath(), resourceName);
```
sobolk, line 17: what does this do? (apart from adding extra line, i.e. why adding line is important?)

sobolk (approved):

packages/amplify-cli/src/extensions/amplify-helpers/update-tracked-files.ts
lazpavel: it enforces a resource change to be pushed

danielleadams (approved):

PR #11403, opened 2022-11-15
chore: update update-data-dependencies.sh script to allow for bumping to a tag version

alharris-at:
#### Description of changes
I made these changes to make supporting the `feature/rds-support` branch a bit easier. Change is already present there.

#### Issue #, if available
N/A

Tested on the `feature/rds-support` branch where the package.json file specifies the tag to use.

https://github.com/aws-amplify/amplify-cli/blob/feature/rds-support/package.json#L49

#### Checklist
- [x] PR description included

### DIFF
--- a/scripts/update-data-dependencies.sh
+++ b/scripts/update-data-dependencies.sh
@@ -12,12 +12,14 @@ if [ $# -eq 0 ]
     npx ncu \
         --deep \
         --upgrade \
+        --dep "prod,dev,peer,bundle,optional" \
         --filter "$FILTER"
   else 
     echo "Updating to $1 tag"
     npx ncu \
         --deep \
         --upgrade \
+        --dep "prod,dev,peer,bundle,optional" \
         --filter "$FILTER" \
         --target $1
 fi
--- a/scripts/update-data-dependencies.sh
+++ b/scripts/update-data-dependencies.sh
@@ -1,6 +1,23 @@
 #!/usr/bin/env sh
 
-npx ncu \
-    --deep \
-    --upgrade \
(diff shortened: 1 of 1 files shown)


sobolk (approved):

scripts/update-data-dependencies.sh
```diff
@@ -1,6 +1,23 @@
...
+        --deep \
+        --upgrade \
+        --filter "$FILTER"
+  else 
+    echo "Updating to $1 tag"
+    npx ncu \
+        --deep \
+        --upgrade \
+        --filter "$FILTER" \
+        --target $1
```
sobolk, line 22: @alharris-at 
thing to consider. This doesn't update peer deps by default. should we include them here?

alharris-at: Yeah, I'll update to specify the peer deps as well. I don't think it'll make an impact in CLI, but we want that in the codegen/api category repos for sure, and I'd like to keep the functionality consistent.

sobolk (approved):

scripts/update-data-dependencies.sh
```diff
@@ -1,6 +1,25 @@
...
+#      `npx scripts/update-data-dependencies.sh update-data-packages @rds-support`
+
+FILTER="amplify-codegen @aws-amplify/amplify-category-api @aws-amplify/graphql-auth-transformer @aws-amplify/graphql-default-value-transformer @aws-amplify/graphql-function-transformer @aws-amplify/graphql-http-transformer @aws-amplify/graphql-index-transformer @aws-amplify/graphql-maps-to-transformer @aws-amplify/graphql-model-transformer @aws-amplify/graphql-predictions-transformer @aws-amplify/graphql-relational-transformer @aws-amplify/graphql-schema-test-library @aws-amplify/graphql-searchable-transformer @aws-amplify/graphql-transformer-core @aws-amplify/graphql-transformer-interfaces @aws-amplify/graphql-transformer-migrator graphql-auth-transformer graphql-connection-transformer graphql-dynamodb-transformer graphql-elasticsearch-transformer graphql-function-transformer graphql-http-transformer graphql-key-transformer graphql-mapping-template graphql-predictions-transformer graphql-relational-schema-transformer graphql-transformer-common graphql-transformer-core graphql-transformers-e2e-tests graphql-versioned-transformer"
```
AaronZyLee, line 7: Just one question. The filter also includes `amplify-codegen` which might not have a tagged release at the same time with API category. Will this script ignore the update if the tag does not exist?

AaronZyLee (approved):

scripts/update-data-dependencies.sh
alharris-at: Yep! It'll only pull if there's a relevant update.

Issue #10180, opened 2022-04-10
Can't add third-party NPM plugins since Amplify CLI v8.0.0


LaurensBrinker:
### Before opening, please confirm:

- [X] I have installed the latest version of the Amplify CLI (see above), and confirmed that the issue still persists.
- [X] I have [searched for duplicate or closed issues](https://github.com/aws-amplify/amplify-cli/issues?q=is%3Aissue+).
- [X] I have read the guide for [submitting bug reports](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#bug-reports).
- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

14.19.1 (but I checked and the error is also there for v16.14.2, so seems Node version doesn't affect it)

### Amplify CLI Version

8.0.0

### What operating system are you using?

Ubuntu, Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No manual changes made

### Amplify Categories

Not applicable

### Amplify Commands

add

### Describe the bug

I'm one of the authors of this third-party Amplify plugin: https://www.npmjs.com/package/amplify-graphql-seed-plugin

In previous versions of the Amplify CLI, our users could install our plugin using NPM directly following the method described in the [following official Amplify documentation](https://docs.amplify.aws/cli/plugins/authoring/#step-5-install-and-use):
```
npm install -g amplify-graphql-seed-plugin
amplify plugin add amplify-graphql-seed-plugin
```

However, since Amplify CLI version 8.0.0, this is no longer working. Instead, when running the same commands, the following error emerges:
```
$ npm install -g amplify-graphql-seed-plugin
$ amplify plugin add amplify-graphql-seed-plugin
Can not locate the plugin package.
? Enter the absolute path for the root of the plugin directory:
```

This is not an issue with our Plugin specifically, as e.g. [this plugin](https://www.npmjs.com/package/amplify-category-data-importer) can't be added either.

Two workarounds described below:
* Downgrade Amplify version
* Clone source-code of Plugin from GitHub to project, and add plugin by providing absolute path (undesirable since this increases the complexity of adoption of plugin, especially in CI/CD pipelines)

### Expected behavior

```
$ npm install @aws-amplify/cli
[...]

PR #11408, opened 2022-11-16
fix(global-prefix): remove global-prefix package

MorCohenAres:
Addressing this issue:
https://github.com/aws-amplify/amplify-cli/issues/10180

#### Description of changes

The global-prefix package, which is the source of the issue, is not supported anymore and we cannot merge [this PR](https://github.com/jonschlinkert/global-prefix/pull/29).
I moved it's (tiny) logic into `amplify-cli` **with fixing the [bug](https://github.com/jonschlinkert/global-prefix/issues/30)**

ran `yarn build` & `pkg-all-local` before and after the changes, and see if it recognizes my `amplify-category-video` plugin.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


katieklein (dismissed):

packages/amplify-cli/src/utils/global-prefix.ts
```diff
@@ -21,3 +25,62 @@ function getYarnPrefix() {
...
+  if (home) {
+    const prefix = tryConfigPath(path.resolve(home, '.npmrc'));
+    if (prefix) return prefix;
+  }
+
+  // Find global NPM path
+  const npm = tryWhich('npm');
+  if (npm) {
+    // Check the built-in npm config file
+    const prefix = tryConfigPath(path.resolve(npm, '..', '..', 'npmrc'));
```
evcodes, line 45: Can you elaborate a little on what this is doing? [From the docs](https://nodejs.org/api/path.html#pathresolvepaths), path.resolve it seems like there's an extraneous '..' in the array of paths.

packages/amplify-cli/src/utils/global-prefix.ts
```diff
@@ -21,3 +25,62 @@ function getYarnPrefix() {
...
+    if (process.platform === 'win32' || OSTYPE === 'msys' || OSTYPE === 'cygwin') {
+      return APPDATA ? path.join(APPDATA, 'npm') : path.dirname(nodePath);
+    }
+  
+    // UNIX
+    // /usr/local/bin/node --> prefix=/usr/local
+    const prefix = path.dirname(path.dirname(nodePath));
+    if (DESTDIR) {
+      return path.join(DESTDIR, prefix);
+    }
```
evcodes, line 68: I would try to follow the same pattern that you implemented for windows in line 59 where you check the ```process.platform === linux```

evcodes: Also feel free to elaborate on what your intentions are

eddy2048 (dismissed):
please feel free to give me additional context into the issue that influenced your design/code choices :)

packages/amplify-cli/src/utils/global-prefix.ts
MorCohenAres: Sure!
What do you mean by " extraneous '..' "? I will happily make an explanation.
The goal in this line is to get the [built-in npm config](https://docs.npmjs.com/cli/v9/configuring-npm/npmrc#built-in-config-file), where you have the ini setting called `prefix`, which may have the desired value of your global `npm` path. 
According to the docs, this file can be located in the `npm` folder, which is 2 levels above `npm`'s executable (hence 2 `..`). 

If the file exists, you can find something in it's `prefix` key, which indicates the global prefix.
Then you check for the [global npm config](https://docs.npmjs.com/cli/v9/configuring-npm/npmrc#global-config-file)'s `prefix` key, and if that key doesn't exist, or the global config file itself doesn't exist, there's a fallback to your already-found `prefix`.

In a default installtion of NPM, the `npmrc` file doesn't exist. But `global-prefix` package added support for it, since it's a valid option for the user to customize according to the docs, and so did I.

MorCohenAres:
> please feel free to give me additional context into the issue that influenced your design/code choices :)

@evcodes Thank you for your review!
I highly preferred not to change the `amplify-cli` package, and fix the source of the problem which is the `global-prefix` package.

Unfortunately, I tried contacting the authors in many ways, but no response. The last commit was several years ago.
So I made this request to remove the dependency, and implement it directly in `amplify-cli`, along with fixing the bug described in [the issue](https://github.com/aws-amplify/amplify-cli/issues/10180).

The issue that affects my team is the CLI not recognizing any 3rd party plugin since `v8.0.0`update.
We use `amplify-category-video`.
There's a workaround if you interactively tell `amplify-cli` where your plugin is, but that's problematic with a CI/CD process like we have.

packages/amplify-cli/src/utils/global-prefix.ts
MorCohenAres: Do you think the right thing to do is to specifcally check for `macOS || linux`?
I'm afraid to miss some UNIX options that way.

evcodes:
will take a look and update you by EoD

sapir-muallem:
@evcodes Hey, having the same issue. any progress with this PR?

MorCohenAres:
Waiting for response as well :)

MorCohenAres:
> will take a look and update you by EoD

@evcodes Waiting for your updates.

packages/amplify-cli/src/utils/global-prefix.ts
```diff
@@ -1,12 +1,16 @@
 import * as os from 'os';
 import * as path from 'path';
-import globalNpmPrefix from 'global-prefix';
+import * as fs from 'fs';
```
jhockett, line 3:
```suggestion
import * as fs from 'fs-extra';
```

packages/amplify-cli/src/utils/global-prefix.ts
```diff
@@ -21,3 +25,62 @@ function getYarnPrefix() {
...
   return yarnPrefix;
 }
+
+// NOTE: This issue: https://github.com/jonschlinkert/global-prefix/issues/30 prevents us from using the global-prefix package
+function getGlobalNpmPrefix() {
+  // System-wide prefix
+  if (process.env.PREFIX) return process.env.PREFIX;
+
+  // $HOME variable
+  const home = os.homedir();
```
jhockett, line 35: for the various variables that represent a path, can you append "Path" to the end for readability?

```ts
  const homeDirPath = os.homedir();
```

packages/amplify-cli/src/utils/global-prefix.ts
jhockett: The way this is written is correct and in line with the rest of our code where we only special case for Windows.

jhockett (dismissed):
A few small comments, but LGTM. Thank you for the contribution @MorCohenAres!

jhockett (approved):

MorCohenAres:
@jhockett Is this ready for merge?

eddy2048 (approved):
lgtm, approved!

PR #11418, opened 2022-11-17
chore: updated pagination to ensure account selection

awsluja:
#### Description of changes
The JS AWS SDK paginates by default, so it would return only the first 20 accounts. This is not the case for our child account selection script (that one is OK because it doesn't paginate - so we'll get to use all 27+ accounts there).

Here is a working run showing that the script successfully cleans up all accounts, even past 20.

image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


edwardfoyle (approved):
apart from the fact that this is one of the rare occasions to flex your `do ... while` chops, LGTM

sobolk (approved):

PR #11371, opened 2022-11-10
chore: fix npx pkg arm64 binary creation

akshbhu:
#### Description of changes
-  fix for arm64 binary pkg fail from tagged release branch : https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/12428/workflows/c68e39e3-441d-4f14-923e-953a2a51fa62/jobs/476066
- Inspired from this : https://github.com/vercel/pkg#other-considerations 

#### Issue #, if available
https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/12428/workflows/c68e39e3-441d-4f14-923e-953a2a51fa62/jobs/476066

#### Description of how you validated changes
- fix tested on tagged-release branch : https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/12445/workflows/302d61c1-9c93-45f3-adb6-f41f64f77d4a
#### Checklist

- [x] PR description included
- [ ] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #11424, opened 2022-11-17
chore: add binary validation in all builds

sobolk:
#### Description of changes

**This PR solves the following:**
- Builds all binaries in every build including pull requests. I.e. this can be caught earlier than in e2e tests https://github.com/aws-amplify/amplify-cli/pull/11371 .
- It verifies all binaries sizes against hardcoded threshold.
- Parallelizes binary builds, so it's no longer needed to skip some to save time. And to make sure PR build time is kept same.

**PR validation**

Before:
image

After:
image

**E2E tests**

Before:
image

After:
image

**Verification step**
image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/local_publish_helpers.sh
+++ b/.circleci/local_publish_helpers.sh
@@ -107,26 +107,26 @@ function verifyPkgCli {
     echo "Human readable sizes"
     du -h out/*
     echo "Sizes in bytes"
-    du out/*
+    wc -c out/*
 
-    linux_compressed_binary_threshold_in_bytes=12
+    linux_compressed_binary_threshold_in_bytes=$((200 * 1024 * 1024))
     linux_compressed_binary_size=$(wc -c out/amplify-pkg-linux-x64.tgz | awk '{print $1}')
     if (( linux_compressed_binary_size > linux_compressed_binary_threshold_in_bytes )); then
-      echo "Linux binary size has grown over $linux_compressed_binary_threshold_in_bytes bytes"
+      echo "Linux compressed binary size has grown over $linux_compressed_binary_threshold_in_bytes bytes"
       exit 1
     fi
 
-    macos_compressed_binary_threshold_in_bytes=12
+    macos_compressed_binary_threshold_in_bytes=$((200 * 1024 * 1024))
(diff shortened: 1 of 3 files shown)


awsluja:
Can you provide a negative & negative run? or at least confirm it fails the job if file size is over 1mb (for example)

.circleci/local_publish_helpers.sh
```diff
@@ -78,23 +72,65 @@ function generatePkgCli {
 
   # Build pkg cli
   cp package.json ../build/node_modules/package.json
-  if [[ "$CIRCLE_BRANCH" == "release" ]] || [[ "$CIRCLE_BRANCH" =~ ^run-e2e-with-rc\/.* ]] || [[ "$CIRCLE_BRANCH" =~ ^release_rc\/.* ]] || [[ "$CIRCLE_BRANCH" =~ ^tagged-release ]]; then
-    # This will generate a file our arm64 binary
+
+  if [[ "$@" =~ 'arm' ]]; then
     npx pkg --no-bytecode --public-packages "*" --public -t node14-linux-arm64 ../build/node_modules -o ../out/amplify-pkg-linux-arm64
```
akshbhu, line 77: can you try removing `--public` flag and check for the arm binary

.circleci/local_publish_helpers.sh
```diff
@@ -78,23 +72,65 @@ function generatePkgCli {
...
   cd ..
 }
+
+function verifyPkgCli {
+    echo "Human readable sizes"
+    du -h out/*
+    echo "Sizes in bytes"
+    wc -c out/*
+
+    linux_compressed_binary_threshold_in_bytes=$((200 * 1024 * 1024))
```
akshbhu, line 105: default size is chosen based on previous data ?

akshbhu (approved):
lgtm

.circleci/local_publish_helpers.sh
sobolk: By looking at current state of art. I.e. here.
image

.circleci/local_publish_helpers.sh
sobolk: I can but not here. This is not the goal of this PR.

jhockett (approved):

PR #11433, opened 2022-11-17
fix: resource usage cost

awsluja:
#### Description of changes
Currently, our E2E tests execute using Linux VM - Large instances, but we only utilize about 33% CPU/RAM on average.

This updates our E2E to use medium executors for all of our E2E jobs - except those where CPU is important, such as our build/integtests/buildPkgBinaries jobs.

This will hopefully reduce our costs by about 50%, from 200k credits per run, to 100k credits per run.
image
I have also addressed a possible concern that some tests might require more resources. For those tests, we can simple add them to an exception list, so that they get to use the Large instances.

E2E Runs: https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli?branch=run-e2e%2Faluja%2Freduce-ci-costs
(all tests pass, geo-update is flaky on all other branches as well)

Here are the results of some of the most recent runs that utilize this optimization:
image
image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -36,12 +36,21 @@ executors:
     environment:
       AMPLIFY_DIR: C:/home/circleci/repo/out
       AMPLIFY_PATH: C:/home/circleci/repo/out/amplify.exe
-
-  l: &linux-e2e-executor
+# If you update this name, make sure to update it in split-e2e-tests.ts as well
+  l_large: &linux-e2e-executor-large
     docker:
       - image: public.ecr.aws/j4f5f3h7/amplify-cli-e2e-base-image-repo-public:latest
     working_directory: ~/repo
     resource_class: large
+    environment:
+      AMPLIFY_DIR: /home/circleci/repo/out
+      AMPLIFY_PATH: /home/circleci/repo/out/amplify-pkg-linux-x64
+# If you update this name, make sure to update it in split-e2e-tests.ts as well
+  l_medium: &linux-e2e-executor-medium
+    docker:
+      - image: public.ecr.aws/j4f5f3h7/amplify-cli-e2e-base-image-repo-public:latest
+    working_directory: ~/repo
+    resource_class: medium
     environment:
       AMPLIFY_DIR: /home/circleci/repo/out
       AMPLIFY_PATH: /home/circleci/repo/out/amplify-pkg-linux-x64
@@ -51,7 +60,7 @@ defaults: &defaults
   parameters:
     os:
       type: executor
-      default: l
+      default: l_medium
   executor: << parameters.os >>
 
 clean_e2e_resources: &clean_e2e_resources
@@ -89,8 +98,8 @@ jobs:
     parameters:
       os:
         type: executor
-        default: l
-    executor: l
+        default: l_large
+    executor: l_large
     steps:
       - checkout
       - run: yarn run production-build
@@ -132,7 +141,7 @@ jobs:
                   - .cache
                   - .ssh
   test:
-    <<: *linux-e2e-executor
+    <<: *linux-e2e-executor-large
     steps:
       - restore_cache:
           key: amplify-cli-repo-{{ .Branch }}-{{ .Revision }}
@@ -148,7 +157,7 @@ jobs:
       NODE_OPTIONS: --max-old-space-size=4096
 
   lint:
-    <<: *linux-e2e-executor
+    <<: *linux-e2e-executor-large
     steps:
       - restore_cache:
           key: amplify-cli-repo-{{ .Branch }}-{{ .Revision }}
@@ -157,7 +166,7 @@ jobs:
       - run: chmod +x .circleci/lint_pr.sh && ./.circleci/lint_pr.sh
 
   mock_e2e_tests:
-    <<: *linux-e2e-executor
+    <<: *linux-e2e-executor-large
     steps:
       - restore_cache:
           key: amplify-cli-repo-{{ .Branch }}-{{ .Revision }}
@@ -176,7 +185,7 @@ jobs:
           path: packages/amplify-util-mock/
 
   publish_to_local_registry:
-    <<: *linux-e2e-executor
+    <<: *linux-e2e-executor-large
     steps:
       - restore_cache:
           key: amplify-cli-repo-{{ .Branch }}-{{ .Revision }}
@@ -214,7 +223,7 @@ jobs:
             - ~/repo/.amplify-pkg-version
 
   upload_pkg_binaries:
-    <<: *linux-e2e-executor
+    <<: *linux-e2e-executor-large
     steps:
       - restore_cache:
           key: amplify-cli-repo-{{ .Branch }}-{{ .Revision }}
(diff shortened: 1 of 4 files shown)


scripts/split-e2e-tests.ts
```diff
@@ -85,6 +85,10 @@ const WINDOWS_TEST_ALLOWLIST: string[] = [
   'auth_5f_pkg',
 ];
 
+// some tests may require a larger executor, specify those here
+const USE_LINUX_LARGE_VM: string[] = [
+];
```
evcodes, line 90: are we ever adding any tests to this array?

scripts/split-e2e-tests.ts
```diff
@@ -334,13 +338,15 @@ function splitTests(
           if (typeof workflowJob === 'string') {
             return newJobName;
           } else {
+            // for the most up-to-date list of executors for e2e, see the config.base.yml file
+            let linuxVMSize = USE_LINUX_LARGE_VM.includes(newJobName) ? 'l_large' : 'l_medium';
```
evcodes, line 342: ah yes i see

scripts/split-e2e-tests.ts
```diff
@@ -334,13 +338,15 @@ function splitTests(
...
             return {
               [newJobName]: {
                 ...Object.values(workflowJob)[0],
                 requires: [...(requires ? [requires] : workflowJob[jobName].requires || [])],
                 matrix: {
                   parameters: {
-                    os: WINDOWS_TEST_ALLOWLIST.includes(newJobName) ? ['l', 'w'] : ['l'],
+                    os: WINDOWS_TEST_ALLOWLIST.includes(newJobName) ? [linuxVMSize, 'w'] : [linuxVMSize],
```
evcodes, line 349: I would be a fan of changing this 'w' to something more descriptive 🙉

evcodes (commented):
small change

packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
```diff
@@ -538,7 +538,8 @@ const deleteCfnStack = async (account: AWSAccountInfo, accountIndex: number, sta
   try {
     const cfnClient = new aws.CloudFormation(getAWSConfig(account, region));
     await cfnClient.deleteStack({ StackName: stackName, RetainResources: resourceToRetain }).promise();
-    await cfnClient.waitFor('stackDeleteComplete', { StackName: stackName }).promise();
+    // we'll only wait up to 10 minutes before moving on
+    await cfnClient.waitFor('stackDeleteComplete', { StackName: stackName, $waiter: { 'maxAttempts': 20 }}).promise();
```
sobolk, line 542: nice !

.circleci/config.base.yml
```diff
@@ -36,22 +36,31 @@ executors:
     environment:
       AMPLIFY_DIR: C:/home/circleci/repo/out
       AMPLIFY_PATH: C:/home/circleci/repo/out/amplify.exe
-
-  l: &linux-e2e-executor
+# If you update this name, make sure to update it in split-e2e-tests.ts as well
```
sobolk, line 39: does push hook catch it ?

.circleci/config.base.yml
```diff
@@ -148,7 +157,7 @@ jobs:
       NODE_OPTIONS: --max-old-space-size=4096
 
   lint:
-    <<: *linux-e2e-executor
+    <<: *linux-e2e-executor-large
```
sobolk, line 160: does lint need large?
I think that we could try medium for:
- lint
- publish_to_local_registry
- upload_pkg_binaries
- verify_pkg_binaries

They're not very CPU intensive I think.

.circleci/config.base.yml
```diff
@@ -749,7 +780,7 @@ jobs:
             version=$(cat .amplify-pkg-version)
             yarn ts-node scripts/github-release.ts $version $commit
   cleanup_resources:
-    <<: *linux-e2e-executor
+    <<: *linux-e2e-executor-large
```
sobolk, line 783: clenup_resources and cleanup_resources_after_e2e_runs are likely IO bound. could you try scaling down?

scripts/split-e2e-tests.ts
```diff
@@ -85,6 +85,22 @@ const WINDOWS_TEST_ALLOWLIST: string[] = [
   'auth_5f_pkg',
 ];
 
+// some tests may require a larger executor, specify those here
+const USE_LINUX_LARGE_VM: string[] = [
```
sobolk, line 89: verbs are usually reserved for functions/methods.

```suggestion
const JOBS_RUNNING_ON_LINUX_LARGE_VM: string[] = [
```

.circleci/config.base.yml
awsluja: it never has; that script only checks if this file is valid/compiles

.circleci/config.base.yml
awsluja: we are only billed by the minute used; those jobs all take less than 5 minutes combined (costs penies) & run in series, so I would rather keep those where they are if the 4 cores help reduce linear time

.circleci/config.base.yml
awsluja: similar tradeoff here; we can downscale but the savings are minimal; prefer to speed this job up during spinup/teardown

awsluja: for reference, 1 hour of large executor cost about $.72 vs $.36 for small executor

packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
awsluja: good question; I considered it but did 10 minutes to be safe, this keeps the job as a whole within 10-15 minutes each time

.circleci/config.base.yml
sobolk: I wouldn't trade time for pennies either.
The suggestion is to try running these jobs (and from other comment) on smaller size and see if size makes any difference for them. If there isn't then downsize. if there is then keep it large.

scripts/split-e2e-tests.ts
awsluja: just updated it!

scripts/split-e2e-tests.ts
awsluja: done!

evcodes (approved):

scripts/split-e2e-tests.ts
```diff
@@ -474,29 +382,29 @@ function main(): void {
   const splitPkgTests = splitTests(
     config,
     'amplify_e2e_tests_pkg',
-    'build_test_deploy',
+    'build_test_deploy_v3',
```
jhockett, line 385: Should this be changed back before merging to dev?

jhockett (commented):
LGTM, just 1 question

scripts/split-e2e-tests.ts
awsluja: I did this intentionally, since our current build_test_deploy job is cluttered & freezes up due to the number of old job names that are associated with it. Github & other branch rules will automatically apply (per Ed's comments, and I've verified that checks are automatically picked up here:
image

jhockett (approved):

PR #11460, opened 2022-11-23
fix(api): add null placeholder for nested stack during api rebuild

AaronZyLee:
#### Description of changes
Fixed the failure of `amplify api rebuild` for the following scenarios:

- Multiple models (appsync function dependencies between table stacks)
- Relational models (connection stack)
- Models with `@searchable` defined (searchable stack)

**Note: for `@searchable` rebuild the open search domain will be kept but the other resources (IAM roles, streaming lambda functions) will be removed and regenerated.**

The root cause is that some parameters of nested stack refers to the other nested stacks(including tables, connection and searchable), whereas the `Outputs` of them are dropped during the first stage of deployment. The change is to use null placeholder values to replace the references mentioned above in the api root template.

### Example
Given the following schema
```GraqhQL
input AMPLIFY { globalAuthRule: AuthRule = { allow: public } } # FOR TESTING ONLY!

type Todo @model @searchable{
  id: ID!
  name: String!
  description: String
  tasks: [Task] @hasMany
}
type Task @model {
  id: ID!
  todo: Todo @belongsTo
}
```
During the rebuild process, the following stack templates will be generated. The DynamoDB table stack will remain the same behavior as previous.
#### `SearchableStack.json`
```json
{
  "Description": "An auto-generated nested stack for searchable.",
  "AWSTemplateFormatVersion": "2010-09-09",
  "Mappings": {
        ...
  },
  "Conditions": {
      ...
  },
  "Resources": {
    //keep domain only and remove the rest
    "OpenSearchDomain": {
      "Type": "AWS::Elasticsearch::Domain",
      "Properties": {
            ...
       }
    }
  },
  "Outputs": {
    //empty output
  },
  "Parameters": {
    //remain unchanged
    ...
  }
}
```
#### `ConnectionStack.json`
```json
{
  "Resources": {
   //remove all resources but leave one placeholder
    "PlaceholderNullResource": {
        "Type": "AWS::CloudFormation::WaitConditionHandle"
    }
  },
  "Outputs": {
    //remove all outputs
  }
  "Parameters": {
    //remain unchanged
    ...
  }
}
```
#### `cloudformation-template.json` (API root template)
Only `Paramters` referencing the `Outputs` of nested stack will be changed to the placeholder value. The previous behavior will cause null reference error (as outputs are dropped during rebuild)
```json
{
  "Resources": {
    "Task": {
      "Properties": {
        "Parameters": {
          "refertotodonestedstackoutput":
            "Fn::GetAtt": [ "Todo", "Outputs.todoParamOutput"] // this will be replaced by following line
            "TemporaryPlaceholderValue"
        }
      }
    }
  }
}
```

#### Issue #, if available
fix https://github.com/aws-amplify/amplify-category-api/issues/964
fix https://github.com/aws-amplify/amplify-category-api/issues/56

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-core/src/init/deleteProject.ts
+++ b/packages/amplify-e2e-core/src/init/deleteProject.ts
@@ -6,7 +6,12 @@ import { getBackendAmplifyMeta } from '../utils';
  * Runs `amplify delete`
  */
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
-export const deleteProject = async (cwd: string, profileConfig?: any, usingLatestCodebase = false): Promise<void> => {
+export const deleteProject = async (
+  cwd: string,
+  profileConfig?: any,
+  usingLatestCodebase = false,
+  noOutputTimeout: number = 1000 * 60 * 20,
+): Promise<void> => {
   // Read the meta from backend otherwise it could fail on non-pushed, just initialized projects
   try {
     const { StackName: stackName, Region: region } = getBackendAmplifyMeta(cwd).providers.awscloudformation;
@@ -15,7 +20,6 @@ export const deleteProject = async (cwd: string, profileConfig?: any, usingLates
       (stack) => stack.StackStatus.endsWith('_COMPLETE') || stack.StackStatus.endsWith('_FAILED'),
     );
 
-    const noOutputTimeout = 1000 * 60 * 20; // 20 minutes;
     await spawn(getCLIPath(usingLatestCodebase), ['delete'], { cwd, stripColors: true, noOutputTimeout })
       .wait('Are you sure you want to continue?')
       .sendYes()
@@ -25,18 +29,3 @@ export const deleteProject = async (cwd: string, profileConfig?: any, usingLates
     console.log('Error on deleting project at:', cwd);
   }
 };
-
-export const amplifyDeleteWithLongerTimeout = (cwd: string, usingLatestCodebase = false): Promise<void> => new Promise((resolve, reject) => {
-  const noOutputTimeout = 1000 * 60 * 30; // 30 minutes;
-  spawn(getCLIPath(usingLatestCodebase), ['delete'], { cwd, stripColors: true, noOutputTimeout })
-    .wait('Are you sure you want to continue?')
-    .sendConfirmYes()
-    .wait('Project deleted locally.')
-    .run((err: Error) => {
-      if (!err) {
-        resolve();
-      } else {
-        reject(err);
-      }
-    });
-});
--- a/packages/amplify-e2e-core/src/init/deleteProject.ts
+++ b/packages/amplify-e2e-core/src/init/deleteProject.ts
@@ -1,6 +1,7 @@
 /* eslint-disable import/no-cycle */
 import { nspawn as spawn, retry, getCLIPath, describeCloudFormationStack } from '..';
 import { getBackendAmplifyMeta } from '../utils';
+import { $TSAny } from 'amplify-cli-core';
 
 /**
  * Runs `amplify delete`
@@ -8,7 +9,7 @@ import { getBackendAmplifyMeta } from '../utils';
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export const deleteProject = async (
   cwd: string,
-  profileConfig?: any,
+  profileConfig?: $TSAny,
   usingLatestCodebase = false,
   noOutputTimeout: number = 1000 * 60 * 20,
 ): Promise<void> => {
--- a/packages/amplify-e2e-core/src/init/deleteProject.ts
+++ b/packages/amplify-e2e-core/src/init/deleteProject.ts
@@ -25,3 +25,18 @@ export const deleteProject = async (cwd: string, profileConfig?: any, usingLates
     console.log('Error on deleting project at:', cwd);
   }
 };
+
+export const amplifyDeleteWithLongerTimeout = (cwd: string, usingLatestCodebase = false): Promise<void> => new Promise((resolve, reject) => {
+  const noOutputTimeout = 1000 * 60 * 30; // 30 minutes;
+  spawn(getCLIPath(usingLatestCodebase), ['delete'], { cwd, stripColors: true, noOutputTimeout })
+    .wait('Are you sure you want to continue?')
+    .sendConfirmYes()
+    .wait('Project deleted locally.')
+    .run((err: Error) => {
+      if (!err) {
+        resolve();
+      } else {
+        reject(err);
+      }
+    });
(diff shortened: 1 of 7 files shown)


packages/amplify-e2e-tests/src/__tests__/api_6a.test.ts
```diff
@@ -9,36 +9,61 @@ import {
   scanTable,
   rebuildApi,
   getProjectMeta,
+  updateApiSchema,
 } from '@aws-amplify/amplify-e2e-core';
 
 const projName = 'apitest';
+
+export const testTableBeforeRebuildApi = async (apiId: string, region: string, modelName: string) => {
```
phani-srikar, line 17: nit: I see these methods `testTableBeforeRebuildApi` and `testTableAfterRebuildApi` could be used in both e2e tests, can you move these to a shared test utils?

packages/amplify-e2e-tests/src/__tests__/api_6c.test.ts
```diff
@@ -0,0 +1,48 @@
...
+    await initJSProjectWithProfile(projRoot, { name: projName });
+    await addApiWithoutSchema(projRoot, { transformerVersion: 2 });
+    await updateApiSchema(projRoot, projName, 'searchable_model_v2.graphql');
+    await amplifyPush(projRoot);
+    const projMeta = getProjectMeta(projRoot);
+    const apiId = projMeta?.api?.[projName]?.output?.GraphQLAPIIdOutput;
+    const region = projMeta?.providers?.awscloudformation?.Region;
+    expect(apiId).toBeDefined();
+    expect(region).toBeDefined();
+    const tableName = `Todo-${apiId}-integtest`;
```
phani-srikar, line 40: re-use `testTableBeforeRebuildApi` and `testTableAfterRebuildApi`

packages/amplify-provider-awscloudformation/src/graphql-resource-manager/amplify-graphql-resource-manager.ts
```diff
@@ -304,16 +317,67 @@ export class GraphQLResourceManager {
...
+        const paramObj = stackParamsMap[stackParamKey];
+        const paramObjKeys = Object.keys(paramObj);
+        if ( paramObjKeys.length === 1 && paramObjKeys[0] === 'Fn::GetAtt') {
+          const paramObjValue  = paramObj[paramObjKeys[0]]
+          if ( Array.isArray(paramObjValue)
+            && paramObjValue.length === 2
+            && recreatedNestedStackNames.includes(paramObjValue[0])
+            && paramObjValue[1].startsWith('Outputs.') ) 
+          {
+            stackParamsMap[stackParamKey] = 'TemporaryPlaceholderValue';
```
phani-srikar, line 374: I don't fully understand the fix here. Can you post a screenshot of before and after state of a nested child stack parameters? How are we ensuring that these defaults will not show up in the final templates deployed via `push`?
Instead of adding a default can we just remove these parameters?

phani-srikar (changes requested):

packages/amplify-provider-awscloudformation/src/graphql-resource-manager/amplify-graphql-resource-manager.ts
AaronZyLee: The parameters cannot be removed but faking a placeholder value. These parameter values are defined in api root stack (`cloudformation-template.json` => `Resources.[nestedStack].Properties.Parameters`). I update a example in the PR description.

phani-srikar (approved):

evcodes:
I am going to repro and test these changes. Getting a review by EOD

packages/amplify-e2e-core/src/init/deleteProject.ts
```diff
@@ -23,3 +23,18 @@ export const deleteProject = async (cwd: string, profileConfig?: any, usingLates
     .wait('Project deleted locally.')
     .runAsync();
 };
+
+export const amplifyDeleteWithLongerTimeout = (cwd: string, usingLatestCodebase = false): Promise<void> => new Promise((resolve, reject) => {
```
sundersc, line 27: Can we override the existing method instead of creating a new one?

packages/amplify-e2e-tests/src/__tests__/api_6a.test.ts
```diff
@@ -5,40 +5,51 @@ import {
...
 afterEach(async () => {
   await deleteProject(projRoot);
   deleteProjectDir(projRoot);
 });
 
 describe('amplify rebuild api', () => {
-  it('recreates all model tables', async () => {
+  it('recreates single table', async () => {
+    await initJSProjectWithProfile(projRoot, { name: projName });
+    await addApiWithoutSchema(projRoot, { transformerVersion: 2 });
```
sundersc, line 28: Can these 2 lines stay inside 'beforeEach' block?

packages/amplify-provider-awscloudformation/src/graphql-resource-manager/amplify-graphql-resource-manager.ts
sundersc: It looks like we are looking for a very specific CFN template string. Could this break if there is a change in the way we pass this parameter value in the future?

danielleadams:
The CI passed: https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli?branch=run-e2e%2Flizeyu%2Ffix-rebuild-api, but the GH checks were not updated.

packages/amplify-provider-awscloudformation/src/graphql-resource-manager/amplify-graphql-resource-manager.ts
AaronZyLee: The phrase it looks for is the `Fn:GetAtt:[Outputs.*, ]` which is an intrinsic function and reserved word in the Cloudformation. The words should be unlikely to change unless the breaking change occurs in CFN. 
Currently it is constructed by the CDK sync operation. If it changes its form, this will be a breaking change for CDK users and the CDK should convey this message to us.

packages/amplify-e2e-core/src/init/deleteProject.ts
evcodes: +1 on this

evcodes (changes requested):
Please review the changes that have been suggested

phani-srikar (approved):

edwardfoyle (approved):

lazpavel (approved):

PR #11465, opened 2022-11-23
chore: optimize bottlenecks in CI 

awsluja:
#### Description of changes
This PR addresses some bottlenecks that we have in our E2E suite, and reduces the time spent scanning artifacts by over 95%.

1. Scanning E2E artifacts happens after ALL tests, and takes 8 minutes to run. Currently, it scans the entire directory, even though only specified paths are storing those artifacts. We can scan those artifacts
- This PR enforces that store_artifacts/store_test_results paths are explicitly defined, so we can scan only those directories.
- It also provides developer friendly output if a violation is found, and provides resolution steps to address violations.
- This will also fail the PUSH if this validation fails

Old runtime: 8 minutes
New Runtime : 5 seconds
image

Windows:
image

Validation example:
image

2. Searchable / open search tests take the longest to run (~54minutes), and they are currently the slowest test, by a lot
- delegating the cleanup step to the cleanup script fixes this (I verified that the open search stacks 'delete' is being triggered in our cleanup script; so there is no need to wait 15 minutes for teardown in this test
Screen Shot 2022-11-22 at 3 52 13 PM
image

CI Run: https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/12894/workflows/708e8bff-05a9-4698-a1ae-8bc930f850cc

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.circleci/config.base.yml
```diff
@@ -194,7 +194,7 @@ jobs:
           environment:
             JEST_JUNIT_OUTPUT: 'reports/junit/js-test-results.xml'
       - store_test_results:
-          path: packages/amplify-util-mock/
+          path: ~/repo/packages/amplify-util-mock/
```
sobolk, line 197: what does this (and similar) change do ?

.circleci/scan_artifacts.ts
```diff
@@ -1,20 +1,32 @@
...
+  folder: string,
+  excludeFolder = '{node_modules,.cache,.git,\.cache,\.git}',
 ): boolean => {
+  console.log("Scanning folder:", folder);
   const patternParam = patterns.reduce<string[]>((acc, v) => [...acc, '-e', v], []);
+
+  let actualFolder = folder;
+  if(folder.startsWith("~/") || folder.startsWith("~\\")){
+    actualFolder = folder.replace("~", REPO_FOLDER);
+  }
```
sobolk, line 16: ah, now I understand.
I'd replace `~` with something more self explanatory like `[repo_folder]` or some other more verbose token/placeholder.
`~` is reserved for user's home in linux shells.

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-datastore.test.ts
```diff
@@ -26,8 +26,13 @@ describe('transformer model searchable migration test', () => {
   });
 
   afterEach(async () => {
-    await deleteProject(projRoot);
-    deleteProjectDir(projRoot);
+    if (process.env.CIRCLECI) {
+      console.log('Skipping cloud deletion since we are in CI, and cleanup script will delete this stack in cleanup step.');
+      deleteProjectDir(projRoot);
+    } else {
```
sobolk, line 32: would kicking off fire and forget deletion make sense instead?

sobolk (commented):
LGTM.

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-datastore.test.ts
awsluja: i tried this initially, but jest will wait for any open handles to finish before finishing the test;
ie, 
'amplify delete' runs inside of a terminal spawn, which makes it part of the tests (indirectly, because our tests are configured so that these terminal spawns must succeed/exit successfully, otherwise test fails)

I also tried to manually exit/throw & try catch the error, but the spawn that we use is also responsible for reporting the success of the test; so if any error/process exits the spawn early, it will mark the job/test as failed

awsluja: here is the code, this exitHandler is registered on the terminal that executes each amplify command; so if it exists early from any external process it will still throw a failure
image

awsluja: nexpect.ts
image

.circleci/config.base.yml
awsluja: ~ is the most reliable way to resolve paths in the CircleCI executor, and is consistent between linux & windows (they use some kind of interpretation to resolve ~ in both linux and windows)

The scan_artifacts.ts file exists in ~/repo/.circleci/scan_artifacts.ts, and we can easily determine the absolute value of ~ by generating a path that moves up 2 directories:
const REPO_FOLDER = path.normalize(path.join(__dirname, '..', '..')); // same as ~

I am also using normalize to make sure that slashes are correct, based on the OS:
const normalizedFolder = path.normalize(folder);

Using ~ is mainly for simplicity; as I don't want people to be concerned with absolute paths, since all they need to know is that the working directory is ~, and everything exists inside of repo 
repo/src/packages/amplify-e2e-tests/.. etc

.circleci/scan_artifacts.ts
awsluja: in CircleCI it is also the working directory (both linux & windows)

.circleci/scan_artifacts.ts
```diff
@@ -1,20 +1,32 @@
...
-  folder = DEFAULT_SEARCH_FOLDER,
-  excludeFolder = '{*/node_modules,.git}',
+  folder: string,
+  excludeFolder = '{node_modules,.cache,.git,\.cache,\.git}',
 ): boolean => {
+  console.log("Scanning folder:", folder);
   const patternParam = patterns.reduce<string[]>((acc, v) => [...acc, '-e', v], []);
+
+  let actualFolder = folder;
+  if(folder.startsWith("~/") || folder.startsWith("~\\")){
```
lazpavel, line 14: nit (spacing): `if (folder.startsWith("~/") || folder.startsWith("~\\")) {`

.circleci/scan_artifacts.ts
awsluja: we cannot include special characters like [repo_folder] in the config.yaml paths

awsluja: paths in the config.yaml are 1 to 1 with the allow-list ; and circleci can't interpret [repo_folder], so we'd also have to implement an override script to replace [repo_folder] in config.yaml with ~ or similar

lazpavel (dismissed):
LGTM (small spacing nits that can be be fixed with auto-formatting)

scripts/split-e2e-tests.ts
```diff
@@ -380,6 +381,37 @@ function verifyConfig() {
...
   const config = loadConfig();
+  
+  // make sure that only valid paths are used to store artifacts/results
+  const storagePathsUsedInConfig = new Set();
+  const invalidPaths = new Set();
+  for(let key of Object.keys(config.jobs)) {
+    const job = config.jobs[key];
+    const steps = job.steps;
+    
+    for(let i = 0; i < steps.length; i ++){
```
lazpavel, line 392: can we use `for (const step of steps)`

awsluja: steps is a custom type, and doesn't implement the iterator interface unfortunately

.circleci/scan_artifacts.ts
sobolk: Okay, I think I understand the constraints now. Circle CI must be then doing similar trick internally on windows (~ is meaningless in windows shell I recall).
Can you change naming or add comments ? or even extract this logic to `normalizeFolderPath` function and mention those things in the comments?
If REPO_FOLDER is absolute path we can name it accordingly ?

packages/amplify-e2e-tests/src/__tests__/graphql-v2/searchable-datastore.test.ts
sobolk: I see, I was hoping that this is just some CFN delete call that could be fired and forgotten.

.circleci/scan_artifacts.ts
awsluja: sure; that make sense, ill update it and add comments

sobolk (approved):

lazpavel (approved):

PR #11484, opened 2022-11-28
Convert auth questions to ts

sdstolworthy:
#### Description of changes

Converts `packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts` and `packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts` to Typescript. Code depended on by these files is also updated.

As a general rule, disabling linting rules was preferred over making overly broad changes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -129,19 +136,19 @@ const updateConfigOnEnvInit = async (context, category, service) => {
 
   // legacy headless mode (only supports init)
   if (isInHeadlessMode(context)) {
-    const envParams = {};
-    let mergedValues;
+    const envParams: {[key: string]: any} = {};
+    let mergedValues: {[key: string]: any} | undefined;
```
jhockett, line 140: nit: use `$TSObject` from `amplify-cli-core`

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -182,32 +190,32 @@ const updateConfigOnEnvInit = async (context, category, service) => {
...
   const existingAuth = context.migrationInfo.amplifyMeta.auth || {};
-  if (!Object.keys(existingAuth).length > 0) {
+  if (!(Object.keys(existingAuth).length > 0)) {
     return;
   }
   const { provider, cfnFilename, defaultValuesFilename } = getSupportedServices().Cognito;
   const defaultValuesSrc = `${__dirname}/assets/${defaultValuesFilename}`;
 
-  const { roles } = require(defaultValuesSrc);
+  const { roles } = await import(defaultValuesSrc);
```
jhockett, line 203: The imported files are older JS. Please make sure this change doesn't break the `migrate` flow.

jhockett (approved):
LGTM! Left a couple of comments

packages/amplify-category-auth/API.md
```diff
@@ -110,7 +110,7 @@ export function initEnv(context: any): Promise<void>;
 export function isSMSWorkflowEnabled(context: any, resourceName: any): Promise<boolean>;
 
 // @public (undocumented)
-export function migrate(context: any): Promise<void>;
```
sobolk, line 113: why did we go from `Promise<void>` to `Promise<any>` ?

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -1,46 +1,51 @@
...
+  serviceWalkthroughFilename: any,
+  serviceMetadata: any): Promise<any> => {
   const serviceWalkthroughSrc = `${__dirname}/service-walkthroughs/${serviceWalkthroughFilename}`;
-  const { serviceWalkthrough } = require(serviceWalkthroughSrc);
+  const { serviceWalkthrough } = await import(serviceWalkthroughSrc);
   return serviceWalkthrough(context, defaultValuesFilename, stringMapsFilename, serviceMetadata);
 };
 
-const addResource = async (context, service, skipNextSteps = false) => {
+export const addResource = async (context: any, service: any): Promise<any> => {
```
sobolk, line 28: Do we have types for context and service? I recall we have `$TSContext` for example.
Is return type correct? should it be void?
Please revisit other places like this if this is valid concern.

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -218,12 +226,13 @@ const getHeadlessParams = context => {
   }
 };
 
-const getOAuthProviderKeys = (currentEnvSpecificValues, resourceParams) => {
-  const oAuthProviders = JSON.parse(resourceParams.hostedUIProviderMeta).map(h => h.ProviderName);
+/* eslint-disable no-param-reassign */
```
sobolk, line 229: I think we should disable this rule.

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -1,46 +1,51 @@
-// TODO enable eslint after converting to TS
-/* eslint-disable */
-const inquirer = require('inquirer');
-const _ = require('lodash');
-const { stateManager, open } = require('amplify-cli-core');
-const { ensureEnvParamManager } = require('@aws-amplify/amplify-environment-parameters');
-const { getAuthResourceName } = require('../../utils/getAuthResourceName');
-const { copyCfnTemplate, saveResourceParameters } = require('./utils/synthesize-resources');
-const {
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
sobolk, line 1: I'd disable no-explicit-any globally. We have `$TSAny` but that's just a cheat.

packages/amplify-cli-core/src/serviceSelection.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jsdoc/require-jsdoc */
```
sobolk, line 1: I'd disable this rule as well.

sobolk (commented):
The change looks good to me.

Couple of suggestions.
- we should consider disabling couple of rules globally if we end up disabling them in many places anyway. (e.g. explicit any, comments, no param re-assign).
- in places where we introduced `any` have we looked around for existing types that could work?

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
sdstolworthy: Changed

packages/amplify-category-auth/API.md
sdstolworthy: Changed this back to Promise<void>

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
sdstolworthy: Imported $TSContext here and used it for all `context` parameters. On further investigation, it looks like the correct return type here is `Promise<string>`. I updated the method signature to reflect that.

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
sdstolworthy: Broadly, I don't know that I agree that no-param-reassign should be disabled. I disabled it here just because the code modifies the parameter in this block, and the objective is to make minimal changes for the migration from JS to TS, and refactoring this to not reassign the parameter feels out of scope.

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
sdstolworthy: I agree it should be disabled globally, but I would prefer to do that in a separate PR. If it's ok, I would like to keep the explicit local disable in this PR, and then disable the rule in a follow up PR.

packages/amplify-cli-core/src/serviceSelection.ts
sdstolworthy: Agree, but same thoughts as above, I would like to disable in a follow up PR

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
sobolk: I have mixed feelings about this particular rule, because it doesn't always lead to more readable code.

I think the rule is useful in code blocks that span across a lot of lines (i.e. a screen or more), but on the other hand one should not write code like that.
In smaller blocks defining new var/const just for the sake of satisfying this rule often leads to weird naming, especially if it's series of transforms.

Besides, regardless of how do we like this rule we seem to use it quite often, see https://github.com/aws-amplify/amplify-cli/search?q=no-param-reassign .

Let's talk about this when we trim rulset.

sobolk (dismissed):

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
sdstolworthy: Confirmed that this is working as expected with a migration from v6 -> 10.5.1

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -517,31 +527,31 @@ function structureOAuthMetadata(coreAnswers, context, defaults, amplify) {
...
 
   return coreAnswers;
-}
+};
 
 /*
   Deserialize oAuthData for CLI update flow
 */
-function parseOAuthMetaData(previousAnswers) {
+const parseOAuthMetaData = (previousAnswers: any): any => {
```
evcodes, line 535: is this return type void?

sdstolworthy: indeed it is. Changing.

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -237,18 +248,19 @@ const getOAuthProviderKeys = (currentEnvSpecificValues, resourceParams) => {
...
     const previousResult = JSON.parse(currentEnvSpecificValues.hostedUIProviderCreds);
     if (resourceParams.hostedUIProviderMeta) {
-      const currentProviders = JSON.parse(resourceParams.hostedUIProviderMeta).map(h => h.ProviderName);
-      currentProviders.forEach(c => {
-        const previousProvider = previousResult.find(p => p.ProviderName === c);
-        const resultProvider = inputResult.find(r => r.ProviderName === c);
+      const currentProviders = JSON.parse(resourceParams.hostedUIProviderMeta).map((h: any) => h.ProviderName);
+      currentProviders.forEach((currentProvider: any) => {
+        const previousProvider = previousResult.find((p: any) => p.ProviderName === currentProvider);
+        const resultProvider = inputResult.find((r: any) => r.ProviderName === currentProvider);
```
evcodes, line 263: nit: more descriptive variable names
currentProvider -> provider
p -> previous
r-> result

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -315,15 +329,16 @@ const getRequiredParamsForHeadlessInit = (projectType, previousValues) => {
...
     }
+    // eslint-disable-next-line spellcheck/spell-checker
     if (previousValues.authProviders.includes('appleid.apple.com')) {
       requiredParams.push('appleAppId');
     }
   }
 
   if (previousValues.hostedUIProviderMeta) {
-    const oAuthProviders = JSON.parse(previousValues.hostedUIProviderMeta).map(h => h.ProviderName);
+    const oAuthProviders = JSON.parse(previousValues.hostedUIProviderMeta).map((h: any) => h.ProviderName);
```
evcodes, line 339: h -> hostedProvider

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -412,48 +427,40 @@ const getCognitoOutput = amplifyMeta => {
...
-const openUserPoolConsole = async (context, region, userPoolId) => {
+const openUserPoolConsole = async (context: $TSContext, region: any, userPoolId: any): Promise<any> => {
   const userPoolConsoleUrl = `https://${region}.console.aws.amazon.com/cognito/users/?region=${region}#/pool/${userPoolId}/details`;
   await open(userPoolConsoleUrl, { wait: false });
   context.print.info('User Pool console:');
   context.print.success(userPoolConsoleUrl);
 };
 
-const openIdentityPoolConsole = async (context, region, identityPoolId) => {
+const openIdentityPoolConsole = async (context: $TSContext, region: any, identityPoolId: any): Promise<any> => {
```
evcodes, line 446:
```suggestion
const openIdentityPoolConsole = async (context: $TSContext, region: any, identityPoolId: any): Promise<void> => {
```

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -412,48 +427,40 @@ const getCognitoOutput = amplifyMeta => {
...
-  const providerPlugin = require(context.amplify.getProviderPlugins(context).awscloudformation);
+  const providerPlugin = await import(context.amplify.getProviderPlugins(context).awscloudformation);
   const baseUrl = providerPlugin.adminBackendMap[region].amplifyAdminUrl;
   const adminUrl = `${baseUrl}/admin/${appId}/${envName}/auth`;
   await open(adminUrl, { wait: false });
   context.print.success(adminUrl);
 };
 
-const openUserPoolConsole = async (context, region, userPoolId) => {
+const openUserPoolConsole = async (context: $TSContext, region: any, userPoolId: any): Promise<any> => {
```
evcodes, line 439:
```suggestion
const openUserPoolConsole = async (context: $TSContext, region: any, userPoolId: any): Promise<void> => {
```

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -31,11 +40,13 @@ async function serviceWalkthrough(context, defaultValuesFilename, stringMapsFile
     const q = await parseInputs(questionObj, amplify, defaultValuesFilename, stringMapsFilename, coreAnswers, context);
 
     // ASK QUESTION
-    const answer = await inquirer.prompt(q);
+    const answer: any = await inquirer.prompt(q);
```
evcodes, line 43:
```suggestion
    const answer: any = await inquirer.prompt(question);
```

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -179,17 +190,15 @@ async function serviceWalkthrough(context, defaultValuesFilename, stringMapsFile
   }
 
   const isPullOrEnvCommand = context.input.command === 'pull' || context.input.command === 'env';
-  if (coreAnswers.authSelections !== 'identityPoolOnly' && context.input.command != 'init' && !isPullOrEnvCommand) {
+  if (coreAnswers.authSelections !== 'identityPoolOnly' && context.input.command !== 'init' && !isPullOrEnvCommand) {
```
evcodes, line 193: nice catch

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -370,15 +380,16 @@ async function updateAdminQuery(context, userPoolGroupList) {
...
 /*
   Create key/value pairs of third party auth providers,
   where key = name accepted by updateIdentityPool API call and value = id entered by user
 */
-function identityPoolProviders(coreAnswers, projectType) {
+/* eslint-disable no-param-reassign */
+export const identityPoolProviders = (coreAnswers: any, projectType: any): any => {
   coreAnswers.selectedParties = {};
-  authProviders.forEach(e => {
+  authProviders.forEach((e: any) => {
```
evcodes, line 392: e-> provider

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -370,15 +380,16 @@ async function updateAdminQuery(context, userPoolGroupList) {
...
     // don't send google value in cf if native project, since we need to make an openid provider
     if (projectType === 'javascript' || e.answerHashKey !== 'googleClientId') {
       if (coreAnswers[e.answerHashKey]) {
```
evcodes, line 395:
```suggestion
      if (coreAnswers[provider.answerHashKey]) {
```

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -405,31 +416,31 @@ function identityPoolProviders(coreAnswers, projectType) {
...
   const answers = Object.assign(prevAnswers || {}, coreAnswers);
   const attributesForMapping = answers.requiredAttributes
     ? JSON.parse(JSON.stringify(answers.requiredAttributes)).concat('username')
     : ['email', 'username'];
-  const res = {};
+  const res: {[key: string]: any} = {};
   if (answers.hostedUI) {
     res.hostedUIProviderMeta = JSON.stringify(
-      oAuthProviders.map(el => {
+      oAuthProviders.map((el: any) => {
```
evcodes, line 437: what is el here?

evcodes: seems like it's a providerName from the code below

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -439,18 +450,18 @@ function userPoolProviders(oAuthProviders, coreAnswers, prevAnswers) {
             scopes.unshift('openid');
           }
           if (attributeKey && attributeKey[`${lowerCaseEl}`] && attributeKey[`${lowerCaseEl}`].attr) {
-            maps[a] = attributeKey[`${lowerCaseEl}`].attr;
+            maps[attribute] = attributeKey[`${lowerCaseEl}`].attr;
```
evcodes, line 453: 👍🏼

evcodes (dismissed):
I made some suggestions that are not blocking for the migration but should be considered for a future PR.

jhockett (approved):

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
sdstolworthy: I went through `auth-questions.ts` and added more descriptive return types where possible.

jhockett (approved):

jhockett (approved):

sobolk (dismissed):

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -158,19 +167,20 @@ const updateConfigOnEnvInit = async (context, category, service) => {
   const isPullingOrEnv = context.input.command === 'pull'
     || (context.input.command === 'env' && context.input.subCommands && !context.input.subCommands.includes('add'));
   // don't ask for env_specific params when checking out env or pulling
-  srvcMetaData.inputs = srvcMetaData.inputs.filter(
-    input => ENV_SPECIFIC_PARAMS.includes(input.key) && !Object.keys(currentEnvSpecificValues).includes(input.key) && !isPullingOrEnv,
+  serviceMetadata.inputs = serviceMetadata.inputs.filter(
+    (input: any) => ENV_SPECIFIC_PARAMS.includes(input.key)
+      && !Object.keys(currentEnvSpecificValues).includes(input.key) && !isPullingOrEnv,
```
evcodes, line 172: is this a new line here?

packages/amplify-category-auth/src/provider-utils/awscloudformation/index.ts
```diff
@@ -412,48 +427,40 @@ const getCognitoOutput = amplifyMeta => {
   return cognitoOutput;
 };
 
-const openAdminUI = async (context, appId, region) => {
+const openAdminUI = async (context: $TSContext, appId: any, region: any): Promise<any> => {
```
evcodes, line 430: I think appId and region are both strings. We should have some type for both of these to have some validation for each of these params.

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -517,31 +528,31 @@ function structureOAuthMetadata(coreAnswers, context, defaults, amplify) {
...
+/**
+ * Deserialize oAuthCredentials for CLI update flow
+ */
+export const parseOAuthCreds = (providers: any, metadata: any, envCreds: any): Record<string, unknown> => {
+  const providerKeys: Record<string, unknown> = {};
   try {
     const parsedMetaData = JSON.parse(metadata);
     const parsedCreds = JSON.parse(envCreds);
-    providers.forEach(el => {
+    providers.forEach((el: any) => {
```
evcodes, line 551: what is el here?

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -517,31 +528,31 @@ function structureOAuthMetadata(coreAnswers, context, defaults, amplify) {
...
       const lowerCaseEl = el.toLowerCase();
       try {
-        const provider = parsedMetaData.find(i => i.ProviderName === el);
-        const creds = parsedCreds.find(i => i.ProviderName === el);
+        const provider = parsedMetaData.find((i: any) => i.ProviderName === el);
```
evcodes, line 554: do we know of a better variable name for i?

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
```diff
@@ -553,19 +564,20 @@ function parseOAuthCreds(providers, metadata, envCreds) {
         }
         providerKeys[`${lowerCaseEl}AuthorizeScopes`] = provider.authorize_scopes.split(',');
       } catch (e) {
-        return null;
+        // ignore
```
evcodes, line 567: should we use AmplifyError with this error?

evcodes (changes requested):
small changes but overall looks good. Thank you

packages/amplify-category-auth/src/provider-utils/awscloudformation/service-walkthroughs/auth-questions.ts
sdstolworthy:
```typescript
        providerKeys[`${lowerCaseEl}AuthorizeScopes`] = provider.authorize_scopes.split(',');
```
This is the only line where it is even possible to throw an error in this block, and it would be a null access error (`authorize_scopes` not found on `provider`, which is nullable from the find on line 554. The current behavior is a try catch, with the catch ignored. The last change to this line was made 4 years ago. Before introducing an AmplifyError here, I think a good first step would just be an optional chaining operator on `provider` on line 565. So line 565 would read `providerKeys[`${lowerCaseEl}AuthorizeScopes`] = provider?.authorize_scopes?.split?.(',');`

sdstolworthy: I lied, `creds` can also be null in that block. The reasoning remains the same though. The catch behavior was to simply pass, so an optional chaining operator in each applicable place should carry the same behavior.

evcodes (approved):

jhockett (approved):

PR #11496, opened 2022-11-30
chore: cdk v2 migration tests part 1

awsluja:
#### Description of changes
This is a draft; as I'm still working through some hiccups with the migration tests scaffold to support v10 -> v11.
The general idea is to make use of the pre-existing migration test framework, and then we can extend this with more specific test scenarios.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -102,6 +102,9 @@ jobs:
     executor: l_large
     steps:
       - checkout
+      - restore_cache:
+          key: >-
+            amplify-cli-yarn-deps-{{ .Branch }}-{{ checksum "yarn.lock" }}
       - run: yarn run production-build
       - run:
           name: Build tests
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -541,12 +541,12 @@ jobs:
       - restore_cache:
           key: amplify-pkg-binaries-{{ .Branch }}-{{ .Revision }}
       - run:
-          name: Run tests migrating from CLI v10.5.0
+          name: Run tests migrating from CLI v10.5.1
           command: |
             source .circleci/local_publish_helpers.sh
             changeNpmGlobalPath
             cd packages/amplify-migration-tests
-            retry yarn run migration_v10.5.0 --maxWorkers=3 $TEST_SUITE
+            retry yarn run migration_v10.5.1 --maxWorkers=3 $TEST_SUITE
           no_output_timeout: 90m
       - run: *scan_e2e_test_artifacts
       - store_test_results:
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -1049,6 +1049,20 @@ workflows:
           requires:
             - upload_pkg_binaries
             - build_windows_workspace_for_e2e
+      - amplify_migration_tests_v10:
+          context:
+            - e2e-auth-credentials
+            - cleanup-resources
+            - e2e-test-context
+          filters:
+            branches:
+              only:
+                - dev
+                - /run-e2e-with-rc\/.*/
+                - /tagged-release\/.*/
+                - /run-e2e\/.*/
+          requires:
+            - upload_pkg_binaries
       - amplify_migration_tests_v6:
           context:
             - e2e-auth-credentials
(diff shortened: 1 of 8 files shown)


.circleci/config.base.yml
```diff
@@ -525,6 +528,36 @@ jobs:
...
+      - restore_cache:
+          key: amplify-pkg-binaries-{{ .Branch }}-{{ .Revision }}
+      - run:
+          name: Run tests migrating from CLI v10.5.1
+          command: |
+            source .circleci/local_publish_helpers.sh
+            changeNpmGlobalPath
+            cd packages/amplify-migration-tests
+            unset IS_AMPLIFY_CI
+            echo $IS_AMPLIFY_CI
```
awsluja, line 551: this is a small hack, unfortunately, in v10 there is custom logic that checks for this parameter, and will try to install the wrong version using a commit hash. That feature interferes with us trying to test a specific version, such as 10.5.1, but this was not an issue in v6/v5
In order for the script to download the exact version, we must disable this "commit hash" feature manually.
image

.circleci/config.base.yml
```diff
@@ -525,6 +528,38 @@ jobs:
           path: ~/repo/packages/amplify-migration-tests/amplify-migration-reports
     working_directory: ~/repo
 
+  amplify_migration_tests_v10:
+    environment:
+      AMPLIFY_PATH: /home/circleci/.amplify/bin/amplify
```
awsluja, line 533: this is where the amplify cli is installed when using npm -g install, for v10 and up (or anything that uses the packaged cli); thats why its different here

packages/amplify-e2e-core/src/index.ts
```diff
@@ -91,9 +94,19 @@ export async function installAmplifyCLI(version = 'latest') {
...
+  console.log("SETTING PATH:");
+  if(version === '10.5.1'){
+    process.env.AMPLIFY_PATH = process.platform === 'win32'
+    ? path.join(os.homedir(), '.amplify', 'bin', 'amplify')
+    : path.join(os.homedir(), '.amplify', 'bin', 'amplify');
+  } else {
+    process.env.AMPLIFY_PATH = process.platform === 'win32'
     ? path.join(os.homedir(), '..', '..', 'Program` Files', 'nodejs', 'node_modules', '@aws-amplify', 'cli', 'bin', 'amplify')
     : path.join(os.homedir(), '.npm-global', 'bin', 'amplify');
+  }
```
awsluja, line 107: I had to update this because during setup, it would set AMPLIFY_PATH to the old cli location (before we moved to packaged cli), and this would ultimately fail because on v10 the cli is not installed to .npm-global

scripts/split-e2e-tests.ts
```diff
@@ -215,20 +215,6 @@ function splitTests(
...
-        ...newJob.environment,
-        ...(isMigration
-          ? {
-              AMPLIFY_PATH: '/home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify',
-            }
-          : {
-              AMPLIFY_DIR: '/home/circleci/repo/packages/amplify-cli/bin',
-              AMPLIFY_PATH: '/home/circleci/repo/packages/amplify-cli/bin/amplify',
-            }),
-      };
```
awsluja, line 230: previously, the split-e2e-tests script was responsible for specifying the AMPLIFY_PATH to use when running tests. We are also partially doing this in config.base.yml; so I updated that and got rid of this logic.
This logic only works for migration tests v4,v5,v6. For v10, a different AMPLIFY_PATH must be used. So this logic now exists in the config file instead

.circleci/config.base.yml
```diff
@@ -496,13 +501,13 @@ jobs:
...
   amplify_migration_tests_v6:
-    environment:
-      AMPLIFY_PATH: /home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify
     parameters:
       os:
         type: executor
         default: l_large
     executor: << parameters.os >>
+    environment:
+      AMPLIFY_PATH: /home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify
```
awsluja, line 510: because AMPLIFY_PATH is specified by the 'executor: <<..>>' step, we must provide it AFTER if we want to override the value. This is the case for migration tests.

.circleci/config.base.yml
```diff
@@ -473,6 +476,8 @@ jobs:
...
```
awsluja, line 480: this migration tests was missing this, all other migration tests were already specifying this

awsluja:
closing this in favor of: https://github.com/aws-amplify/amplify-cli/pull/11535

PR #11505, opened 2022-12-01
fix: add fallpack for undefined env var

danielleadams:
#### Description of changes
Fixes a parsing error that happens when running e2e tests locally due to a missing environment variable that is supplied by CI.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


sobolk (dismissed):

jhockett (dismissed):

danielleadams:
Failing test is unrelated to changes.

danielleadams:
@sobolk @jhockett this needs another review because there was a merge conflict

packages/amplify-e2e-core/src/utils/index.ts
```diff
@@ -30,7 +30,8 @@ export * from './git-operations';
 /**
  * Whether the current environment is CircleCI or not
  */
-export const isCI = (): boolean => process.env.CI && process.env.CIRCLECI && JSON.parse(process.env.CI) && JSON.parse(process.env.CIRCLECI);
+
+export const isCI = (): boolean => JSON.parse(process.env.CI || 'false') && JSON.parse(process.env.CIRCLECI || 'false');
```
jhockett, line 34: Is this change still needed? Looks like `isCI` was updated to handle the cases where these variables are undefined.

danielleadams: The previous statement returns undefined when `process.env.CI` and `process.env.CIRCLECI` are undefined. The above statement will always return a boolean, which is the defined return type. Typescript compiler doesn't seem to mind. I'm partial to returning the correct type for better definition of the function, but I don't feel too strongly since the compiler passes.

sobolk (approved):

jhockett (approved):

PR #11509, opened 2022-12-01
chore: filter out usage data from dev instances

lazpavel:
#### Description of changes
- injects `production` into the `process.env.NODE_ENV` value in our build
- filters out the usage data for non production builds

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/local_publish_helpers.sh
+++ b/.circleci/local_publish_helpers.sh
@@ -64,6 +64,9 @@ function generatePkgCli {
   # Restore .d.ts files required by @aws-amplify/codegen-ui at runtime
   cp ../node_modules/typescript/lib/*.d.ts node_modules/typescript/lib/
 
+  # Set NODE_ENV to production
+  sed -i "2s/^/process.env.NODE_ENV = 'production';\n/" node_modules/@aws-amplify/cli-internal/bin/amplify
+
   # Transpile code for packaging
   npx babel node_modules --extensions '.js,.jsx,.es6,.es,.ts' --copy-files --include-dotfiles -d ../build/node_modules
 
--- a/.circleci/local_publish_helpers.sh
+++ b/.circleci/local_publish_helpers.sh
@@ -64,8 +64,8 @@ function generatePkgCli {
   # Restore .d.ts files required by @aws-amplify/codegen-ui at runtime
   cp ../node_modules/typescript/lib/*.d.ts node_modules/typescript/lib/
 
-  # Set NODE_ENV to production
-  sed -i "2s/^/process.env.NODE_ENV = 'production';\n/" node_modules/@aws-amplify/cli-internal/bin/amplify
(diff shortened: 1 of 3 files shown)


.circleci/local_publish_helpers.sh
```diff
@@ -64,6 +64,9 @@ function generatePkgCli {
   # Restore .d.ts files required by @aws-amplify/codegen-ui at runtime
   cp ../node_modules/typescript/lib/*.d.ts node_modules/typescript/lib/
 
+  # Set NODE_ENV to production
+  sed -i "2s/^/process.env.NODE_ENV = 'production';\n/" node_modules/@aws-amplify/cli-internal/bin/amplify
+
```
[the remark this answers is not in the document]
danielleadams: I agree - I think this would be a good opportunity to create an either build time or runtime injection mechanism so that there is dev/prod parity. We can either configure the build environments with the correct env vars or inject the Node process with environment variables at runtime.

There is the `dotenv` package: https://www.npmjs.com/package/dotenv that does something like this, but it is only supposed to be used for development so as not to encourage committing secrets. I like the idea of having non-private configurations near the code base though, and we inject the the environment variables based on the build target. Or we can configure the CI builds with the build targets env vars.

packages/amplify-cli/src/domain/amplify-usageData/getUsageDataUrl.ts
```diff
@@ -7,8 +7,12 @@ export const prodUrl = `https://api.cli.amplify.aws/${version}/metrics`;
...
  */
 export const getUrl = (): UrlWithStringQuery => {
-  if (process.env.AMPLIFY_CLI_BETA_USAGE_TRACKING_URL && typeof process.env.AMPLIFY_CLI_BETA_USAGE_TRACKING_URL === 'string') {
-    return url.parse(process.env.AMPLIFY_CLI_BETA_USAGE_TRACKING_URL || '');
+  if (isProduction() && !useBetaUrl()) {
+    return url.parse(prodUrl);
   }
-  return url.parse(prodUrl);
+
+  return url.parse(process.env.AMPLIFY_CLI_BETA_USAGE_TRACKING_URL || '');
```
sobolk, line 14: I'd suggest to tweak this code a bit to make sure that we `url.parse` once per process lifespan.
I.e. we cache result of this logic somewhere. I don't think we need to compute it every time we resolve `getUrl`.
Otherwise change looks good.

sobolk (approved):

packages/amplify-cli/src/domain/amplify-usageData/getUsageDataUrl.ts
lazpavel: updated

sobolk (approved):

danielleadams (approved):

PR #11512, opened 2022-12-01
Go 1.17 function compile

MtthwBrwng:
#### Description of changes
Convert version to `SemVer` before comparing, and supplying a version on go module installed

#### Issue #, if available
#11306

Followed [CONTRIBUTING.md](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#getting-started) to setup local environment. Built & tested CLI locally

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-go-function-runtime-provider/src/runtime.ts
```diff
@@ -172,8 +172,8 @@ const winZip = async (src: string, dest: string, print: any) => {
   // get lambda zip tool with the fix of https://go.dev/doc/go-get-install-deprecation
   const version = getGoVersion();
   try {
-    if (gte(version, '1.17')) {
-      await execa(executableName, ['install', 'github.com/aws/aws-lambda-go/cmd/build-lambda-zip']);
+    if (gte(version, <SemVer>coerce('1.17'))) {
```
jhockett, line 175: What does `<SemVer>` do to `coerce`?

I can't find any examples of this in the docs: https://github.com/npm/node-semver#coercion

jhockett (commented):
Hi @MtthwBrwng, thanks for your contribution!

packages/amplify-go-function-runtime-provider/src/runtime.ts
MtthwBrwng: `<SemVer>` converts the return type of `coerce` to be a non-null value

```
declare function coerce(
    version: string | number | SemVer | null | undefined,
    options?: semver.CoerceOptions,
): SemVer | null;
```

Since coerce can return either `SemVer` or `null`, by not converting the type it throws a type error for the `gte` function from the `semver` package which expects the values to all be `non-null`. Which I think in this instance should be fine since it's a hardcoded version string that meets `semver` requirements.

jhockett: I didn't know you could write type assertions with angle brackets, thanks for providing such a detailed explanation!

I didn't test this code, but does TypeScript resolve the type to be defined if the value is checked to not be null or undefined?

```suggestion
    const goSemVer1Dot17 = coerce('1.17');
    if (!goSemVer1Dot17) { 
      throw new Error('SemVer coercion failed');
    }
    if (gte(version, goSemVer1Dot17)) {
```

MtthwBrwng: Yes. Typescript resolves the type to be defined if the value is checked.

There are 3 other instances where that style of type assertion is being used. Lines 23, 24, and 126 in the same file. Should I look at changing those over as well? Two of them are setting a global variable for the minimum and maximum supported versions of Go. Looks like they're only referenced in one function call, so I can check for value there.

jhockett: In that case, I think refactoring the code to avoid the type assertions is out of scope for this PR. Thanks for going back and forth with me on this.

jhockett (approved):

sobolk (approved):

Issue #11066, opened 2022-09-24
"The local environment configuration does not allow modifying the backend." when running `amplify env checkout <env_name>`


manu2504:

ykethan:
Hey, thank you for the information. I was able to reproduce this, marking this as bug.

Note: ran the following steps:

1. `amplify pull --envName dev --appId <app id>` to pull in dev environment and select modify backend as `no`.
2. then ran `amplify pull --envName test` to pull in test environment.
3. `amplify env list` shows the following:
```
| Environments |
| ------------ |
| dev          |
| *test        |
```
4. running `amplify env checkout dev` shows the following error.
```
The local environment configuration does not allow modifying the backend.
Use amplify env pull --envName dev
```
5. running `amplify pull --envName dev --appId <app id>` also throws the error.
```
- Creating Zip
File at path: 'C:\test\amplify\backend\backend-config.json' does not exist
× Creating Zip
🛑 Cannot set property 'config' of undefined
TypeError: Cannot set property 'config' of undefined
```

manu2504:
@ykethan any update on this ticket? I'm still facing this really annoying bug as an AWS customer.

I had to rollback to @aws-amplify/cli@9 as v10 is broken on this aspect.

wcricchi:
Is there an update on this issue? The same thing happened to me when I pulled a different env, chose "No" to modifying the backend for that env, but now I cannot make changes to any of my other environments.

Rolando-Barbella:
This is also happening to me when I do amplify push

GitNothing:
A quick fix is do "amplify configure". It will basically reset your amplify configuration and later prompt if you're planning on pushing local changes. Hit YES this time.

FliCrz:
To fix this issue just go to folder amplify > config and in  file local-env-info.json change the value of the parameter: "noUpdateBackend" to false.

Umoren:
> To fix this issue just go to folder amplify > config and in file local-env-info.json change the value of the parameter: "noUpdateBackend" to false.

Yes this is the solution you're looking for. 
If you can't find `amplify/.config` it means in your code editor `settings.json` the folder is excluded. Something like this:
```json
{
  "files.exclude": {
    "amplify/.config": false,
    ...
  }
}
```
Then head to `amplify/.config/local-env-info.json` and change `noUpdateBackend` to `false`:
```json 
{
 ...
  "noUpdateBackend": false
}

```
Run `amplify push` to confirm it's working

PR #11513, opened 2022-12-01
fix: diagnose error if backend not pulled

jhockett:
#### Description of changes
Currently when no backend is pulled (Answer No to "Do you plan on modifying this backend?), diagnose flows fail. Either by triggering an error to be thrown, or by running `amplify diagnose`.

The changes remove assumptions that a backend is always present.

#### Issue #, if available
https://github.com/aws-amplify/amplify-cli/issues/11066

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/diagnose.ts
+++ b/packages/amplify-cli/src/commands/diagnose.ts
@@ -126,7 +126,7 @@ const createZip = async (context: Context, error: Error | undefined): Promise<st
   if (!rootPath) {
     throw projectNotInitializedError();
   }
-  const backend = stateManager.getBackendConfig(rootPath);
+  const backend = stateManager.getBackendConfig(rootPath, { throwIfNotExist: false, default: {} });
   const resources: { category: string; resourceName: string; service: string }[] = [];
   const categoryResources = Object.keys(backend).reduce((array, key) => {
     Object.keys(backend[key]).forEach(resourceKey => {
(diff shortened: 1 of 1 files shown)


packages/amplify-cli/src/commands/diagnose.ts
```diff
@@ -240,6 +240,6 @@ const hashedProjectIdentifiers = (): { projectIdentifier: string; projectEnvIden
 };
 
 const getAppId = (): string => {
-  const meta = stateManager.getMeta();
-  return _.get(meta, ['providers', 'awscloudformation', 'AmplifyAppId']);
+  const meta = stateManager.getMeta(undefined, { throwIfNotExist: false });
+  return meta?.providers?.awscloudformation?.AmplifyAppId ?? 'noAppIdFound';
```
akshbhu, line 244: Are we checking this some where 'noAppIdFound' ?

jhockett: It's used to create the projectIdentifier: https://github.com/aws-amplify/amplify-cli/pull/11513/files#diff-71cabcaa3379af3b9458f6802b25c65f6c00768606d5ef7f8c010094d8720aa1R238-R239

But there's no `amplify-meta.json` file when no backend is pulled, so the existing code always fails on ln 243.

akshbhu (approved):

PR #11518, opened 2022-12-01
chore: add codeql scanning

sdstolworthy:
#### Description of changes

Add Github CodeQL scanning in preparation for the sunset of LGTM. This PR does not remove LGTM from the current developer workflow. LGTM will be removed in a subsequent PR.

CodeQL provides insight into potential vulnerability issues in the codebase by performing static analysis of the code. These issues are visible to maintainers in the Security tab on Github.

Changes were validated in my own fork of Amplify CLI.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.github/workflows/codeql.yaml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ["javascript"]
```
evcodes, line 20: I'm curious if we wanna include typescript here too since we may have some problems with types that can be caught by codeql. I know that TypeScript compiles down to Javascript but I think there may be some value in adding that to the language matrix.

evcodes (changes requested):
Only one suggestion. From what I understood in the codeql docs, it seems that having typescript there would be valuable.

.github/workflows/codeql.yaml
sdstolworthy: https://github.com/github/codeql-action/issues/365

Amplifiyer:
Is the section `Unchanged files with check annotations` added by the CodeQL and will be present in subsequent PRs?

evcodes (approved):
thanks for the explanation, LGTM

sdstolworthy:
@Amplifiyer Yes

Amplifiyer:
> @Amplifiyer Yes

Is there a way to turn them off for unrelated files (files that didn't get changed in the PR)? Right now it seems noise.

sdstolworthy:
@Amplifiyer I was looking into that yesterday afternoon. I didn't see any clear option to do that. I'll do some more digging, but my current thought is no. That said, it shouldn't fail a PR due to files not included in that change. I tested on my own fork and PRs were passing after merging this change into my own `dev` branch.

jhockett (approved):

PR #11535, opened 2022-12-05
chore: migration tests refactor

awsluja:
#### Description of changes
This set of changes does the following:
1. cleans up existing migration tests so that we know which versions of CLI e2e tooling is being used (init v4, init v6, etc)
2. updates split-e2e-tests so that migration tests can be created for a specific version, and specific tests are maintained in a list (instead of using jest regex to ignore no applicable tests - this used to create a bunch of unnecessary jobs that just ran without any tests)
3. adds an initial test that points to 10.5.1 , which will be used for cdkv2 migration testing

E2E run: https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/13378/workflows/6ed9f8ed-ec1b-450b-b727-c669dc669c2a
All test passing 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.circleci/config.base.yml
```diff
@@ -462,6 +462,8 @@ jobs:
         type: executor
         default: l_large
     executor: << parameters.os >>
+    environment:
+      AMPLIFY_PATH: /home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify
```
awsluja, line 466: this migration tests was missing this, all other migration tests were already specifying this.
There was a 'hack' in the split-e2e-tests script that allowed v5 migration tests to work without this line, but now thats been removed.

.circleci/config.base.yml
```diff
@@ -485,13 +487,13 @@ jobs:
...
   amplify_migration_tests_v6:
-    environment:
-      AMPLIFY_PATH: /home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify
     parameters:
       os:
         type: executor
         default: l_large
     executor: << parameters.os >>
+    environment:
+      AMPLIFY_PATH: /home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify
```
awsluja, line 496: because AMPLIFY_PATH is also specified by the 'executor: <<..>>' step, we must provide it AFTER if we want to override the value. This is the case for migration tests.

.circleci/config.base.yml
```diff
@@ -514,6 +516,38 @@ jobs:
...
+      - restore_cache:
+          key: amplify-pkg-binaries-{{ .Branch }}-{{ .Revision }}
+      - run:
+          name: Run tests migrating from CLI v10.5.1
+          command: |
+            source .circleci/local_publish_helpers.sh
+            changeNpmGlobalPath
+            cd packages/amplify-migration-tests
+            unset IS_AMPLIFY_CI
+            echo $IS_AMPLIFY_CI
```
awsluja, line 541: this is a small hack, unfortunately, in v10 there is custom logic that checks for this parameter, and will try to install the wrong version using a commit hash. That feature interferes with us trying to test a specific version, such as 10.5.1, but this was not an issue in v6/v5
In order for the script to download the exact version, we must disable this "commit hash" feature manually.

image

packages/amplify-e2e-core/src/index.ts
```diff
@@ -91,9 +94,19 @@ export async function installAmplifyCLI(version = 'latest') {
...
+  console.log("SETTING PATH:");
+  if(version === '10.5.1'){
+    process.env.AMPLIFY_PATH = process.platform === 'win32'
+    ? path.join(os.homedir(), '.amplify', 'bin', 'amplify')
+    : path.join(os.homedir(), '.amplify', 'bin', 'amplify');
+  } else {
+    process.env.AMPLIFY_PATH = process.platform === 'win32'
     ? path.join(os.homedir(), '..', '..', 'Program` Files', 'nodejs', 'node_modules', '@aws-amplify', 'cli', 'bin', 'amplify')
     : path.join(os.homedir(), '.npm-global', 'bin', 'amplify');
+  }
```
awsluja, line 107: I had to update this because during setup, it would set AMPLIFY_PATH to the old cli location (before we moved to packaged cli), and this would ultimately fail because on v10 the cli is not installed to .npm-global

packages/amplify-migration-tests/package.json
```diff
@@ -18,10 +18,11 @@
...
-    "migration_v4.52.0_multienv_layers": "npm run setup-profile 4.52.0 && jest src/__tests__/migration_tests/lambda-layer-migration/layer-migration.test.ts --verbose --passWithNoTests",
-    "migration_v5.2.0": "npm run setup-profile 5.2.0 && jest --testPathIgnorePatterns=auth.deployment.secrets\\|layer-migration\\|overrides --verbose --passWithNoTests",
-    "migration_v6.1.0": "npm run setup-profile 6.1.0 && jest --testPathIgnorePatterns=auth.deployment.secrets\\|layer-migration --verbose --passWithNoTests",
+    "migration_v4.28.2_nonmultienv_layers": "npm run setup-profile 4.28.2 && jest src/__tests__/migration_tests/lambda-layer-migration/layer-migration.test.ts --verbose",
+    "migration_v4.52.0_multienv_layers": "npm run setup-profile 4.52.0 && jest src/__tests__/migration_tests/lambda-layer-migration/layer-migration.test.ts --verbose",
+    "migration_v5.2.0": "npm run setup-profile 5.2.0 && jest --verbose",
+    "migration_v6.1.0": "npm run setup-profile 6.1.0 && jest --verbose",
+    "migration_v10.5.1": "npm run setup-profile 10.5.1 && jest --verbose",
```
awsluja, line 25: previously, all of these JEST commands were using regex to exclude tests, this is the wrong way to do it, as we were still creating jobs for these tests but Jest would skip the tests and mark the jobs as passing. 

I moved this logic to our split-e2e-test script. Now, jobs are only created if the test should exist.

packages/amplify-migration-tests/src/__tests__/migration_tests_v10/scaffold.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    projRoot = await createNewProjectDir('scaffoldTest');
+  });
+
+  afterEach(async () => {
+    await deleteProject(projRoot, null, true);
+    deleteProjectDir(projRoot);
+  });
+
+  it('...should init a project', async () => {
+    await initJSProjectWithProfileV10(projRoot, { name: 'scaffoldTest' });
```
awsluja, line 31: in v10, we have custom init flow that is different than in v4/v5/v6.. so this test showcases that

packages/amplify-migration-tests/src/configure_tests.ts
```diff
@@ -9,6 +9,8 @@ import { amplifyConfigure as configure, isCI, installAmplifyCLI, injectSessionTo
 async function setupAmplify(version: string = 'latest') {
   // install CLI to be used for migration test initial project
   await installAmplifyCLI(version);
+  
+  console.log("INSTALLED CLI:", version);
```
awsluja, line 13: sanity check :)

packages/amplify-migration-tests/src/migration-helpers-v10/init.ts
```diff
@@ -0,0 +1,94 @@
+
```
awsluja, line 1: I created this folder so that we can place v10 e2e tools here.. such as for init, and other custom e2e flows

scripts/split-e2e-test-filters.ts
```diff
@@ -0,0 +1,45 @@
+export const migrationFromV4_28_2Tests = [
+    "src/__tests__/migration_tests/lambda-layer-migration/layer-migration.test.ts",
+];
+export const migrationFromV4_52_0Tests = [
+    "src/__tests__/migration_tests/lambda-layer-migration/layer-migration.test.ts",
+]
```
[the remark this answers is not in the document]
awsluja: This explicit list is easier to maintain, and does not produce jobs in circleCI for tests that will be skipped by Jest anyway. (here is an example of this:
image

scripts/split-e2e-tests.ts
```diff
@@ -215,20 +219,6 @@ function splitTests(
...
-        ...(isMigration
-          ? {
-              AMPLIFY_PATH: '/home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify',
-            }
-          : {
-              AMPLIFY_DIR: '/home/circleci/repo/packages/amplify-cli/bin',
-              AMPLIFY_PATH: '/home/circleci/repo/packages/amplify-cli/bin/amplify',
-            }),
-      };
-    }
```
awsluja, line 231: previously, the split-e2e-tests script was responsible for specifying the AMPLIFY_PATH to use when running tests. We are also partially doing this in config.base.yml; so I updated that and got rid of this logic.
This logic only works for migration tests v4,v5,v6. For v10, a different AMPLIFY_PATH must be used. So this logic now exists in the config file instead

.circleci/config.base.yml
sobolk: is `/home/circleci/.amplify/bin/amplify` different from `/home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify` change set wise?

awsluja: the split-e2e-tests script was inserting this line for all migration tests;
This resolves to the binary that is installed via NPM prior to packaged CLI (v6 and earlier), which was at the time installed to .npm-globa/.../bin

sobolk: gotcha, so if I run `/home/circleci/.npm-global/lib/node_modules/@aws-amplify/cli/bin/amplify version` on that circle ci VM this will print v5 or v6 right ?

.circleci/config.base.yml
sobolk: I would recommend to do this:
```
  if (process.env.IS_AMPLIFY_CI && process.env.IS_AMPLIFY_CI_USE_LATEST_BINARY) { // exact naming of extra env var tbd.
    url = url.replace('.tgz', `-${getCommitHash()}.tgz`);
  }
```
or
```
  if (process.env.IS_AMPLIFY_CI && process.env.BINARY_URL_OVERRIDE) {
    url = BINARY_URL_OVERRIDE;
  }
```

using just `IS_AMPLIFY_CI` to work around this is going to be problematic in the future if there is another logic that would like to use that to branch out.

sobolk: fwiw I like the `BINARY_URL_OVERRIDE` more. and I'd shift this getComitHash logic to the caller (script).

packages/amplify-e2e-core/src/index.ts
```diff
@@ -91,9 +90,19 @@ export async function installAmplifyCLI(version = 'latest') {
     env: process.env,
     stdio: 'inherit',
   });
-  process.env.AMPLIFY_PATH = process.platform === 'win32'
+  
+  console.log("SETTING PATH:");
+  if(version === '10.5.1'){
```
sobolk, line 95: consider using https://www.npmjs.com/package/semver .
e.g.
`if (semver.gt(version, '10.0.0')) `

packages/amplify-migration-tests/package.json
sobolk: should we think about failing a test run if zero tests have been found?

packages/amplify-migration-tests/src/configure_tests.ts
sobolk: nit: should we also print result of amplify version ? to verify that installAmplifyCLI actually installs requested version.

sobolk (commented):
lgtm

packages/amplify-migration-tests/src/migration-helpers-v10/init.ts
```diff
@@ -0,0 +1,94 @@
...
+    const providerConfigSpecified = !!s.providerConfig && typeof s.providerConfig === 'object';
+    if (providerConfigSpecified) {
+      cliArgs.push('--providers', JSON.stringify(s.providerConfig));
+    }
+  
+    if (s.permissionsBoundaryArn) {
+      cliArgs.push('--permissions-boundary', s.permissionsBoundaryArn);
+    }
+  
+    if (s?.name?.length > 20) console.warn('Project names should not be longer than 20 characters. This may cause tests to break.');
```
jhockett, line 48: Should we throw an error here to enforce we don't run into this instead?

packages/amplify-migration-tests/src/migration-helpers-v10/init.ts
```diff
@@ -0,0 +1,94 @@
...
+    profileName: EOL,
+    region: process.env.CLI_REGION,
+    local: false,
+    disableAmplifyAppCreation: true,
+    disableCIDetection: false,
+    providerConfig: undefined,
+    permissionsBoundaryArn: undefined,
+};
+  
+export function initJSProjectWithProfileV10(cwd: string, settings?: Partial<typeof defaultSettings>): Promise<void> {
```
jhockett, line 26: nit: this can be simplified to an async function that returns the Promise from `.runAsync()`

.circleci/config.base.yml
awsluja: yeah thats right

packages/amplify-migration-tests/package.json
awsluja: Yeah ; I did this already, previously it said "--allow pass if no tests found" , thats been removed

packages/amplify-migration-tests/src/migration-helpers-v10/init.ts
awsluja: I copied this method as a snapshot from the current v10 e2e-core utils; i can modify but seems out of scope for this pr

packages/amplify-migration-tests/src/__tests__/migration_tests_v10/scaffold.test.ts
```diff
@@ -0,0 +1,34 @@
...
+describe('amplify migration test scaffold v10', () => {
+  let projRoot: string;
+
+  beforeAll(async () => {
+    const migrateFromVersion = { v: 'unintialized' };
+    const migrateToVersion = { v: 'unintialized' };
+    await versionCheck(process.cwd(), false, migrateFromVersion);
+    await versionCheck(process.cwd(), true, migrateToVersion);
+    console.log(`Test migration from: ${migrateFromVersion} to ${migrateToVersion}`);
+    // expect(migrateFromVersion.v).not.toEqual(migrateToVersion.v); // uncomment this once we are in v11 for local codebase
```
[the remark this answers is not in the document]
awsluja: thanks, just created this: https://app.asana.com/0/0/1203504441618194/f

packages/amplify-migration-tests/src/migration-helpers-v10/init.ts
awsluja: similarly here; didn't want to refactor since it is just a snapshot of the e2e-core tooling from v10

.circleci/config.base.yml
awsluja: we can't modify this in 10.5.1 or previous versions; the logic executes from the existing/old release as it's already part of the binary files; if we change this now in v10.5.2/v10.6 we will still need this custom logic if we add migration tests for v9 or v10.4

sobolk: gotcha. good to go as is in this PR. I'd log follow up task to get this straight for future releases.

we can keep IS_AMPLIFY_CI as is. I'd just add extra knob to control this particular piece of code and use it in migration tests, i.e. `IS_AMPLIFY_CI && URL_SPECIFIC_VAR`

sobolk (approved):

jhockett (approved):

PR #11543, opened 2022-12-06
fix: add version bumps logic to changelog creation

goldbez:
#### Description of changes
I added logic to `unified-changelog.ts` that checks if a release is comprised of all version bumps and, if so, includes version bumps in the changelog. If there are changes in the changelog that are not version bumps, then all version bumps are omitted from the changelog. This is to avoid release notes from ever being completely empty, as happened with 10.5.1.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/scripts/unified-changelog.ts
+++ b/scripts/unified-changelog.ts
@@ -56,16 +56,19 @@ const extractChangelogInfo = async (changelogPath: string): Promise<ChangelogInf
 
 const formatUnifiedChangelog = (infos: ChangelogInfo[]): string => {
   console.log('Formatting all changes into single changelog');
-  const formattedChangelog = infos
+  var formattedChangelogComponents = infos
     .filter(info => info.latestChanges)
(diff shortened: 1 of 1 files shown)


edwardfoyle (approved):

scripts/unified-changelog.ts
```diff
@@ -56,16 +56,19 @@ const extractChangelogInfo = async (changelogPath: string): Promise<ChangelogInf
 
 const formatUnifiedChangelog = (infos: ChangelogInfo[]): string => {
   console.log('Formatting all changes into single changelog');
-  const formattedChangelog = infos
+  var formattedChangelogComponents = infos
```
jhockett, line 59: nit: I prefer `let` because I think block scoping is easier to reason about.

```suggestion
  let formattedChangelogComponents = infos
```

jhockett (approved):
LGTM, 1 small nit to consider

akshbhu (approved):

PR #11548, opened 2022-12-06
feature: graphiql explorer added clear data button

ncarvajalc:
#### Description of changes

Added a clear data button when using `amplify mock api` in the Graphiql explorer:

image
image

#### Issue #, if available

aws-amplify/amplify-category-api#261

`yarn test` passes and created different schemas and deleted all of it's data with the newly created button. I wasn't able to delete open search indexes as they are not generated by the appsync simulator.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


evcodes:
Thanks for the contribution, this seems like a great feature. I am working with internal teams to help get this over the line 👍🏼

packages/amplify-appsync-simulator/src/data-loader/dynamo-db/index.ts
```diff
@@ -60,6 +63,37 @@ export class DynamoDBDataLoader implements AmplifyAppSyncSimulatorDataLoader {
...
+    try {
+      const items = await this.getAllItems();
+      for await (const item of items) {
+        await this.client
+          .deleteItem({
+            TableName: this.tableName,
+            Key: { id: item.id },
+            ReturnValues: 'ALL_OLD',
+          }).promise();
+      }
```
alharris-at, line 76: Code style suggestion. Instead of nesting awaits, I'd try to think about the promises you're trying to spin up, and whether they need to be sequenced for any particular reason. As written this code will delete one item at a time, but it may be more readable (and faster) to just do something like:
```ts
await Promise.all(items.map(item => this.client.deleteItem({
  TableName: this.tableName,
  Key: { id: item.id },
  ReturnValues: 'ALL_OLD',
})));
```

And kick off all deletes in parallel, then yield until all deletes to complete.

Note: for large tables, this could spin up a huge number of promises and requests to the client, so you instead may want to batch these by say 20 at a time (or some other tested value), or maybe we can actually chunk to a set of ids, and use [BatchWriteItems](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/DynamoDB.html#batchWriteItem-property) to delete in batches at the ddb level, instead of in client code.

packages/amplify-appsync-simulator/src/data-loader/dynamo-db/index.ts
```diff
@@ -60,6 +63,37 @@ export class DynamoDBDataLoader implements AmplifyAppSyncSimulatorDataLoader {
...
+    } catch (e) {
+      throw new Error(`Error while deleting all items from ${this.tableName}`);
+    }
+    return [this.tableName];
+  }
+
+  private async getAllItems(): Promise<Array<AttributeMap> | null> {
```
alharris-at, line 83: Similarly, a comment on this method wouldn't go amiss.

alharris-at: As I read this method, kind of a silly thought that may be simpler.

Can you just continually run a scan on page 1, verify you have items.length > 0 && LastEvaluatedKey !== undefined, then call BatchWriteItems w/ deletes for those items, and break when you get no records back? Then you don't need to worry about holding a ton of items in-memory and passing them between these methods.

packages/amplify-appsync-simulator/src/index.ts
```diff
@@ -154,6 +155,19 @@ export class AmplifyAppSyncSimulator {
     return this.resolvers.get(`${typeName}:${fieldName}`);
   }
 
+  async clearData(): Promise<object> {
```
alharris-at, line 158: Comment here please :D

packages/amplify-appsync-simulator/src/index.ts
```diff
@@ -154,6 +155,19 @@ export class AmplifyAppSyncSimulator {
...
+    const it = this.dataSources.values();
```
alharris-at, line 159: n.p. on naming, but I initially read this like 'it' in kotlin, which is usually just used in a closure to represent 'the thing I'm operating on' but I realize you probably meant it for 'iterator'. I'd prefer to either call this dataSources, or dataSourceIterator, or something a little more verbose.

packages/amplify-appsync-simulator/src/server/operations.ts
```diff
@@ -27,9 +27,22 @@ export class OperationServer {
...
 
+  private handleClearDBData = async (request: express.Request, response: express.Response) => {
+    console.log('Clearing DB data...');
+    try {
+      const deletedItems = await this.simulatorContext.clearData();
+      console.log('DB data cleared');
+      return response.status(200).send({ message: `Successfully deleted ${JSON.stringify(deletedItems)} tables` });
+    } catch (e) {
+      console.log('Error clearing DB data');
+      return response.status(500).send({ message: e.message });
```
alharris-at, line 42: I'd include the error message in your console log here, and should use `console.error` instead.

packages/amplify-graphiql-explorer/src/App.tsx
```diff
@@ -49,6 +50,10 @@ function getAPIInfo() {
   return fetch('/api-config').then(response => response.json());
 }
 
+function clearData() {
```
alharris-at, line 53: Comment :D

packages/amplify-graphiql-explorer/src/App.tsx
```diff
@@ -299,9 +312,26 @@ class App extends Component<{}, State> {
...
+          this.setState({ clearDataModalVisible: false });
+        }}
+        onClear={async () => {
+          this.setState({ clearDataModalVisible: false });
+          this.setState({ clearResponse: JSON.stringify(await clearData(), undefined, 2) });
+          setTimeout(() => {
+            this.setState({ clearResponse: '' });
+          }
+          , 8000);
+        }}
```
alharris-at, line 328: I'd pull these methods out into their own definition, just to separate the functionality and binding to the component here.

Also, where's the 8s delay coming from?

packages/amplify-graphiql-explorer/src/App.tsx
```diff
@@ -310,13 +340,14 @@ class App extends Component<{}, State> {
             onRunOperation={operationName => this._graphiql?.handleRunQuery(operationName)}
             explorerIsOpen={this.state.explorerIsOpen}
             onToggleExplorer={this._handleToggleExplorer}
-          />
+            />
```
alharris-at, line 343: whitespace change

packages/amplify-graphiql-explorer/src/ClearDataModal.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+type Props = {
+  onClose: Function;
+  onClear: Function;
+};
+export function ClearDataModal(props: Props) {
+  const [isOpen, setIsOpen] = useState(true);
+  const actionText = "Clear";
+  const cancelText = "Cancel";
+  const confirmationText = "Do you want to clear the data of your mock server?";
```
alharris-at, line 12: These can be pulled out of the definition, since they don't need to be redefined on every instance of this component.

packages/amplify-graphiql-explorer/src/ClearDataModal.tsx
```diff
@@ -0,0 +1,39 @@
+import { useState } from "react";
+import { Button, Modal } from "semantic-ui-react";
+
+type Props = {
+  onClose: Function;
+  onClear: Function;
+};
```
alharris-at, line 7: Could you add a comment about the intention for these callbacks? It's not obvious to me what the difference between these should be, since you're calling them both on reset, for example.

alharris-at (commented):
Overall contribution makes sense, and looks good! Just a few stylistic questions/comments, and a few on the interactions w/ DDB.

danielleadams:
@ncarvajalc have you had some time to look at the Amplify team's review? Thanks!

packages/amplify-appsync-simulator/src/data-loader/dynamo-db/index.ts
ncarvajalc: Tried to do it this way and records weren't getting deleted for some reason

packages/amplify-graphiql-explorer/src/App.tsx
ncarvajalc: It was an arbitrary delay. The message was still showing up even after running other queries. So after 8 seconds the deletion message gets cleared

evcodes (approved):

PR #11562, opened 2022-12-08
fix: dotnet lambda templates compilation issues

sobolk:
#### Description of changes

This PR fixes existing dotnet templates and makes sure that they work (i.e. compile and return something).

Current samples (except Hello World) don't compile due to dependency conflicts.
image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-dotnet-function-template-provider/src/utils/dynamoDBWalkthrough.ts
```diff
@@ -85,7 +85,7 @@ export async function getTableParameters(context: any, dynamoAnswers: any): Prom
   } // Looking for table parameters on local configuration
   const projectBackendDirPath = context.amplify.pathManager.getBackendDirPath();
   const resourceDirPath = path.join(projectBackendDirPath, 'storage', dynamoAnswers.resourceName);
-  const parametersFilePath = path.join(resourceDirPath, 'parameters.json');
+  const parametersFilePath = path.join(resourceDirPath, 'build', 'parameters.json');
```
sobolk, line 88: this miss showed up in new test coverage.

packages/amplify-e2e-core/src/categories/lambda-function.ts
```diff
@@ -359,6 +359,56 @@ export const selectTemplate = (chain: ExecutionContext, functionTemplate: string
   moveUp(chain, templateChoices.indexOf('Hello World'));
 
   singleSelect(chain, functionTemplate, templateChoices);
+  if (functionTemplate === 'CRUD function for DynamoDB (Integration with API Gateway)') {
+    configureCRUDTemplate(chain);
+  } else if (functionTemplate === 'Trigger (DynamoDb, Kinesis)') {
+    configureDDBTrigger(chain);
+  }
```
sobolk, line 366: these lambda templates require follow up steps to set up extra resources.

packages/amplify-e2e-tests/src/__tests__/function_3b.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      'dotnetCore31',
+    );
+    await functionMockAssert(projRoot, {
+      funcName,
+      successString: serverlessSuccessString,
+      eventFile: 'src/event.json',
+    }); // will throw if successString is not in output
+  });
+
+  it('add dotnet crud function and invoke in the cloud', async () => {
```
sobolk, line 76: crud template is not able to use `amplify mock function`. it requires ddb table reference at runtime. Hence this test pushes and calls real lambda.

packages/amplify-e2e-tests/src/__tests__/function_3b.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      },
+      'dotnetCore31',
+    );
+    const payload = '{"key1":"value1","key2":"value2","key3":"value3"}';
+    await amplifyPushAuth(projRoot);
+    const response = await functionCloudInvoke(projRoot, { funcName, payload });
+    expect(JSON.parse(response.Payload.toString())).toEqual(helloWorldSuccessObj);
+  });
+
+  it('add dotnet serverless function and mock locally', async () => {
```
sobolk, line 60: new coverage is  base on mock as it compiles and runs dotnet code locally. this is best effort, some templates like crud don't work locally.

packages/amplify-dotnet-function-template-provider/resources/lambda/Crud/event.json.ejs
```diff
@@ -1,7 +1,7 @@
 {
   "body": null,
   "resource": "<%= props.functionTemplate.parameters.path %>/{proxy+}",
-  "path": "<%= props.functionTemplate.parameters.path %>/",
+  "path": "<%= props.functionTemplate.parameters.path %>/<%= props.functionTemplate.parameters.database.tableDefinition.partitionKeyName %>:foo",
```
sobolk, line 4: while we don't use this event in test I was playing a bit with this in aws console to invoke resulting lambda.
previous version was leading to error response asking for missing path parameter - this change fixes this.

packages/amplify-e2e-core/src/categories/lambda-function.ts
sobolk: nvm, I refactored that.

packages/amplify-dotnet-function-template-provider/src/utils/dynamoDBWalkthrough.ts
jhockett: Line 86 to 96 can be simplified down to a single stateManager call. Technically, the try/catch handles additional error scenarios, like if permissions is denied reading the file, but I think we should throw in that case.

```ts
return stateManager.getResourceParametersJson(undefined, AmplifyCategories.STORAGE, dynamoAnswers.resourceName, { throwIfNotExists: false, default: {} });
```

jhockett (approved):
LGTM, I left 1 suggestion for consideration, but not blocking

packages/amplify-dotnet-function-template-provider/src/utils/dynamoDBWalkthrough.ts
sobolk: @jhockett done.

jhockett (approved):

lazpavel (approved):

PR #11564, opened 2022-12-08
chore: convert storeCurrentCloudBackend from promise-based chain to async await for easier reading

lazpavel:
#### Description of changes
- converts promise-based chain to async await for easier reading

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/initializer.ts
+++ b/packages/amplify-provider-awscloudformation/src/initializer.ts
@@ -323,13 +323,13 @@ const storeCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
     const archive = await archiver.run(currentCloudBackendDir, zipFilePath, undefined, cliJSONFiles);
     const s3Key = `${archive.zipFilename}`;
     const s3Instance = await S3.getInstance(context);
-    spinner.stop('Deployment bucket fetched.');
     const s3Params = {
       Body: fs.createReadStream(archive.zipFilePath),
       Key: s3Key,
     };
     logger('storeCurrentCloudBackend.s3.uploadFile', [{ Key: s3Key }])();
     await s3Instance.uploadFile(s3Params);
+    spinner.stop('Deployment state saved successfully.');
   } catch (ex) {
     spinner.stop('Deployment state save failed.', false);
     throw new AmplifyFault('DeploymentFault', {
--- a/packages/amplify-provider-awscloudformation/src/initializer.ts
+++ b/packages/amplify-provider-awscloudformation/src/initializer.ts
@@ -318,30 +318,26 @@ const storeCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
   const zipFilePath = path.normalize(path.join(tempDir, zipFilename));
   const spinner = new AmplifySpinner();
 
-  return archiver
-    .run(currentCloudBackendDir, zipFilePath, undefined, cliJSONFiles)
-    .then(result => {
-      const s3Key = `${result.zipFilename}`;
-      spinner.start('Saving deployment state.');
-      return S3.getInstance(context).then(s3 => {
-        spinner.stop('Deployment bucket fetched.');
-        const s3Params = {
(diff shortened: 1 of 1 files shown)


packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -318,30 +318,26 @@ const storeCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
...
-      }, ex);
-    })
-    .then(() => {
-      fs.removeSync(tempDir);
-    });
+  try {
+    spinner.start('Saving deployment state.');
+    const archive = await archiver.run(currentCloudBackendDir, zipFilePath, undefined, cliJSONFiles);
+    const s3Key = `${archive.zipFilename}`;
+    const s3Instance = await S3.getInstance(context);
```
evcodes, line 325: nice find

packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -318,30 +318,26 @@ const storeCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
...
-    .catch(ex => {
-      spinner.stop('Deployment state save failed.', false);
-      throw new AmplifyFault('DeploymentFault', {
-        message: ex.message,
-      }, ex);
-    })
-    .then(() => {
-      fs.removeSync(tempDir);
-    });
+  try {
```
sobolk, line 321: the new code looks good.

I read old code and the promise chain there seems ok-ish. Do you know which line leaked a promise there?

packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -318,30 +318,26 @@ const storeCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
...
+    spinner.start('Saving deployment state.');
+    const archive = await archiver.run(currentCloudBackendDir, zipFilePath, undefined, cliJSONFiles);
+    const s3Key = `${archive.zipFilename}`;
+    const s3Instance = await S3.getInstance(context);
+    spinner.stop('Deployment bucket fetched.');
```
sobolk, line 326: The spinner is kind of stopped in a very weird place and has quite surprising message.

I suggest we move it after `s3Instance.uploadFile` and change messaging to `Deployment state saved successfully` or so.
(I know old code had this behavior, but we should make it better if we're touching it).

packages/amplify-provider-awscloudformation/src/initializer.ts
lazpavel: I believe the `uploadFile` was not awaited

lazpavel: nevermind, confirmed offline that the refactored code has the same flow

sobolk (approved):

packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -318,30 +318,26 @@ const storeCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
...
+    spinner.start('Saving deployment state.');
+    const archive = await archiver.run(currentCloudBackendDir, zipFilePath, undefined, cliJSONFiles);
+    const s3Key = `${archive.zipFilename}`;
+    const s3Instance = await S3.getInstance(context);
+    const s3Params = {
+      Body: fs.createReadStream(archive.zipFilePath),
+      Key: s3Key,
+    };
+    logger('storeCurrentCloudBackend.s3.uploadFile', [{ Key: s3Key }])();
+    await s3Instance.uploadFile(s3Params);
```
jhockett, line 331: I think this might be missing a `.promise()` call.

```suggestion
    await s3Instance.uploadFile(s3Params).promise();
```

jhockett: Disregard, not needed because the underlying class handles this.

jhockett (approved):

PR #11565, opened 2022-12-08
fix: enforce cli-internal and cli-npm package are always on the same …

goldbez:
…version

#### Description of changes
I added logic to the publish.sh script that checks the versions of the cli-internal and cli-npm packages in their respective package.json files. If these versions do not match, then the script fails before anything is published to verdaccio.

I ran the added block of code on from the .circleci directory with matching versions and non-matching versions in the package.json files and ensured that it worked as intended.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -186,11 +186,16 @@ jobs:
           key: amplify-cli-repo-{{ .Branch }}-{{ .Revision }}
       - restore_cache:
           key: amplify-cli-yarn-deps-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      - restore_cache:
+          key: amplify-verdaccio-cache-{{ .Branch }}-{{ .Revision }}
       - run:
           name: verify versions match
           command: |
             source .circleci/local_publish_helpers.sh
-            verifyVersionsMatch
+            startLocalRegistry "$(pwd)/.circleci/verdaccio.yaml"
+            setNpmRegistryUrlToLocal
+            changeNpmGlobalPath
+            checkPackageVersionsInLocalNpmRegistry
 
   mock_e2e_tests:
     <<: *linux-e2e-executor-large
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -179,6 +179,19 @@ jobs:
           command: |
             yarn verify-api-extract
 
+  verify-versions-match:
+    <<: *linux-e2e-executor-medium
+    steps:
+      - restore_cache:
+          key: amplify-cli-repo-{{ .Branch }}-{{ .Revision }}
+      - restore_cache:
+          key: amplify-cli-yarn-deps-{{ .Branch }}-{{ checksum "yarn.lock" }}
+      - run:
+          name: verify versions match
+          command: |
+            source .circleci/local_publish_helpers.sh
+            verifyVersionsMatch
+
   mock_e2e_tests:
     <<: *linux-e2e-executor-large
     steps:
@@ -874,6 +887,16 @@ workflows:
                 - /release_rc\/.*/
                 - /tagged-release-without-e2e-tests\/.*/
       - verify-api-extract:
+          requires:
+            - build
+          filters:
+            branches:
+              ignore:
+                - beta
+                - release
+                - /release_rc\/.*/
+                - /tagged-release-without-e2e-tests\/.*/
+      - verify-versions-match:
           requires:
             - build
           filters:
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -898,14 +898,7 @@ workflows:
                 - /tagged-release-without-e2e-tests\/.*/
       - verify-versions-match:
           requires:
-            - build
-          filters:
-            branches:
-              ignore:
-                - beta
-                - release
-                - /release_rc\/.*/
-                - /tagged-release-without-e2e-tests\/.*/
+            - publish_to_local_registry
       - test:
           requires:
             - build
--- a/.circleci/local_publish_helpers.sh
+++ b/.circleci/local_publish_helpers.sh
@@ -280,17 +280,16 @@ function runE2eTest {
     fi
 }
 
-function verifyVersionsMatch {
-  internal_cli_package_json_path="./packages/amplify-cli/package.json"
-  npm_cli_package_json_path="./packages/amplify-cli-npm/package.json"
+function checkPackageVersionsInLocalNpmRegistry {
+    cli_internal_version=$(npm view @aws-amplify/cli-internal version)
+    cli_version=$(npm view @aws-amplify/cli version)
 
-  internal_cli_version=$(jq '.version' $internal_cli_package_json_path)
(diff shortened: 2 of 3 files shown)


.circleci/publish.sh
```diff
@@ -85,6 +85,19 @@ elif [[ "$CIRCLE_BRANCH" =~ ^run-e2e-with-rc\/.* ]] || [[ "$CIRCLE_BRANCH" =~ ^r
...
+  internal_cli_version=$(jq '.version' $internal_cli_package_json_path)
+  npm_cli_version=$(jq '.version' $npm_cli_package_json_path)
+
+  if [[ $internal_cli_version != $npm_cli_version ]]; then
+    echo "Versions did not match."
+    echo "amplify-cli version: $internal_cli_version"
+    echo "amplify-cli-npm version: $npm_cli_version"
+    exit 1
+  fi
+
```
sobolk, line 100: Does publish run in PR validation workflow ?

sobolk: Ok I see this here https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/13340/workflows/5b8e6c6d-0cf5-4cdf-a5bc-337b33d80cba/jobs/582609 .

My suggestion is to extract this code to separate function (and perhaps move to local_publish_helpers).
Then add explicit step to PR job that validates this. Adding this here makes it hard to discover and reason about for future readers.

awsluja: is the pathing correct? looks like its not finding the file:

image

goldbez: The path should be `./packages/amplify-cli/package.json`

goldbez: @sobolk I will add an explicit step in CircleCI

.circleci/config.base.yml
```diff
@@ -883,6 +896,16 @@ workflows:
...
+      - verify-versions-match:
+          requires:
+            - build
+          filters:
+            branches:
+              ignore:
+                - beta
+                - release
+                - /release_rc\/.*/
+                - /tagged-release-without-e2e-tests\/.*/
```
sobolk, line 908: couple of things here.
1. I'd remove filters. we want that validation everywhere.
2. we should run this after publish_to_local_registry, i.e. after lerna bumps versions - we want to verify that this process happens correctly.

.circleci/local_publish_helpers.sh
```diff
@@ -279,3 +279,18 @@ function runE2eTest {
         yarn run e2e --force-exit --detectOpenHandles --maxWorkers=3 $TEST_SUITE
     fi
 }
+
+function verifyVersionsMatch {
+  internal_cli_package_json_path="./packages/amplify-cli/package.json"
+  npm_cli_package_json_path="./packages/amplify-cli-npm/package.json"
```
sobolk, line 285: when lerna bumps versions, does it modify these paths?

.circleci/config.base.yml
goldbez: Ok. I will revise. Thank you.

.circleci/local_publish_helpers.sh
```diff
@@ -280,17 +280,16 @@ function runE2eTest {
...
+    cli_version=$(npm view @aws-amplify/cli version)
+
+    if [[ $cli_internal_version != $cli_version ]]; then
+        echo "Versions did not match."
+        echo "@aws-amplify/cli-internal version: $cli_internal_version"
+        echo "@aws-amplify/cli version: $cli_version"
+        exit 1
+    else
+        echo "Versions matched ($cli_internal_version)"
+    fi
```
sobolk, line 294: I suggest to print both versions to console regardless of this if/else condition. So that it's known what this script compares when reading circle ci output.

```suggestion
    cli_internal_version=$(npm view @aws-amplify/cli-internal version)
    cli_version=$(npm view @aws-amplify/cli version)
    echo "@aws-amplify/cli-internal version: $cli_internal_version"
    echo "@aws-amplify/cli version: $cli_version"

    if [[ $cli_internal_version != $cli_version ]]; then
        echo "Versions did not match."
        exit 1
    else
        echo "Versions matched ($cli_internal_version)"
    fi
```

sobolk (approved):

sobolk (approved):

.circleci/local_publish_helpers.sh
```diff
@@ -279,3 +279,18 @@ function runE2eTest {
...
+    cli_version=$(npm view @aws-amplify/cli version)
+
+    echo "@aws-amplify/cli-internal version: $cli_internal_version"
+    echo "@aws-amplify/cli version: $cli_version"
+
+    if [[ $cli_internal_version != $cli_version ]]; then
+        echo "Versions did not match."
+        exit 1
+    else
+        echo "Versions matched."
```
awsluja, line 294: looks good, but usually if we fail we provide resolution steps if possible, are there any resolution steps available?

akshbhu: I agree with @awsluja . 
Well one way is to update this file https://github.com/aws-amplify/amplify-cli/blob/2b81f92bb5d53b4da1705c47fdf8035c26be3caa/packages/amplify-cli-npm/index.ts#L19

Can we have some logic in ```cli-npm```  package that can fetch correct hash of the cli-internal package published ?

sobolk: The scope of this PR is to catch mismatch in PR checks, i.e. shorten feedback loop. not to develop fully automatic solution which is more expensive.

The resolution step today is to create PR that has proper conventional commit message and touches either cli or cli-internal depending on which was left behind.

akshbhu (approved):

.circleci/local_publish_helpers.sh
akshbhu: We can update this message in script

awsluja (approved):
nice work!

.circleci/local_publish_helpers.sh
```diff
@@ -279,3 +279,19 @@ function runE2eTest {
...
+function checkPackageVersionsInLocalNpmRegistry {
+    cli_internal_version=$(npm view @aws-amplify/cli-internal version)
+    cli_version=$(npm view @aws-amplify/cli version)
+
+    echo "@aws-amplify/cli-internal version: $cli_internal_version"
+    echo "@aws-amplify/cli version: $cli_version"
+
+    if [[ $cli_internal_version != $cli_version ]]; then
+        echo "Versions did not match."
+        echo "Manual fix: add a commit with a 'feat' prefix that touches the cli-npm package to bump its version."
```
sobolk, line 292: There are edge cases there where somebody could have used `feat!` in cli-internal. 
```suggestion
        echo "Manual fix: add a proper conventional commit that touches the amplify-cli-npm package to correct its version. For example https://github.com/aws-amplify/amplify-cli/commit/6f14792d1db424aa428ec4836fed7d6dd5cccfd0"
```

sobolk (approved):

PR #11566, opened 2022-12-08
chore: feature flag for Lazy Loading and Custom Selection Set support

lawmicha:
#### Table of contents
- Custom Selection Set -> https://github.com/aws-amplify/amplify-swift/pull/2526
- LazyReference -> https://github.com/aws-amplify/amplify-swift/pull/2583
- Codegen -> https://github.com/aws-amplify/amplify-codegen/pull/504 
- CLI feature flag -> https://github.com/aws-amplify/amplify-cli/pull/11566  (You are here)
- Docs --> https://github.com/aws-amplify/docs/pull/4934

#### Description of changes
This PR adds a new feature flag used by codegen to generate iOS models containing **LazyReference** usage and **ModelPath** extension. Related `amplify-codegen` PR: https://github.com/aws-amplify/amplify-codegen/pull/504

This feature flag is needed to allow developers a controlled and sequenced upgrade of the Amplify CLI and Amplify Swift library. The latest Amplify Swift 2.x.x library introduces the definition of the new types generated by the codegen process, namely `LazyReference` and `ModelPath`. 

This feature flag can also be reused if Android requires a feature flag for the same feature. See Android implementation details and manual codegenerated model changes here https://github.com/aws-amplify/amplify-android/pull/2171

#### Existing CLI project / Amplify Swift Library 2.x.x
A developer with an existing Amplify CLI project to upgrade the CLI to take on the latest features, fixes, security updates, and allowing them to perform schema upgrades, without taking on this feature yet. They can, for example, add a new model, remove a model, changes model fields, etc using the latest CLI/codegen version without having to install the latest Amplify Swift library and without having to make app code changes due related to the feature, because the feature flag does not exist in the `cli.json` and **the library is backwards compatible with previous codegenerated swift models/schemas**. This allows the developer a sequenced upgrade by first reviewing the migration guide (TBD link), adding and setting the feature flag to `true`, installing the latest Amplify Swift library, and making the code changes to migrate the code to use the new features. Without this feature flag, developers have to immediately install the latest Amplify Swift library or roll back to a previous codegen version (DX and process unknown to me, whether this is possible at all).

#### Existing CLI project / Amplify Swift Library 1.x.x
The feature flag is even more important in this use case since it will be a much larger API breaking change to go from Amplify Swift Library 1.x.x to 2.x.x as all APIs have been replaced with the latest Swift structured concurrency Async APIs. Existing developers can perform the sequenced upgrade by 
1. Upgrading CLI to the latest, in which this feature flag will exist
2. Upgrading Amplify Swift 1.x.x to Amplify Swift 2.x.x and make the necessary call pattern changes to use Async Await APis
3. Add the feature flag `generateModelsForLazyLoadAndCustomSelectionSet ` to the `cli.json` and set it to `true`.
4. Run `amplify codegen models` to generate the latest types
5. Use the migration guide as guidance for making the code changes to existing use cases and to enable new use cases.

#### New CLI project / Amplify Swift Library 2.x.x
This is the "just works" DX where developers create a new CLI project, where the feature flag will be `true` by default for new projects, and they install the Amplify Swift library's latest version of 2.x.x. The Amplify documentation provides guidance and tutorials of the features available in the latest codegen and library.

#### New CLI project / Amplify Swift Library 1.x.x
In some cases, some developers may decide to use the older Amplify Swift Library if there are some features not yet available (such as Predictions, or a particular not-yet-supported auth flow). These developers will have to actively toggle the feature flag to false to generate the swift models and schema files that work with the older Amplify Swift Library.

#### Release sequence
1. The CLI feature flag can be released first, with the "default for new projects" false to make the flag available for use by codegen code. 
2. Amplify library changes next, to include the new type definitions used by codegen. The library code is backwards compatible with previous codegen files. Developers can install the latest library and everything continues to work.
3. The codegen changes should be released, thus if feature flag is set to "true", then it begin generating the new types.
4. The CLI feature flag can be released with the final desired state of "default for new projects" to true, thus new projects will work automatically with 2.x.x library, and require explicit flip of the switch when using 1.x.x library.

**NOTE**: Since this feature flag is shared across frontend platforms, the feature flag can only be set to `true` for default new projects once all platforms that require this feature flag have been released with the feature.

#### JS lazy loading feature and alternative considerations
JS recently released lazy loading in JS 5.x.x (https://aws.amazon.com/blogs/mobile/new-lazy-loading-nested-query-predicates-for-aws-amplify-datastore/) without a feature flag in the CLI/codegen. This is possible because the latest JS codegen works with the previous JS library (4.x.x). This is not possible for iOS due to its codegen process slightly differs from JS. JS's codegen process generates the types at buildtime and constructs the models based on the JSON at runtime. iOS generates the model types directly and so the latest iOS codegen cannot work with previous Amplify library version, because those type definitions (LazyReference/ModelPath) simply does not exist.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


AaronZyLee (dismissed):

phani-srikar (approved):

AaronZyLee (approved):

alharris-at:
Is this intended to be for iOS only, or used for all native platforms?

lawmicha:
> Is this intended to be for iOS only, or used for all native platforms?

This was originally specific to iOS, and after further discussion with @phani-srikar , we renamed it from `generateLazyReferenceAndModelPath` to `generateModelsForLazyLoadAndCustomSelectionSet` and is intended to be reused by Android as well. As Android is not yet supported, we can call that out in the documentation for the FF that it currently only supports iOS frontends. I imagine Flutter can also reuse this feature flag if circumstances require them to, since the rename makes it more common to all frontends in codegen

alharris-at:
> Hi @alharris-at, this was originally specific to iOS, and after further discussion with @phani-srikar , we renamed it from generateLazyReferenceAndModelPath to generateModelsForLazyLoadAndCustomSelectionSet and is intended to be reused by Android as well. As it is not yet supported for Android yet, we can call that out in the documentation for the FF that it currently only supports iOS frontends. I imagine Flutter can also reuse this feature flag if circumstances require them to, since the rename makes it more common to all frontends in codegen

As long we leave the flag defaulting to false, I'm okay with that, but the risk it introduces is that the flag will be defaulting to true for new projects when we enable for ios, then when we add codegen functionality for android it'll be a breaking change for customers with old projects using android (since they're not opting in, their types will just break unless they also update the runtime lib, or their callsite code, or whatever the impact of breaking here is).

lawmicha:
> As long we leave the flag defaulting to false, I'm okay with that, but the risk it introduces is that the flag will be defaulting to true for new projects when we enable for ios, then when we add codegen functionality for android it'll be a breaking change for customers with old projects using android (since they're not opting in, their types will just break unless they also update the runtime lib, or their callsite code, or whatever the impact of breaking here is).

@alharris-at, that sounds good. I added a note over on the Android PR (https://github.com/aws-amplify/amplify-android/pull/2171) that this feature flag can be used for gating the codegen changes. I believe a shared flag is the right way to go, which will reduce polluting the number of feature flags available. We will enable to true once the feature exists on all platforms (that requires using the feature flag).

Added note to the PR description:
> NOTE: Since this feature flag is shared across frontend platforms, the feature flag can only be set to true for default new projects once all platforms that require this feature flag have been released with the feature.

lazpavel (approved):

PR #11569, opened 2022-12-08
feat: use .NET 6 in lambdas

sobolk:
#### Description of changes

This PR bumps .NET to 6 is lambda runtime and templates.
Existing .NET Core 3.1 functions keep working and cli emits warning.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


jhockett (dismissed):

packages/amplify-dotnet-function-runtime-provider/amplify-plugin.json
```diff
@@ -11,8 +11,8 @@
     },
     "runtimes": [
       {
-        "name": ".NET Core 3.1",
-        "value": "dotnetcore3.1"
+        "name": ".NET 6",
```
akshbhu, line 14: Can we support both runtimes and can detect runtime based on Cx project ?

Also we can warn to Cx that ```NET Core 3.1``` is getting deprecated.

sobolk: How about this ?
image

sobolk: btw. this `amplify-plugin.json`'s `runtimes` section doesn't seem to be used for anything. I'd keep just .net6 here and not advertise deprecated tech. @jhockett @akshbhu do you see any reasons we should keep net core here? we're removing ability to generate new functons from netcore templates.

akshbhu: yeah I agree , it's getting deprecated , we shouldn't allow to create ```dotnetcore31 ``` runtime.

sobolk: rebased to dev and added migration test.

jhockett (approved):

akshbhu (approved):

PR #11574, opened 2022-12-09
chore: delegate cleanup script

awsluja:
#### Description of changes
Currently, our e2e-cleanup script that runs after each e2e workflow takes about 10-20 minutes to run, but we can't trigger a new job or re-run without cancelling the entire workflow.
Cancelling the workflow prevents the cleanup script from running to completion, and we don't want to do that.

As shown below, 95% of cleanup jobs run 36 minutes or less (P95 metric), this will bring the P95 down to ~40 seconds.
image

This change **delegates the cleanup step to a separate job, that can run independent of the E2E run**, so that we can allocate time to our cleanup script without preventing re-runs of E2E.
This will also allow us to expand on our cleanup script without compromising due to how long it takes to run the cleanup.

Additionally, **tests that don't run on windows no longer depend on the windows job**. This allows linux tests to start whenever the binaries are ready without having to wait for the windows build to finish. This should help reduce hotspots/throttling during the start of testing.

Here is the update workflow:
image

That triggers the e2e_resource_cleanup job:
image

Here is an example of being able to re-run while the e2e cleanup is still running for the last workflow:
image

Additional details on how this works:
1. The trigger is an API call to CircleCI that triggers the pipeline with some parameters.
2. The pipeline reads those parameters, including the workflow ID, and can delete resources specific to that workflow.
3. The Token used for triggering the pipeline has been added to the cleanup e2e context, this can be any personal CCI token; it will not work with the current 'project' token.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -1245,7 +1245,7 @@ commands:
     steps:
       - run:
           name: Scan E2E artifacts
-          no_output_timeout: 90m
+          no_output_timeout: 2m
           command: |
             if ! yarn ts-node .circleci/scan_artifacts.ts; then
               echo "Cleaning the repository"
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -9,9 +9,6 @@ parameters:
   nightly_console_integration_tests:
     type: boolean
     default: false
-  e2e_resource_cleanup:
-    type: boolean
-    default: false
   setup:
     type: boolean
     default: true
@@ -70,14 +67,6 @@ defaults: &defaults
       type: executor
       default: l_medium
   executor: << parameters.os >>
-
-clean_e2e_resources: &clean_e2e_resources
-  name: Cleanup resources
-  command: |
-    pwd
-    cd packages/amplify-e2e-tests
-    yarn clean-e2e-resources job ${CIRCLE_BUILD_NUM}
-  working_directory: ~/repo
 
 scan_e2e_test_artifacts: &scan_e2e_test_artifacts
   name: Scan And Cleanup E2E Test Artifacts
@@ -461,8 +450,6 @@ jobs:
           path: ~/repo/packages/amplify-e2e-tests/
       - store_artifacts:
           path: ~/repo/packages/amplify-e2e-tests/amplify-e2e-reports
-      # - clean_e2e_resources:
-      #     os: << parameters.os >>
 
   amplify_migration_tests_v5:
     parameters:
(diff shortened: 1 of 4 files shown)


.circleci/config.base.yml
```diff
@@ -819,23 +806,18 @@ jobs:
...
-      - run: *scan_e2e_test_artifacts
-      - store_artifacts:
-          path: ~/repo/packages/amplify-e2e-tests/amplify-e2e-reports
-    working_directory: ~/repo
+            echo "Triggering cleanup for Workflow $CIRCLE_WORKFLOW_ID"
+            curl --request POST \
+              --url https://circleci.com/api/v2/project/gh/aws-amplify/amplify-cli/pipeline \
+              --header "Circle-Token: $AMPLIFY_CLI_CIRCLECI_TRIGGER_TOKEN" \
+              --header "content-type: application/json" \
+              --data '{"branch":"'$CIRCLE_BRANCH'","parameters":{"e2e_workflow_cleanup":true,"setup":false,"e2e_workflow_cleanup_workflow_id":"'$CIRCLE_WORKFLOW_ID'"}}'
```
sobolk, line 820: could you check if this step fails if HTTP Response is not successful ?

sobolk (approved):
very nice.

.circleci/config.base.yml
awsluja: I can follow up with different behaviors on this; I don't think it should be blocking or fail the build, since this step is kind of an after-effect, but we should be alerted to fix this issue somehow without marking the builds as 'failed'. I can do a follow up

awsluja: here is what happens if you get an error response (forced it)
image

awsluja: fwiw, this should only fail if circleci is having issues or if their API changes, but they have versioned APIs so we should be good

sobolk: the reason I'm asking is that we should be warned if this integration stops working for some reason.
looks like `--fail-with-body` might be cheap solution to this problem.
see https://superuser.com/questions/590099/can-i-make-curl-fail-with-an-exitcode-different-than-0-if-the-http-status-code-i

.circleci/config.base.yml
```diff
@@ -1263,25 +1245,11 @@ commands:
     steps:
       - run:
           name: Scan E2E artifacts
-          no_output_timeout: 90m
+          no_output_timeout: 2m
```
evcodes, line 1248: will this scan always run in under 2m? 90m -> 2m seems like a big jump

evcodes (approved):
one question but not a blocker.

PR #11610, opened 2022-12-14
fix: show warning then delete detached forms

letsbelopez:
fix: delete detached forms and show warning

#### Description of changes

On a pre-push show a warning that the detached forms will be deleted. Delete the forms on a post-push event.

#### Issue #, if available

https://app.asana.com/0/1202185248783861/1203039443118078/f

1. Add a custom field to a form in Studio Form Builder based on an existing data model.
2. Rename the model in the schema.graphql file.
3. Do an `amplify push`.
4. Verify a warning shows that the form with the custom field will be deleted. See attached image.
5. Verify the push.
6. Post push the form is deleted.

And I added/updated unit tests.

Screen Shot 2022-12-14 at 10 31 11 AM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-util-uibuilder/src/__tests__/generateComponents.test.ts
```diff
@@ -37,15 +40,23 @@ describe('can generate components', () => {
...
       ],
     };
     mockedExport = jest.fn().mockReturnValue({
       entities: schemas.entities,
     });
-    awsMock.AmplifyUIBuilder = jest.fn().mockReturnValue({
-      exportComponents: jest.fn().mockReturnValue({
+    mockedDeleteForm = jest.fn(() => true);
+    awsMock.AmplifyUIBuilder = jest.fn(() => ({
+      exportComponents: jest.fn(() => ({
```
jhockett, line 57: This is fine as is, but as an FYI `jest.fn().mockResolvedValue` is helpful syntactic sugar: https://jestjs.io/docs/mock-function-api#mockfnmockresolvedvaluevalue

packages/amplify-util-uibuilder/src/__tests__/generateComponents.test.ts
```diff
@@ -63,29 +77,29 @@ describe('can generate components', () => {
...
-    utilsMock.getAmplifyDataSchema = jest.fn().mockReturnValue(undefined);
-    utilsMock.generateAmplifyUiBuilderIndexFile = jest.fn().mockReturnValue(true);
-    utilsMock.generateAmplifyUiBuilderUtilFile = jest.fn().mockReturnValue(true);
+    }));
+    utilsMock.generateUiBuilderComponents = jest.fn().mockImplementation(() => schemas.entities);
+    utilsMock.generateUiBuilderThemes = jest.fn().mockImplementation(() => schemas.entities);
+    utilsMock.generateUiBuilderForms = jest.fn().mockImplementation(() => schemas.entities);
+    utilsMock.getAmplifyDataSchema = jest.fn().mockImplementation(() => undefined);
+    utilsMock.generateAmplifyUiBuilderIndexFile = jest.fn().mockImplementation(() => true);
+    utilsMock.generateAmplifyUiBuilderUtilFile = jest.fn().mockImplementation(() => true);
```
jhockett, line 86: `jest.mockReturnValue` should be the same as passing an arrow function to `jest.mockImplementation` as far as I know

packages/amplify-util-uibuilder/src/clients/amplify-studio-client.ts
```diff
@@ -77,7 +77,7 @@ export default class AmplifyStudioClient {
   /**
    * Used to configure the AWS Amplify clients.
    */
-  static async setClientInfo(context: $TSContext, envName?: string, appId?: string): Promise<AmplifyStudioClient> {
+  static async setClientInfo(context: $TSContext, envName?: string, appId?: string): Promise<AmplifyStudioClient & {isSupportedGraphQL?: boolean}> {
```
jhockett, line 80: Can the type for `AmplifyStudioClient` be extended? Because you're directly attaching this key value pair to the client object. For this type to be accurate, I'd expect a return below like:

```ts
    let isGraphQLSupported = false;
    if ((await getTransformerVersion()) === 2) {
      await client.loadMetadata();
      isGraphQLSupported = true;
    }
    return { ...client, isGraphQLSupported };
```

Side note, this method should really not be called `setClientInfo` if it both configures and returns a client, but it doesn't need to be addressed in this PR.

packages/amplify-util-uibuilder/src/utils/printDetachedFormsWarning.ts
```diff
@@ -0,0 +1,54 @@
+
+
+import { $TSContext } from 'amplify-cli-core';
```
jhockett, line 3:
```suggestion
import { $TSContext } from 'amplify-cli-core';
```

packages/amplify-util-uibuilder/src/utils/printDetachedFormsWarning.ts
```diff
@@ -0,0 +1,54 @@
...
+  }>, Promise<Schema>]>([
+    studioClient.listForms(),
+    context.amplify.invokePluginMethod(
+      context,
+      'codegen',
+      undefined,
+      'getModelIntrospection',
+      [context],
+    ),
+  ]);
```
jhockett, line 30: This is a lot of heavy lifting for a function called `printDetachedFormsWarning`. Can this be broken into separate functions?

jhockett (changes requested):

packages/amplify-util-uibuilder/src/utils/printDetachedFormsWarning.ts
letsbelopez: wow how did I miss this.

packages/amplify-util-uibuilder/src/utils/printDetachedFormsWarning.ts
letsbelopez: What if I rename it to handlePrePush?

packages/amplify-util-uibuilder/src/clients/amplify-studio-client.ts
letsbelopez: I added this because of a TS error. Not sure how it wasn't added before.
The question mark makes it optional, which works for the current implementation.

letsbelopez: Actually I'm able to resolve this another way.

.eslint-dictionary.json
```diff
@@ -89,6 +90,7 @@
   "createcipheriv",
   "creds",
   "cron",
+  "customizations",
```
jhockett, line 93: This is not an English word, can `customization` be used instead?

packages/amplify-util-uibuilder/src/utils/printDetachedFormsWarning.ts
letsbelopez: Is it just the name that's an issue? As of right now, the only thing this function does is print the warning. What other helper functions were you thinking? We have them for getting the forms, schema, isFormDetachedFromModel and isFormSchemaCustomized. Trying to think of another helper that wouldn't just be a thin wrapper...

jhockett: Sorry for not being clearer in my feedback. I was thinking of separating the printing logic into it's own functions that takes `formSchemas` as a parameter.

Something like:
```ts
export const prePushHandler = async (context: $TSContext): Promise<void> => {
  if (!(await shouldRenderComponents(context))) {
    return;
  }
  try {
    const studioClient = await AmplifyStudioClient.setClientInfo(context);
    const [formSchemas, localSchema] = await Promise.all<[Promise<{
      entities: AmplifyUIBuilder.Form[];
  }>, Promise<Schema>]>([
    studioClient.listForms(),
    context.amplify.invokePluginMethod(
      context,
      'codegen',
      undefined,
      'getModelIntrospection',
      [context],
    ),
  ]);

    if (!localSchema) {
      printer.debug('Local schema not found');
      return;
    }
    printDetachedFormsWarning(formSchema, localSchema)
  } catch (error) {
    printer.debug(error);
  }
}

const printDetachedFormsWarning = (formSchema: SchemaType, localSchema: SchemaType): void => {
  const modelNames = new Set(Object.keys(localSchema.models));
  const detachedCustomForms: string[] = [];

  formSchemas.entities.forEach(form => {
    // Only show warnings for schema's that have customizations, otherwise it's the same as
    // an autogen form.
    if (isFormDetachedFromModel(form, modelNames) && isFormSchemaCustomized(form)) {
      detachedCustomForms.push(form.name);
    }
  });

  if (detachedCustomForms.length > 0) {
    printer.warn(`The following form${detachedCustomForms.length === 1 ? '' : 's'} will no longer be available because the connected data model no longer exists: ${detachedCustomForms.join(', ')}`);
  }
};
```

packages/amplify-util-uibuilder/src/commands/utils/detachedFormCleanup.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  // Check style
+  if (!isEmpty(style)) {
+    // Can be empty but each style can be undefined
+    for (const styleConfig of Object.values(style)) {
+      if (styleConfig !== undefined) return true;
+    }
+  }
+
+  return false;
```
jhockett, line 34: nit: I prefer using [Arrays.prototype.some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some) here, but not a blocker

```suggestion
  if (isEmpty(style)) {
    return true; // Can be empty
  }
  // If not empty, at least one style must be defined
  return Object.values(style).some(styleConfig => styleConfig !== undefined);
```

.eslint-dictionary.json
letsbelopez: Updated

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
```diff
@@ -44,22 +46,48 @@ export const run = async (context: $TSContext): Promise<void> => {
...
     };
 
     const successfulSchemas: StudioSchema[] = [];
     let hasSuccessfulForm = false;
     const failedResponseNames: string[] = [];
+    const modelNames = dataSchema?.models ? new Set(Object.keys(dataSchema.models)) : new Set<string>();
 
     Object.entries(generatedResults).forEach(([key, results]) => {
-      results.forEach(result => {
+      results.forEach(async result => {
```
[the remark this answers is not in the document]
letsbelopez: updated. moved async stuff to its own function.

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
```diff
@@ -13,12 +13,15 @@ import {
...
+  isStudioForm,
+  isFormDetachedFromModel,
+  deleteDetachedForms,
 } from './utils';
 
 /**
  * Pulls ui components from Studio backend and generates the code in the user's file system
  */
-export const run = async (context: $TSContext): Promise<void> => {
+export const run = async (context: $TSContext, eventType: 'PostPush' | 'PostPull'): Promise<void> => {
```
evcodes, line 24: I think eventType should be its own type and pull 'PostPush' | 'PostPull' out of the parameters. This will allow for more extensibility when adding other event types in the future such as PrePush, PrePull, PostInit, and others.

packages/amplify-util-uibuilder/src/event-handlers/handle-PrePush.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * The code to run before a push.
+ */
+export const run = async (context: $TSContext): Promise<void> => {
+  try {
+    await prePushHandler(context);
+  } catch (e) {
+    // Swallow all errors from the uibuilder plugin
+    printer.debug(e);
+  }
```
evcodes, line 14: Along the same lines as my previous comment, there seems to be a lot of duplicated code here and it seems like extensibility would get exponentially more difficult for each event. There's probably an opportunity to dry up some of this code.

packages/amplify-util-uibuilder/src/__tests__/handleEvents.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    context = {
+      amplify: {
+        invokePluginMethod: jest.fn(),
+      },
+    };
+  });
+
+  it('handles postPull', async () => {
+    await runPostPull(context);
+    expect(generateComponents).toHaveBeenCalledWith(context, 'PostPull');
```
evcodes, line 20: When creating the EventType, you can also make POST_PULL a constant instead of a string. This will allow for some nice help from IDE

packages/amplify-util-uibuilder/src/commands/utils/detachedFormCleanup.ts
```diff
@@ -0,0 +1,58 @@
+import { StudioForm, StudioSchema } from '@aws-amplify/codegen-ui';
+import { Form } from 'aws-sdk/clients/amplifyuibuilder';
+import { AmplifyStudioClient } from '../../clients';
+import { printer } from 'amplify-prompts';
+import asyncPool from 'tiny-async-pool';
+
+/**
+ * Does the form's reference DataStore type exist in the list of models
+ */
+export const isFormDetachedFromModel = (schema: StudioForm | Form, models: Set<string>): boolean => {
```
evcodes, line 10: I would also suggest that the models have stricter types instead of allowing a set of Strings and checking in the functions, create a ModelType that models the information better.

packages/amplify-util-uibuilder/src/commands/utils/syncAmplifyUiBuilderComponents.ts
```diff
@@ -1,3 +1,4 @@
+// es-lint-disable-file spellcheck/spell-checker
```
evcodes, line 1: We are actively working on improving our linter. Please do not include this and instead add words into the dictionary here: https://github.com/aws-amplify/amplify-cli/blob/dev/.eslint-dictionary.json

evcodes (changes requested):
Thanks for these changes, I have some comments that I am happy to discuss

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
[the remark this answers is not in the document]
jhockett: My only remaining question is the use of `tiny-async`. I don't have a specific issue with the package, but we try to be conservative about introducing new dependencies. Is `Promise.all` too concurrent for the use-case?

How was `yarn.lock` updated? I see a lot of changes that I wouldn't anticipate given that `tiny-async` doesn't declare any [dependencies](https://github.com/rxaviers/async-pool/blob/master/package.json#L40).

packages/amplify-util-uibuilder/src/commands/utils/detachedFormCleanup.ts
letsbelopez: models are names of models, they can only be strings. If I created a ModelType it would look like 
```
{name: string}
```
is that what you mean? If so, I'm not sure what the benefit is. What's wrong with allowing a set of strings?

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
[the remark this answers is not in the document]
letsbelopez: I used `yarn add`

packages/amplify-util-uibuilder/src/commands/utils/detachedFormCleanup.ts
jhockett: Perhaps a slight change in naming can clear up any confusion. Thoughts on changing `models` => `modelNames`?

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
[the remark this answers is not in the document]
letsbelopez: I tried looking through other packages and I didn't see anything for handling a limited pool concurrent async tasks. We can potentially be deleting 1,000 forms, that's why I wanted to limit the async tasks running at once. The other alternative is using the promise sequential that other packages are using. I can look into that one more.

packages/amplify-util-uibuilder/src/event-handlers/handle-PrePush.ts
jhockett: While I agree with the sentiment, I don't think refactoring all of the event handlers is in-scope for this PR. It's a significant enough refactor to warrant it's own PR.

packages/amplify-util-uibuilder/src/utils/prePushHandler.ts
```diff
@@ -0,0 +1,55 @@
...
+import { isFormDetachedFromModel, isFormSchemaCustomized, shouldRenderComponents } from '../commands/utils';
+
+/**
+ * Handles showing a warning for detached forms pre-push.
+ */
+export const prePushHandler = async (context: $TSContext): Promise<void> => {
+  if (!(await shouldRenderComponents(context))) {
+    return;
+  }
+  try {
```
jhockett, line 15: This try/catch seems redundant with the caller's behavior: https://github.com/aws-amplify/amplify-cli/pull/11610/files/3be18a90df5c548545f8cf96c58faee8d722a1f6#diff-29aed445c842f7a58104e2fc95802749904534fe0680381ce758b1536a34eb36R10

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
jhockett: Again, I agree with the sentiment here, but the events are hooks defined by the platform, so I think this is also out of scope. This is a gap in our own API. I'll create a task to track this in our backlog.

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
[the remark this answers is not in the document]
jhockett: That's okay, promise-sequential is a dependency we'd like to move away from. This use-case makes sense, I just wanted to be sure it was needed. Marking this resolved.

jhockett (approved):

evcodes (approved):
Approved. My comments can be followed up on a separate PR.

PR #11614, opened 2022-12-15
chore: add storage migration test

sobolk:
#### Description of changes

This adds storage migration test, from v10 -> v11.
Test:
- provisions ddb and s3 resources (and auth needed by storage).
- pushes with v10
- pulls with v11
- checks for parameter and cfn template drift
- pushes with v11
-  checks for parameter and cfn template drift

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-migration-tests/src/__tests__/migration_tests_v10/__snapshots__/storage.migration.test.ts.snap
```diff
@@ -0,0 +1,24 @@
...
+[+] Output LambdaExecutionRoleArn: {\\"Value\\":{\\"Fn::GetAtt\\":[\\"LambdaExecutionRole\\",\\"Arn\\"]}}
+
+Resources
+[~] AWS::Lambda::Function UserPoolClientLambda 
+ └─ [~] Runtime
+     ├─ [-] nodejs14.x
+     └─ [+] nodejs16.x
+
+"
+`;
```
awsluja, line 24: is this the output from the CFN diff comparison? 

does this test suggest that these differences are valid (ie, node14 and node16 difference is known , and OK)

And that LamdbaExecutionRoleArn is new in v11?

packages/amplify-migration-tests/tsconfig.tests.json
```diff
@@ -3,6 +3,7 @@
     "target": "es6",
     "module": "commonjs",
     "sourceMap": true,
+    "esModuleInterop": true,
```
akshbhu, line 6: why this one?

packages/amplify-migration-tests/src/migration-helpers/utils.ts
```diff
@@ -8,3 +20,73 @@ export const getShortId = (): string => {
...
+          const cliInputs2 = getCLIInputs(projectRoot2, categoryKey, resourceKey);
+          expect(cliInputs1)
+            .toEqual(cliInputs2);
+        }
+        if (parametersExists(projectRoot1, categoryKey, resourceKey)) {
+          const parameters1 = getParameters(projectRoot1, categoryKey, resourceKey);
+          const parameters2 = getParameters(projectRoot2, categoryKey, resourceKey);
+          expect(parameters1)
+            .toEqual(parameters2);
+        }
```
awsluja, line 46: nice

packages/amplify-migration-tests/src/migration-helpers/utils.ts
```diff
@@ -1,4 +1,16 @@
 import { v4 as uuid } from 'uuid';
+import {
+  cliInputsExists,
+  getBackendConfig,
+  getCLIInputs,
+  getCloudFormationTemplate,
+  getParameters,
+  parametersExists,
+} from '@aws-amplify/amplify-e2e-core';
+import * as cfnDiff from '@aws-cdk/cloudformation-diff';
```
awsluja, line 10: Is the assumption that Cdkv1 and Cdkv2 output is standardized CFN templates - so there is no need to worry that the cfnDiff utility only works when comparing v2 <-> v2 or v1 <-> v1?

I guess what I'm asking is: since this tool is imported from aws-cdk, it works when diffing cdkv1 & cdkv2 CFN templates? Seems like it should, but thats my guess

akshbhu (approved):

packages/amplify-migration-tests/tsconfig.tests.json
sobolk: Compiler said I need that if I want `import strip from 'strip-ansi';` . 
This flag is pretty common https://github.com/aws-amplify/amplify-cli/search?q=%22esModuleInterop%22%3A+true%2C .

packages/amplify-migration-tests/src/__tests__/migration_tests_v10/__snapshots__/storage.migration.test.ts.snap
sobolk: Yes, these are outputs from CFN diff comparison.

The approach I took originally was to assert that cfn diff is empty, which worked well for single ddb. But as I progressed in adding more resources to the mix I realized that this goal is not reasonable, because there are and will be changes that are expected to happen. 
Therefore I changed strategy to what's present in this PR, i.e.:
- dump cfn diff output to snapshot
- inspect diff and check if changes are expected (like nodejs runtime bump) OR not breaking (like that extra output).

Than having snapshot as baseline any new change will show up in PR diff and both author and reviewer can assess if new change is acceptable.

packages/amplify-migration-tests/src/migration-helpers/utils.ts
sobolk: Yes, that's the assumption. CFN template doesn't know which CDK generated it. It has to conform to CFN spec which is not something CDK have control on.

jhockett (approved):
LGTM! When the latest dev commits get merged into this branch, any cdkv2 migration test jobs need to be added to the Windows skip list (until migration tests are fixed for Windows)

awsluja (approved):

sobolk:
> LGTM! When the latest dev commits get merged into this branch, any cdkv2 migration test jobs need to be added to the Windows skip list (until migration tests are fixed for Windows)

Thanks for heads up.

PR #11618, opened 2022-12-15
chore: Handle rejected promises that weren't caught or awaited anywhere in the code

Amplifiyer:
#### Description of changes

As per https://nodejs.org/api/process.html#event-unhandledrejection the event `unhandledRejection` is emitted whenever a Promise is rejected and no error handler is attached to the promise. However the promise can be rejected with just a simple message as well like `Promise.reject("message")` instead of an Error object like `Promise.reject(new Error("message")`.

This PR fixes that such that rejection with any message will yield the right message in `AmplifyError` instead of `UnknownFault`

Unit tests for the new method.

However testing the unhandledRejection event handler is blocked on https://github.com/facebook/jest/issues/5620
#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/amplify-exception-handler.ts
+++ b/packages/amplify-cli/src/amplify-exception-handler.ts
@@ -79,6 +79,21 @@ export const handleException = async (exception: unknown): Promise<void> => {
   process.exit(1);
 };
 
+/**
+ * Handle rejected promises that weren't caught or awaited anywhere in the code.
+ */
+export const handleUnhandledRejection = (reason: Error | $TSAny): void => {
+  if (reason instanceof Error) {
+    throw reason;
+  } else if (reason !== null && typeof reason === 'string') {
+    throw new Error(reason);
+  } else if (reason !== null && typeof reason === 'object') {
+    throw new Error(JSON.stringify(reason));
+  } else {
+    throw new Error('Unhandled promise rejection');
+  }
+};
+
 const getDeepestAmplifyException = (amplifyException: AmplifyException): AmplifyException => {
   let deepestAmplifyException = amplifyException;
   while (deepestAmplifyException.downstreamException && deepestAmplifyException.downstreamException instanceof AmplifyException) {
--- a/packages/amplify-cli/src/amplify-exception-handler.ts
+++ b/packages/amplify-cli/src/amplify-exception-handler.ts
@@ -87,7 +87,7 @@ export const handleUnhandledRejection = (reason: Error | $TSAny): void => {
     throw reason;
   } else if (reason !== null && typeof reason === 'string') {
     throw new Error(reason);
-  } else if (reason !== null && typeof reason === 'object') {
+  } else if (reason !== null) {
     throw new Error(JSON.stringify(reason));
   } else {
     throw new Error('Unhandled promise rejection');
--- a/packages/amplify-cli/src/__tests__/amplify-exception-handler.test.ts
+++ b/packages/amplify-cli/src/__tests__/amplify-exception-handler.test.ts
@@ -2,7 +2,7 @@ import { AmplifyError } from 'amplify-cli-core';
 import { printer } from 'amplify-prompts'; // eslint-disable-line import/no-extraneous-dependencies
 import { reportError } from '../commands/diagnose';
 import { Context } from '../domain/context';
-import { init, handleException } from '../amplify-exception-handler';
+import { init, handleException, handleUnhandledRejection } from '../amplify-exception-handler';
 
 const printerMock = printer as any;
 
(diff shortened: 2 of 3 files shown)


packages/amplify-cli/src/amplify-exception-handler.ts
```diff
@@ -79,6 +79,21 @@ export const handleException = async (exception: unknown): Promise<void> => {
...
+export const handleUnhandledRejection = (reason: Error | $TSAny): void => {
+  if (reason instanceof Error) {
+    throw reason;
+  } else if (reason !== null && typeof reason === 'string') {
+    throw new Error(reason);
+  } else if (reason !== null && typeof reason === 'object') {
+    throw new Error(JSON.stringify(reason));
+  } else {
+    throw new Error('Unhandled promise rejection');
+  }
```
sobolk, line 94: Two things here.

`console.log(typeof 1)` prints `number`. 
I suggest to simplify it and use JSON.stringify if not error and not null. it should handle both objects and primitives.

Should we throw `Error` or `AmplifyError`?

sobolk (commented):
Does this change DX anyhow? If yes then could you add screenshot of console before/after change to description?

packages/amplify-cli/src/amplify-exception-handler.ts
Amplifiyer: 1. This is done to prevent double quotation when printing strings (most common way of rejecting promises). E.g.
```
console.log(JSON.stringify("Message"));  // ""Message""
console.log("Message");                  // "Message"
console.log(JSON.stringify(1));          // "1"
console.log(1);                          // 1
```

2. This gets wrapped in `AmplifyError` by `handleException` which wraps a generic 'Error` to an `UnknownFault` which is the right categorization since we don't really know what caused the error. This effort ensures to capture the promise's rejected message. I'll add a comment here to make it clear.

Amplifiyer:
> Does this change DX anyhow? If yes then could you add screenshot of console before/after change to description?
Not in any substantial way. Instead of getting `Unknown Error` printed to the console, it will now print the message that was given when promise was rejected.

I don't have reproduction steps for this. This change is driven from the error metrics and currently they don't have any information to identify the root cause. This change is to remedy exactly that.

packages/amplify-cli/src/amplify-exception-handler.ts
sobolk: makes sense, but we should still make sure that all primitives (i.e. number) are captured.

sobolk:
> > Does this change DX anyhow? If yes then could you add screenshot of console before/after change to description?
> > Not in any substantial way. Instead of getting `Unknown Error` printed to the console, it will now print the message that was given when promise was rejected.
> 
> I don't have reproduction steps for this. This change is driven from the error metrics and currently they don't have any information to identify the root cause. This change is to remedy exactly that.

have you tried to artificially inject fault like this in local dev? for example in one of the simpler commands. it might not be the exact repro, but might be useful to understand if/how new logic handles it.

Amplifiyer:
> > > Does this change DX anyhow? If yes then could you add screenshot of console before/after change to description?
> > > Not in any substantial way. Instead of getting `Unknown Error` printed to the console, it will now print the message that was given when promise was rejected.
> > 
> > 
> > I don't have reproduction steps for this. This change is driven from the error metrics and currently they don't have any information to identify the root cause. This change is to remedy exactly that.
> 
> have you tried to artificially inject fault like this in local dev? for example in one of the simpler commands. it might not be the exact repro, but might be useful to understand if/how new logic handles it.

Yes! artificially inducing the error (e.g. at the beginning of any CLI command doing `Promise.reject(<String>)` generate the following before/after

Before
 
Screen Shot 2022-12-21 at 5 17 04 PM

After 

Screen Shot 2022-12-21 at 5 16 46 PM

packages/amplify-cli/src/amplify-exception-handler.ts
Amplifiyer: Makes sense. Update to capture everything and added a case for numbers.

sobolk (approved):

sdstolworthy (approved):

packages/amplify-cli/src/amplify-exception-handler.ts
```diff
@@ -79,6 +79,21 @@ export const handleException = async (exception: unknown): Promise<void> => {
   process.exit(1);
 };
 
+/**
+ * Handle rejected promises that weren't caught or awaited anywhere in the code.
+ */
+export const handleUnhandledRejection = (reason: Error | $TSAny): void => {
```
sdstolworthy, line 85: Should we report to analytics here? Or is that handled elsewhere?

Amplifiyer: This is going to analytics! Its handled in 'uncaughtException' node handler.

PR #11629, opened 2022-12-19
feat: added Node package dependency detector

akshbhu:
#### Description of changes

- added class to parse lock files
- detects packages in lock file with a specific dependency
- added warnings if helper packages (@aws-amplify/cli-extensibility-helper and @aws-cdk/core) contains cdkV1 dependency

- added unit tests for detector class
- unit tests for warning messages
- cloud e2e : https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/15059/workflows/d50e1d73-4c09-429e-9962-33dad8b072aa

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/amplify-cli-core/package.json
```diff
@@ -26,6 +26,7 @@
   },
   "dependencies": {
     "@aws-amplify/graphql-transformer-interfaces": "2.1.0-beta.2",
+    "@yarnpkg/lockfile": "^1.1.0",
```
sobolk, line 29: does this handle both yarn 1 and 2 ?

packages/amplify-cli-core/API.md
```diff
@@ -1187,6 +1223,14 @@ export class PathManager {
...
 export const pathManager: PathManager;
 
+// @public (undocumented)
+export type PkgJsonType = {
+    name: string;
+    version: string;
+    dependencies: Record<string, string>;
+    devDependencies?: Record<string, string>;
+};
+
```
sobolk, line 1233: this shouldn't be exported from cli core.

packages/amplify-cli-core/API.md
```diff
@@ -179,6 +179,42 @@ export enum AmplifyFrontend {
...
+    // (undocumented)
+    packageManager: PackageManager;
+    // Warning: (ae-forgotten-export) The symbol "Lockfile" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    parseLockFile(): Lockfile;
+    // (undocumented)
+    parsePkgJson: (projectRoot: string) => PkgJsonType;
+    // (undocumented)
+    pkgJsonObj: PkgJsonType;
```
sobolk, line 215: all of this should not be public (outside of cli-core).

packages/amplify-cli-core/API.md
```diff
@@ -179,6 +179,42 @@ export enum AmplifyFrontend {
...
+    constructor(amplifyDetectorProps: AmplifyNodeJsDetectorProps);
+    // (undocumented)
+    dependencyToSearch: string;
+    // (undocumented)
+    dependencyVersion: string;
+    // Warning: (ae-forgotten-export) The symbol "YarnLockDependencyType" needs to be exported by the entry point index.d.ts
+    // Warning: (ae-forgotten-export) The symbol "PackageLockDep" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    getDependentPackage(packageName: string): YarnLockDependencyType | PackageLockDep | undefined;
```
sobolk, line 201: like I mentioned in the other comment, package manager detection can be made internal given that `AmplifyNodePkgDetector` knows projectRoot path. When we do that we should change return type of `getDependentPackage` (or whatever comes out from next iteration) to something that just carries detected dependency version and name, i.e. information about package manager should not leak through this abstraction.

packages/amplify-provider-awscloudformation/src/print-cdk-migration-warning.ts
```diff
@@ -0,0 +1,104 @@
...
+export const printCdkMigrationWarning = async (context: $TSContext): Promise<void> => {
+  const resourcesToBuild: IAmplifyResource[] = [];
+  const { allResources } = await context.amplify.getResourceStatus();
+  allResources.forEach(resource => {
+    resourcesToBuild.push({
+      service: resource.service as string,
+      category: resource.category as string,
+      resourceName: resource.resourceName as string,
+    });
+  });
```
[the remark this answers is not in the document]
akshbhu: returns any type. So I added this type

packages/amplify-cli-core/package.json
akshbhu: only yarn 1 . I will check for yarn2 if it is supported by our projects

packages/amplify-cli-core/src/amplify-lockfile-dependency-detector/lock-file-types.ts
```diff
@@ -0,0 +1,17 @@
...
+}
+
+/**
+* package.json type
+*/
+export type PkgJsonType = {
+    name : string,
+    version: string,
+    dependencies: Record<string, string>,
+    devDependencies?: Record<string, string>
```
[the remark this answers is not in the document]
akshbhu: I don't think for amplify we need dev/peer dependencies check.

packages/amplify-cli-core/API.md
```diff
@@ -182,6 +182,25 @@ export enum AmplifyFrontend {
...
+export class AmplifyNodePkgDetector {
+    constructor(amplifyDetectorProps: AmplifyNodePkgDetectorProps);
+    // Warning: (ae-forgotten-export) The symbol "DetectedDependencies" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    detectAffectedDirectDependencies(): Array<DetectedDependencies> | undefined;
+    // Warning: (ae-forgotten-export) The symbol "Lockfile" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    parseLockFile(): Lockfile;
```
sobolk, line 195: this shouldn't be public.

packages/amplify-provider-awscloudformation/src/print-cdk-migration-warning.ts
```diff
@@ -0,0 +1,112 @@
...
+  // check for override.ts file enabled
+  const migrationString = getMigrationMessage(resourcesToBuild);
+  if (!_.isEmpty(migrationString)) {
+    printer.warn(migrationString);
+  }
+};
+
+const getOverridesWarningObj = (resourcesToBuild: IAmplifyResource[]): AmplifyWarning | undefined => {
```
sobolk, line 36: nit: please drop `Obj` suffix from all these methods. We know it's object by looking at return type.

sobolk (commented):
the API looks much cleaner now, thank you.

packages/amplify-cli-core/src/amplify-lockfile-dependency-detector/lock-file-types.ts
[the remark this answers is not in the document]
sobolk: I think we don't need this check for dev dependencies.
But I'm not sure about peer deps. Can you check what happens if you move extensibility helper/cdk from dependencies to peer deps ? I.e. whether override/custom mini project compiles and works as expected or not? If it does then we should check peers.

jhockett (commented):
There's a lot of non-null assertions that should be removed, partly because the new lint rules don't allow them, but also they remove a some of the confidence that `tsc`'s type checking would otherwise provide.

packages/amplify-cli-core/API.md
akshbhu: I have kept this public for 2 reasons:
1) if someone want to use the detector to parse lock files in future
2) helps in testing the detector class for both parsers (though this I can change)

packages/amplify-cli-core/src/amplify-node-pkg-detector/yarn-lock-parser.ts
```diff
@@ -0,0 +1,122 @@
...
+     * traverses dependency tree
+     */
+    private dfs(dependency: string, lockFileDependenciesMap: YarnLock, dependencyToSearch: string): void {
+      const dependencyPkgKey = dependency.substring(0, dependency.lastIndexOf('@'));
+      const dependencyObj = lockFileDependenciesMap.dependencies![dependency];
+      if (!_.isEmpty(dependencyObj) && !_.isEmpty(dependencyObj.dependencies)) {
+        const dependencyObjDeps = dependencyObj.dependencies!;
+        if (!_.isEmpty(dependencyObjDeps)) {
+          for (const nestedDependency of Object.keys(dependencyObjDeps)) {
+            const nestedDependencyActual = this.getDependencyKey(nestedDependency, `${dependencyObjDeps![nestedDependency]}`);
```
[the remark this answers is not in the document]
akshbhu: Its not a child dependency , I am just converting nestedDependency to whats the actual key present in lockfile obj.

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
```diff
@@ -0,0 +1,109 @@
...
+       return this.lockFileParser.parseLockFile(this.lockFileContents);
+     }
+
+     /**
+    * parses package.json project files
+    */
+      private parsePkgJson = (projectRoot: string): PackageJson => {
+        const pkgJsonFullPath = path.resolve(projectRoot, 'package.json');
+        return <PackageJson>JSON.parse(fs.readFileSync(pkgJsonFullPath, 'utf-8'));
+      }
```
Amplifiyer, line 70: Some indentation looks off here

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
```diff
@@ -0,0 +1,109 @@
...
+
+     /**
+      * get file content as string
+      */
+     private getFileContent = (projectRoot: string) : string => {
```
Amplifiyer, line 75:
```suggestion
     private getLockFileContent = (projectRoot: string) : string => {
```

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
```diff
@@ -0,0 +1,109 @@
...
+       const lockFileFullPath = path.resolve(projectRoot, this.packageManager.lockFile);
+       if (!fs.existsSync(lockFileFullPath)) {
+         throw new AmplifyFault('FileNotFoundFault', {
```
Amplifiyer, line 78:
```suggestion
         throw new AmplifyFault('LockFileNotFoundFault', {
```

Should this be error or fault?

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
```diff
@@ -0,0 +1,109 @@
...
+           message: 'Lockfile not found at location: ${lockFileFullPath}',
+         });
+       }
+       return fs.readFileSync(lockFileFullPath, 'utf-8');
+     };
+
+     /**
+     * returns  explicit dependencies from package.json if lock file package depends on passed dependency else undefined
+     */
+     detectAffectedDirectDependencies(): Array<DetectedDependencies> | [] {
```
Amplifiyer, line 88: For consistency, let's use the same style of defining functions
`fnName = () => {}` or `fnName(): {}`

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
```diff
@@ -0,0 +1,109 @@
...
+       if (!fs.existsSync(lockFileFullPath)) {
+         throw new AmplifyFault('FileNotFoundFault', {
+           message: 'Lockfile not found at location: ${lockFileFullPath}',
+         });
+       }
+       return fs.readFileSync(lockFileFullPath, 'utf-8');
+     };
+
+     /**
+     * returns  explicit dependencies from package.json if lock file package depends on passed dependency else undefined
```
Amplifiyer, line 86: There is nothing passed in this method.

On second read, shouldn't `dependencyToSearch` be passed to this method instead of the constructor of the class?

packages/amplify-cli-core/src/amplify-node-pkg-detector/package-lock-parser.ts
```diff
@@ -0,0 +1,106 @@
...
+    dependencies?: PackageLockDependency;
+    dev?: boolean;
+}
+
+/**
+ * package lock parser
+ */
+export class PackageLockParser {
+  type: LockfileType;
+  dependenciesMap:Record<string, Record<string, PackageLockDependencyType>>;
```
Amplifiyer, line 41: why do we need this as a class attribute?

packages/amplify-cli-core/src/amplify-node-pkg-detector/package-lock-parser.ts
```diff
@@ -0,0 +1,106 @@
...
+      return packageLock;
+    } catch (e) {
+      throw new AmplifyFault('LockFileParsingFault', {
+        message: `package-lock.json parsing failed with an error: ${e.message}`,
+      }, e);
+    }
+  }
+
+  /**
+     * getDependentNpmPackage()
```
Amplifiyer, line 64: Can we add more descriptive comments for public methods

packages/amplify-cli-core/src/amplify-node-pkg-detector/package-lock-parser.ts
```diff
@@ -0,0 +1,106 @@
...
+      [depName: string]: string;
+    };
+    dependencies?: PackageLockDependency;
+    dev?: boolean;
+}
+
+/**
+ * package lock parser
+ */
+export class PackageLockParser {
```
Amplifiyer, line 39: Should this class have `NPM` in it to complement `YarnLockParser`

packages/amplify-provider-awscloudformation/src/print-cdk-migration-warning.ts
```diff
@@ -0,0 +1,112 @@
...
+    }
+  });
+  return customResourcesWarningObject;
+};
+
+/**
+ * returns migration message otherwise undefined
+ */
+export const getMigrationMessage = (resourcesToBuild: IAmplifyResource[]): string => {
+  const migrationBanner = `We detect you are using CDK v1 with custom stacks and overrides.AWS CDK v1 has entered maintenance mode on June 1, 2022`;
```
Amplifiyer, line 88:
```suggestion
  const migrationBanner = `We detected that you are using CDK v1 with custom stacks and overrides. AWS CDK v1 has entered maintenance mode on June 1, 2022`;
```

Amplifiyer (commented):
Left some suggestions (don't accept from github since I suggested some interface changes)

packages/amplify-cli-core/src/amplify-node-pkg-detector/package-lock-parser.ts
akshbhu: this stores all packages present in lockfiles which are dependent on `dependencyToSearch`(i.e. @aws-cdk/core in cdk case)

packages/amplify-cli-core/src/amplify-node-pkg-detector/package-lock-parser.ts
akshbhu: Its present in constructor

Amplifiyer: Sorry I meant in the name of the class.

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
```diff
@@ -0,0 +1,105 @@
...
+import { AmplifyFault } from '../errors/amplify-fault';
+import { getPackageManager, PackageManager } from '../utils/packageManager';
+import { Lockfile, LockfileParser } from './lock-file-interface';
+import { PackageJson } from './lock-file-types';
+import { LockFileParserFactory } from './parser-factory';
+
+/**
+ * return type of detectAffectedDirectDependencies
+ */
+export type DetectedDependencies = {
```
Amplifiyer, line 13: Should this be DetectedDependency since it contains a singular package and the array of it would be dependencies?

packages/amplify-cli-core/src/amplify-node-pkg-detector/yarn-lock-parser.ts
```diff
@@ -0,0 +1,125 @@
...
+      } catch (e) {
+        throw new AmplifyFault('LockFileParsingFault', {
+          message: `yarn.lock parsing failed with an error: ${e.message}`,
+        }, e);
+      }
+    }
+
+    /**
+     * this function takes dependencyTo search and lockFile content as input and
+     * returns a map of < allPackages ,dependencyToSearchPayload>
```
Amplifiyer, line 73:
```suggestion
     * returns a map of <allPackages, dependencyToSearchPayload>
```

packages/amplify-e2e-core/src/init/amplifyPush.ts
```diff
@@ -210,6 +210,23 @@ export const amplifyPushAuth = (cwd: string, testingWithLatestCodebase = false):
...
 });
 
+/**
+ * Function to test amplify push with cdk warninng
+ */
+export const amplifyPushWithCDkWarning = (cwd: string, testingWithLatestCodebase = false): Promise<void> => new Promise((resolve, reject) => {
+  spawn(getCLIPath(testingWithLatestCodebase), ['push'], { cwd, stripColors: true, noOutputTimeout: pushTimeoutMS })
+    .wait('Are you sure you want to continue?')
+    .sendConfirmYes()
+    .wait(/We detect you are using CDK v1 with custom stacks and overrides.AWS CDK v1 has entered maintenance mode on June 1, 2022[[:ascii:]]*/)
```
Amplifiyer, line 220: Probably still needs updating?

packages/amplify-cli-core/src/amplify-node-pkg-detector/package-lock-parser.ts
Amplifiyer: I meant why specifically a _class attribute_. This seems like an output field which we can create in `getDependentPackageMap` and return it from there to avoid it getting manipulated by some other method in future.

Amplifiyer (commented):
Some small nits but otherwise looks good.

packages/amplify-cli-core/src/amplify-lockfile-dependency-detector/lock-file-types.ts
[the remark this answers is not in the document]
akshbhu: added yarn2 parser since amplify project was working with yarn2

packages/amplify-cli-core/package.json
akshbhu: added yarn2 parser since amplify project was working with yarn2

packages/amplify-e2e-core/src/init/amplifyPush.ts
akshbhu: its passed as regex

packages/amplify-cli-core/src/amplify-node-pkg-detector/package-lock-parser.ts
akshbhu: I am manipulating this is class methods so I thought to keep as as class private variable.

packages/amplify-cli-core/src/amplify-lockfile-dependency-detector/lock-file-types.ts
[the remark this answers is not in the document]
akshbhu: I checked with adding helper package to peer dependencies:
Screen Shot 2023-01-18 at 2 52 15 PM

amplify build fails with this error:
Screen Shot 2023-01-18 at 2 53 00 PM

Though I have added a check on peer dependencies as well in package.json and  also detector will be able to detect this package as in lock files, this will be added as dependency

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
```diff
@@ -0,0 +1,106 @@
...
+
+     /**
+      * get lock file content as string
+      */
+     private getLockFileContent = (projectRoot: string) : string => {
+       const lockFileFullPath = path.resolve(projectRoot, this.packageManager.lockFile);
+       if (!fs.existsSync(lockFileFullPath)) {
+         throw new AmplifyFault('LockFileNotFoundFault', {
+           message: 'Lockfile not found at location: ${lockFileFullPath}',
+         });
```
Amplifiyer, line 76: Should this be error?

Amplifiyer (approved):
One minor comment, looks good otherwise

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
akshbhu: We do a `npm install / yarn install `before calling this , so lock file should be there if there is `package.json` ( for this we have checks for package managers.)

So I think this should be fault as lock file should be present according to workflow.

Thoughts?

packages/amplify-cli-core/src/amplify-node-pkg-detector/yarn2-lock-parser.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * parseLockFile
+   */
+  public parseLockFile = (lockFileContents: string): YarnLock => {
+    try {
+      const yarn2LockContent = yaml.parse(lockFileContents);
+      delete yarn2LockContent.__metadata;
+      const object: YarnLockDependency = {};
+      _.forEach(yarn2LockContent, (packageValue, packageKey) => {
```
jhockett, line 30: What's the advantage of lodash here instead of a standard forEach loop?

packages/amplify-cli-core/src/state-manager/pathManager.ts
```diff
@@ -66,6 +66,7 @@ export const PathConstants = {
   ExportManifestJsonFilename: 'amplify-export-manifest.json',
   ExportTagsJsonFileName: 'export-tags.json',
   ExportCategoryStackMappingJsonFilename: 'category-stack-mapping.json',
+  OverrideFilePath: 'override.ts',
```
jhockett, line 69:
```suggestion
  OverrideFileName: 'override.ts',
```

packages/amplify-cli-core/src/amplify-node-pkg-detector/amplify-nodejs-detector.ts
```diff
@@ -0,0 +1,106 @@
...
+               dependentPackage: {
+                 name: dependencyToSearch,
+                 version: allPackagesWithDependency[pkg][dependencyToSearch].version,
+
+               },
+             };
+             return obj;
+           }
+           return {};
+         }).filter(value => Object.keys(value).length !== 0);
```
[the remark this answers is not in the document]
akshbhu: Nope , If you check .map() returns a `DetectedDependency[]` type  which will be of type 
```
type DetectedDependency = {
    packageName?: string | undefined;
    dependentPackage?: {
        name: string;
        version: string;
    } | undefined;
}
```
SO either a empty object {} or a detected dependency

packages/amplify-cli-core/src/amplify-node-pkg-detector/yarn2-lock-parser.ts
akshbhu: No advantage as such but one thing is  it implicitly handles the case where `yarn2.lock` is corrupted as we are reading it through yaml parser

packages/amplify-cli-core/API.md
sobolk: Public api (assuming that we treat semver seriously) is an element of a library/package that's hardest to change. Therefore we shouldn't overexpose it if there's no strong case to do that - tests shouldn't be a driver here. Besides using detector to parse doesn't seem like a cohesive component (i.e. it stops being responsible for single thing).

Consider the following:
- the need of using cli core to parse lock files outside of this package is speculative at this time. therefore publishing such interface should be deferred until needed (following YAGNI philosophy).
- if we were to provide lockfile parsing facility publicly, we'd rather publish parsers (most likely in a form of interface and type factory for the interface, but this would need another review and consideration, we don't do this until a need arises).
- parsing looks quite surprising on "detector". the detector might use parser internally through delegation but exposing parse method seems like violation of srp principle here.

packages/amplify-cli-core/API.md
```diff
@@ -981,6 +1004,12 @@ export class JSONUtilities {
     } | undefined) => void;
 }
 
+// Warning: (ae-forgotten-export) The symbol "PackageLock" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "YarnLock" needs to be exported by the entry point index.d.ts
+//
+// @public (undocumented)
+export type Lockfile = PackageLock | YarnLock;
+
```
sobolk, line 1012: 1. see warnings, they say that PackageLock and YarnLock should also be exported.
2. do we need to export this at all? if we hide `AmplifyNodePkgDetector.parseLockFile` then we could hide this type as well.

packages/amplify-cli-core/API.md
```diff
@@ -1042,10 +1071,11 @@ export type PackageManager = {
     packageManager: PackageManagerType;
     lockFile: string;
     executable: string;
+    yarnrc?: string;
```
sobolk, line 1074: if this is a path consider suffixing, i.e. `yarnrcPath`.

sobolk (approved):

packages/amplify-provider-awscloudformation/src/__tests__/print-cdk-migration-warning.test.ts
```diff
@@ -0,0 +1,200 @@
...
+
+const printerMock = printer as jest.Mocked<typeof printer>;
+
+jest.mock("amplify-cli-core", () => ({
+  ...(jest.requireActual("amplify-cli-core") as $TSObject),
+  pathManager: {
+    getBackendDirPath: jest.fn().mockReturnValue("mockDirPath"),
+    getResourceOverrideFilePath: jest.fn().mockReturnValue("mockDirPath/override.ts")
+  },
+  AmplifyCategories: {
```
jhockett, line 22: Wouldn't line 17 take care of the need to defined this?

packages/amplify-provider-awscloudformation/src/print-cdk-migration-warning.ts
```diff
@@ -0,0 +1,111 @@
...
+    printer.warn(migrationString);
+  }
+};
+
+const getOverridesWarning = (resourcesToBuild: IAmplifyResource[], dependencyToSearch: string): AmplifyWarning | undefined => {
+  let overridesWarningObject;
+  for (const resource of resourcesToBuild) {
+    // since overrides project is backend
+    const overrideFilePath = pathManager.getResourceOverrideFilePath(undefined, resource.category, resource.resourceName);
+    // const overrideFilePath = path.join(backendDir, resource.category, resource.resourceName, 'override.ts');
```
jhockett, line 41: nit: remove if not needed

jhockett (approved):

PR #11669, opened 2022-12-27
feat: generate models on pull only when api changes

dpilch:
#### Description of changes

Set model generation to occur on pull only when there are changes to the API.

#### Issue #, if available

N/A

1. Created Amplify project with DataStore
2. Make change to schema on Studio
3. amplify-dev pull
4. See codegen ran in cli output
5. Add auth on Studio (don't modify the api)
6. amplify-dev pull
7. See codegen did not run in cli output

https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli?branch=run-e2e/dppilche/feat/generate-models-on-change

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/amplify-service-helper.ts
```diff
@@ -44,6 +43,16 @@ export const postPullCodegen = async (context: $TSContext): Promise<void> => {
     return;
   }
   if (await isDataStoreEnabled(path.join(pathManager.getBackendDirPath(), 'api', gqlApiName))) {
-    await context.amplify.invokePluginMethod(context, 'codegen', undefined, 'generateModels', [context]);
+    const { resourcesToBeCreated, resourcesToBeUpdated, resourcesToBeSynced, resourcesToBeDeleted } = await getResourceStatus(
+      'api',
```
jhockett, line 47: please use `AmplifyCategories.API` from `amplify-cli-core`

packages/amplify-cli/src/__tests__/commands/env.test.ts
```diff
@@ -1,8 +1,17 @@
...
-    pathManager: { getAmplifyMetaFilePath: jest.fn().mockReturnValue('test_file_does_not_exist') },
+    pathManager: {
+      getAmplifyMetaFilePath: jest.fn().mockReturnValue('test_file_does_not_exist'),
+      getAWSCredentialsFilePath: jest.fn(),
+      getAWSConfigFilePath: jest.fn(),
+    },
+    FeatureFlags: {
+      getBoolean: jest.fn().mockReturnValue(true),
+      getNumber: jest.fn().mockReturnValue(1),
+    },
```
jhockett, line 14: Why did this need to be added? I don't see any additional uses of FeatureFlags in this PR

dpilch: I'm not entirely sure why the changes necessitated this. Leaving this out resulted in the following.

```
amplify env:  › encountered a declaration exception

    FeatureFlags is not initialized

      211 |   private static ensureInitialized = (): void => {
      212 |     if (!FeatureFlags.instance) {
    > 213 |       throw new Error('FeatureFlags is not initialized');
          |             ^
      214 |     }
      215 |   };
      216 |
```

The requireActual and mocks for path manager were also necessary for separate errors.

I did try mocking amplify-service-helper directly, but it did not prevent the errors.
```
jest.mock('../../amplify-service-helper', () => ({
  ...(jest.requireActual('../../amplify-service-helper') as Record<string, never>),
  postPullCodegen: jest.fn(() => Promise.resolve()),      
}));  
```

jhockett (approved):

packages/amplify-cli/src/amplify-service-helper.ts
```diff
@@ -44,6 +43,16 @@ export const postPullCodegen = async (context: $TSContext): Promise<void> => {
...
   if (await isDataStoreEnabled(path.join(pathManager.getBackendDirPath(), 'api', gqlApiName))) {
-    await context.amplify.invokePluginMethod(context, 'codegen', undefined, 'generateModels', [context]);
+    const { resourcesToBeCreated, resourcesToBeUpdated, resourcesToBeSynced, resourcesToBeDeleted } = await getResourceStatus(
+      AmplifyCategories.API,
+      undefined,
+      undefined,
+      undefined,
+    );
+    const apiHasChanges =
+      resourcesToBeCreated.length || resourcesToBeUpdated.length || resourcesToBeSynced.length || resourcesToBeDeleted.length;
```
akshbhu, line 53: Do we need `resourcesToBeDeleted.length` ?
if api is getting deleted, do we have to call `generateModels` ?

dpilch:
I've found that this approach does not work in all cases. I will need to make a new PR.

PR #11685, opened 2022-12-30
ci: e2e multi-core optimizations

awsluja:
#### Description of changes
**Currently, every single test file runs on its own executor.**
On windows, that means we only utilize 1 of 4 cores available.
On linux, that means we only utilize 1 of 2 cores on Medium instances, and 1 of 4 cores on Large instances.

This PR updates our testing framework so that we utilize every core available, or 1 core for the operating system, and the remaining cores are assigned to a worker in the worker pool.
#### We will use this setup moving forward:
**Windows Test with 4 CPUs -> 1 core for the OS, and 3 Workers
Linux Test with 4 CPUs -> 1 core for the OS, and 3 Workers**
Linux Test with 2 CPUs -> 1 core for the OS, and 1 Worker --- we use this setup sparingly (only for tests that don't run well with other tests)

Here is an example run (99.9% success on the first run, or 798 / 801 tests passing):
https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/14459/workflows/04f7137d-cc29-4997-8d2e-4433530d0825
All linux tests pass on the first run, some windows tests (8 test files containing 25 individual tests) have been moved to linux only due to flakiness in favor of more consistent runs.
You can also get an idea of the efficiency of this change here: https://app.circleci.com/insights/github/aws-amplify/amplify-cli/workflows/build_test_deploy_v3/overview?branch=run-e2e%2Faluja%2Fe2e-multi-core-optimizations&reporting-window=last-30-days
image

#### This PR includes several updates that are collectively required to make this work:

1. Update Config files & Test setup files to use Worker Pools (maxWorkers=3, run in parallel)
2. Update split-e2e-tests to assign multiple test files to each executor, making sure to consider Region specific tests, and Parent Account specific tests. Job names should be clear, and user friendly.
3. Use timing data from CircleCI to match tests with similar durations together (this allows us to make efficient use of each worker pool)
4. Update E2E framework to fix problems introduced with concurrency
5. Update Tests as needed to fix problems introduced with concurrency
6. Update E2E artifacts to work with concurrency (artifacts are now prefixed with the test name)
7. Provide a mechanism to run tests independently if they don't pass otherwise (RUN_SOLO), and use minimal resources while doing so (use medium executors on linux for solo tests)

Here are some details on specific changes that were needed to fix problems with concurrency:
1. Upgrade to Jest 27 (and related dependencies)
2. Fix Tests that broke in Jest 27 (only 1, caused by changes to mock timers)
3. Introduce a mechanism (in createProjectDir) that prevents tests from starting at the same time (stagger them to avoid race conditions when 'init' is called for the very first time on a machine)
4. Update tests so that they call createProjectDir first (this was done for several migration tests)

Other work:
1. Refactor CircleCI utilities into cci-utils.ts, cci-types.ts, cci.api.ts, etc
2. Provide a mechanism to keep timing data up-to-date (yarn update-test-timing-data)

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-e2e-core/src/nexpect-reporter.js
```diff
@@ -76,19 +88,24 @@ class AmplifyCLIExecutionReporter {
...
               sanitizedSections.push(section.replace(/[^a-z0-9]/gi, '_').toLowerCase());
             }
             let suffix = sanitizedSections.join('_');
-            if(suffix.length > 30){
-              suffix = suffix.substring(0, 30);
+            if (suffix.length > 20) {
+              suffix = suffix.substring(0, 20);
             }
-            const castFile = `${new Date().getTime()}_${index}_${suffix}.cast`;
+            const castFile = `${testName}_${index}_${suffix}.cast`;
```
awsluja, line 108: This change works much better with concurrent tests - and looks awesome below:
image

packages/amplify-e2e-core/src/utils/add-circleci-tags.ts
```diff
@@ -12,8 +14,13 @@ declare global {
   /* eslint-enable */
 }
 
-export const addCircleCITags = (projectPath: string): void => {
+function delay(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+export const addCircleCITags = async (projectPath: string): Promise<void> => {
   if (process.env && process.env['CIRCLECI']) {
+    await staggerTestsThatRunOnTheSameMachine(projectPath);
```
awsluja, line 22: This is basic round robin scheduling during test startup.

I implemented it using the OS/filesystem because there is no other way to communicate between tests, and running init multiple times concurrently seems to cause a lot of errors with simulators (they apparently create files, and throw error if someone else has created them)

scripts/split-e2e-tests-v2.ts
```diff
@@ -0,0 +1,314 @@
...
+    const baseJob = baseJobs[baseJobName];
+    
+    let testSuites = getTestFiles(jobRootDir);
+    if(pickTests && typeof pickTests === 'function'){
+        testSuites = pickTests(testSuites);
+    }
+    if(testSuites.length === 0){
+        return output;
+    }
+    const testFileRunTimes = loadTestTimings().timingData;
```
awsluja, line 116: this is important because we want similar timed jobs to be together
image

packages/amplify-e2e-core/src/init/deleteProject.ts
```diff
@@ -10,16 +10,19 @@ import { getBackendAmplifyMeta } from '../utils';
...
-    stack => stack.StackStatus.endsWith('_COMPLETE') || stack.StackStatus.endsWith('_FAILED'),
-  );
-
-  const noOutputTimeout = 1000 * 60 * 20; // 20 minutes;
-  return spawn(getCLIPath(usingLatestCodebase), ['delete'], { cwd, stripColors: true, noOutputTimeout })
-    .wait('Are you sure you want to continue?')
-    .sendConfirmYes()
-    .wait('Project deleted locally.')
-    .runAsync();
+  try {
```
awsluja, line 13: we must exit gracefully even if delete fails; otherwise artifacts won't get stored properly by our test runner

packages/amplify-e2e-core/src/nexpect-reporter.js
```diff
@@ -3,6 +3,17 @@ const path = require('path');
 const { v4: uuid } = require('uuid');
 const localTemplatePath = path.resolve(__dirname, '../dist/index.html');
 
+const getShortNameForTestSuite = (
```
awsluja, line 6: tests file names have a structure like:
"C:\\<path>\\<to>\\mytest.test.ts" on windows
or
"home//<path>//<to>//mytest.test.ts" on linux

packages/amplify-e2e-core/src/utils/add-circleci-tags.ts
```diff
@@ -53,3 +60,45 @@ export const addCircleCITags = (projectPath: string): void => {
 export function sanitizeTagValue(value: string): string {
   return value.replace(/[^ a-z0-9_.:/=+\-@]/gi, '');
 }
+
+/**
+ * Jest runs tests that start at the same time, which can lead to issues 
+ * when running amplify init on 4 tests simultaneously.
```
awsluja, line 66: This issue is caused by DynamoDb & OpenSearch simulators, which create directories that have a fixed path (not specific to each project), leading to conflicts when spinning up. These issues are exaggerated on Windows tests

packages/amplify-e2e-tests/src/__tests__/api_2a.test.ts
```diff
@@ -173,78 +170,4 @@ describe('amplify add api (GraphQL)', () => {
     expect(_.isEmpty(disableDSConfig.ResolverConfig)).toBe(true);
   });
```
awsluja, line 171: this test file was too large, i split it

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -21,8 +21,8 @@ const { initializeProgressBars } = require('./aws-cfn-progress-formatter');
 const { printer } = require('amplify-prompts');
 
 const CFN_MAX_CONCURRENT_REQUEST = 5;
-const CFN_POLL_TIME = 5 * 1000; // 5 secs wait to check if  new stacks are created by root stack
-const CFN_POLL_TIME_MAX = 30 * 1000; // 30 seconds
+const CFN_POLL_TIME = (process.env.IS_AMPLIFY_CI ? 30 : 5) * 1000; // 5 secs wait to check if  new stacks are created by root stack
+const CFN_POLL_TIME_MAX = (process.env.IS_AMPLIFY_CI ? 120 : 30) * 1000; // 30 seconds
```
awsluja, line 25: still experimenting with this, tests were failing on some batches where multiple long running deployments occur;
it seems to have fixed most of the throttling

scripts/artifact-storage-path-allow-list.ts
```diff
@@ -32,6 +32,8 @@
  * will automatically normalize these paths for Windows if it detects it.
  */
 export const ARTIFACT_STORAGE_PATH_ALLOW_LIST = [
+    '~/repo/artifacts',
+    '~/repo/packages/amplify-util-mock/',
```
awsluja, line 36: todo, cleanup duplicate line

scripts/cci-data.sh
```diff
@@ -0,0 +1,13 @@
...
+    --header "Circle-Token: $CIRCLECI_TOKEN" \
+    -o ./scripts/cci-job.data.json
+}
+
+function updateTestData {
+  curl --request GET \
+    --url 'https://circleci.com/api/v2/insights/gh/aws-amplify/amplify-cli/workflows/build_test_deploy_v3/test-metrics?branch=dev&all-branches=false' \
+    --header "Circle-Token: $CIRCLECI_TOKEN" \
+    -o ./scripts/cci-test.data.json
+}
```
awsluja, line 13: todo: move to JS files


I originally wrote these as Bash scripts while I was prototyping

scripts/cci-process-data.ts
```diff
@@ -0,0 +1,32 @@
+import { join } from "path";
+import { getOldJobNameWithoutSuffixes, getTestFiles, getTimingsFromJobsData, REPO_ROOT, saveTestTimings } from "./cci-utils";
+
+function main(): void {
```
awsluja, line 4: todo: add js docs

This function takes all tests suites & checks if we have timing data for them.
If we have timing data, we can use that to organize tests by duration.

This script saves the timing data to a file that is checked in; so we don't need to setup CCI tokens for every dev, someone with a CCI token should just be able to run the data scripts every few months.

scripts/cci-test-timings.data.json
```diff
@@ -0,0 +1,1070 @@
+{
+  "lastUpdated": "2022-12-31T02:03:46.162Z",
+  "totalTestFiles": 266,
+  "timingData": [
```
awsluja, line 4: timing data is the worst of either Linux or Windows median runtime for each file

scripts/cci-workflow-results.ts
```diff
@@ -0,0 +1,95 @@
+import { join } from "path";
+import { getOldJobNameWithoutSuffixes, getTestFiles, getTimingsFromJobsData, REPO_ROOT, saveTestTimings, saveWorkflowResults, saveWorkflowResultsHTML } from "./cci-utils";
+import axios from 'axios';
+
```
awsluja, line 4: todo: refactor and move CCI api calls to a different file - those should be exported/common methods

scripts/cci-workflow-results.ts
```diff
@@ -0,0 +1,95 @@
...
+    const result = await axios.get(`https://circleci.com/api/v2/project/github/aws-amplify/amplify-cli/${jobId}/tests`, {
+        headers: {
+            "Circle-Token": process.env.CIRCLECI_TOKEN 
+        }
+    });
+    return result.data;
+}
+
+
+const runit = async () => {
```
awsluja, line 39: todo: add js docs

This function runs after a workflow has completed, and it aggregates the results of all the failed tests into a single HTML page/JSON doc that you can locate in the 'wait_for_all' job.
Sample: https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/14097/workflows/792d5fbd-f763-435a-b00a-c325c225699b/jobs/663225/artifacts

setup-jest.js
```diff
@@ -0,0 +1 @@
+jest.retryTimes(2);
```
awsluja, line 1: Each batch of tests is executed by a worker pool containing 3 workers, so if a test fails we want to be able to re-run that tests without having to wait for all of the other tests to finish up. 
This will allow a failed test to run on a worker that frees up.

.circleci/config.base.yml
```diff
@@ -947,18 +936,6 @@ workflows:
                 - beta
                 - release
                 - /tagged-release-without-e2e-tests\/.*/
-      - graphql_e2e_tests:
```
awsluja, line 950: these tests were left over after data repo split - we don't use them

.circleci/config.yml
```diff
@@ -159,7 +159,7 @@ jobs:
             source .circleci/local_publish_helpers.sh
             amplify -v
             cd packages/amplify-console-integration-tests
-            retry yarn run console-integration --maxWorkers=3
```
awsluja, line 162: --no-cache is used so that Jest doesn't ignore worker pools (by default it will try to use cache info to determine test ordering, so it won't always run tests in parallel)

.circleci/local_publish_helpers.sh
```diff
@@ -250,8 +250,8 @@ function setAwsAccountCredentials {
     export AWS_ACCESS_KEY_ID_ORIG=$AWS_ACCESS_KEY_ID
     export AWS_SECRET_ACCESS_KEY_ORIG=$AWS_SECRET_ACCESS_KEY
     export AWS_SESSION_TOKEN_ORIG=$AWS_SESSION_TOKEN
-    # introduce a delay of up to 5 minutes to allow for more even spread aws list-accounts calls due to throttling
-    sleep $[ ( $RANDOM % 300 )  + 1 ]s
```
awsluja, line 254: i've reduced this a lot because the number of executors has come down drastically, and it appears to be working fine

from ~490 to ~280

.circleci/local_publish_helpers.sh
```diff
@@ -276,11 +276,10 @@ function runE2eTest {
 
     if [ -f  $FAILED_TEST_REGEX_FILE ]; then
         # read the content of failed tests
-        failedTests=$(<$FAILED_TEST_REGEX_FILE)
-        # adding --force-exit per https://github.com/facebook/jest/issues/9473
-        yarn run e2e --force-exit --detectOpenHandles --maxWorkers=3 $TEST_SUITE -t "$failedTests"
+        failedTests=$(<$FAILED_TEST_REGEX_FILE)=
+        yarn run e2e --no-cache --maxWorkers=3 $TEST_SUITE -t "$failedTests"
     else
-        yarn run e2e --force-exit --detectOpenHandles --maxWorkers=3 $TEST_SUITE
```
awsluja, line 283: --detectOpenHandles implicitly forces tests to run in band (in series) so we must remove this to allow multi core utilization

.gitignore
```diff
@@ -1,5 +1,9 @@
 **/.DS_Store
 .circleci/generated_config.yml
+scripts/cci-job.data.json
+scripts/cci-test.data.json
+artifacts/cci-workflow-results.json
+artifacts/cci-workflow-results.html
```
awsluja, line 6: these files are intermediates (we only use them temporarily, and discard after processing)

package.json
```diff
@@ -28,6 +28,10 @@
     "link-win": "node ./scripts/link-bin.js packages/amplify-cli/bin/amplify amplify-dev",
     "link-aa-win": "node ./scripts/link-bin.js packages/amplify-app/bin/amplify-app amplify-app-dev",
     "setup-dev-win": "yarn dev-build && yarn link-win && yarn link-aa-win",
+    "update-test-data": "source ./scripts/cci-data.sh && updateTestData",
+    "update-job-data": "source ./scripts/cci-data.sh && updateJobData",
+    "process-job-data": "yarn update-job-data && ts-node ./scripts/cci-process-data.ts",
+    "workflow-results": "yarn ts-node ./scripts/cci-workflow-results.ts",
```
awsluja, line 34: These scripts can be run manually a significant change in test timings has occurred.
"process-job-data" will fetch the latest cci job data & update the cci-timings data

package.json
```diff
@@ -95,7 +99,7 @@
     "@commitlint/prompt": "^16.2.3",
     "@microsoft/api-extractor": "^7.33.5",
     "@types/glob": "^7.1.1",
-    "@types/jest": "^26.0.20",
+    "@types/jest": "^27.0.0",
```
awsluja, line 102: the upgrade to Jest 27 fixes several issues with worker pools:
https://github.com/facebook/jest/issues/11143

packages/amplify-cli/src/commands/push.ts
```diff
@@ -43,11 +43,11 @@ const syncCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
       securityChangeNotified = await notifyListQuerySecurityChange(context);
     }
 
-    spinner.start(`Fetching updates to backend environment: ${currentEnv} from the cloud.`);
+    printer.info(`Fetching updates to backend environment: ${currentEnv} from the cloud.`);
```
awsluja, line 46: spinners tend to crash on windows after a long running deployment, and does not appear to be thread safe; this replaces spinners with printer until we find an alternative

packages/amplify-cli/src/initialize-env.ts
```diff
@@ -79,8 +77,8 @@ export const initializeEnv = async (
       }
     }
 
-    spinner.start(
-      isPulling ? `Fetching updates to backend environment: ${currentEnv} from the cloud.` : `Initializing your environment: ${currentEnv}`,
+    printer.info(
+      isPulling ? `Fetching updates to backend environment: ${currentEnv} from the cloud.` : `Initializing your environment: ${currentEnv}`
```
awsluja, line 81: same here - spinners appear to crash on windows

packages/amplify-e2e-core/src/index.ts
```diff
@@ -111,14 +112,16 @@ export async function createNewProjectDir(
   // eslint-disable-next-line spellcheck/spell-checker
   prefix = path.join(fs.realpathSync(os.tmpdir()), amplifyTestsDir),
 ): Promise<string> {
-  const currentHash = execSync('git rev-parse --short HEAD', { cwd: __dirname }).toString().trim();
   let projectDir;
   do {
-    const randomId = await global.getRandomId();
-    projectDir = path.join(prefix, `${projectName}_${currentHash}_${randomId}`);
+    projectDir = path.join(prefix, `${projectName}_${Math.floor(Math.random() * 1000000)}`);
```
awsluja, line 117: this became more important as we started running multiple tests concurrently; long project names would overlap because the first 20 characters generated by this could be identical - due to the commit hashes being identical (currentHash)

The primary change here is that random number is moved to the 2nd portion of the directory name, instead of the last portion of it.

packages/amplify-e2e-core/src/index.ts
```diff
@@ -111,14 +112,16 @@ export async function createNewProjectDir(
...
   } while (fs.existsSync(projectDir));
 
   fs.ensureDirSync(projectDir);
+
+  const initialDelay = Math.floor(Math.random() * 180 * 1000);
+  await sleep(initialDelay);
```
awsluja, line 123: createProjectDir(..) is something that nearly every test uses; and we need a way to stagger the tests to avoid hotspots on a specific account+region.
This makes it a perfect candidate for staggering test start times, which I had previously attempted using addCircleCITags(..), but I found that method was over complicated- and this does the job just fine.

packages/amplify-e2e-tests/src/__tests__/schema-auth-11-a.test.ts
```diff
@@ -10,7 +10,7 @@ describe('api directives @auth batch 3a', () => {
   let projectDir: string;
 
   beforeEach(async () => {
-    projectDir = await createNewProjectDir('auth3');
+    projectDir = await createNewProjectDir('auth11a');
```
awsluja, line 13: The directory names should align with the test file name, so I updated these

packages/amplify-e2e-tests/src/setup-tests.ts
```diff
@@ -7,3 +7,4 @@ expect.extend({ toBeAS3Bucket });
 const JEST_TIMEOUT = 1000 * 60 * 60; // 1 hour
 
 jest.setTimeout(JEST_TIMEOUT);
+jest.retryTimes(1);
```
awsluja, line 10: this change allows a test to retry before all other tests are done, which is ideal because each executor has a worker pool of 3 threads (so there is no need to wait before retrying)

packages/amplify-migration-tests/src/__tests__/migration_tests/lambda-layer-migration/layer-migration.test.ts
```diff
@@ -28,14 +28,14 @@ describe('test lambda layer migration flow introduced in v5.0.0', () => {
   let projRoot: string;
   let versionToMigrateFrom: string;
 
-  beforeAll(async () => {
+  beforeEach(async () => {
```
awsluja, line 31: this change, and similar changes to the files in src/__tests__/ are done to ensure that "createNewProjectDir(..)" is the first thing that happens in a test.

There were only a handful of tests that needed this, specifically - migration tests.

package.json
```diff
@@ -119,8 +121,8 @@
     "execa": "^5.1.1",
     "glob": "^7.2.0",
     "husky": "^3.1.0",
-    "jest": "^26.6.3",
-    "jest-circus": "^26.6.3",
+    "jest": "^27.0.0",
```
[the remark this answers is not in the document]
awsluja: just to limit scope for this PR for the minimum incremental change needed

packages/amplify-e2e-core/src/utils/add-circleci-tags.ts
```diff
@@ -53,3 +58,45 @@ export const addCircleCITags = (projectPath: string): void => {
...
+ * 
+ * @param projectPath 
+ */
+const staggerInitCalls = async (projectPath: string) => {
+  const lock = path.join(projectPath, '..', 'init-lock.txt');
+  // one test will create the lock first, 15 seconds should be enough to allow 1 test to do this first without collision risk
+  const initialDelay = Math.floor(Math.random() * 15 * 1000);
+  await sleep(initialDelay);
+  while(true){
+    if(fs.existsSync(lock)) {
```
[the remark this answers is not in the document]
awsluja: going to remove this code - so thats good :D

packages/amplify-e2e-core/src/utils/index.ts
```diff
@@ -42,7 +42,9 @@ config();
  * delete project directory
  */
 export const deleteProjectDir = (root: string): void => {
-  rimraf.sync(root);
+  if(fs.existsSync(root)){
```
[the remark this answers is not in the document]
awsluja: in this case, projectDir paths are unique between the tests on the same machine - so we should be good

packages/amplify-e2e-tests/src/__tests__/schema-auth-11-a.test.ts
awsluja: we do generate random uuids after the project name - so these are just suffixes

scripts/cci-test-timings.data.json
```diff
@@ -0,0 +1,1074 @@
+{
+  "lastUpdated": "2022-12-31T02:03:46.162Z",
+  "totalTestFiles": 266,
```
jhockett, line 3: Is this the same as the length of `timingData`? If so, I would think computing the length is better to avoid having to ensure both are in sync.

jhockett (commented):
Overall, really nice work!

Why is the empty file `setup-jest.js` needed? 

I recommend installing prettier and fixing the indentation in the scripts so we avoid a large amount of white-space changes in future PRs

packages/amplify-e2e-tests/src/__tests__/schema-auth-11-a.test.ts
sdstolworthy: Ok, instead of manually updating these, we could create a helper function that does this for us. Something like
```typescript
function getBasename(filename = ''): string {
  const filename = path.basename(filename);
  const extensionIndex = filename.lastIndexOf('.')
  if (extensionIndex < 0) {
      return filename ?? 'test';
  }
  return filename.slice(0, extensionIndex);
}
projectDir = await createNewProjectDir(getBasename(__filename));
```

awsluja: I really like this idea, but we'll need to find a way to isolate that into a PR, it will affect a lot of files (300ish)

packages/amplify-provider-awscloudformation/src/__tests__/iterative-deployment/stack-event-monitor.test.ts
```diff
@@ -15,14 +15,16 @@ const cfn = ({
 } as unknown) as CloudFormation;
 
 jest.useFakeTimers();
+jest.spyOn(global, 'setTimeout');
```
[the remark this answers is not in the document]
awsluja: Good question - from Jest 26 -> Jest 27 there was a change to the default Spy list; we now need to manually spy on global's setTimeout in order to track the calls made to them

package.json
```diff
@@ -28,6 +28,8 @@
     "link-win": "node ./scripts/link-bin.js packages/amplify-cli/bin/amplify amplify-dev",
     "link-aa-win": "node ./scripts/link-bin.js packages/amplify-app/bin/amplify-app amplify-app-dev",
     "setup-dev-win": "yarn dev-build && yarn link-win && yarn link-aa-win",
+    "update-test-timing-data": "ts-node ./scripts/cci-get-job-metrics.ts && ts-node ./scripts/cci-extract-test-timings-from-job-metrics.ts",
+    "workflow-results": "yarn ts-node ./scripts/cci-workflow-results.ts",
```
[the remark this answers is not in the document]
awsluja: most of the scripts use yarn ts-node (i believe just out of habit) - but its completely optional because the scripts already execute in yarn context

scripts/artifact-storage-path-allow-list.ts
awsluja: Yes - I updated it to remove the duplicate line

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
awsluja: Hoping to follow up on this:
1. create a follow up PR that fixes “too many requests” exceptions caused by CFN polling but maintains the same logic on E2E as w/ regular customers
- get rid of custom polling rate for e2e
- implement error handling for “too many requests” instead

result: 
logic is the same for both customers & our e2e tests

scripts/split-e2e-tests-v2.ts
```diff
@@ -0,0 +1,371 @@
...
+        if (typeof j === 'string') {
+            return j === baseJobName;
+        } else {
+            const name = Object.keys(j)[0];
+            return name === baseJobName;
+        }
+        });
+
+        if (workflowJob) {
+        Object.values(jobByRegion).forEach((regionJobs: any) => {
```
sdstolworthy, line 276: Formatting. The formatting in the whole file is non-standard and should be reformatted.

sdstolworthy (changes requested):

scripts/split-e2e-tests-v2.ts
```diff
@@ -0,0 +1,371 @@
...
+                    }
+                }
+            }
+            if(j.os === 'l'){
+                counts.l ++;
+            } else {
+                counts.w ++;
+            }
+        }
+    }
```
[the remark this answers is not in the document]
sdstolworthy: https://eslint.org/docs/latest/rules/no-plusplus

packages/amplify-cli/src/initialize-env.ts
awsluja: thanks , fixed!

scripts/split-e2e-tests-v2.ts
awsluja: thanks - just ran formatting on all files

scripts/split-e2e-tests-v2.ts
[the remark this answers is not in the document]
awsluja: fixed

scripts/cci-test-timings.data.json
awsluja: this is just some metadata to provide at a glance; its helpful to quickly see when the last updated time was & how many files we having timing data for

sdstolworthy (approved):

jhockett (approved):

PR #11709, opened 2023-01-04
chore: remove datastore Schema type dependency

letsbelopez:
#### Description of changes

Update code to remove the datastore type. We will use the type from codegen when it's exported in the future. As for now, there's no need to use datastore and we'd like to remove the import.

This also removes the datastore dependency.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-util-uibuilder/src/utils/prePushHandler.ts
+++ b/packages/amplify-util-uibuilder/src/utils/prePushHandler.ts
@@ -1,9 +1,13 @@
 import { $TSContext } from 'amplify-cli-core';
-import type { Schema } from '@aws-amplify/datastore';
 import { printer } from 'amplify-prompts';
 import AmplifyUIBuilder from 'aws-sdk/clients/amplifyuibuilder';
 import { AmplifyStudioClient } from '../clients';
 import { isFormDetachedFromModel, isFormSchemaCustomized, shouldRenderComponents } from '../commands/utils';
+
+// TODO: replace this type when amplify-codegen package export official types
+type Schema = {
+  models: Record<string, unknown>
+};
 
 /**
  * Handles showing a warning for detached forms pre-push.
--- a/packages/amplify-util-uibuilder/src/__tests__/prePushHandler.test.ts
+++ b/packages/amplify-util-uibuilder/src/__tests__/prePushHandler.test.ts
@@ -24,7 +24,7 @@ describe('handlePrePush', () => {
   beforeEach(() => {
     context = {
       amplify: {
-        invokePluginMethod: () => ({ models: { Comment: {} } }),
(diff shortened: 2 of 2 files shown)


packages/amplify-util-uibuilder/src/utils/prePushHandler.ts
```diff
@@ -26,15 +23,23 @@ export const prePushHandler = async (context: $TSContext): Promise<void> => {
...
     return;
   }
-  printDetachedFormsWarning(formSchemas, localSchema);
 };
 
-const printDetachedFormsWarning = (formSchemas: { entities: AmplifyUIBuilder.Form[] }, localSchema: Schema): void => {
-  const modelNames = new Set(Object.keys(localSchema.models));
+const printDetachedFormsWarning = (formSchemas: { entities: AmplifyUIBuilder.Form[] }, localSchema: {models: unknown}): void => {
+  const models = localSchema.models;
+  if (!models || typeof models !== 'object') {
```
lazpavel, line 38: nit: can't this use !hasModels helper?

letsbelopez: @lazpavel I talked with codegen and they will eventually be exporting types. So this is an intermediate update to remove the datastore import.

lazpavel (approved):

packages/amplify-util-uibuilder/src/utils/prePushHandler.ts
```diff
@@ -1,10 +1,14 @@
 import { $TSContext } from 'amplify-cli-core';
-import type { Schema } from '@aws-amplify/datastore';
 import { printer } from 'amplify-prompts';
 import AmplifyUIBuilder from 'aws-sdk/clients/amplifyuibuilder';
 import { AmplifyStudioClient } from '../clients';
 import { isFormDetachedFromModel, isFormSchemaCustomized, shouldRenderComponents } from '../commands/utils';
 
+// TODO: replace this type when amplify-codegen package export official types
```
lazpavel, line 7: I would suggest to create an asana task blocked on amplify-codegen export types task, otherwise TODOs tend to remain orphaned in the code

letsbelopez: asana task created. thanks

danielleadams (commented):
Is the dependency being removed in a future PR? (Or is it being used in the dependency tree? It is still in the `yarn.lock` file)

letsbelopez:
> Is the dependency being removed in a future PR? (Or is it being used in the dependency tree? It is still in the `yarn.lock` file)

There may be another package with a dependency on datastore, but there are no other datastore references in this util-uibuilder package.

danielleadams (approved):

PR #11711, opened 2023-01-05
chore: Persist downstream exceptions in usage metrics

Amplifiyer:
#### Description of changes

Persist downstream exceptions in usage metrics that are not of type 'AmplifyException`

Pending items:
- [ ] Update transformation lambda function to handle the new field similar to `error` [In progress]
- [ ] Update db with the new columns generated by lambda 

unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
+++ b/packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
@@ -19,7 +19,7 @@ export class UsageDataPayload implements IUsageDataPayload {
   inputOptions: InputOptions;
   timestamp: string;
   error!: SerializableError;
-  downStreamException!: SerializableError;
+  downstreamException!: SerializableError;
   payloadVersion: string;
   osPlatform: string;
   osRelease: string;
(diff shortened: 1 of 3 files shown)


packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
```diff
@@ -19,6 +19,7 @@ export class UsageDataPayload implements IUsageDataPayload {
   inputOptions: InputOptions;
   timestamp: string;
   error!: SerializableError;
+  downstreamException!: SerializableError;
```
akshbhu, line 22: will there always be downstream exception when we initialize ```UsageDataPayload``` ?

akshbhu (approved):

packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
Amplifiyer: No, it's not a non null assertion operator here but a definitive assignment assertion operator since it's not being initialized in the constructor(see the `OkGreeter` class [here](https://www.typescriptlang.org/docs/handbook/2/classes.html#fields))

It can be null, and there is a unit test for that as well (works exactly similar to `Error` which can also be null)

lazpavel (approved):

PR #11737, opened 2023-01-10
feat: update amplify delete to also delete all PS values for the project

goldbez:
#### Description of changes
These changes makes the CLI delete all environment parameters from Parameter Store when deleting an Amplify project. The function in `delete-from-parameter-store.ts` interacts with the AWS SSM to delete the parameters. The function retrieves all paramaters for a single environment that have been uploaded to Parameter Store and then deletes these parameters from Parameter Store. This function is invoked for each environment in `delete-project.ts`.

Unit tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are changed and added (https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)


packages/amplify-provider-awscloudformation/src/utils/delete-from-parameter-store.ts
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Higher order function for deleting CloudFormation parameters from the service
+ */
+export const getEnvParametersDeleteHandler = async (context: $TSContext, envName: string): Promise<(keys: Array<string>) => Promise<void>> => {
+  const appId = resolveAppId(context);
+  const { client } = await SSM.getInstance(context);
+  return deleteParametersFromParameterStore(appId, envName, client);
+};
+
+const deleteParametersFromParameterStore = (
```
jhockett, line 16: Similar comment here about the naming. Perhaps `deleteParametersFromService`

packages/amplify-cli/src/extensions/amplify-helpers/delete-project.ts
```diff
@@ -43,16 +42,44 @@ export const deleteProject = async (context: $TSContext): Promise<void> => {
           printer.warn('Amplify App cannot be deleted, other environments still linked to Application');
         }
       }
+
+      // delete paramater store parameters for each environment
+      const CloudFormationProviderName = 'awscloudformation';
+      const deleteParametersFromParameterStoreFn: (
```
sdstolworthy:
```suggestion
      const deleteParametersFromParameterStore: (
```
avoid hungarian-esque notation

packages/amplify-environment-parameters/src/environment-parameter-manager.ts
```diff
@@ -127,6 +127,22 @@ class EnvironmentParameterManager implements IEnvironmentParameterManager {
...
+
+  async deleteAllEnvParametersFromPs(
+    environmentName: string,
+    deleteParametersFromParameterStoreFn: (envName: string, keys: Array<string>) => Promise<void>,
+  ): Promise<void> {
+    const parameterNamesToDelete: Array<string> = [];
+    Object.entries(this.resourceParamManagers).forEach(([resourceKey, paramManager]) => {
+      const [category, resourceName] = splitResourceKey(resourceKey);
+      const resourceParams = paramManager.getAllParams();
+      Object.entries(resourceParams).forEach(([paramName]) => {
```
sdstolworthy, line 139:
```suggestion
      Object.keys(resourceParams).forEach((paramName) => {
```

packages/amplify-provider-awscloudformation/src/__tests__/utils/delete-from-parameter-store.test.ts
```diff
@@ -0,0 +1,35 @@
...
+mockSSM.getInstance = jest.fn().mockResolvedValue({
+  client: {
+    deleteParameters: jest.fn().mockImplementation(() => ({
+      promise: deleteParametersPromiseMock,
+    })),
+  },
+});
+
+describe('parameters-delete-handler', () => {
+  it('check if returned function is defined', async () => {
```
sdstolworthy, line 17: what happens if it's not defined?

packages/amplify-provider-awscloudformation/src/__tests__/utils/delete-from-parameter-store.test.ts
```diff
@@ -0,0 +1,35 @@
+import { $TSContext } from 'amplify-cli-core';
+import { getEnvParametersDeleteHandler } from '../../utils/delete-from-parameter-store';
+import { SSM } from '../../aws-utils/aws-ssm';
+
+jest.mock('../../aws-utils/aws-ssm');
+const mockSSM = SSM as jest.Mocked<typeof SSM>;
+const deleteParametersPromiseMock = jest.fn().mockImplementation(() => Promise.resolve());
```
sdstolworthy, line 7: Can we test what happens when this promise is rejected?

packages/amplify-provider-awscloudformation/src/__tests__/utils/delete-from-parameter-store.test.ts
```diff
@@ -0,0 +1,35 @@
...
+        inputParams: {
+          amplify: {
+            appId: '1',
+          },
+        },
+      },
+    };
+
+    const deleteParametersFromParameterStoreFn = await getEnvParametersDeleteHandler((contextStub as unknown) as $TSContext, envName);
+    expect(deleteParametersFromParameterStoreFn).toBeDefined();
```
sdstolworthy: This expect isn't really asserting anything, because `mockSSM.getInstance` is returning an explicitly mocked value that is always defined. Therefore, this expectation will necessarily be true always. Avoid expectations for mocked conditions because they can lead to false positives in coverage.

packages/amplify-provider-awscloudformation/src/__tests__/utils/delete-from-parameter-store.test.ts
```diff
@@ -0,0 +1,35 @@
...
+    await deleteParametersFromParameterStoreFn(keys);
+    expect(deleteParametersPromiseMock).toBeCalledTimes(1);
```
sdstolworthy, line 33: This seems to be the meat of this unit test.
So, could we rework this test to essentially assert:
Given we have an `appId`, `envName`, and `ssmClient`,
When I call the function returned from `getEnvParametersDeleteHandler`,
`deleteParametersPromiseMock` should be called with a parameter of X shape.

packages/amplify-provider-awscloudformation/src/__tests__/utils/delete-from-parameter-store.test.ts
goldbez: If it is not defined, then the parameters cannot be deleted from Parameter Store

packages/amplify-provider-awscloudformation/src/__tests__/utils/delete-from-parameter-store.test.ts
goldbez: Good catch. Thank you.

sdstolworthy: updated language for clarity

packages/amplify-provider-awscloudformation/src/__tests__/utils/delete-from-parameter-store.test.ts
sdstolworthy: Sorry, poor wording. In this test, it is never possible for the returned function to be undefined. Once the test is reworked, the description of the test should match what is being asserted in the test.

sdstolworthy (approved):

PR #11751, opened 2023-01-11
chore: upgrade to graphql 15 and upgrade data deps

dpilch:
#### Description of changes

In some rare cases both graphql 14 and 15 were resolved in the package leading to errors.

```
Error: Cannot use GraphQLScalarType \\"String\\" from another module or realm.

Ensure that there is only one instance of \\"graphql\\" in the node_modules
directory. If different versions of \\"graphql\\" are the dependencies of other
relied on modules, use \\"resolutions\\" to ensure only one version is installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate \\"graphql\\" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.
```

amplify-codegen and amplify-category-api will also need to be upgraded to GraphQL 15. 

https://github.com/aws-amplify/amplify-category-api/pull/1072
https://github.com/aws-amplify/amplify-codegen/pull/524

Run e2e tests on circle ci

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-function/package.json
```diff
@@ -35,7 +35,7 @@
     "folder-hash": "^4.0.2",
     "fs-extra": "^8.1.0",
     "globby": "^11.0.3",
-    "graphql-transformer-core": "^7.6.9",
+    "graphql-transformer-core": "7.6.10-upgrade-graphql15-2.1",
```
dpilch, line 38: These will all be changed to the production release with GraphQL 15 when it is available.

packages/amplify-appsync-simulator/src/__tests__/utils/graphql-runner/query-and-mutation.test.ts
```diff
@@ -144,7 +144,7 @@ describe('runQueryAndMutation', () => {
       }
     `);
 
-    const error = new GraphQLError('An error from template');
+    const error = new Error('An error from template');
```
phani-srikar, line 147: Why is this replacement necessary?

phani-srikar (changes requested):
Is there a reason why we are removing the `^` in API dependencies? I think you need to just run `yarn update-data-packages` and commit the changes.

dpilch:
Oops, I ran `yarn update-data-packages` but from the tagged release versions. I guess it does not add the caret back.

packages/amplify-appsync-simulator/src/__tests__/utils/graphql-runner/query-and-mutation.test.ts
dpilch: Was a placeholder and I forgot to commit the real fix. I'll push it shortly.

jhockett (approved):

phani-srikar (approved):

PR #11809, opened 2023-01-19
feat: prompt for missing backend parameters

jhockett:
#### Description of changes
Add a generic fallback prompt for missing CloudFormation parameter values. Any prompts that handle specific parameters, such as environment variables in the function category, will be executed first.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/execution-manager.ts
+++ b/packages/amplify-cli/src/execution-manager.ts
@@ -353,8 +353,8 @@ export const raiseEvent = async (context: Context, args: AmplifyEventArgs): Prom
         };
         return eventHandler;
       });
-    for (const eventHanler of eventHandlers) {
-      await eventHanler();
+    for (const eventHandler of eventHandlers) {
+      await eventHandler();
     }
   }
 };
--- a/packages/amplify-cli/src/execution-manager.ts
+++ b/packages/amplify-cli/src/execution-manager.ts
@@ -1,6 +1,5 @@
 import * as fs from 'fs-extra';
 import * as path from 'path';
-import sequential from 'promise-sequential';
 import {
   stateManager, executeHooks, HooksMeta,
 } from 'amplify-cli-core';
@@ -354,7 +353,9 @@ export const raiseEvent = async (context: Context, args: AmplifyEventArgs): Prom
         };
         return eventHandler;
       });
-    await sequential(eventHandlers);
+    for (const eventHanler of eventHandlers) {
+      await eventHanler();
+    }
   }
 };
 
--- a/packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
@@ -88,12 +88,14 @@ export const pushResources = async (
   // Verify any environment parameters before push operation
   const envParamManager = (await ensureEnvParamManager()).instance;
 
-  const promptMissingParameter = async (parameter: string, envParamManager: IEnvironmentParameterManager): Promise<void> => {
-    printer.warn(`Could not find value for parameter ${parameter}`);
-    const [, categoryName, resourceName, parameterName] = parameter.split('_');
-
+  const promptMissingParameter = async (
+    categoryName: string,
+    resourceName: string,
+    parameterName: string,
+    envParamManager: IEnvironmentParameterManager,
+  ): Promise<void> => {
+    printer.warn(`Could not find value for parameter ${parameterName}`);
     const value = await prompter.input(`Enter a value for ${parameterName} for the ${categoryName} resource: ${resourceName}`);
-
     const resourceParamManager = envParamManager.getResourceParamManager(categoryName, resourceName);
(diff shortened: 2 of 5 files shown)


packages/amplify-cli/src/execution-manager.ts
```diff
@@ -354,7 +353,9 @@ export const raiseEvent = async (context: Context, args: AmplifyEventArgs): Prom
         };
         return eventHandler;
       });
-    await sequential(eventHandlers);
+    for (const eventHanler of eventHandlers) {
```
jhockett, line 356: This change was required to fix some strange behavior in the prePushHook event. Without this change, `ctrl+c` during the prompt for a missing env var (existing prompt) would not exit the CLI, instead exiting the prompt and the CLI continued executing.

sdstolworthy: Spelling: hanler->handler

packages/amplify-environment-parameters/src/environment-parameter-manager.ts
```diff
@@ -122,31 +121,38 @@ class EnvironmentParameterManager implements IEnvironmentParameterManager {
...
-        allEnvParams.add(paramName);
+        allEnvParams.add(`${resourceKey}_${paramName}`);
       }
     }
 
     Object.keys(expectedParameters).forEach(expectedParameter => {
-      const paramName = getParamaterNameFromParameterStoreKey(expectedParameter);
-      if (!allEnvParams.has(paramName)) {
-        missingParameterNames.push(paramName);
+      const [categgoryName, resourceName, parameterName] = getNamesFromParameterStoreKey(expectedParameter);
```
sdstolworthy, line 140: spelling: categgory -> category

packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
```diff
@@ -87,7 +87,28 @@ export const pushResources = async (
 
   // Verify any environment parameters before push operation
   const envParamManager = (await ensureEnvParamManager()).instance;
-  envParamManager.verifyExpectedEnvParameters();
+
+  const promptMissingParameter = async (parameter: string, envParamManager: IEnvironmentParameterManager): Promise<void> => {
+    printer.warn(`Could not find value for parameter ${parameter}`);
+    const [, categoryName, resourceName, parameterName] = parameter.split('_');
```
sdstolworthy, line 93: This kind of magic destructuring is happening in enough places in this change that I think it's worth it to factor it out into a helper method. It's not explicit why the first value is being ignored (presumably because it equals `AMPLIFY`?). It's also not explicit why in `packages/amplify-environment-parameters/src/environment-parameter-manager.ts` line 201, the final value is ignored.

Factoring this destructuring out to another function that returns a predictable value is probably only half of a solution. If we are implicitly expecting our parameter names to match this naming structure, we should probably have a type that explicitly defines that structure in the first place, rather than implicitly capturing it in magic strings.

packages/amplify-environment-parameters/src/environment-parameter-manager.ts
```diff
@@ -190,4 +197,14 @@ const getParameterStoreKey = (
...
 ): string => `AMPLIFY_${categoryName}_${resourceName}_${paramName}`;
 
-const getParamaterNameFromParameterStoreKey = (fullParameter: string) => fullParameter.split('_').slice(3).join('_');
+const getNamesFromParameterStoreKey = (fullParameter: string) => {
+  const [, categoryName, resourceName] = fullParameter.split('_'); // Ignores the AMPLIFY prefix
+  const parameterName = fullParameter.split('_').slice(3).join('_'); // In case parameterName contains underscores
+  return [categoryName, resourceName, parameterName];
+}
+
+type ResourceParameter = {
```
sdstolworthy, line 206: This should probably be exported

packages/amplify-environment-parameters/API.md
```diff
@@ -19,6 +19,11 @@ export type IEnvironmentParameterManager = {
     hasResourceParamManager: (category: string, resource: string) => boolean;
     getResourceParamManager: (category: string, resource: string) => ResourceParameterManager;
     save: (serviceUploadHandler?: ServiceUploadHandler) => Promise<void>;
+    getMissingParameters: () => Promise<{
```
sdstolworthy, line 22: If you export `ResourceParameter`, this file will show a named type as the resolution value of the promise instead of an inline object.

sdstolworthy (approved):

packages/amplify-environment-parameters/API.md
jhockett: I didn't want to commit to a name for the type (insert hardest problems in cs joke). Given that it's only 3 values, I figured it was better to avoid polluting the namespace with something generic like `ResourceParameter`.

goldbez: I think that this should be a named type.

goldbez (approved):

PR #11814, opened 2023-01-19
test: e2e test for parameter store upload/delete

goldbez:
#### Description of changes
This PR adds an E2E test that tests whether PS parameters are correctly uploaded and deleted when a function is added, a new environment is added, the environment is deleted, and the project is deleted. 

Ran E2E test locally. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [added](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)


packages/amplify-e2e-core/src/utils/sdk-calls.ts
```diff
@@ -401,42 +403,127 @@ export const getSSMParameters = async (region: string, appId: string, envName: s
...
+  }
+  return await ssmClient
+    .getParameters({
+      Names: parameterNames.map(name => `/amplify/${appId}/${envName}/AMPLIFY_function_${funcName}_${name}`),
+      WithDecryption: true,
+    })
+    .promise();
+};
+
+export const getAllSSMParamatersFromAppId = async (appId: string, region: string): Promise<Array<string>> => {
```
jhockett, line 425: naming nit: I think something like `getAllSSMParamatersForAppId` makes more sense

packages/amplify-e2e-core/src/utils/sdk-calls.ts
```diff
@@ -401,42 +403,127 @@ export const getSSMParameters = async (region: string, appId: string, envName: s
...
+  const ssmClient = new SSM({ region });
+  const retrievedParameters: Array<string> = [];
+  let recievedNextToken = '';
```
jhockett, line 428: nit: spelling ~~recieved~~ => received

packages/amplify-e2e-tests/src/environment/env.ts
```diff
@@ -20,6 +20,28 @@ export function addEnvironment(cwd: string, settings: { envName: string; numLaye
   });
 }
 
+export function addEnvironmentCarryOverEnvVars(cwd: string, settings: { envName: string }): Promise<void> {
```
jhockett, line 23: Please make this an async function and use the newer `.runAsync()` method at the end of the chain returned by `spawn(...)`

packages/amplify-e2e-tests/src/__tests__/guardrail-test.ts
```diff
@@ -0,0 +1,96 @@
+import {
```
jhockett, line 1: Change the test file name to describe what's being tested and ensure it ends in `.test.ts`. Perhaps something like `parameter-store.test.ts`?

packages/amplify-e2e-core/src/utils/sdk-calls.ts
```diff
@@ -401,42 +403,127 @@ export const getSSMParameters = async (region: string, appId: string, envName: s
...
+  do {
+    const ssmArgument = getSsmSdkParametersGetParametersByPath(appId, recievedNextToken);
+    const data = await ssmClient.getParametersByPath(ssmArgument).promise();
+    retrievedParameters.push(...data.Parameters.map(returnedParameter => returnedParameter.Name));
+    recievedNextToken = data.NextToken;
+  } while (recievedNextToken);
+  return retrievedParameters;
+};
+
+export const expectParametersOptionalValue = async (
```
jhockett, line 438: Is there a more descriptive name that can be used? I don't understand the optional value part, because all the function parameters are required.

goldbez: The optional value part means that you can pass in a key-value pair to expect, or just a key. So the value is optional. This is for parameters where the value cannot be predicted.

jhockett (approved):

PR #11838, opened 2023-01-24
chore: jest config normalization

sdstolworthy:
#### Description of changes

This PR moves all jest configurations to the `"jest"` block of each project's `package.json`. Additionally, a `collectCoverageFrom` option is added to each Jest config.

In some projects, Jest configs were stored in both a `jest.config.js` file and the `package.json` `"jest"` block. This is a no-op. Since the majority of projects were using the Jest config in `package.json`, all packages were standardized on that norm.

Additionally, a `collectCoverageFrom` option was added to each Jest config file. In the absence of the `collectCoverageFrom` option, coverage statistics were only collected on files that are actually loaded during testing. This can lead to underreporting coverage.

See [the jest documentation](https://jestjs.io/docs/cli#--collectcoveragefromglob).

Ran tests, verified tests passing. Verified coverage was collected from a sampling of previously missing files.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-dynamodb-simulator/package.json
+++ b/packages/amplify-dynamodb-simulator/package.json
@@ -42,8 +42,7 @@
     "testEnvironment": "node",
     "collectCoverage": true,
     "collectCoverageFrom": [
-      "src/**/*.{ts,tsx,js,jsx}",
-      "!src/__tests__/"
+      "index.js"
     ]
   }
 }
--- a/packages/amplify-dynamodb-simulator/package.json
+++ b/packages/amplify-dynamodb-simulator/package.json
@@ -40,6 +40,10 @@
   },
   "jest": {
     "testEnvironment": "node",
-    "collectCoverage": true
+    "collectCoverage": true,
+    "collectCoverageFrom": [
+      "src/**/*.{ts,tsx,js,jsx}",
+      "!src/__tests__/"
+    ]
   }
 }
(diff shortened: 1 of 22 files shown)


lazpavel (dismissed):

edwardfoyle (dismissed):

packages/amplify-dynamodb-simulator/package.json
```diff
@@ -40,6 +40,10 @@
   },
   "jest": {
     "testEnvironment": "node",
-    "collectCoverage": true
+    "collectCoverage": true,
+    "collectCoverageFrom": [
+      "src/**/*.{ts,tsx,js,jsx}",
+      "!src/__tests__/"
+    ]
```
jhockett, line 47: This package has a non-standard folder structure (no src folder), so this needs to be modified

sdstolworthy: Done, good catch

packages/amplify-frontend-ios/package.json
```diff
@@ -29,6 +29,10 @@
   },
   "jest": {
     "collectCoverage": true,
+    "collectCoverageFrom": [
```
jhockett, line 32: Sorry didn't think to check earlier, but most of the FE packages have source code in `lib` and the package root

lazpavel (approved):
Please trigger e2e tests

evcodes (approved):
LGTM!

jhockett (approved):

PR #11840, opened 2023-01-24
fix: skip deleting from parameter store if no app id

goldbez:
#### Description of changes
This change skips deleting parameters from parameter store upon deleting an environment or an Amplify project if no app ID can be resolved.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/utils/delete-from-service.ts
+++ b/packages/amplify-provider-awscloudformation/src/utils/delete-from-service.ts
@@ -8,14 +8,14 @@ import { getSsmSdkParametersDeleteParameters, getSsmSdkParametersGetParametersBy
  * Delete CloudFormation parameters from the service
  */
 export const deleteEnvironmentParametersFromService = async (context: $TSContext, envName: string): Promise<void> => {
-  let appId;
   try {
-    appId = resolveAppId(context);
+    const appId = resolveAppId(context);
+    const { client } = await SSM.getInstance(context);
+    await deleteParametersFromParameterStore(appId, envName, client);
   } catch {
     return;
   }
-  const { client } = await SSM.getInstance(context);
-  await deleteParametersFromParameterStore(appId, envName, client);
+  
 };
 
 const deleteParametersFromParameterStore = async (appId: string, envName: string, ssmClient: SSMType): Promise<void> => {
--- a/packages/amplify-provider-awscloudformation/src/utils/delete-from-service.ts
+++ b/packages/amplify-provider-awscloudformation/src/utils/delete-from-service.ts
@@ -11,7 +11,7 @@ export const deleteEnvironmentParametersFromService = async (context: $TSContext
   let appId;
   try {
     appId = resolveAppId(context);
-  } catch (error) {
+  } catch {
     return;
   }
   const { client } = await SSM.getInstance(context);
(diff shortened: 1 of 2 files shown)


packages/amplify-provider-awscloudformation/src/utils/delete-from-service.ts
```diff
@@ -8,7 +8,12 @@ import { getSsmSdkParametersDeleteParameters, getSsmSdkParametersGetParametersBy
  * Delete CloudFormation parameters from the service
  */
 export const deleteEnvironmentParametersFromService = async (context: $TSContext, envName: string): Promise<void> => {
-  const appId = resolveAppId(context);
+  let appId;
+  try {
+    appId = resolveAppId(context);
+  } catch (error) {
+    return;
+  }
```
jhockett, line 16: nit: If error is ignored, no need to capture it

```suggestion
  } catch {
    return;
  }
```

jhockett (approved):

packages/amplify-provider-awscloudformation/src/utils/delete-from-service.ts
```diff
@@ -8,7 +8,12 @@ import { getSsmSdkParametersDeleteParameters, getSsmSdkParametersGetParametersBy
...
 export const deleteEnvironmentParametersFromService = async (context: $TSContext, envName: string): Promise<void> => {
-  const appId = resolveAppId(context);
+  let appId;
+  try {
+    appId = resolveAppId(context);
+  } catch {
+    return;
+  }
   const { client } = await SSM.getInstance(context);
   await deleteParametersFromParameterStore(appId, envName, client);
```
sdstolworthy, line 18:
```suggestion
  try {
    const appId = resolveAppId(context);
    const { client } = await SSM.getInstance(context);
    await deleteParametersFromParameterStore(appId, envName, client);
  } catch {
    return;
  }
  
```

sdstolworthy: Is suppressing the error the desired behavior here?

jhockett: `resolveAppId` will throw if no appId can be determined, but we do not want to prevent delete from succeeding because not having an appId is a valid state (such as running `amplify pull` but not modifying the backend).

sdstolworthy (approved):

packages/amplify-provider-awscloudformation/src/utils/delete-from-service.ts
```diff
@@ -8,9 +8,13 @@ import { getSsmSdkParametersDeleteParameters, getSsmSdkParametersGetParametersBy
...
 export const deleteEnvironmentParametersFromService = async (context: $TSContext, envName: string): Promise<void> => {
-  const appId = resolveAppId(context);
-  const { client } = await SSM.getInstance(context);
-  await deleteParametersFromParameterStore(appId, envName, client);
+  try {
+    const appId = resolveAppId(context);
+    const { client } = await SSM.getInstance(context);
+    await deleteParametersFromParameterStore(appId, envName, client);
+  } catch {
+    return;
```
akshbhu, line 16: nit: we can have a debug log here, for more info in our debug logs 
```
printer.debug(`no AppId ${appId} found`);
```

jhockett (approved):

PR #11922, opened 2023-02-07
chore: prepend @aws-amplify to amplify-category-analytics

evcodes:
#### Description of changes
prepending @aws-amplify scope across packages to prevent others from impersonating the Amplify team

running e2e test suite and unit test suite

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-analytics/package.json
```diff
@@ -14,6 +14,9 @@
     "amplify",
     "aws"
   ],
+  "publishConfig": {
```
lazpavel, line 17: Do we want it public?

lazpavel (changes requested):
This PR looks that it needs to be rebased

packages/amplify-category-analytics/package.json
evcodes: It's already public here: https://www.npmjs.com/package/amplify-category-function. 
If we don't do this then when release it, the change wouldn't be reflected on npm

sobolk: yes, every package that is scoped must be declared public.
fwiw, we had a release fiasco in the past when new package got introduced without that config, see https://github.com/aws-amplify/amplify-cli/pull/11029 .

sobolk (approved):

lazpavel (approved):

sobolk (approved):

PR #11937, opened 2023-02-08
chore: move context-extensions to amplify-cli-core

sdstolworthy:
#### Description of changes

Moves `context-extensions` to `amplify-cli-core`.

Context Extensions attaches most of the properties to the Amplify Context. The file has no amplify dependencies except a reference to the context interface (in its previous state, the `Context` class in `amplify-cli/`, now `$TSContext`).

With this file in `amplify-cli`, mocking context is difficult because depending on `amplify-cli/` in other projects creates a circular dependency. Moving this file to `amplify-cli-core` makes it easier to mock.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli-core/API.md
+++ b/packages/amplify-cli-core/API.md
@@ -27,7 +27,9 @@ export type $TSContext = {
     migrationInfo: $TSAny;
     projectHasMobileHubResources: boolean;
     prompt: $TSAny;
-    patching: $TSAny;
+    patching: {
+        replace(filePath: string, oldContent: string, newContent: string): Promise<string>;
+    };
     exeInfo: $TSAny;
     input: $TSAny;
     parameters: $TSAny;
--- a/packages/amplify-cli-core/API.md
+++ b/packages/amplify-cli-core/API.md
@@ -27,6 +27,7 @@ export type $TSContext = {
     migrationInfo: $TSAny;
     projectHasMobileHubResources: boolean;
     prompt: $TSAny;
+    patching: $TSAny;
     exeInfo: $TSAny;
     input: $TSAny;
     parameters: $TSAny;
@@ -242,6 +243,9 @@ export class ApiCategoryFacade {
     // (undocumented)
     static transformGraphQLSchema(context: $TSContext, options: $TSAny): Promise<DeploymentResources | DeploymentResources_2 | undefined>;
 }
+
+// @public (undocumented)
+export function attachExtensions(context: $TSContext): void;
 
 // @public (undocumented)
 export const AWS_AMPLIFY_DEFAULT_BANNER_URL = "https://aws-amplify.github.io/amplify-cli/banner-message.json";
@@ -860,7 +864,7 @@ export type IContextTemplate = {
         target: string;
         props: $TSObject;
         directory: string;
-    }) => string;
+    }) => Promise<string>;
 };
 
 // @public (undocumented)
@@ -1241,6 +1245,22 @@ export enum PluginAPIError {
 }
 
 // @public (undocumented)
+const print_2: {
+    info: typeof info;
+    fancy: typeof fancy;
+    warning: typeof warning;
+    error: typeof error;
+    success: typeof success;
+    table: typeof table;
+    debug: typeof debug;
+    green: typeof green;
+    yellow: typeof yellow;
+    red: typeof red;
+    blue: typeof blue;
+};
+export { print_2 as print }
+
+// @public (undocumented)
 export const projectNotInitializedError: () => AmplifyError;
 
 // @public (undocumented)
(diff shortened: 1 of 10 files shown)


packages/amplify-cli-core/src/types.ts
```diff
@@ -23,6 +23,7 @@ export type $TSContext = {
   migrationInfo: $TSAny;
   projectHasMobileHubResources: boolean;
   prompt: $TSAny;
+  patching: $TSAny;
```
sdstolworthy, line 26: Meaningful change ^

packages/amplify-cli-core/API.md
```diff
@@ -1565,6 +1585,17 @@ export type WriteCFNTemplateOptions = {
...
+// src/context/context-extensions.ts:257:19 - (ae-forgotten-export) The symbol "error" needs to be exported by the entry point index.d.ts
+// src/context/context-extensions.ts:257:19 - (ae-forgotten-export) The symbol "success" needs to be exported by the entry point index.d.ts
+// src/context/context-extensions.ts:257:19 - (ae-forgotten-export) The symbol "table" needs to be exported by the entry point index.d.ts
+// src/context/context-extensions.ts:257:19 - (ae-forgotten-export) The symbol "debug" needs to be exported by the entry point index.d.ts
+// src/context/context-extensions.ts:257:19 - (ae-forgotten-export) The symbol "green" needs to be exported by the entry point index.d.ts
+// src/context/context-extensions.ts:257:19 - (ae-forgotten-export) The symbol "yellow" needs to be exported by the entry point index.d.ts
+// src/context/context-extensions.ts:257:19 - (ae-forgotten-export) The symbol "red" needs to be exported by the entry point index.d.ts
+// src/context/context-extensions.ts:257:19 - (ae-forgotten-export) The symbol "blue" needs to be exported by the entry point index.d.ts
```
sobolk, line 1598: are these warnings legit ?

packages/amplify-cli-core/src/types.ts
sobolk: what does `patching` mean ?

goldbez (approved):

packages/amplify-cli-core/API.md
sdstolworthy: I am not sure what is causing those. I've worked through a few ways to try to get rid of them, but was not successful. The `print` is exported and its members are accessible in outside packages. I tried scoping the exports from `context-extensions` and that did not work either.

See https://github.com/sdstolworthy/amplify-cli/blob/e1111304ffc89d7bec691df85823c331a4e8e50a/packages/amplify-cli/src/plugin-manager.ts#L140

packages/amplify-cli-core/src/types.ts
sdstolworthy: https://github.com/aws-amplify/amplify-cli/blob/608455b8f76fc2ac18d316ebb2cb901a15b58231/packages/amplify-cli-core/src/context/context-extensions.ts#L167

packages/amplify-cli-core/API.md
sdstolworthy: I just updated the `$TSContext` with a more meaningful type than `any`

packages/amplify-cli-core/src/types.ts
sdstolworthy: I just updated `$TSContext` with a more meaningful type

packages/amplify-cli-core/API.md
sobolk: Looks like either these (see screenshot) should get exported.
Or (better?) `export const print` should get properties typed as `(message: string) => void` (or so) and those private functions assigned to them.

image

packages/amplify-cli-core/src/types.ts
sobolk: yeah. I saw that definition. I still have no idea what it does.
btw. should property be named patchingHandler or patcher ? it's kind of verb-ish for the property.

sdstolworthy: It doesn't actually look like it's used anywhere in the application. So this can likely be marked as deprecated, then removed in the future. Agreed the name is weird, but I think renaming is out of scope of this PR.

sobolk: if that's dead code why not remove it now ?

packages/amplify-cli-core/API.md
sdstolworthy: Thanks. Did that. Those warnings are removed.

packages/amplify-cli-core/src/types.ts
sdstolworthy: Fair enough. Removed.

packages/amplify-cli/src/context-manager.ts
```diff
@@ -1,8 +1,8 @@
-import { $TSAny, stateManager } from 'amplify-cli-core';
+import { $TSAny, $TSContext, stateManager } from 'amplify-cli-core';
 import * as _ from 'lodash';
 import { init } from './app-config';
 // eslint-disable-next-line spellcheck/spell-checker
```
Amplifiyer, line 4: nit: we can possibly remove this now.

Amplifiyer (approved):

sobolk (approved):

Amplifiyer (approved):

PR #11946, opened 2023-02-08
chore: enforce new line at the end of files

sobolk:
#### Description of changes

Enforces end of line at end of files.
See
https://eslint.org/docs/latest/rules/eol-last
https://stackoverflow.com/questions/44704968/visual-studio-code-insert-newline-at-the-end-of-files

#### Description of how you validated changes

Failure injection. (tested also via SSH run in circle ci)
image

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies
- [ ] [Pull request labels](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#labels) are added

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #11947, opened 2023-02-08
chore: dont lint in e2e runs

sobolk:
#### Description of changes

When both PR validation job and e2e tests are run for the same PR then e2e linting result wins and is show in GH checks.
However in e2e run lint is no-op... https://github.com/aws-amplify/amplify-cli/blob/4316a610477594f11b7e9b35d97554e5a839e1e9/.circleci/lint_pr.sh#L5-L8

Therefore, we should exclude linting from e2e run.

Unfortunately circle ci doesn't have a good mechanism to only run lint on PR jobs (see https://ruarfff.com/circleci-pr-only/ ). But hence we do run pr validation and e2e tests on PRs usually. Just excluding lint from e2e seems like a cheap and effective solution.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/.circleci/lint_pr.sh
+++ b/.circleci/lint_pr.sh
@@ -1,4 +1,4 @@
-set -xeo pipefail
+set -xe
 # extract the PR number from the PR link
 PR_NUM=${CIRCLE_PULL_REQUEST##*/}
 
--- a/.circleci/lint_pr.sh
+++ b/.circleci/lint_pr.sh
@@ -8,5 +8,5 @@ if [ -z "$PR_NUM" ]; then
 fi
 
 # get PR file list, filter out removed files, filter only JS/TS files, then pass to the linter
-curl -fsSL https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PR_REPONAME/pulls/$PR_NUM/files | jq -r '.[] | select(.status!="removed") | .filename' | grep -E '\.(js|jsx|ts|tsx)$' | xargs yarn eslint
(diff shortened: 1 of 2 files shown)


.circleci/lint_pr.sh
```diff
@@ -8,5 +8,5 @@ if [ -z "$PR_NUM" ]; then
 fi
 
 # get PR file list, filter out removed files, filter only JS/TS files, then pass to the linter
-curl -fsSL https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PR_REPONAME/pulls/$PR_NUM/files | jq -r '.[] | select(.status!="removed") | .filename' | grep -E '\.(js|jsx|ts|tsx)$' | xargs yarn eslint
+curl -fsSL https://api.github.com/repos/$CIRCLE_PROJECT_USERNAME/$CIRCLE_PR_REPONAME/pulls/$PR_NUM/files | jq -r '.[] | select(.status!="removed") | .filename' | grep -E '\.(js|jsx|ts|tsx)$' | xargs --no-run-if-empty yarn eslint
```
sobolk, line 11: This PR triggered failure here.
The `--no-run-if-empty` skips if stdin is empty.

Tested also with https://github.com/aws-amplify/amplify-cli/pull/11946 to see if it still works with non empty file list 

image

sdstolworthy (dismissed):

jhockett (dismissed):

.circleci/lint_pr.sh
```diff
@@ -1,4 +1,4 @@
-set -xeo pipefail
+set -xe
```
sobolk, line 1: Verified in ssh session that script exists with non 0 exit code if there's lint failure.
image

the pipefail was too aggressive for what we do below.

jhockett (approved):

sdstolworthy (approved):

PR #11962, opened 2023-02-09
fix: improved typing for $TSContext

sdstolworthy:
#### Description of changes

Adds additional typing for $TSContext.

As a general rule, types were not renamed unless there was a logical conflict or additional clarity was needed.
All types/interfaces used to define $TSContext were moved into `amplify-cli-core`.

In many cases across the codebase, nullable values are treated as non-nullable. In cases where this occurred, the type was defined as nullable, and null checks were added where needed. In cases where non-null assertions were removed and a null value would throw a type error in the existing code, a null check is added that specifically throws a type error.

Where possible, types and interfaces were moved to `amplify-cli-core`, and implementers are left in `amplify-cli` (or `-auth`).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli-core/src/types.ts
```diff
@@ -401,3 +599,13 @@ interface AmplifyToolkit {
...
+export interface IOptionFlowHeadlessData {
+  input: string;
+  timestamp: number;
+}
+export interface IOptionFlowCLIData {
+  prompt: string;
+  input: unknown;
+  timestamp: number;
+}
+export type TypeOptionFlowData = IOptionFlowHeadlessData | IOptionFlowCLIData;
```
sobolk, line 611: Few things here.
- I'm on the fence regarding the `I` prefix for interfaces. In general I'd rather name implementations with more detailed name or use something for default impl like `DefaultFoo implements Foo` or `$Foo implements Foo`. However, we have plenty of existing interfaces with`I` so we should probably be consistent.
- I would definietly drop `Type` from `TypeOptionFlowData`
- why is `input` string in one shape and `unknown` in the other. Isn't this creating diamond problem in `TypeOptionFlowData` ?
- should `IOptionFlowCLIData` extend `IOptionFlowHeadlessData` ?

packages/amplify-cli-core/src/types.ts
```diff
@@ -201,7 +391,15 @@ export type GetPackageAssetPaths = () => Promise<string[]>;
...
  */
-export type $IPluginManifest = $TSAny;
+export type $IPluginManifest = {
+  name: string;
+  type: string;
+  commands?: string[];
+  services?: string[];
+  functionRuntime?: {
+    pluginId: string;
+  };
```
sobolk, line 401: Does every plugin define `functionRuntime` ? What does `functionRuntime` mean in this context?
If this is plugin specific we should find a way to subtype.

packages/amplify-cli-core/src/types.ts
```diff
@@ -201,7 +391,15 @@ export type GetPackageAssetPaths = () => Promise<string[]>;
 /**
  * Placeholder type
  */
-export type $IPluginManifest = $TSAny;
+export type $IPluginManifest = {
```
sobolk, line 394: This is neither interface nor implements interface, why is it prefixed with `$` and `I` ?

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  name: string;
+  directory: string;
+  pluginName: string;
+  pluginType: string;
+  commands: string[];
+};
+
+export type Runtime = {
+  plugins: Plugin[];
 };
```
sobolk, line 226: should this be called `PluginCollection` or `Plugins` ? Why `Runtime` ?

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  raw: $CommandLineInput['argv'];
+  array: $CommandLineInput['subCommands'];
+  first?: string;
+  second?: string;
+  third?: string;
+}
+
+export type $CommandLineInput = {
+  argv: Array<string>;
+  plugin?: string;
```
sobolk, line 190:
```suggestion
  pluginName?: string;
```

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+    quickstart?: boolean;
+    app?: string | boolean;
+    timeout?: string;
+    event?: string;
+    minify?: boolean;
+    help?: boolean;
+    localEnvFilePath?: string;
+    yes?: boolean;
+    appId?: string;
+  } & Record<string, string | boolean>;
```
sobolk, line 213: do all explicit options apply to every command ?

if not then perhaps we need `CliCommand` type and subtypes with coherent set of options?

Or we type options just as `Record<string, string | boolean>` and we let plugins to cast it down into something more specific.

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  emitError: (error: Error | null) => Promise<void>;
+  emitSuccess: () => Promise<void>;
+  init: (
+    installationUuid: string,
+    version: string,
+    input: $CommandLineInput,
+    accountId: string,
+    projectSettings: ProjectSettings,
+    processStartTimeStamp: number,
+  ) => void;
```
sobolk, line 147: instead of init this should be `ctor` on implementing type.
also there should probably be separate shape that keeps these things as read only properties.
and from the content it seems that such shape should be created at process start and cached somewhere as singleton, i.e. global constant.

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  cmd: string;
+  subCmd: string | undefined;
+  optionFlowData: Array<TypeOptionFlowData>; //IOptionFlowHeadlessData | IOptionFlowCLIData
+  input: $CommandLineInput;
+  timestamp: string;
+  projectEnvIdentifier?: string; // hash(ProjectName + Amplify AppId + EnvName)
+  projectIdentifier?: string; // hash( ProjectName + Amplify App Id)
+}
+export type TimedCodePath = ManuallyTimedCodePath | UntilExitTimedCodePath | FromStartupTimedCodePaths;
+export interface IUsageMetricsData {
```
sobolk, line 136: This should likely be called `IUsageMetricsCollector` or something like this. `Data` would be more appropriate for plain data structure.

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  emitSuccess: () => Promise<void>;
+  init: (
+    installationUuid: string,
+    version: string,
+    input: $CommandLineInput,
+    accountId: string,
+    projectSettings: ProjectSettings,
+    processStartTimeStamp: number,
+  ) => void;
+  getUsageDataPayload: (error: Error | null, state: string) => IUsageDataPayload;
```
sobolk, line 148:
```suggestion
  getUsageData: (error: Error | null, state: string) => IUsageData;
```

or perhaps `UsageMetrics`

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  executable: string;
+  category: string;
+  isHeadless: boolean;
+  cmd: string;
+  subCmd: string | undefined;
+  optionFlowData: Array<TypeOptionFlowData>; //IOptionFlowHeadlessData | IOptionFlowCLIData
+  input: $CommandLineInput;
+  timestamp: string;
+  projectEnvIdentifier?: string; // hash(ProjectName + Amplify AppId + EnvName)
+  projectIdentifier?: string; // hash( ProjectName + Amplify App Id)
```
sobolk, line 133: should these be read only ?
(this question applies to all types, would it make sense to seal some of them?)

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  state: string;
+  isCi: boolean;
+  accountId: string;
+  projectSetting: ProjectSettings;
+  codePathDurations: Partial<Record<TimedCodePath, number>>;
+  flowReport: IFlowReport;
+  pushNormalizationFactor: number;
+}
+
+type Trace = {
```
sobolk, line 107:
```suggestion
type StackTraceElement = {
```

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  file: string;
+  lineNumber: string;
+  columnNumber: string;
+};
+export type SerializableError = {
+  name: string;
+  message: string;
+  details?: string;
+  code?: string;
+  trace?: Trace[];
```
sobolk, line 118: if we export SerializableError we should export Trace

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  frontend?: string;
+  editor?: string;
+  framework?: string;
+};
+export interface IUsageDataPayload {
+  sessionUuid: string;
+  installationUuid: string;
+  amplifyCliVersion: string;
+  input: $CommandLineInput | null;
+  inputOptions: InputOptions;
```
sobolk, line 90: are these redundant ?

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+}
+export interface IUsageData extends IUsageMetricsData, IFlowData {}
+export type ProjectSettings = {
+  frontend?: string;
+  editor?: string;
+  framework?: string;
+};
+export interface IUsageDataPayload {
+  sessionUuid: string;
+  installationUuid: string;
```
sobolk, line 87: if we decide to create type for `IUsageMetricsData` implementor ctor then we could maybe extend from it or embed it here as property.

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  pushHeadlessFlow: (headlessFlowDataString: string, input: $CommandLineInput) => void;
+  pushInteractiveFlow: (prompt: string, input: unknown) => void;
+  getFlowReport: () => IFlowReport | Record<string, never>;
+  assignProjectIdentifier: (envName?: string) => string | undefined;
+}
+export interface IUsageData extends IUsageMetricsData, IFlowData {}
+export type ProjectSettings = {
+  frontend?: string;
+  editor?: string;
+  framework?: string;
```
sobolk, line 83: should we create enums for those ? they have closed set of possible values.

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  projectPath?: string;
+  defaultEditor?: string;
+  frontendHandler?: unknown;
+} & Record<T, string>;
+export interface LocalEnvInfo {
+  projectPath: string;
+  defaultEditor: string;
+  envName: string;
+}
+export interface IFlowData {
```
sobolk, line 72: This shouldn't be called `Data` it's not a data structure. `IFlowReporter` perhaps ?

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  frontend: T;
+  version: string;
+  providers: string[];
+  projectPath?: string;
+  defaultEditor?: string;
+  frontendHandler?: unknown;
+} & Record<T, string>;
+export interface LocalEnvInfo {
+  projectPath: string;
+  defaultEditor: string;
```
sobolk, line 69: defaultEditor and projectPath sound like candidates to go to `ProjectSettings` no ?

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+  NONE: unknown;
+};
+export type ProjectConfig<T extends string = ''> = {
+  frontend: T;
+  version: string;
+  providers: string[];
+  projectPath?: string;
+  defaultEditor?: string;
+  frontendHandler?: unknown;
+} & Record<T, string>;
```
sobolk, line 66: should this be `ProjectSettings` too ?

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
+}
+export type TeamProviderEnvironment = {
+  categories: Record<string, unknown>;
+};
+export type TeamProviderInfo = {
+  [envName: string]: Record<string, unknown>;
+};
+export type LocalAwsInfo = {
+  NONE: unknown;
+};
```
sobolk, line 58: what does this do ?

packages/amplify-cli-core/src/types.ts
```diff
@@ -20,19 +20,209 @@ export type $TSContext = {
...
-  parameters: $TSAny;
-  usageData: $TSAny;
-  runtime: $TSAny;
+  input: $CommandLineInput;
+  parameters: ContextParameters;
+  usageData: IUsageData;
+  runtime: Runtime;
   pluginPlatform: IPluginPlatform;
-  newUserInfo?: $TSAny;
+  newUserInfo?: string;
```
sobolk, line 32: is this serialized json ?

sobolk (commented):
This is going into right direction I think.

packages/amplify-category-auth/src/commands/auth.ts
```diff
@@ -16,10 +16,13 @@ type AuthCommandType = {
  * @returns auth command response
  */
 export const run = async (context: $TSContext): Promise<$TSAny> => {
-  if (context.parameters.options.help) {
+  if (context.parameters.options?.help) {
     return runHelp(context);
   }
   try {
+    if (!context.parameters.first) {
+      throw new TypeError('Missing command');
```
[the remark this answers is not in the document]
sdstolworthy: Throwing a `TypeError` here (and in other places this is called out), preserves the existing behavior. In the original code, on line 23, `parameters.first` is passed to `path.join` as if it were non-nullable. If `path.join` is passed `undefined`, it throws a `TypeError`. This code is functionally identical to the previous revision, but explicitly handles the null case.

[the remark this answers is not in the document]
sdstolworthy: image

packages/amplify-category-storage/src/commands/storage.ts
```diff
@@ -8,12 +8,15 @@ export { categoryName as name } from '../constants';
...
 export async function run(context: $TSContext): Promise<void> {
-  if (context.parameters.options.help) {
+  if (context.parameters.options?.help) {
     runHelp(context);
     return;
   }
   if (/^win/.test(process.platform)) {
     try {
+      if (!context.parameters.first) {
+        throw new TypeError('Missing command');
```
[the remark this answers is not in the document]
sdstolworthy: While the context is slightly different here, the result is the same as in the previous comment. In the existing code, `TypeError` is what would have already been thrown. It is handled in the catch below.

packages/amplify-cli-core/src/types.ts
sdstolworthy: As best I could see, `newUserInfo` is the set to the awaited return value of `packages/amplify-provider-awscloudformation/src/setup-new-user.ts`'s `run` function.

image

packages/amplify-cli-core/src/types.ts
sdstolworthy: https://github.com/sdstolworthy/amplify-cli/blob/0f94a470070a5e4ff6f2b4a1212f9f6a81e0fbd6/packages/amplify-cli/src/migrate-project.ts#L243

It's not clear to me. But the best I could tell was that this value in the context of MigrationInfo will always return `{ NONE: <contents of awsInfo.json>}`.

packages/amplify-cli-core/src/types.ts
sdstolworthy: I don't think these should be enums, because, at least in theory, we could have infinite number of frontend possibilities, supported by plugins or extensions of existing plugins. The code itself is structured to support unknown values for the frontend/framework/editor fields.

packages/amplify-cli-core/src/types.ts
sdstolworthy: It sure seems like it. Input options just references `input.options`. I've updated the type, but I'm going to leave the class structure the same for this change.

packages/amplify-cli-core/src/types.ts
sdstolworthy: Fixed

packages/amplify-cli-core/src/types.ts
sdstolworthy: Agreed. I'll leave this as is in this request, just to minimize potentially breaking changes. But I agree this is an inappropriate impl leak, and should be fixed.

packages/amplify-cli-core/src/types.ts
sdstolworthy: I think it's best to not change this right now. This is one of those words that is used everywhere in the project. This refactor should be done in its own PR.

packages/amplify-cli-core/src/types.ts
sdstolworthy: It gets pretty convoluted. In some cases, plugins are overwriting the options with values that go against their type. See https://github.com/sdstolworthy/amplify-cli/blob/0f94a470070a5e4ff6f2b4a1212f9f6a81e0fbd6/packages/amplify-cli/src/init-steps/postInitSetup.ts#L19.

Ultimately, I think the right course is to define options on `CommandLineInput` simply as a `Record<string, string (or unknown?)>`. Then, allow the plugins to downcast to a more specific type. Ultimately, `options` is just a collection of flags that the `cli` parses as a convenience for itself and its plugins. (see: https://github.com/sdstolworthy/amplify-cli/blob/0f94a470070a5e4ff6f2b4a1212f9f6a81e0fbd6/packages/amplify-cli/src/input-manager.ts#L40)

I think this defining the used flags here is a first step for now, but I could be talked out of that.

packages/amplify-cli-core/src/types.ts
sdstolworthy: - The `I` prefix is considered unidiomatic pretty widely in the Typescript world. Personally, I dislike it, as it encourages a somewhat lazy naming pattern, and also implicitly endorses the notion that an interface is just a class with its guts ripped out. I think they should be removed. I can make that change in this PR.
- Good points on the shapes of the types. I copied those verbatim from `cli`. Let me take a closer look at them and see if I can narrow them down.

packages/amplify-cli-core/src/types.ts
sdstolworthy: I didn't go down the readonly road on this change, but I absolutely agree.

packages/amplify-cli-core/src/types.ts
sobolk: seems that
```
export type LocalAwsInfo = {
  NONE: Record<string, unknown>;
};
```
would be more appropriate here.

packages/amplify-cli-core/src/types.ts
sobolk: I'd rather follow the downcast route and create more concrete types and mapping in plugins.

packages/amplify-cli-core/src/types.ts
sobolk: we should get this rule going to encourage sealing https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/prefer-readonly.md .

packages/amplify-cli-core/src/types.ts
sobolk: makes sense.

packages/amplify-cli-core/src/types.ts
sobolk: see https://github.com/aws-amplify/amplify-cli/blob/66f65adf2280a433fff933ff40a7632bf21e804d/packages/amplify-provider-awscloudformation/src/setup-new-user.ts#L123 .
Seems that right name for this prop is `profileName` or `awsProfileName`.
See https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-profiles.html .

packages/amplify-cli-core/src/types.ts
sdstolworthy: It's not clear what this references. There is a `FunctionRuntime` defined here: https://github.com/sdstolworthy/amplify-cli/blob/f4d6faa880fd4b316672a04e54a7c13aa9440a37/packages/amplify-function-plugin-interface/src/index.ts#L180. However, there is not a `pluginId` defined on that interface.

I hesitate to assume those are actually intended to be the same type. As far as I can tell, `functionRuntime` in this context is only ever referenced here: https://github.com/sdstolworthy/amplify-cli/blob/f4d6faa880fd4b316672a04e54a7c13aa9440a37/packages/amplify-cli/src/extensions/amplify-helpers/load-runtime-plugin.ts#L7.

sobolk: Meanwhile we bumped into this with @evcodes . It seems to be related to `XXX-function-runtime-provider` packages. See https://github.com/aws-amplify/amplify-cli/pull/11985 for a bit of breadcrumbs.
The function runtime providers seem to form their own plugin family (plugins within plugins?...).
Now I wonder if `$IPluginManifest` is only for categories or "any plugin"... within cli. I.e. category function is a plugin that has runtime plugins inside...
I'd say that lumping any plugin into single type isn't cohesive and we should probably have two types `CategoryPlugin` and `FunctionRuntimeProvider` (or so).

sdstolworthy: https://github.com/sdstolworthy/amplify-cli/blob/tscontext-typing/packages/amplify-cli/src/extensions/amplify-helpers/read-breadcrumbs.ts#L20

This file shows the shape that structure is supposed to have. Thanks for the tip. I've created a type in `types.ts` to describe the shape. I think that distinction will work for the purposes of this PR.

packages/amplify-cli-core/src/types.ts
sdstolworthy: I opted against making the change in this PR. I will handle renaming these in a follow-up. The PR has grown quite large, and this would make reviewing the import changes more difficult for reviewers.

packages/amplify-category-analytics/src/commands/analytics/update.ts
```diff
@@ -43,7 +43,7 @@ export const run = async (context: $TSContext) : Promise<$TSAny> => {
     .catch(err => {
       printer.info(err.stack);
       printer.error(`There was an error updating the ${category} resource`);
-      context.usageData.emitError(err);
+      void context.usageData.emitError(err);
```
sobolk, line 46: do we ever await `emitError` ? perhaps it shouldn't return promise.

packages/amplify-cli/API.md
```diff
@@ -61,24 +68,18 @@ export class UsageData implements IUsageData {
...
     // (undocumented)
     getSessionUuid(): string;
-    // Warning: (ae-forgotten-export) The symbol "UsageDataPayload" needs to be exported by the entry point index.d.ts
-    //
     // (undocumented)
-    getUsageDataPayload(error: Error | null, state: string): UsageDataPayload;
+    getUsageDataPayload(error: Error | null, state: string): IUsageDataPayload;
     // (undocumented)
-    init(installationUuid: string, version: string, input: CommandLineInput, accountId: string, projectSettings: ProjectSettings, processStartTimeStamp: number): void;
+    init(installationUuid: string, version: string, input: CLIInput, accountId: string, projectSettings: ProjectSettings, processStartTimeStamp: number): void;
```
sobolk, line 74: does this init initialize CLI or just metrics collector? if the latter it might use better naming.

packages/amplify-cli-core/API.md
```diff
@@ -491,6 +694,9 @@ export class CloudformationProviderFacade {
     static prePushCfnTemplateModifier(context: $TSContext, template: Template_2): Promise<(template: Template_2) => Promise<void>>;
 }
 
+// @public (undocumented)
+export type CognitoConfiguration = ServiceQuestionsBaseResult & OAuthResult & SocialProviderResult & IdentityPoolResult & PasswordPolicyResult & PasswordRecoveryResult & MfaResult & AdminQueriesResult & Triggers;
+
```
sobolk, line 699: We shouldn't move category specific types into core. I.e. category auth _using_ core is fine, but category auth _implementing_ (and also _using_...) core doesn't seem to be right component split.

I think we should:
- either create a new package for category api interfaces (best and it creates nice pattern for other categories too)
- or move these types to amplify-shared-interfaces instead (not ideal, but much better than moving them to core).

sobolk (commented):
The cli-internal - cli-core moves are looking good.
I am concerned about moving auth types to core.

packages/amplify-cli-core/src/types.ts
sobolk: Sounds good. Could you record refactorings that you leave for follow up in github issue or asana task, so that we don't lose them.

packages/amplify-cli/src/init-steps/preInitSetup.ts
```diff
@@ -27,8 +27,11 @@ export const preInitSetup = async (context: $TSContext): Promise<$TSContext> =>
  *
  * @throws error if url is not a valid remote github url
  */
-const validateGithubRepo = async (repoUrl: string): Promise<void> => {
+function validateGithubRepo(repoUrl: string | boolean): asserts repoUrl is string {
```
jhockett, line 30: We made a decision in the past to convert standard functions to arrow functions. Is there a reason for converting it back?

packages/amplify-category-analytics/src/commands/analytics/update.ts
sdstolworthy: I think it's still appropriate for it to return a promise, because it is asynchronous. This allows the caller to decide what to do with it. But, it in all cases that it is changed in this PR, I believe it is appropriate to discard the promise, since we are exiting the process in the same error handler.

packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/handlers/resource-handlers.test.ts
```diff
@@ -1,6 +1,5 @@
-import { $TSContext, stateManager } from 'amplify-cli-core';
+import { $TSContext, stateManager, CognitoConfiguration } from 'amplify-cli-core';
```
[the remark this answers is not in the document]
sdstolworthy: Taking another look through the code, `updatingAuth` is only used in `amplify-cli` and `amplify-category-auth`. I agree with you that it is best not to move this into core for that reason. I moved it back into `-auth`, and removed `updatingAuth` from the `$TSContext` exported in `-core`. I created a new union type in `-auth` that describes the `updatingAuth` property within the `auth` category. Since `amplify-cli` already depends on `-auth`, `-cli` can import the union type as necessary.

packages/amplify-cli-core/API.md
```diff
@@ -323,6 +514,18 @@ export function attachExtensions(context: $TSContext): void;
...
+export enum AttributeType {
+    // (undocumented)
+    EMAIL = "email",
+    // (undocumented)
+    EMAIL_AND_PHONE_NUMBER = "email, phone_number",
+    // (undocumented)
+    PHONE_NUMBER = "phone_number",
+    // (undocumented)
+    PREFERRED_USERNAME = "preferred_username"
+}
```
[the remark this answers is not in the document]
sdstolworthy: This is moved back to auth. See https://github.com/aws-amplify/amplify-cli/pull/11962#discussion_r1108848638

packages/amplify-cli-core/API.md
sdstolworthy: See https://github.com/aws-amplify/amplify-cli/pull/11962#discussion_r1108848638

packages/amplify-category-auth/src/provider-utils/awscloudformation/auth-stack-builder/auth-stack-transform.ts
```diff
@@ -181,12 +181,12 @@ export class AmplifyAuthTransform extends AmplifyCategoryTransform {
...
 
       // handle dependsOn data
-      const dependsOnKeys = Object.keys(this._cliInputs.cognitoConfig.triggers).map(
+      const dependsOnKeys = Object.keys(this._cliInputs.cognitoConfig.triggers ?? {}).map(
         (i) => `${this._cliInputs.cognitoConfig.resourceName}${i}`,
       );
       const dependsOn = context.amplify.dependsOnBlock(context, dependsOnKeys, 'Cognito');
       // generate trigger config
-      const keys = Object.keys(this._cliInputs.cognitoConfig.triggers);
+      const keys = Object.keys(this._cliInputs.cognitoConfig.triggers as string);
```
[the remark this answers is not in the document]
sdstolworthy: There are a couple of properties that are initialized in this way. Basically, they are initially strings (typically of input from the CLI -- think the json objects that are passed in headless mode). So the property is set as a string of the raw input. They are then parsed to JSON and the resultant object is set back to the same property. I made the decision to represent them in their object form where possible in `$TSContext`, and cast them as strings where necessary in the code.

That said, this really should be refactored. It's not good practice, and invites problems. In this PR, I avoided actually changing the behavior of the code to the extent possible, so I considered this change out of scope.

packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/handlers/resource-handlers.test.ts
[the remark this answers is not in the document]
sobolk: Thank you.

packages/amplify-cli/src/init-steps/preInitSetup.ts
sdstolworthy: [Assertion ](https://github.com/aws-amplify/amplify-cli/pull/11962#discussion_r1108848638) functions in typescript cannot use arrow syntax. I converted this to an assertion function to make the code in the calling function simpler. 

Why an assertion function here? `repoUrl` is one of those odd properties that is misused in the application. There is a single line of code that sets the property to `true` instead of a string. Therefore, the value can either be `true`, or a url to a git repo. In order to avoid type casting (potentially erroneously) in the calling method, I refactored this validation method to have a typescript assertion return value. This makes it so that we don't need to use coercion in the caller to call repoUrl a string.

sdstolworthy: https://github.com/sdstolworthy/amplify-cli/blob/4b0076b29146e57609057d5ca870ed049cbc21f6/packages/amplify-cli/src/init-steps/postInitSetup.ts#L19

packages/amplify-cli/API.md
sobolk: Gotcha. I was tricked to believe this is global export.

packages/amplify-cli/src/init-steps/preInitSetup.ts
jhockett: Makes perfect sense, thanks for the explanation! Also, TIL about assertion functions in TS: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/auth-stack-builder/auth-cognito-stack-builder.test.ts
```diff
@@ -4,7 +4,7 @@ import * as iam from '@aws-cdk/aws-iam';
 import * as cdk from '@aws-cdk/core';
 import { AmplifyAuthCognitoStack } from '../../../../provider-utils/awscloudformation/auth-stack-builder/auth-cognito-stack-builder';
 import { AuthStackSynthesizer } from '../../../../provider-utils/awscloudformation/auth-stack-builder/stack-synthesizer';
-import { AttributeType } from '../../../../provider-utils/awscloudformation/service-walkthrough-types/awsCognito-user-input-types';
+import { AttributeType } from 'amplify-cli-core';
```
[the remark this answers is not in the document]
sdstolworthy: https://github.com/aws-amplify/amplify-cli/pull/11962#discussion_r1108848638

packages/amplify-cli-core/src/types.ts
edwardfoyle: Another vote in favor of defining this type as `Record<string, string | boolean>`

packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
```diff
@@ -13,7 +12,7 @@ export class UsageDataPayload implements IUsageDataPayload {
   installationUuid: string;
   amplifyCliVersion: string;
   input: CommandLineInput | null;
-  inputOptions: InputOptions;
+  inputOptions: Pick<CommandLineInput, 'options'>;
```
edwardfoyle, line 15: Just confirming you don't want `CommandLineInput['options']` instead? ie this object still has everything nested under "options"

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
```diff
@@ -920,7 +920,7 @@ async function determineAuthFlow(context: $TSContext, projectConfig?: ProjectCon
     const docsUrl = 'https://docs.amplify.aws/cli/usage/headless';
     context.print.error(errorMessage);
     context.print.info(`Access keys for continuous integration can be configured with headless parameters: ${chalk.green(docsUrl)}`);
-    await context.usageData.emitError(errorMessage);
+    await context.usageData.emitError(new Error(errorMessage));
```
edwardfoyle, line 923: should this be `AmplifyError` or `AmplifyFault`?

edwardfoyle (commented):
A couple small comments but overall looks like a great step in the right direction!

packages/amplify-cli/src/domain/amplify-usageData/UsageDataPayload.ts
sdstolworthy: You're absolutely right, great catch.

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
sdstolworthy: I'm not sure it matters here, let me check.

Amplifiyer: The `emitError` shouldn't be called here ideally and handled with https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-cli/src/amplify-exception-handler.ts#L42 which is central place for error reporting. 

That might be outside the scope of this PR, but when that happens (i.e. throw error instead of emit), the `Error` **needs** to be an `AmplifyError` else it gets wrapped as `UnknownError` by the centralized error handling

sdstolworthy: @Amplifiyer Perfect. I've got this logged in my doc for tracking additional tasks from this PR. Agreed that this is out of scope for this PR. I'll link you the doc.

packages/amplify-cli-core/src/types.ts
sobolk: That's still elevating command details to generic scope right ?. I.e. IDE would show all options in every plugin if I write `options.`.

Is it possible to add `getOptionsAs<T>()` instead of exposing property? (In Java that would be possible I recall).
So that plugins could use it to downcast to their specific type ?

If that's too hard to do in one go I'd recommend to:
- keep options and CLIOptionFlags but mark them deprecated (to not spread)
- expose new accessor/method/property that allows downcasting and make it work same way as `options`

sobolk: or maybe just deprecating that `CLIOptionFlags` right away is good enough.

packages/amplify-provider-awscloudformation/src/configuration-manager.ts
jhockett: This looks like something I wrote a while back before `AmplifyError` existed. Ideally, this should throw an `AmplifyError` going forward for consistency and accurate error reporting.

packages/amplify-cli-core/src/types.ts
sdstolworthy: I'm still noodling on this. I don't mind the idea of deprecating the CLIOptionsFlags off the bat. But, like you said, I hesitate to bring them into the globally scoped package in the first place. A few of the flags are treated as global flags (yes, headless, frontend, version). [See here](https://github.com/sdstolworthy/amplify-cli/blob/tscontext-typing/packages/amplify-cli/src/input-manager.ts). 

I'm trying to think of a good way to do this that doesn't completely derail this PR.

sobolk: truely global flags might be defined in core and then plugins extend that type.

sdstolworthy: Here's my current thinking: A union type between `CLIGlobalFlags & Record<string, any>`. This allows us to pull out the global flags, but not pollute core with an interface. Then, I can punt to a future PR a better implementation of a method that grabs a value from the options record and makes assertions about its type.

I think any kind of method to allow the plugins to assert types around the flags they expect needs to be given its own consideration in a separate request.

sobolk: If `CLIGlobalFlags` contains only global flags then I think this is good .

jhockett (approved):

Amplifiyer (approved):

sobolk (approved):

jhockett (approved):

PR #12058, opened 2023-02-21
chore: filter api category errors from deployment / Unknown fault

akshbhu:
#### Description of changes
- added a class to filter api category errors from deployment / Unknown fault. Default is previously amplify Fault.

- unit testing of the module

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


packages/amplify-provider-awscloudformation/src/graphql-resource-manager/amplify-graphql-resource-manager.ts
```diff
@@ -119,14 +122,11 @@ export class GraphQLResourceManager {
...
-            message: err.message,
-            link: AMPLIFY_SUPPORT_DOCS.CLI_GRAPHQL_TROUBLESHOOTING.url,
-          },
-          err,
-        );
+        // change here for adding Errors
+        throw new AmplifyErrorFactory(new AmplifyFault('UnknownFault', {
+          message: err.message,
+          link: AMPLIFY_SUPPORT_DOCS.CLI_GRAPHQL_TROUBLESHOOTING.url,
+        }, err)). create(err);
```
sobolk, line 129: Couple of things:
- if the purpose of this change is to map category api errors then `AmplifyErrorFactory` should be called `AmplifyCategoryApiErrorMapper or Converter`.
- that thing could be singleton/module method
- I don't think this design is scalable, i.e. trying to intercept api category errors in selected places. Did we consider intercepting errors at the place we report them? I.e. make this transformation part of metrics collection?

Btw. why not go after category api repo and refactor errors there ?

akshbhu: I checked  the api category repo first , There are a lot of errors thrown for one  ```invalidDirectiveError``` with multiple classes defined for that. for eg: https://github.com/aws-amplify/amplify-category-api/search?p=1&q=InvalidDirectiveError&type=code. I guess someone from api team can handle the task well of converting error to faults.
 
 I am open to changes on this. This is just an initial implementation to filter api errors from faults.

sobolk: @alharris-at what do you think ?

alharris-at: I think I agree, this is similar to the 'minify' concern we had last week, where we need to be more intentional about where we're loading up these concerns. I think having the last step of error reporting ONLY concerned about the generic cases where they're otherwise unmapped is right, and burning down generation of other errors types at the source will lead to less jumbled concerns down the road.

akshbhu:
Closing as already merged

PR #12084, opened 2023-02-22
fix: handle fault for function stack not found in stack resources

lazpavel:
#### Description of changes
- handle the case where StackResources is empty or undefined and print a better message than `'Cannot read property 'PhysicalResourceId' of undefined'`

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/disconnect-dependent-resources/utils.ts
+++ b/packages/amplify-provider-awscloudformation/src/disconnect-dependent-resources/utils.ts
@@ -8,6 +8,7 @@ import { S3 } from '../aws-utils/aws-s3';
 import { fileLogger } from '../utils/aws-logger';
 import { getPreviousDeploymentRecord } from '../utils/amplify-resource-state-utils';
 import { DeploymentOp, DeploymentStep } from '../iterative-deployment';
+import { AmplifyFault } from '../../../amplify-cli-core/src/errors/amplify-fault';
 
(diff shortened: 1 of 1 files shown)


goldbez (approved):

packages/amplify-provider-awscloudformation/src/disconnect-dependent-resources/utils.ts
```diff
@@ -129,6 +130,13 @@ const generateIterativeFuncDeploymentOp = async (
   const funcStack = await cfnClient
     .describeStackResources({ StackName: rootStackId, LogicalResourceId: `function${functionName}` })
     .promise();
+
+  if (!funcStack.StackResources || funcStack.StackResources.length === 0) {
+    throw new AmplifyFault('ResourceNotFoundFault', {
+      message: `Could not find function ${functionName} in root stack ${rootStackId}`,
+    });
+  }
```
Amplifiyer, line 138: 1. Do we know why this situation can happen? 
2. And what customers can do to unblock themselves?

lazpavel: we do not know, I didn't find any information in our usage data to help, so this is a first step for us to get more information if it still happens so we can fix the root cause

Amplifiyer (approved):

PR #12088, opened 2023-02-23
fix: update error details in AmplifyFault during category push

Amplifiyer:
#### Description of changes

update error details in AmplifyFault during category push

#### Issue #, if available

N/A

N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
+++ b/packages/amplify-cli/src/extensions/amplify-helpers/push-resources.ts
@@ -124,6 +124,7 @@ export const pushResources = async (
           'PushResourcesFault',
           {
             message: err.message,
+            details: err.details,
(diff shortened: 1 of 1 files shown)


evcodes (approved):
LGTM

danielleadams (commented):
Do we have tests for this behavior? If so, would it be possible to add one or change one so that it fails if this is breaks?

Amplifiyer:
> Do we have tests for this behavior? If so, would it be possible to add one or change one so that it fails if this is breaks?

Unfortunately there are no tests for this method (except for some trivial context tests). 

I can refactor this method in smaller units and export them to be testable, but ideally they shouldn't really be made public. I don't think I have an easy way to unit test this without exploding the scope of this PR.

danielleadams (approved):
That's what I figured. I'll approve since the level of effort to write a test overcomplicates the change.

PR #12094, opened 2023-02-23
chore: add types to exeInfo in context

kuhlmanp:
#### Description of changes
Initial changes to provide first layer parameters for ExeInfo type in context objects. Out of scope for this PR are changes that reduce the areas of instantiation as an empty object and ensuring that all properties are non-null. Properties whose types are known primitives have been assigned.

Ran some basic manual testing and unit tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli-core/API.md
```diff
@@ -747,6 +747,32 @@ export type EventPrefix = 'pre' | 'post';
...
+    metaData?: $TSAny;
+    pinpointApp?: $TSAny;
+    pinpointClient?: $TSAny;
+    pinpointInputParams?: $TSAny;
+    projectConfig?: $TSAny;
+    restoreBackend?: boolean;
+    serviceMeta?: $TSAny;
+    sourceEnvName?: string;
+    teamProviderInfo?: $TSAny;
+};
```
sobolk, line 774: This type looks like set of randomly associated properties. Which might also be redundant with other data that's already in context.
Should we attempt to split it ? Or at least create types for property groupings that make some sense together ?

kuhlmanp: This work is primarily here to facilitate the code coverage work Spencer and I are working on. Refactoring it better should absolutely be a future task. Right now I'm writing things almost solely to understand and be able to mock what's in exeInfo.

sobolk: Yes, but this type is going to spread around pretty fast, so we should get this right.
I suggest we spend some time understanding consumption patterns of this and if they're orthogonal we should type it as union of smaller types.

kuhlmanp: Updated types for better bucketing derived from explicitly logical and derived from an analysis of instantiation across all files. I don't think this is perfect, but I think it's better than what I had.

packages/amplify-cli-core/src/types.ts
```diff
@@ -10,6 +10,39 @@ import { Tag } from './tags';
  */
 export type $TSAny = any; // eslint-disable-line  @typescript-eslint/no-explicit-any
 
+export type PinpointInfo = {
```
jhockett, line 13: Instead of using `$TSAny`, I think `Record<string, any>` would provide a tad more insight into these various types.

There's also a few specific types like `$TSMeta` for `amplifyMeta` and `$TSTeamProviderInfo` for `teamProviderInfo`. Those types are essentially aliases of `$TSAny`, but I think it would be good to change those to `Record<string, any>` too.

kuhlmanp: These are things that Spencer and I are considering moving forward. The next PR will address non-primitive types and do more re-typing where I will verify whether $TSMeta, $TSTeamProviderInfo, etc are applicable to these properties. The reason I want to do this in multiple PRs is to keep things moving relatively quickly, do due diligence since we don't actually know what's inside exeInfo (it may be $TSMeta, but it may be a $TSMeta with other fields added as well) and, finally, to hopefully have real types instead of $TSMeta -> $TSAny.

Record<string, any> is something I could do here - but it wasn't clear to met yet what the benefits would be.

jhockett: I suggested it because `Record<string, any>` would narrow the type definition to objects and prevent accidentally assigning something like a boolean to one of the fields currently marked as $TSAny. Not a blocker for this PR though.

packages/amplify-cli-core/API.md
```diff
@@ -1133,6 +1140,13 @@ export interface INotificationsResourceMeta {
 // @public (undocumented)
 export type InputOptions = Record<string, string | boolean>;
 
+// @public (undocumented)
+export type InputParameters = {
+    forcePush?: boolean;
+    iterativeRollback?: boolean;
+    restoreBackend?: boolean;
+};
```
sobolk, line 1148: I think I saw something similar on @sdstolworthy PR - the one we were discussing `Record<string, string` and "core parameters" vs category specific parameters. Could you check with @sdstolworthy if this isn't duplicating his work ?

packages/amplify-cli-core/src/types.ts
kuhlmanp: I see, that makes sense.

packages/amplify-cli-core/API.md
sdstolworthy: @sobolk For these three properties, I think it's appropriate to have them separate here. Looking through the code, they do all certainly relate to similar input properties, they are explicitly set on `exeInfo` in `amplify-cli`, and referenced in one or two other packages. I think it's fine in this case to definitely type them here. But I agree, @kuhlmanp, I would look for a name other than `InputParameters` for this grouping.  `ExecutionInputOptions`, `ExecutionInputFlags` maybe?

sobolk: If typescript has anonymous types maybe that's an option, so we don't claim good names prematurely.
Otherwise I think prefixing is good option. I.e. `ExecInfoInputParameters`.
Or maybe we create namespace for these types.

packages/amplify-cli-core/src/types.ts
kuhlmanp: Oh! Yeah, that makes sense. But I hope to have solid types for objects as I go. I will do this for other types in my next PR. Thanks!

packages/amplify-cli-core/API.md
kuhlmanp: Took the namespacing option. These are good points. Thanks!

sobolk (approved):

sdstolworthy (approved):

PR #12087, opened 2023-02-23
fix: depend on inquirer in amplify-cli-core

sdstolworthy:
#### Description of changes
Adds a dependency on `inquirer@^7.3.3` in `amplify-cli-core`.

#### Description of how you validated changes

Built an application that leverages `AmplifyHelpers.addResourceDependency` using a version of `@aws-amplify/cli-extensibility-helper` on local dev that depended on the patched `amplify-cli-core`.

image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #12095, opened 2023-02-23
chore: add dependency to custom resource test

sobolk:
#### Description of changes

This is follow up on https://github.com/aws-amplify/amplify-cli/pull/12087 .

This PR:
- adds usage of `AmplifyHelpers.addResourceDependency` in custom resources e2e test
- adds swapping `cli-extensiblity-helper` for `latest` and pulls it from verdaccio, so if changes are made to that package or it's dependencies it will get tested.

**Running e2e tests before absorbing fix from https://github.com/aws-amplify/amplify-cli/pull/12087 .
Test run fails**
image
image

**Then testing with the fix.** (i.e. after merging latest dev)

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-e2e-core/src/categories/custom.ts
```diff
@@ -42,3 +47,10 @@ export function buildCustomResources(cwd: string) {
       });
   });
 }
+
+export const useLatestExtensibilityHelper = (projectRoot: string, customResourceName: string) => {
+  const packageJsonPath = path.join(projectRoot, 'amplify', 'backend', 'custom', customResourceName, 'package.json');
+  const packageJson: Record<string, Record<string, string>> = JSONUtilities.readJson(packageJsonPath);
+  packageJson.dependencies['@aws-amplify/cli-extensibility-helper'] = 'latest';
```
sobolk, line 54: This is accompanied by starting verdaccio on VM (we already pull verdaccio cache). In that case latest is either from source (if there was any change) or from npm otherwise.

.circleci/config.base.yml
```diff
@@ -1300,6 +1300,9 @@ commands:
                 command: |
                   source .circleci/local_publish_helpers.sh
                   source $BASH_ENV
+                  startLocalRegistry "$(pwd)/.circleci/verdaccio.yaml"
```
akshbhu, line 1303: Is this needed for adding test ?

akshbhu (approved):

.circleci/config.base.yml
sobolk: yes, see my comment above ^

packages/amplify-e2e-core/src/categories/custom.ts
akshbhu: understood , is this needed if we have set it with major version to simulate customer pkg.json?

sobolk: Yes. publish to local registry step does this https://github.com/aws-amplify/amplify-cli/blob/f03d170ae01d42a934f7b1392762f4d231ab687a/.circleci/publish.sh#L38-L41 .

But. The version that lands in registry is "tagged" for example `2.4.11-rc.021a139b2.0` . Soo '^2.0.0' won't match it. And figuring exact version is PITA, plus the package is not always published to local registry (if there's no change there or it's dependencies). `latest` covers all cases with minimum energy.

image

.circleci/config.base.yml
sobolk: Btw. I was hesitating to add this across the board. But the dent on run time is `0.05 s` . So it seems that trying to being precise and do this in just one job is not worth the effort.

sdstolworthy (approved):

PR #12109, opened 2023-02-24
fix: bucket already exists error

goldbez:
#### Description of changes
Customers sometimes get a deployment error that states that an S3 bucket already exists. More than 1.75% of customers (55 out of 3148) ran into this issue while running push/publish. This error is caused because Amplify storage buckets have a retain policy to prevent accidental data loss and are named by appending the stack ID to the given name. So the issue arrises if a user removes storage and then add it again with the same bucket name. 

This PR adds more robust error handeling by throwing an AmplifyError with a suggested remediation. 

This PR also adds logic to serialize/deserialize cloudformation error detail stacks. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -655,7 +655,32 @@ const updateCloudFormationNestedStack = async (
...
+    const s3Indicator = '(AWS::S3::Bucket)';
+    const alreadyExistsSuffix = 'already exists';
+    if (error.details.includes(s3Indicator) && error.details.includes(alreadyExistsSuffix)) {
+      const prefix = `Name: ${
+        error.details.includes('CustomMessageConfirmationBucket') ? 'CustomMessageConfirmationBucket' : 'S3Bucket'
+      } (AWS::S3::Bucket), Event Type: create, Reason:`;
+      const bucketName = error.details.slice(prefix.length, error.details.length - alreadyExistsSuffix.length);
+      throw new AmplifyError('ResourceAlreadyExistsError', {
+        message: `The S3 bucket ${bucketName} already exists.`,
+        resolution: 'Please delete this bucket in the AWS S3 console and try again.',
```
akshbhu, line 670: Can we throw a link to the s3 bucket in resolution ?

akshbhu (changes requested):
Can we also add unit tests for the functionality ?

akshbhu (dismissed):

akshbhu (dismissed):

packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -78,6 +78,7 @@ import { buildOverridesEnabledResources } from './build-override-enabled-resourc
 
 import { invokePostPushAnalyticsUpdate } from './plugin-client-api-analytics';
 import { minifyJSONFile } from './utils/minify-json';
+import { handleCloudFormationError } from './handle-cloud-formation-error';
```
sobolk, line 81:
```suggestion
import { handleCloudFormationError } from './cloud-formation-error-handler';
```

sobolk (dismissed):

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
```diff
@@ -0,0 +1,34 @@
...
+      err,
+    );
+  }
+
+  const s3Indicator = '(AWS::S3::Bucket)';
+  const alreadyExistsSuffix = 'already exists';
+  if (err?.details?.includes(s3Indicator) && err?.details?.includes(alreadyExistsSuffix)) {
+    const prefix = `Name: ${
+      err.details.includes('CustomMessageConfirmationBucket') ? 'CustomMessageConfirmationBucket' : 'S3Bucket'
+    } (AWS::S3::Bucket), Event Type: create, Reason:`;
```
evcodes, line 25: Should this error handling be in the handleCloudFormation Error? This seems more specific to S3 errors. Would it be more maintainable to add a separate function that handles S3 errors and branch out from this function if those come up?

evcodes: Further, is there an error code associated with this error message? It seems like searching for that error string can lead to a breaking change in the future if S3 changes their error message?

eddy2048 (commented):
This is a great addition! I have a few comments to make this change more maintainable and scalable to other cloudformation errors that will come up

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
goldbez: This is a CloudFormation error when deploying an S3 bucket. I think that it should go in this function since this is the function where all deployment errors are passed. I agree with you that if the error message is changed it could cause this error to go uncaught, but that is all the information in the error object that there is to work from.

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
```diff
@@ -0,0 +1,34 @@
...
+    const bucketName = err.details.slice(prefix.length + 1, err.details.length - alreadyExistsSuffix.length - 1);
```
Amplifiyer, line 26: 1. This won't work in few cases since `err.details` is a `\n` separated string of all CFN errors. The s3 bucket error may or may not be the first one. e.g a sample `err.details` 
```
Name: <stackName> (AWS::CloudFormation::Stack), Event Type: update, Reason: Parameters: [a1] must have values
Name: S3Bucket (AWS::S3::Bucket), Event Type: create, Reason: <bucketName> already exists
```
2. There are errors with S3 bucket's resource name as DeploymentBucket e.g. `Name: DeploymentBucket (AWS::S3::Bucket), Event Type: create, Reason: <bucketName> already exists`. Is the root cause not the same for this occurrence?

We should look into adding a type for serializing and deserializing the CFN errors. The serialization happens [here](https://github.com/aws-amplify/amplify-cli/blob/137b629c9e74e4ab8ec90f4519f947ed748f47cf/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js#L145-L155)

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
```diff
@@ -0,0 +1,40 @@
...
+        message: 'Deployment is already in progress.',
+        resolution: 'Wait for the other deployment to finish and try again.',
+        code: (err as $TSAny).code,
+      },
+      err,
+    );
+  }
+
+  const s3Indicator = '(AWS::S3::Bucket)';
+  if (err?.details?.includes(s3Indicator)) {
```
evcodes, line 35:
```suggestion
  if (err?.details?.includes('(AWS::S3::Bucket)')) {
```

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
evcodes: This is an interesting idea, maybe we can deserialize it into something that is more predictable and consistent?

goldbez: 1. Thank you for catching this. I missed the possibility of multi-line details fields. I just updated the PR to first find the line that contains this error and then parse the bucket name from just this line.
2. I was not able to reproduce the DeploymentBucket error yet. Though if it occurs when pulling an existing project, I do not think that it is the same root cause.

goldbez: I implemented functions to serialize/deserialize, as well as types to hold the deserialized data.

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const handleS3Error = (err: Error & { details?: string }): void => {
+  const alreadyExistsSuffix = 'already exists';
+  const deserializedErrorMessages: ErrorMessages = deserializeErrorMessages(err.details);
+  const bucketExistsLines: Array<ErrorMessage> = deserializedErrorMessages.messages.filter(
+    (message) => message.name.includes(s3Indicator) && message.reason.includes(alreadyExistsSuffix),
+  );
+  if (bucketExistsLines.length) {
+    const messageWithError: ErrorMessage = bucketExistsLines[0];
+    const bucketName = messageWithError.reason.slice(0, messageWithError.reason.length - alreadyExistsSuffix.length - 1);
```
Amplifiyer, line 14: A slightly better way would be a regex matching, something like

```ts
const bucketError = 'Name: S3Bucket (AWS::S3::Bucket), Event Type: create, Reason: some-bucket-dev already exists\n'
const bucketRegex = /Reason: (.*) already exists/;
console.log(bucketError.match(bucketRegex)![1]);
```

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
```diff
@@ -0,0 +1,44 @@
...
+    throw new AmplifyError('ResourceAlreadyExistsError', {
+      message: `The S3 bucket ${bucketName} already exists.`,
+      resolution: `Please delete this bucket in the AWS S3 console and try again. The bucket can be found at: https://s3.console.aws.amazon.com/s3/buckets/${bucketName}.`,
+    });
```
Amplifiyer, line 18: I believe there are three different types of buckets causing this issue (and possibly 3 different root causes): `CustomMessageConfirmationBucket`, `DeploymentBucket` and `S3Bucket` from the storage category.

Is this the right error message and resolution for all of them?

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
```diff
@@ -0,0 +1,46 @@
...
+import { deserializeErrorMessages, CFNErrorMessage, CFNErrorMessages } from './aws-utils/cloudformation-error-serializer';
+
+const s3Indicator = '(AWS::S3::Bucket)';
+
+const handleS3Error = (err: Error & { details?: string }): void => {
+  const alreadyExistsSuffix = 'already exists';
+  const deserializedErrorMessages: CFNErrorMessages = deserializeErrorMessages(err.details);
+  const bucketExistsLines: Array<CFNErrorMessage> = deserializedErrorMessages.messages.filter(
+    (message) =>
+      message.name.includes(s3Indicator) && message.reason.includes(alreadyExistsSuffix) && !message.name.includes('DeploymentBucket'),
```
Amplifiyer, line 11: nit: Similar to below a regex can be used instead of multiple includes
```ts
const bucketError = 'Name: S3Bucket (AWS::S3::Bucket), Event Type: create, Reason: some-bucket-dev already exists\n'
const bucketRegex = /Name: (S3Bucket|CustomMessageConfirmationBucket) \(AWS::S3::Bucket\), .* already exists.*/;
console.log(bucketRegex.test(bucketError));
```

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
```diff
@@ -0,0 +1,46 @@
+import { $TSAny, AmplifyError } from 'amplify-cli-core';
+import { deserializeErrorMessages, CFNErrorMessage, CFNErrorMessages } from './aws-utils/cloudformation-error-serializer';
+
+const s3Indicator = '(AWS::S3::Bucket)';
+
+const handleS3Error = (err: Error & { details?: string }): void => {
```
Amplifiyer, line 6: Can you add a quick comment on this method on why this is needed?

Amplifiyer (dismissed):
Looks good, left minor suggestions

Amplifiyer (approved):

packages/amplify-provider-awscloudformation/src/cloud-formation-error-handler.ts
```diff
@@ -0,0 +1,46 @@
...
+        message: 'Deployment is already in progress.',
+        resolution: 'Wait for the other deployment to finish and try again.',
+        code: (err as $TSAny).code,
+      },
+      err,
+    );
+  }
+
+  if (err?.details?.includes(s3Indicator)) {
+    handleS3Error(err);
```
evcodes, line 42: much nicer and cleaner! Eventually when we run into more CFN errors that are not correctly handled we can move this to be a switch statement. Overall, awesome work!

eddy2048 (approved):

PR #12119, opened 2023-02-27
fix: remove unnecessary update of pluginDirectories with 'packaged-node-modules'

auvred:
#### Description of changes
`amplify plugin scan` every time adds `"packaged-node-modules"` to `pluginDirectories` list in `~/.amplify/plugins.json`. So after several plugin scanning it looks like this:
```json
"pluginDirectories": [
  "cli-local-node-modules",
  "cli-parent-directory",
  "global-node-modules",
  "packaged-node-modules",
  "packaged-node-modules",
  "packaged-node-modules",
  "packaged-node-modules",
  "packaged-node-modules",
  "packaged-node-modules",
  "packaged-node-modules"
]
```
 I've added check if `"packaged-node-modules"` already exists in this list

I've build cli binary and tested it locally. I also ran `yarn test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/plugin-helpers/scan-plugin-platform.ts
+++ b/packages/amplify-cli/src/plugin-helpers/scan-plugin-platform.ts
@@ -38,12 +38,16 @@ export async function scanPluginPlatform(pluginPlatform?: PluginPlatform): Promi
     await sequential(scanUserLocationTasks);
   }
 
-  if (isPackaged && !pluginPlatform!.pluginDirectories.includes(constants.PACKAGED_NODE_MODULES)) {
-    pluginPlatform!.pluginDirectories.push(constants.PACKAGED_NODE_MODULES);
+  if (!pluginPlatform.pluginDirectories) {
+    pluginPlatform.pluginDirectories = new PluginPlatform().pluginDirectories;
   }
 
-  if (pluginPlatform!.pluginDirectories.length > 0 && pluginPlatform!.pluginPrefixes.length > 0) {
-    const scanDirTasks = pluginPlatform!.pluginDirectories.map((directory) => async () => {
+  if (isPackaged && !pluginPlatform.pluginDirectories.includes(constants.PACKAGED_NODE_MODULES)) {
+    pluginPlatform.pluginDirectories.push(constants.PACKAGED_NODE_MODULES);
+  }
+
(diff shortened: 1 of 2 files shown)


auvred:
For some reason, the `build_pkg_binaries_win` throws an [error](https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/16627/workflows/67cc9201-b68d-42cf-8f28-085a273c6842/jobs/807718?invite=true#step-104-69), I'm not sure if it's related to my changes...

sobolk:
@auvred not related to your PR. Other builds show this as well.

packages/amplify-cli/src/plugin-helpers/scan-plugin-platform.ts
```diff
@@ -38,12 +38,16 @@ export async function scanPluginPlatform(pluginPlatform?: PluginPlatform): Promi
     await sequential(scanUserLocationTasks);
   }
 
-  if (isPackaged) {
-    pluginPlatform!.pluginDirectories.push(constants.PACKAGED_NODE_MODULES);
+  if (!pluginPlatform.pluginDirectories) {
+    pluginPlatform.pluginDirectories = new PluginPlatform().pluginDirectories;
   }
```
sobolk, line 43: This might be removing non-null assertion warning, but it's not great way of achieving that. Previous version of the change was better.
I suggest to 
- either, revert to previous version + increment max warning threshold
- or, address the warning in this whole function (this would be preferred).

as for the later if we do this
```
export async function scanPluginPlatform(pluginPlatform?: PluginPlatform): Promise<PluginPlatform> {
  const pluginPlatformLocal = pluginPlatform || readPluginsJsonFile() || new PluginPlatform();

  pluginPlatformLocal.plugins = new PluginCollection();
```
and use `pluginPlatformLocal` below, this will eliminate plenty of those eslint warnings.

sobolk (approved):

packages/amplify-cli/src/plugin-helpers/scan-plugin-platform.ts
auvred: Fixed it, thanks!

akshbhu (approved):

sobolk:
Thanks @auvred for contribution.

Issue #12118, opened 2023-02-27
Help prompt should not allow you to select a resource when removing a category


evcodes:
### How did you install the Amplify CLI?

npm install -g @aws-amplify/cli

### If applicable, what version of Node.js are you using?

16.18.0

### Amplify CLI Version

10.7.3

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

Added hosting to my project and then tried to delete the hosting category with command `amplify delete hosting`. I had the option of selecting a plugin module to execute.
![Screenshot 2023-02-27 at 11 32 45 AM](https://user-images.githubusercontent.com/22283303/221622932-8130b496-3725-48a6-9738-a33121ef5572.png)
[](url)

### Describe the bug

Allowing me to select a module to execute when inputting the wrong command.

### Expected behavior

Just go straight from the Warning to the help usage commands without allowing me to select anything from hosting. Customer might get confused thinking that delete 

### Reproduction steps

1. npm install -g @aws-amplify/cli
2. amplify init
4. amplify add hosting
5. amplify delete hosting

### Project Identifier

2d42ef56-95b2-4e97-85af-8bd1019b1252

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Before submitting, please confirm:

- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

josefaidt:
This is related to https://github.com/aws-amplify/amplify-cli/issues/12065, where `amplify override hosting` has a similar result

evcodes:
Closed with merged PR

PR #12132, opened 2023-02-28
fix: do not route to plugin when command is invalid

goldbez:
#### Description of changes
When commands are deemed invalid by the input-manager, the command is set to "help" but the plugin is not changed. This leads to invalid commands still being routed to non-core plugins, which sometimes results in users being prompted, even after a message showing that the command is invalid is printed. This change stops commands from being routed to plugins by setting the plugin to core. The plugin that executes the help command does not matter, as all plugins yield the same top-level help. See the first photo below for an example of a prompt appearing after an invalid command message and the second photo for the new behavior.

image

image

#### Issue #, if available
#12118 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)

### DIFF
--- a/packages/amplify-cli/src/index.ts
+++ b/packages/amplify-cli/src/index.ts
@@ -101,6 +101,7 @@ export const run = async (startTime: number): Promise<void> => {
   if (!verificationResult.verified) {
     if (verificationResult.helpCommandAvailable) {
       input.command = constants.HELP;
+      input.plugin = constants.CORE;
     } else {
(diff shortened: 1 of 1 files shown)


evcodes:
Excellent fix!

evcodes (approved):

jhockett (approved):

PR #12148, opened 2023-03-02
chore: migrate finish-release script to typescript

sdstolworthy:
#### Description of changes

This migrates the `finish-release` bash script to Typescript. 

There are some changes to the behavior in the script:

1. Instead of pushing to `dev`, the script creates a new branch `dev-main-merge-RELEASE_SHORT_SHA`, and pushes that branch to main repository. 
2. It then prompts the user to open a pull request between the merge branch
3. It does not delete the branch at the end of the script. Instead, the script fails if the `dev-main-merge-SHA` branch already exists
4. It introduces a `--continue` flag. This takes the place of using `kill` to suspend the process. The script attempts to merge `main` and `dev` together on the merge branch. If there are conflicts, the script exits. It prompts the user to resolve the conflicts, continue the merge, and rerun the script with the `--continue` flag.

This PR would coincide with a change in the release runbook that encourages developers to use a PR to resolve conflicts between `main` and the working branch.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


scripts/finish-release.test.ts
```diff
@@ -0,0 +1,52 @@
+import * as finishRelease from './finish-release';
```
sobolk, line 1: should we have `scripts/__tests__/` dir?

scripts/finish-release.ts
```diff
@@ -0,0 +1,210 @@
...
+    DEV_BRANCH,
+    mergeBranch,
+  )}. To finish the release, open a pull request between ${mergeBranch} into ${DEV_BRANCH}
+`);
+
+  process.exit(0);
+}
+
+if (require.main === module) {
+  main().catch(() => process.exit(1));
```
sobolk, line 209: could we try this https://stackoverflow.com/questions/46515764/how-can-i-use-async-await-at-the-top-level ?

scripts/finish-release.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  if (!git.isCleanWorkingTree()) {
+    console.error('Please run the script from a clean working tree');
+    process.exit(2);
+  }
+
+  const releaseSha = git.getShortSha('main');
+  const DEV_BRANCH = 'dev';
+  const RELEASE_BRANCH = 'main';
+  const REPO_NAME = 'aws-amplify/amplify-cli';
```
sobolk, line 151: might be worth parametrizing all references to `aws-amplify/amplify-cli` in this script (with `aws-amplify/amplify-cli` as default). So that one can dry run scripts in our personal forks.
Btw. have you validated this script somehow?

scripts/finish-release.ts
edwardfoyle: my preferred way to do this is:
```
const main = async () => { /* stuff */ }

main().catch(console.error)
```

edwardfoyle (commented):
My only high level concern here is how do we ensure (or reduce the risk of) the PR getting squash merged? We need a merge commit or fast-forward merge to happen otherwise lerna will not be able to properly compute version bumps for the next release

scripts/finish-release.ts
```diff
@@ -0,0 +1,221 @@
...
+  }
+
+  git.push(upstreamName, mergeBranch);
+
+  console.info(`
+You can compare changes between the branches here: ${getCompareLink(
+    repository,
+    devBranch,
+    mergeBranch,
+  )}. To finish the release, open a pull request between ${mergeBranch} into ${devBranch}
```
sobolk, line 210: we should try https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#create-a-pull-request

scripts/finish-release.ts
sdstolworthy: I added additional parameters. I verified behavior with this command:

`yarn finish-release --release-branch update-finish-release-script --repository sdstolworthy/amplify-cli`

scripts/finish-release.ts
sdstolworthy: I have to add the `require.main == module` statement in there so I can import the file for unit tests. Otherwise the script executes when importing for tests.

scripts/finish-release.test.ts
sdstolworthy: Added

scripts/finish-release.ts
sobolk: Could you share links to commit shas/test PR in the fork ?

sdstolworthy:
image

scripts/finish-release.ts
sdstolworthy: https://github.com/sdstolworthy/amplify-cli/pull/10

sobolk (dismissed):
LGTM.

sdstolworthy:
> My only high level concern here is how do we ensure (or reduce the risk of) the PR getting squash merged? We need a merge commit or fast-forward merge to happen otherwise lerna will not be able to properly compute version bumps for the next release

https://github.com/sdstolworthy/amplify-cli/pull/11

Here's my first pass solution at this. Simply a default PR body that warns the developer to use a merge commit. Not great, but not awful. 

We can customize the template and get a bit fancier with it if we want.

https://github.com/aws-amplify/amplify-cli/pull/12148/files#diff-4615dcf33588816e74ed47d4e51da3212aabd72d2bd24e49f0cf913180d0ed1eR109

scripts/finish-release.ts
```diff
@@ -83,7 +83,7 @@ export class Git {
     if (branch) {
       command.push(branch);
     }
-    execSync(`git fetch ${remote} ${branch}`);
+    execSync(command.join(' '));
```
sobolk, line 86: :-)

sobolk (approved):

scripts/finish-release.ts
```diff
@@ -0,0 +1,244 @@
...
+  message: string;
+  mode: 'ff-only';
+}
+export class Git {
+  remote(verbose?: boolean): string {
+    let command = ['git', 'remote'];
+    if (verbose) {
+      command.push('-v');
+    }
+    return execSync(command.join(' ')).toString();
```
edwardfoyle, line 21: for all of these I would prefer to use execaSync from https://www.npmjs.com/package/execa (we already have dependency on this package). That way we don't have to worry about args with spaces in them in the future (which would break this logic)

sobolk (approved):

scripts/finish-release.ts
```diff
@@ -0,0 +1,247 @@
...
+  message: string;
+  mode: 'ff-only';
+}
+export class Git {
+  remote(verbose?: boolean): string {
+    let command = ['git', 'remote'];
+    if (verbose) {
+      command.push('-v');
+    }
+    return execa.sync(command[0], command.slice(1)).stdout;
```
edwardfoyle, line 21: nit: since it looks like all of these methods execute git commands, could we just define 'git' as a private readonly string and reuse everywhere rather than all this array indexing?

edwardfoyle (approved):
One more nit but I won't hold this up any longer if you don't want to address

scripts/finish-release.ts
sdstolworthy: I have a follow on PR queued up that will extract the git module to its own file. I'll address this in that change.

PR #12167, opened 2023-03-03
chore: automate smoke testing

sdstolworthy:
#### Description of changes

A script that automates smoke testing of the CLI.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


scripts/smoketest.ts
```diff
@@ -0,0 +1,397 @@
...
+  fs.ensureDirSync(projectDirectory);
+}
+
+export function removeProjectDirectory(projectDirectory: string) {
+  fs.removeSync(projectDirectory);
+}
+
+export type Evaluatable = {
+  evaluate(data: string, ptyProcess: pty.IPty): boolean;
+};
```
sobolk, line 74: This (and this file) looks a lot like nexpect and our e2e tests.
I'm wondering if we could just throw couple of e2e tests after `deploy` step in release/release_rc pipelines but use amplify installed from npm.

sobolk (commented):
Have you considered running this in circle ci after `deploy` step ?

scripts/smoketest.ts
sobolk: Perhaps nexpect should be extracted from e2e core OR this should become lightweight replacement for nexpect.

sdstolworthy: Yeah, the basic implementation pattern is from `nexpect`. I wanted to avoid having to tinker with nexpect and lift it out of `-e2e`, so I ended up just doing a lightweight reimplementation. I would personally vote for avoiding extracting `nexpect`, because I think that may be a larger change than it is worth. However, I think turning this into an e2e test is a sane path to take if we want to go that route.

sdstolworthy:
> Have you considered running this in circle ci after `deploy` step ?

I've considered it, but have not implemented it :) I could try my hand at that next week.

scripts/smoketest.ts
sobolk: We shouldn't have too many nexpect clones though. I'm a bit afraid this will grow organically into e2e core clone. So if we don't like it maybe we should start nexpectv2 outside of e2e core and then adopt it in e2e and smoke tests.

Given size and usefulness I'd say we should merge it. Let's see what others think as well.

sobolk:
> > Have you considered running this in circle ci after `deploy` step ?
> 
> I've considered it, but have not implemented it :) I could try my hand at that next week.

That would make this even more awesome and simplify runbooks.

danielleadams:
@sdstolworthy are you still working on this?

PR #12169, opened 2023-03-04
Refactor: exeInfo typing input params local env info

kuhlmanp:
Minimalist changes to get typing for inputParams and localEnvInfo. Where possible, using cast as unknown as <type> was used to minimize change footprint, retain current behavior, and make instantiation sites easy to identify.

Part of this strategy you may see as `<some object> as unknown as <some type>`. This prevents the compiler from complaining about required but uninstantiated properties. The gains from this practice are not altering current application behavior (by not instantiating required properties) and identifying instantiation sites that will need further work later on.

#### Issue #, if available
n/a

Ran manual basic testing, ran unit tests locally

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-analytics/src/analytics-resource-api.ts
```diff
@@ -22,6 +22,7 @@ import { invokeNotificationsAPIGetAvailableChannelNames } from './plugin-client-
 import { pinpointHasInAppMessagingPolicy } from './utils/pinpoint-helper';
 import { getAnalyticsResources } from './utils/analytics-helper';
 import { analyticsMigrations } from './migrations';
+import { LocalEnvInfo } from 'amplify-cli-core/src/types';
```
sdstolworthy, line 25:
```suggestion
import { LocalEnvInfo } from 'amplify-cli-core';
```

packages/amplify-cli-core/API.md
```diff
@@ -529,7 +529,7 @@ export type CommandLineInput = {
     plugin?: string;
     command: string;
     subCommands?: string[];
-    options?: CLIGlobalFlags & Record<string, any>;
+    options?: CLIGlobalFlags & Record<string, $TSAny>;
```
sdstolworthy, line 532: Why make this change from `any` to `$TSAny`?

packages/amplify-cli-core/API.md
```diff
@@ -779,6 +779,7 @@ export const executeHooks: (hooksMetadata: HooksMeta) => Promise<void>;
 
 declare namespace ExeInfo {
     export {
+        ExeInfo_2 as ExeInfo,
```
sdstolworthy, line 782: This is going to result in a type `ExeInfo.ExeInfo`. Try finding a more precise name for `ExeInfo.ExeInfo`.

packages/amplify-cli-core/API.md
```diff
@@ -1777,8 +1781,8 @@ type ProjectInfo = {
     existingProjectConfig?: $TSAny;
     existingTeamProviderInfo?: $TSAny;
     isNewProject?: boolean;
-    inputParams?: $TSAny;
-    localEnvInfo?: $TSAny;
+    inputParams: $TSObject;
```
sdstolworthy, line 1784: Why `$TSObject` vs `Record<string, unknown>`?

packages/amplify-category-storage/src/provider-utils/awscloudformation/index.ts
```diff
@@ -164,10 +164,8 @@ const isInHeadlessMode = (context: $TSContext) => {
 };
 
 const getHeadlessParams = (context: $TSContext) => {
-  const { inputParams } = context.exeInfo;
   try {
-    // If the input given is a string validate it using JSON parse
-    const { categories = {} } = typeof inputParams === 'string' ? JSONUtilities.parse(inputParams) : inputParams;
+    const { categories = {} } = context.exeInfo.inputParams;
```
sobolk, line 168: Could you:
- merge dev to pull latest eslint threshold
- address codeql warning
- if possible lower eslint warning threshold
?

packages/amplify-cli-core/src/exeInfo.ts
```diff
@@ -1,7 +1,10 @@
-import { $TSAny } from './types';
+import { $TSAny, $TSObject, LocalEnvInfo } from './types';
+
+// base type
+export type ExeInfo = EnvironmentInfo & InputParameters & PinpointInfo & ProjectInfo;
```
sdstolworthy, line 4: This is the line that is causing `ExeInfo.ExeInfo`. Because you are namespacing the types in the `ExeInfo` file as `ExeInfo` in `amplify-cli-core`'s `index.ts`, this is going to require a consumer to write code like this:

```typescript
import { ExeInfo } from 'amplify-cli-core';

let exeInfo: ExeInfo.ExeInfo = {};

```

That's not necessarily awful, but it does feel a bit repetitive. A clearer name, or possibly exporting it as default could be a path forward here.

sdstolworthy (changes requested):

sdstolworthy:
Resolve unused imports. Also, re-run `yarn extract-api` and repush.

packages/amplify-cli-core/API.md
kuhlmanp: I need to look into this. I had some code that I had created for a test, but it should have been deleted.

packages/amplify-cli-core/API.md
kuhlmanp: inputParams is initialized as {} in several places and the doc in types.ts mentioned that this should be used for those initializers. It seemed like the right thing to do at the time

packages/amplify-cli-core/src/exeInfo.ts
kuhlmanp: Yeah, this is something I meant to delete. I was speculating that it might be easier to have a type that encompasses the joined classes, but didn't end up using it anywhere. I think I should just delete it?

packages/amplify-cli-core/API.md
kuhlmanp: TLDR; wanted to take care of the warning. Posted in brainstorming about wanting to discuss proper path forward.

packages/amplify-cli-core/API.md
kuhlmanp: Was created as a way to easily type exeInfo but was not needed. Deleted for now - we can always re-add if we discover a use for it.

packages/amplify-category-storage/src/provider-utils/awscloudformation/index.ts
kuhlmanp: Should be resolved now.

sdstolworthy (approved):

packages/amplify-category-notifications/src/pinpoint-helper.ts
```diff
@@ -310,7 +310,7 @@ export const ensurePinpointApp = async (
   let pinpointApp: Partial<ICategoryMeta> | undefined;
   let resourceName;
   const amplifyMeta = context.exeInfo.amplifyMeta || stateManager.getMeta();
-  const envName = appEnvName || context.exeInfo.localEnvInfo.envName || stateManager.getCurrentEnvName();
+  const envName: string = appEnvName || context.exeInfo.localEnvInfo.envName || stateManager.getCurrentEnvName() || '';
```
sobolk, line 313: is fallback to `''` safe? would this change how this fails downstream ?

sdstolworthy: Looks like it is. It is only referenced subsequently in functions where a fallback blank string will be handled safely.

https://github.com/sdstolworthy/amplify-cli/blob/9fc6e16163f65fa3cf5e0b495e15a7e2b96d0ebe/packages/amplify-category-notifications/src/pinpoint-helper.ts#L169

https://github.com/sdstolworthy/amplify-cli/blob/9fc6e16163f65fa3cf5e0b495e15a7e2b96d0ebe/packages/amplify-category-notifications/src/pinpoint-name.ts#L11

sobolk (approved):

PR #12190, opened 2023-03-07
feat: add api status acm(access control matrix) subcommand and flag to help command

ykethan:
`amplify api status` had a flag `acm` which allows users to print a sub command matrix. the present help command does not contain this.

but it appears the `yarn build` is not picking up the help command changes. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


danielleadams (changes requested):
@ykethan are you able to add a test to this?

ykethan:
> @ykethan are you able to add a test to this?

@danielleadams this is just adding to the help doc or the help command reference.

danielleadams:
Right - but are there any tests that confirm the expected behavior from the code? Any tests that verify what is in the `commandsInfo` array? If this code gets deleted, how are we able to know?

josefaidt:
Hey @danielleadams here is the test for the help functionality https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-cli-core/src/__tests__/help.test.ts. Ideally this documentation will be generated off of a command definition, which would ensure help documentation is always in sync with the command definition. However in the meantime we are maintaining this blob to print the known commands, subcommands, options, etc.

edwardfoyle (approved):

danielleadams (approved):

PR #12218, opened 2023-03-13
chore: handle simple-git merge error

sdstolworthy:
#### Description of changes

Handles the error thrown by `simple-git`'s when attempting to merge when there are conflicts.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/scripts/finish-release.ts
+++ b/scripts/finish-release.ts
@@ -217,10 +217,12 @@ export async function main() {
   }
 
   console.info(`Merging ${upstreamName}/${releaseBranch} into ${mergeBranch}`);
-  git.merge(`${upstreamName}/${releaseBranch}`, { message: 'chore: merge release commit from main to dev' });
-
-  if (!git.isCleanWorkingTree()) {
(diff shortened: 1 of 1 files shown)


scripts/finish-release.ts
```diff
@@ -217,10 +217,12 @@ export async function main() {
   }
 
   console.info(`Merging ${upstreamName}/${releaseBranch} into ${mergeBranch}`);
-  git.merge(`${upstreamName}/${releaseBranch}`, { message: 'chore: merge release commit from main to dev' });
-
-  if (!git.isCleanWorkingTree()) {
```
evcodes, line 222: should we still keep this conditional?

sobolk (approved):

scripts/finish-release.ts
sdstolworthy: No, it was conditional previously because I was using a custom wrapper around `git`. I brought in `simple-git`, which throws an error if the merge fails due to conflicts. The try/catch here has effectively the same behavior as the `if` statement would have otherwise.

goldbez (approved):

PR #12231, opened 2023-03-15
feat:  replace role map lambda with cfn code

danielleadams:
#### Description of changes
This removes the Lambda callout custom resource that creates the role mapping for the identity pool, and adds it to the CFN code for the identity pool role attachments. The Lambda callout only runs when a user pool group is created, so the role attachment is only added to the CFN code with a user pool group.

- updated snapshots for unit tests
- added an e2e test to check that the identity pool has the expected role mappings
- added a migration e2e test to confirm the role mappings stay the same after migration

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-core/src/categories/auth.ts
```diff
@@ -1340,6 +1340,63 @@ export function updateAuthAddUserGroups(projectDir: string, groupNames: string[]
...
+      .sendCarriageReturn()
+      .wait('"amplify publish" will build all your local backend and frontend resources');
+
+    chain.run((err: Error) => {
+      if (!err) {
+        resolve();
+      } else {
+        reject(err);
+      }
+    });
```
jhockett, line 1396: nit: This can simplify to `chain.runAsync();`

With that change, the function can become an async function instead of returning a `new Promise(...)`.

jhockett (approved):
1 nit, but not a blocker. LGTM!

packages/amplify-migration-tests/src/__tests__/migration_tests_v10/__snapshots__/auth-add-all.migration.test.ts.snap
danielleadams, line 1: incorrect snapshot was generated and needs to be fixed

lazpavel (approved):

packages/amplify-e2e-core/src/utils/sdk-calls.ts
```diff
@@ -144,6 +145,20 @@ export const getUserPool = async (userpoolId, region) => {
...
+export const getIdentityPoolRoles = async (identityPoolId: string, region: string) => {
+  config.update({ region });
+
+  let res;
+
+  try {
+    res = await new CognitoIdentity().getIdentityPoolRoles({ IdentityPoolId: identityPoolId }).promise();
+  } catch (e) {
+    console.log(e);
+  }
```
sobolk, line 157: I noticed this is a pattern in this file. But is this a good practice?
What happens at the caller in this case? Does it handle undefined response gracefully?
What does undef response mean for the caller ? If it's handled gracefully does this mean CLI would produce different outcomes if transient error happen?

I think we should either have retry loop with exponential backoff OR such behavior should be configured when we spin up CongnitoIdentity client. In either case transient errors. if they persist, should propagate and stop user's command.

sobolk (commented):
LGTM. One commend wrt error handling.

packages/amplify-e2e-core/src/utils/sdk-calls.ts
danielleadams: > I noticed this is a pattern in this file. But is this a good practice?

I don't think this is a good pattern because it swallows the error and does not surface errors that arise from SDK, including on errors from the caller payload. I also don't want this pattern to leak into the production code where it's called from the CLI.

> What does undef response mean for the caller ?

Since these are test helpers, I imagine the desired/expected behavior would be that the test fails, but again, we aren't surfacing the error, so there isn't a good understanding of why the test failed which can cause more pain.

> If it's handled gracefully does this mean CLI would produce different outcomes if transient error happen?

I don't know if this is relevant because this is just the test calling straight to the SDK (or at least that's how I am using it) and not going through CLI.

> I think we should either have retry loop with exponential backoff OR such behavior should be configured when we spin up CongnitoIdentity client. In either case transient errors. if they persist, should propagate and stop user's command.

I think we can do this, but I think this is out of scope for this PR. I also am not sure if failed calls to SDK are a huge pain point in tests. I would think that the issues with incorrect payloads and/or failing to match SDK responses cause test failures, which wouldn't be solved by retries.

sobolk: woah, I didn't notice this is tests. nvm me.

sobolk (approved):

packages/amplify-e2e-core/src/utils/sdk-calls.ts
sobolk: But fwiw. I'd just remove try catch and let tests fail fast.

packages/amplify-e2e-core/src/utils/sdk-calls.ts
```diff
@@ -144,6 +145,20 @@ export const getUserPool = async (userpoolId, region) => {
   return res;
 };
 
+export const getIdentityPoolRoles = async (identityPoolId: string, region: string) => {
+  config.update({ region });
```
sobolk, line 149: I'd rather pass this in CognitoIdentity ctor. setting this in config (i.e. globally) may affect other clients and introduce cross test dependnecies.

danielleadams: done

sobolk (approved):

akshbhu (approved):

danielleadams:
There's been follow up tasks to write e2e headless tests

PR #12236, opened 2023-03-15
fix: Show cfn failure error details from api-category nested stack failures

Amplifiyer:
#### Description of changes
Currently we do not show any errors to the users when any of the resources failed from the second level nested stacks such as SearchableStack or DDB Table Stacks. This PR addresses that by building the list of all possible CFN resources and is used when deciding to display CFN failure errors.

Note: This does not address CFN failures caused by iterative deployments, which will be a separate PR.

#### Issue #, if available

All user complaints that amplify doesn't show why push command failed and only prints `Resource is not in the state stackUpdateComplete` such as https://github.com/aws-amplify/amplify-category-api/issues/92

Manually and unit testing

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -87,6 +87,7 @@ const ApiServiceNameElasticContainer = 'ElasticContainer';
 
 const optionalBuildDirectoryName = 'build';
 const cfnTemplateGlobPattern = '*template*.+(yaml|yml|json)';
+const nestedStackTemplateGlobPattern = 'stacks/*.+(yaml|yml|json)';
```
Amplifiyer, line 90: See [this](https://github.com/aws-amplify/amplify-category-api/blob/08b4f168ecb88b5d829f884fb10df3610634643f/packages/amplify-category-api/src/graphql-transformer/utils.ts#L297)

packages/amplify-provider-awscloudformation/src/__tests__/push-resources.test.ts
```diff
@@ -0,0 +1,78 @@
...
+    glob_mock.sync.mockImplementationOnce(() => [...dummyCFNFiles]);
+
+    // call the function under test
+    const { resourceDir, cfnFiles } = getCfnFiles('api', 'api-resource-name');
+    expect(resourceDir).toStrictEqual('backendTest/path/api/api-resource-name/build');
+    expect(fsSpy).toHaveBeenCalledWith('backendTest/path/api/api-resource-name/build');
+    expect(cfnFiles).toEqual(dummyCFNFiles);
+  });
+
+  it.only('does return API category stack from the build directory along with all nested stacks', async () => {
```
sdstolworthy, line 54: You have two tests with `only` and one without. Is this intentional? The test on line 67 will not run.

sdstolworthy (commented):
Double check the `test.only` calls

packages/amplify-provider-awscloudformation/src/__tests__/push-resources.test.ts
Amplifiyer: it's not, good catch. remnant of my testing. Updated

sdstolworthy (dismissed):

jhockett (approved):
LGTM, @Amplifiyer can you run `cloud-e2e` on this?

sobolk (approved):

jhockett (approved):

PR #12262, opened 2023-03-18
fix: skip backend sync on force push

lazpavel:
#### Description of changes
— skips backend sync on force push, as force push is ignoring resource status and always executes the push
— this also helps to override deployed backed in case if the deployed backed zip is corrupted

— manual changes
— unit tests added
— `yarn test` and `yarn e2e` pass

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/push.ts
+++ b/packages/amplify-cli/src/commands/push.ts
@@ -1,52 +1,11 @@
-import { $TSAny, $TSContext, AmplifyError, AmplifyFault, spinner, stateManager } from 'amplify-cli-core';
-import sequential from 'promise-sequential';
-import {
-  notifyFieldAuthSecurityChange,
-  notifyListQuerySecurityChange,
-  notifySecurityEnhancement,
-} from '../extensions/amplify-helpers/auth-notifications';
-import { getProviderPlugins } from '../extensions/amplify-helpers/get-provider-plugins';
+import { $TSAny, $TSContext, AmplifyError } from 'amplify-cli-core';
 import { updateCognitoTrackedFiles } from '../extensions/amplify-helpers/update-tracked-files';
-
-/**
- * Download and unzip deployment bucket contents to #current-cloud-backend so amplify status shows correct state
- */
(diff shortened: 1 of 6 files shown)


packages/amplify-cli/src/commands/push.ts
```diff
@@ -11,7 +11,7 @@ import { updateCognitoTrackedFiles } from '../extensions/amplify-helpers/update-
 /**
  * Download and unzip deployment bucket contents to #current-cloud-backend so amplify status shows correct state
  */
-const syncCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
+export const syncCurrentCloudBackend = async (context: $TSContext): Promise<void> => {
```
Amplifiyer, line 14: I know this is being exported for assertions in the unit tests, but exporting functions that are expected to be 'internal' can cause problems in the future. Is there any way we can skip exporting this and updateTrackedFiles (e.g. you can mock `updateCognitoTrackedFiles` instead)

lazpavel: I did update the updateTrackedFiles, for syncCurrentCloudBackend I prefer to not test on inner methods calls as that might change and miss the point of the test, I did however moved in a separate module

packages/amplify-cli/src/commands/push.ts
```diff
@@ -61,7 +20,12 @@ export const run = async (context: $TSContext): Promise<$TSAny> => {
   if (context.parameters.options?.force) {
     context.exeInfo.forcePush = true;
   }
-  await syncCurrentCloudBackend(context);
+
+  // force flag ignores project status check
+  if (!context.exeInfo.forcePush) {
+    await syncCurrentCloudBackend(context);
+  }
```
sobolk, line 27: Should we add an e2e test where we nuke deployment bucket and then attempt to force push ?

lazpavel: will do

sobolk (approved):

Amplifiyer (approved):

PR #12266, opened 2023-03-19
fix(mock): Swallow stack trace on an un-initialized project upon `amplify mock api`

manaswi223:
#### Issue #, if available
https://github.com/aws-amplify/amplify-category-api/issues/1324

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-cli-core/src/errors/amplify-error.ts
+++ b/packages/amplify-cli-core/src/errors/amplify-error.ts
@@ -20,5 +20,6 @@ export class AmplifyError extends AmplifyException {
    */
   constructor(name: AmplifyErrorType, options: AmplifyExceptionOptions, downstreamException?: Error) {
     super(name, 'ERROR', options, downstreamException);
+    this.stack = undefined;
   }
 }
--- a/packages/amplify-cli-core/src/errors/amplify-exception.ts
+++ b/packages/amplify-cli-core/src/errors/amplify-exception.ts
@@ -10,6 +10,7 @@ export abstract class AmplifyException extends Error {
   public readonly details?: string;
   public readonly link?: string;
   public readonly code?: string;
+  public readonly cause?: Error;
 
   /**
    * You should use AmplifyError or AmplifyFault to throw an exception.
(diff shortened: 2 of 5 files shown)


packages/amplify-cli-core/src/errors/amplify-error.ts
```diff
@@ -20,5 +20,6 @@ export class AmplifyError extends AmplifyException {
    */
   constructor(name: AmplifyErrorType, options: AmplifyExceptionOptions, downstreamException?: Error) {
     super(name, 'ERROR', options, downstreamException);
+    this.stack = undefined;
```
sobolk, line 23: we should not erase stack. we use that in telemetry

packages/amplify-cli-core/src/errors/amplify-exception.ts
```diff
@@ -10,6 +10,7 @@ export abstract class AmplifyException extends Error {
   public readonly details?: string;
   public readonly link?: string;
   public readonly code?: string;
+  public readonly cause?: Error;
```
sobolk, line 13: this property is redundant with `downstreamException`. Please remove it.

packages/amplify-cli/src/extensions/amplify-helpers/get-env-info.ts
```diff
@@ -3,12 +3,23 @@ import { $TSAny, AmplifyError, stateManager } from 'amplify-cli-core';
...
-  });
+    cause: new Error('`amplify init` not done'),
+  }); 
+  // handle error message gracefully by swallowing the stack trace
+} catch(error) {
+  if (!exceptionPrinted) {
+    exceptionPrinted = true;
+    console.error(`Error: ${error.message}\nResolution: ${error.resolution}`);
+  } 
+}
```
sobolk, line 24: We should not make this change here. `getEnvInfo` is called in multiple scenarios - we should not swallow the exception there.
If you're attempting to fix amplify mock api then I suggest you catch that error in the mock package and apply special handling there.

sobolk (changes requested):

packages/amplify-cli/src/extensions/amplify-helpers/get-env-info.ts
manaswi223: yes, I am attempting to make changes to the mock. Sure, thanks @sobolk, will make changes there : )

PR #12268, opened 2023-03-20
chore: mockAll adds capabilities for function and resource selection

ykethan:
#### Description of changes

`amplify mock` will allow the following

1.  mock with functions
2. prompts users to select the resources they would like to mock
3. allows multi select on functions to mock

Example:

image

image 

modified function mock flow and mock all flow to provide the functionality. `amplify mock function ` will now allow users to select functions to run.

manual testing. Working on adding test but could not find the necessary exports to use.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-util-mock/src/mockAll.ts
+++ b/packages/amplify-util-mock/src/mockAll.ts
@@ -1,6 +1,10 @@
 import { start as startAppSyncServer } from './api';
 import { start as startS3Server } from './storage';
+import { start as startLambdaServer } from './func';
+
 import { ServiceName as FunctionServiceName } from '@aws-amplify/amplify-category-function';
+import { prompter } from '@aws-amplify/amplify-prompts';
+
 const MOCK_SUPPORTED_CATEGORY = ['AppSync', 'S3', FunctionServiceName.LambdaFunction];
 
 export async function mockAllCategories(context: any) {
@@ -31,13 +35,25 @@ export async function mockAllCategories(context: any) {
       }
     }
     // Run the mock servers
+    const selectedMockableResources = await prompter.pick<'many', string>(
+      'Select the category',
+      mockableResources.map((r) => r.service).filter((value, index, self) => self.indexOf(value) === index),
+      {
+        returnSize: 'many',
+      },
+    );
+
     const serverPromises = [];
-    if (mockableResources.find((r) => r.service === 'AppSync')) {
+    if (selectedMockableResources.find((service) => service === 'Lambda')) {
+      await startLambdaServer(context);
+    }
+    if (selectedMockableResources.find((service) => service === 'AppSync')) {
       serverPromises.push(startAppSyncServer(context));
     }
-    if (mockableResources.find((r) => r.service === 'S3')) {
+    if (selectedMockableResources.find((service) => service === 'S3')) {
       serverPromises.push(startS3Server(context));
     }
+
     await Promise.all(serverPromises);
   } else {
     context.print.info('No resource in project can be mocked locally.');
--- a/packages/amplify-prompts/package.json
+++ b/packages/amplify-prompts/package.json
@@ -3,6 +3,7 @@
   "version": "2.6.5",
   "description": "Utility functions for Amplify CLI terminal I/O",
   "main": "lib/index.js",
+  "types": "lib/index.d.ts",
   "publishConfig": {
     "access": "public"
   },
--- a/packages/amplify-util-mock/src/func/index.ts
+++ b/packages/amplify-util-mock/src/func/index.ts
@@ -1,60 +1,65 @@
-import { getInvoker, category, isMockable, getBuilder } from '@aws-amplify/amplify-category-function';
 import * as path from 'path';
-import * as inquirer from 'inquirer';
+import { BuildType } from '@aws-amplify/amplify-function-plugin-interface';
+import { getInvoker, category, isMockable, getBuilder } from '@aws-amplify/amplify-category-function';
+import { prompter, printer } from '@aws-amplify/amplify-prompts';
 import { $TSContext, JSONUtilities, pathManager, stateManager } from 'amplify-cli-core';
 import _ from 'lodash';
-import { BuildType } from '@aws-amplify/amplify-function-plugin-interface';
 import { loadLambdaConfig } from '../utils/lambda/load-lambda-config';
 
 const DEFAULT_TIMEOUT_SECONDS = 10;
 
-export async function start(context: $TSContext) {
+export async function start(context: $TSContext): Promise<void> {
   const ampMeta = stateManager.getMeta();
-  let resourceName = context?.input?.subCommands?.[0];
-  if (!resourceName) {
+  const resourceNameInput = context?.input?.subCommands?.[0];
+  const resourceNames = [resourceNameInput];
+  if (!resourceNameInput) {
     const choices = _.keys(_.get(ampMeta, ['function'])).filter((resourceName) => isMockable(context, resourceName).isMockable);
     if (choices.length < 1) {
       throw new Error('There are no mockable functions in the project. Use `amplify add function` to create one.');
-    } else if (choices.length == 1) {
-      resourceName = choices[0];
+    } else if (choices.length === 1) {
+      resourceNames.push(choices[0]);
     } else {
-      const resourceNameQuestion = [
-        {
-          type: 'list',
-          name: 'resourceName',
-          message: 'Select the function to mock',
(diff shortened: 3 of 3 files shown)


packages/amplify-util-mock/src/__tests__/func/index.test.ts
```diff
@@ -53,24 +58,19 @@ describe('function start', () => {
...
-      blue: jest.fn(),
-    },
   };
 
   jest.setTimeout(1000 * 20);
 
   // NOTE: A warning from jest saying that async operations weren't stopped in the test is expected here
   // because the mock function is designed to keep running after the timeout to ensure that the timeout works
-  it('times out function execution at the default time', async () => {
+  it.only('times out function execution at the default time', async () => {
```
sobolk, line 67:
```suggestion
  it('times out function execution at the default time', async () => {
```

packages/amplify-util-mock/src/__tests__/func/multi-function.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import { $TSAny } from 'amplify-cli-core';
+
+jest.mock('../../func');
+jest.mock('@aws-amplify/amplify-prompts');
+jest.mock('amplify-cli-core', () => ({
+  ...jest.requireActual('amplify-cli-core'),
+  FeatureFlags: {
+    getNumber: jest.fn(),
+  },
+}));
```
sobolk, line 12: looks like mocking core requires special handling , see https://github.com/aws-amplify/amplify-cli/blob/2e4cfb92b87499b1cdf2595e72b13b1857a60cb5/packages/amplify-util-mock/src/__e2e__/utils/index.ts#L20-L26 .

packages/amplify-util-mock/src/__tests__/func/multi-function.test.ts
```diff
@@ -0,0 +1,49 @@
...
+          func2: {},
+          func3: {},
+        },
+      }),
+    },
+  };
+
+  it('prompts for multiple functions and selects the resource name', async () => {
+    await lambdaServerStart(context_stub);
+    expect(lambdaServerStartMock).toBeCalledTimes(1);
```
sobolk, line 47: What is this test testing?

packages/amplify-util-mock/src/mockAll.ts
```diff
@@ -31,13 +36,25 @@ export async function mockAllCategories(context: any) {
       }
     }
     // Run the mock servers
+    const selectedMockableResources = await prompter.pick<'many', string>(
+      'Select the category',
+      mockableResources.map((r) => r.service).filter((value, index, self) => self.indexOf(value) === index),
+      {
+        returnSize: 'many',
+      },
+    );
```
sobolk, line 45: how does DX look like for this ?

packages/amplify-util-mock/src/__tests__/func/multi-function.test.ts
ykethan: thank you, i was able to resolve it

packages/amplify-util-mock/src/mockAll.ts
ykethan: thank you for consistency and DX purposes changes the prompt display values from ['AppSync', 'S3', 'Lambda'] to ['GraphQL API','Storage','Function']
image

packages/amplify-util-mock/src/__tests__/func/multi-function.test.ts
ykethan: checking if the lambdaserver has been called , added additional expects to check if the function invoker has been called for each function.

packages/amplify-util-mock/src/__tests__/func/index.test.ts
ykethan: could not change due to the comment provided, as i did not write the original tests. changing it causes the tests to fail.

sobolk: The comment I left is unrelated to the comment in the code. 
if you look at code in `dev` branch there's no `.only` there and the test is passing. If tests are not passing after removing `.only` then either there is regression or tests have to change to accommodate new dx/implementation,

packages/amplify-util-mock/src/__tests__/func/multi-function.test.ts
sobolk: That would make sense if you were calling `lambdaServerStart` indirectly. I.e. 
if your call stack would look like `test -> higher level abstaction -> lambdaServerStart` . If you're calling into `lambdaServerStart` which is at this point mocked implementation then this test doesn't execute any logic.

packages/amplify-util-mock/src/mockAll.ts
sobolk: As far as I can read the code in `mockAll.ts` code `mockableResources` can contain multiple apis and functions.
What happens to this DX in that case? Should we list flat list of resources from each category or have 2 stage selection, i.e. first category and then expand all resources.
This DX becomes tricky if number of resources in each category is large, i.e. more than 10.

ykethan: we cant mock multiple api as a project can only have one graphql api, we dont mock rest api's.
we only mock graphql api, storage which is s3. additionally, i added functions as a option which was present in the file but never invoked(incomplete feature)

sobolk: That explains why assumptions and... shortcuts were made in current code...
What happens at this prompt if project has multiple functions? And what is desired DX?

ykethan: the DX would be as follows 
1. prompt user on the mockable categories present in the project 
2. user selects the categories
3. if function has been selected provide the function name for user to selec
4. start functions

https://user-images.githubusercontent.com/87995712/230987317-dd43e8a8-77a1-46fb-9859-7be2a1797a53.mov


5. start remaining categories

packages/amplify-util-mock/src/__tests__/func/index.test.ts
ykethan: corrected all the test after removing `.only`. ran `yarn test` in the package for check as well.

packages/amplify-util-mock/src/__tests__/func/multi-function.test.ts
ykethan: removed the `multi-function.test.ts` file and added the test to `index.test.ts` as we do have current test to call a single function on project with multiple functions. added test that calls multiple functions in a project improving on the existing test.

packages/amplify-util-mock/src/__tests__/mockAll.test.ts
```diff
@@ -0,0 +1,45 @@
+import { prompter } from '@aws-amplify/amplify-prompts';
+import { mockAllCategories } from '../mockAll';
+import { start as apiServerStart } from '../api';
+import { start as storageServerStart } from '../storage';
+import { start as lambdaServerStart } from '../func';
+import { $TSAny } from 'amplify-cli-core';
```
jhockett, line 6: This should now be `@aws-amplify/amplify-cli-core`. Please make sure you're working off the latest changes from dev.

packages/amplify-util-mock/src/__tests__/func/index.test.ts
```diff
@@ -128,23 +128,44 @@ describe('function start', () => {
...
 
     await start(context_stub_copy);
-
-    expect(inquirer_mock.prompt.mock.calls[0][0][0].choices).toStrictEqual(['func1', 'func2', 'func3']);
+    expect(prompter_mock.pick.mock.calls[0][1]).toStrictEqual(['func1', 'func2', 'func3']);
     expect(getBuilder_mock.mock.calls[0][1]).toBe('func2');
   });
 
+  it('project has multiple functions and we dont specify a specific function, prompts for function names', async () => {
+    const context_stub_copy = _.merge({}, context_stub);
```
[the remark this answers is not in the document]
ykethan: converted to spread operator and fixed another test that was failing

packages/amplify-util-mock/src/__tests__/mockAll.test.ts
ykethan: modified it to `@aws-amplify/amplify-cli-core` and yarn tested for any failures

jhockett (approved):

sobolk (approved):
Merge if you get e2e tests passing.

PR #12276, opened 2023-03-21
chore: increase parallelization in e2e tests

sobolk:
#### Description of changes

Increase parallelization factor in tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.circleci/config.base.yml
+++ b/.circleci/config.base.yml
@@ -515,7 +515,7 @@ jobs:
             source .circleci/local_publish_helpers.sh
             changeNpmGlobalPath
             cd packages/amplify-migration-tests
-            retry yarn run migration_v5.2.0 --no-cache --maxWorkers=3 --forceExit $TEST_SUITE
+            retry yarn run migration_v5.2.0 --no-cache --maxWorkers=4 --forceExit $TEST_SUITE
           no_output_timeout: 65m
       - run: *scan_e2e_test_artifacts
       - store_test_results:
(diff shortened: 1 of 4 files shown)


scripts/split-e2e-tests-v2.ts
```diff
@@ -155,7 +155,7 @@ const TEST_EXCLUSIONS: { l: string[]; w: string[] } = {
     'src/__tests__/pinpoint/notifications-pinpoint-config-util.ts',
   ],
 };
-const MAX_WORKERS = 3;
+const MAX_WORKERS = 4;
```
evcodes, line 158: can we do something like this for the base.yml and config.yml file?

evcodes (approved):
Nice! One nit but LGTM otherwise

scripts/split-e2e-tests-v2.ts
sobolk: Not worth the effort given timeline for this script.

.circleci/config.base.yml
```diff
@@ -515,7 +515,7 @@ jobs:
             source .circleci/local_publish_helpers.sh
             changeNpmGlobalPath
             cd packages/amplify-migration-tests
-            retry yarn run migration_v5.2.0 --no-cache --maxWorkers=3 --forceExit $TEST_SUITE
+            retry yarn run migration_v5.2.0 --no-cache --maxWorkers=4 --forceExit $TEST_SUITE
```
awsluja, line 518: I don't think we can actually use all 4 cores since jest needs 1 cpu to manage the worker pool & OS; maybe its working now ? I noticed when I tried 4 , it just runs the 4th test at the end of the first 3

sobolk: Setting maxWorkers beyond CPU count - 1 would be harmful if our tests were CPU bound. But they mostly spawn CLI and wait for it asynchronously.

awsluja (approved):

PR #12296, opened 2023-03-23
chore: convert generic errors that are user errors to AmplifyErrors

Amplifiyer:
#### Description of changes

chore: convert generic errors that are user errors to AmplifyErrors

Reason for this change:
1. Generic errors thrown if not caught upstream by an `AmplifyException` get converted and reported as `UnknownFaults`. See [this](https://github.com/aws-amplify/amplify-cli/blob/b36078e4408c5b87708cfa936e636e0d032e5239/packages/amplify-cli/src/amplify-exception-handler.ts#L158)
2. In most cases, generic errors are caught and wrapped in `AmplifyFaults` such as this [400 line try catch](https://github.com/aws-amplify/amplify-cli/blob/f39cddb132dbd21cc28ca5e777f33898f212f37c/packages/amplify-provider-awscloudformation/src/push-resources.ts#L474-L487) and starts reporting them as system faults. Converting these downstream errors to `AmplifyErrors` will help to recover it during reporting as implemented [here](https://github.com/aws-amplify/amplify-cli/blob/b36078e4408c5b87708cfa936e636e0d032e5239/packages/amplify-cli/src/amplify-exception-handler.ts#L91C3-L97)

#### Issue #, if available

N/A

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-function/src/commands/function/build.ts
```diff
@@ -31,7 +31,12 @@ export const run = async (context: $TSContext) => {
...
     context.print.info(err.stack);
     context.print.error('There was an error building the function resources');
-    void context.usageData.emitError(err);
+    const amplifyError = new AmplifyError(
+      'PackagingLambdaFunctionError',
+      { message: `There was an error building the function resources ${err.message}` },
+      err,
+    );
+    void context.usageData.emitError(amplifyError);
     process.exitCode = 1;
```
sobolk, line 40: Should we just throw this instead of process.exitCode = 1 ?

Amplifiyer: I was doing that but reverted after seeing that every top level command is doing the same. We need to tackle the `process.exitCode` in a separate PR and make sure it doesn't break anything

sobolk (approved):

akshbhu (commented):
if we are converting all userErrors to amplify Errors, Should we also include resolution step in this errors ?

Amplifiyer:
> if we are converting all userErrors to amplify Errors, Should we also include resolution step in these errors ?

We  definitely should, but likely not with this PR since there are lot many other instances of `AmplifyError` that are missing resolution field. I'll create a task for tracking that.

PR #12298, opened 2023-03-23
fix: handle bucket exists check for s3 buckets in opt in regions

lazpavel:
#### Description of changes
- workaround for aws-sdk bug where calling `headBucket` for a opt-in region bucket will throw and error if s3 client is initialized with a different region
- retry headBucket call on error with the bucket region as s3 client option

- manual testing
- unit and e2e tests pass

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/S3Service.ts
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/S3Service.ts
@@ -1,6 +1,7 @@
 import { $TSContext, AmplifyFault } from 'amplify-cli-core';
 import { IS3Service } from '@aws-amplify/amplify-util-import';
-import S3, { Bucket } from 'aws-sdk/clients/s3';
+import { S3 } from 'aws-sdk';
+import { Bucket } from 'aws-sdk/clients/s3';
 import { AwsSecrets, loadConfiguration } from '../configuration-manager';
 
 export const createS3Service = async (context: $TSContext): Promise<S3Service> => {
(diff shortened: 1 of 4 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/S3Service.ts
```diff
@@ -27,33 +20,25 @@ export class S3Service implements IS3Service {
...
-        this.cachedBucketList.push(...response.Buckets!);
+        this.cachedBucketList.push(...response.Buckets);
       }
     }
 
     return this.cachedBucketList;
   }
 
-  public async bucketExists(bucketName: string): Promise<boolean> {
+  public async bucketExists(bucketName: string, s3?: S3): Promise<boolean> {
```
sobolk, line 30: The point of `S3Service` abstraction is to hide `S3` . We should not pass `S3` as parameter at public API.

Two options:
1. create private `bucketExists(bucketName: string, s3?: S3)` and call into it from public version.
2. when we re-try call into `createS3Service` with overriden region and use it to retry exist call.

packages/amplify-provider-awscloudformation/src/aws-utils/S3Service.ts
```diff
@@ -27,33 +20,25 @@ export class S3Service implements IS3Service {
...
-        .promise();
+      const response = await s3Client.headBucket({ Bucket: bucketName }).promise();
       // If the return object has no keys then it means successful empty object was returned.
       return Object.keys(response).length === 0;
     } catch (error) {
-      if (error.code === 'NotFound') {
-        return false;
+      if (error.region !== s3Client.config.region) {
+        return this.bucketExists(bucketName, new S3({ ...s3Client.config?.credentials, region: error.region }));
       }
```
sobolk, line 39: Please add a comment that describes why are we doing this.

goldbez (approved):
LGTM

sobolk (approved):

PR #12307, opened 2023-03-24
fix: throws a warning for detector

akshbhu:
#### Description of changes
-catching the detector error if it fails

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/push-resources.ts
+++ b/packages/amplify-provider-awscloudformation/src/push-resources.ts
@@ -208,7 +208,12 @@ export const run = async (context: $TSContext, resourceDefinition: $TSObject, re
     await buildOverridesEnabledResources(context, resources);
 
     // print cdk migration warning
-    await printCdkMigrationWarning(context);
+    try{
+      await printCdkMigrationWarning(context);
+    }catch(error){
+      printer.warn(`Amplify detector failed with error ${error}`);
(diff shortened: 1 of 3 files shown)


evcodes (approved):

evcodes:
Is there a test that we should modify to capture this change?

packages/amplify-provider-awscloudformation/src/push-resources.ts
```diff
@@ -208,7 +208,12 @@ export const run = async (context: $TSContext, resourceDefinition: $TSObject, re
     await buildOverridesEnabledResources(context, resources);
 
     // print cdk migration warning
-    await printCdkMigrationWarning(context);
+    try{
+      await printCdkMigrationWarning(context);
+    }catch(error){
+      printer.warn(`Amplify detector failed with error ${error}`);
+      printer.warn(`To migrate to AWS CDK V2 : https://docs.amplify.aws/cli/migration/aws-cdk-migration/`);
+    }
```
sobolk, line 216: if we want to suppress errors from detector (which isn't bad idea) we should:
- move try-catch inside `printCdkMigrationWarning`
- don't print anything - if it's false positive we should just be quiet.

sobolk (dismissed):

evcodes (approved):

PR #12310, opened 2023-03-24
test: add e2e tests for push, init, pull, env with help flag

goldbez:
#### Description of changes
This PR adds e2e tests for push, init, pull, and env with the help flag. The chain.waits expect a certain output. If that output is not given, then the tests will fail.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/help.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/help.test.ts
@@ -5,8 +5,10 @@ import {
   deleteProjectDir,
   statusWithHelp,
   statusForCategoryWithHelp,
-  addS3AndAuthWithAuthOnlyAccess,
-  amplifyPushAuth,
+  pushWithHelp,
+  initWithHelp,
+  pullWithHelp,
+  envWithHelp,
 } from '@aws-amplify/amplify-e2e-core';
 
 describe('help happy paths', () => {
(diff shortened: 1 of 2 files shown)


packages/amplify-e2e-tests/src/__tests__/help.test.ts
```diff
@@ -24,10 +24,26 @@ describe('help happy paths', () => {
   });
 
   test('amplify status', async () => {
-    await statusWithHelp(projRoot, ['USAGE', 'amplify status [-v | --verbose]']);
+    await statusWithHelp(projRoot);
```
evcodes, line 27: what is this testing here? We should have an expect statement that we are testing against

goldbez: The chain.wait expects a certain output. If that output is not given, then the tests will fail.

evcodes (changes requested):
I think we need to add some expect cases that validate the tests

goldbez:
> I think we need to add some expect cases that validate the tests

I don't think that is necessary, since the chain.wait achieves the same thing in a simpler way. This is what it looks like when the chain.wait fails (by giving it the wrong string to expect), causing the whole test to fail:
image

evcodes (approved):
LGTM, thanks for clarifying

sdstolworthy (approved):

kuhlmanp (approved):

PR #12327, opened 2023-03-28
chore: revise net6 baseline

sobolk:
#### Description of changes

This PR lowers required baseline net6 version to `6.0.100` in lambda templates. So that users who install net6 via apt-get on ubuntu can use it without modifying their lambda projects.

This setting does not control what customer's net6 version is it just widens allowed range of versions.

See more here https://devblogs.microsoft.com/dotnet/dotnet-6-is-now-in-ubuntu-2204/

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-dotnet-function-template-provider/resources/lambda/global.json
+++ b/packages/amplify-dotnet-function-template-provider/resources/lambda/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.300",
+    "version": "6.0.100",
     "rollForward": "latestFeature"
   }
 }


packages/amplify-dotnet-function-template-provider/resources/lambda/global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "6.0.300",
+    "version": "6.0.100",
     "rollForward": "latestFeature"
```
sobolk, line 4: these settings together mean that customer can use .net6 that's between [6.0.100 , 7), i.e. any 6.x version after 6.0.100

lazpavel (approved):

packages/amplify-dotnet-function-template-provider/resources/lambda/global.json
akshbhu: Does current code `version": "6.0.300"`, throw error on existing ubuntu based dot net projects ?

sobolk: Then if you install from MSFT feeds (which is more work and recommended way) you'll get successful function build.
The change is meant to cover both cases.

sobolk: to answer the question.
if existing project works on ubuntu then either user had to install dotnet from MS feed or make change after generating function from template to allow 6.0.1xx.

akshbhu: Shall we do a strict checking here to make sure, existing functions build fails fast on existing projects  
https://github.com/aws-amplify/amplify-cli/blob/e914703261c8d4855b09cc0a5bc01a7b5de44591/packages/amplify-dotnet-function-runtime-provider/src/utils/detect.ts#L24. and  we can recommend the way to install ?

akshbhu (approved):

PR #12351, opened 2023-03-30
fix: catch undefined LogicalName when Lambda Layer versions are deleted from the console

sdstolworthy:
#### Description of changes
Provides a more descriptive error when attempting to use a Lambda Layer that was deleted in the AWS Console.

New unit test added.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
```diff
@@ -30,7 +30,12 @@ export function generateLayerCfnObj(isNewVersion: boolean, parameters: LayerPara
     layerCloudState.latestVersionLogicalId = logicalName;
     versionList.unshift({ LogicalName: logicalName, legacyLayer: false });
   } else {
-    logicalName = _.first(versionList).LogicalName;
+    logicalName = _.first(versionList)?.LogicalName;
+    if (!logicalName) {
+      throw new AmplifyError('LambdaLayerNotFoundError', {
+        message: 'No versions were found for the Lambda Layer. Were they deleted?',
```
sobolk, line 36: if this refers to layers stored in the cloud we should somehow convey where the deletion could have happen.
I'm assuming the deletion must have happened outside of amplify - should that be part of the question?

sobolk (approved):

sdstolworthy:
Reverting to draft. I need to do more investigation on this issue.

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/lambda-layer-cloudformation-template.ts
sdstolworthy: Tried to make the message a bit clearer. I was not able to reproduce the specific error, so I can't give much better guidance than to check the console at this point. I'm open to suggestions on messaging.

jhockett (dismissed):

packages/amplify-category-function/src/provider-utils/awscloudformation/utils/layerCloudState.ts
```diff
@@ -103,6 +103,11 @@ export class LayerCloudState {
         // eslint-disable-next-line no-param-reassign
         layerVersion.legacyLayer = layerVersion.LogicalName === undefined || layerVersion.LogicalName === 'LambdaLayer';
       });
+      if (layerVersionList.length === 0) {
+        throw new AmplifyError('LambdaLayerNotFoundError', {
+          message: 'No versions were found for the Lambda Layer. Were they deleted on the AWS Lambda Console?',
+        });
+      }
```
Amplifiyer, line 110: Should we do this before line 45 in this file? Maybe also add the layer name and environment in the error message?

sdstolworthy: Good suggestions. The error message now includes the name of the missing layers.

Amplifiyer (approved):

jhockett (approved):

PR #12378, opened 2023-04-04
chore: add resolution message to missing cloudformation error

lazpavel:
#### Description of changes
- added resolution message to missing CloudFormation template fault
- converted remove and add function from promise based to await async
- converted errors to AmplifyError in remove resource logic
- converted function remove prompt to use prompter library
- unit and e2e test fixes for the above changes

#### Checklist
- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli-core/src/cfnUtilities.ts
```diff
@@ -22,7 +23,11 @@ export function readCFNTemplate(filePath: string, options: Partial<typeof defaul
     if (options.throwIfNotExist === false) {
       return undefined;
     }
-    throw new Error(`No CloudFormation template found at ${filePath}`);
+    throw new AmplifyError('CloudFormationTemplateError', {
```
sobolk, line 26: is this user error or a fault? 
CLI manages these files not customer.

lazpavel: This is what Praveen also pointed out, we do not have enough data to say with confidence, based on the number of distinct customers having this issue it seems to be a fault, but for the cases we've been on calls this file was manually renamed, going to mark this PR as draft until we gain more insights

lazpavel:
remaining e2e failing test will be fixed in dev with this PR: https://github.com/aws-amplify/amplify-cli/pull/12467

Amplifiyer (approved):

PR #9062, opened 2021-11-24
fix(amplify-provider-awscloudformation): fix region mapping

watilde:
#### Description of changes
Update the default region of pinpoint for dedicated regions where need to map to `us-east-1` when uses with Cognito per [Amazon Cognito and Amazon Pinpoint Region mappings](https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-pinpoint-integration.html). This update allows developers to use `ap-northeast-1` for Cognito and `us-east-1` for pinpoint when needed.

#### Issue #, if available
Related to https://github.com/aws-amplify/amplify-cli/issues/1965.

#### Description of how you validated changes
1. `amplify configure` and select `ap-northeast-1` for region as an example
2. Make `latestRegionSupport.pinpoint` 0
3. `amplify add analytics`
4. Check the region and make sure it's `us-east-1`

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/master/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #12379, opened 2023-04-04
fix: update pinpoint regions

lazpavel:
#### Description of changes
- updated pinpoint region maps to include us-east-2: https://docs.aws.amazon.com/general/latest/gr/pinpoint.html
- adds support to map tests to specific regions instead of forcing to us-west-2

#### Checklist
- [x] PR description included
- [x] `yarn test` passes
- [x] migration test added to check that region mapping doesn't affect already created resources


sobolk (dismissed):

Amplifiyer (commented):
Can we confirm this is backwards compatible, specially with customers trying to update their pinpoint resource, will it delete and recreate the resource in the us-east-2 region?

lazpavel:
@Amplifiyer, from the standpoint of the pinpoint resource, we cannot update the existing one so a push will still keep the pinpoint resource from us-east-1 even after update, there are problems trying to add notifications channels as we use the mapping logic to build the pinpoint client instead of getting the region from the pinpoint resource which in turn fails with `Resource not found`. Marking it as draft as it requires some design

Amplifiyer:
We should test and get [this](https://github.com/aws-amplify/amplify-cli/pull/9062) merged as well as part of this PR

packages/amplify-category-notifications/src/pinpoint-helper.ts
```diff
@@ -11,20 +11,32 @@ import {
   $TSMeta,
   AmplifyError,
   AmplifyFault,
+  FeatureFlags,
```
sobolk, line 14: CodeQL says it's unused.

sobolk (approved):

Amplifiyer (approved):
LGTM

PR #12394, opened 2023-04-06
feat: create new NodeJS functions with nodejs18.x runtime

sobolk:
#### Description of changes

This PR:
- Changes NodeJS function templates to generate new functions with nodejs18.x runtime
- Adds test coverage for CRUD template - to test that AWS v3 SDK usage is correct
- Adds nodejs12.x as "from" migration (since it entered deprecation in lambda service)

Please note that this change does not affect existing functions (apart adding node12->node16 migration path).
We're not offering auto-migration from node 14 or 16 to 18 due to AWS v2 and v3 SDK incompatibility.
See https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html for reference.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-nodejs-function-template-provider/resources/lambda/crud/app.js.ejs
```diff
@@ -81,21 +81,20 @@ app.get(path + hashKeyPath, function(req, res) {
...
-    } else {
-      res.json(data.Items);
-    }
-  });
+  try {
+    const data = await ddbDocClient.send(new QueryCommand(queryParams));
+    res.json(data.Items);
+  } catch (err) {
+    res.statusCode = 500;
+    res.json({error: 'Could not load items: ' + err});
```
jhockett, line 89: Should this be `+ err` or `+ err.message`?

Line 133 is different.

yarn.lock
jhockett, line 1: I see multiple 3.x versions of aws sdk being resolved. It may be due to data related dependencies, but if not, can you update existing 3.x aws-sdk dependencies to the newer version?

packages/amplify-nodejs-function-template-provider/resources/lambda/crud/app.js.ejs
sobolk: That came from old promise based code without change. But it seems err.message is more appropriate here. I'll change.

yarn.lock
sobolk: These are not coming from our packages.json. This is all we have in our workspace.
image

packages/amplify-nodejs-function-template-provider/resources/lambda/crud/app.js.ejs
sobolk: updated.

jhockett (approved):

sdstolworthy (approved):

PR #12414, opened 2023-04-07
Smoketest script

sdstolworthy:
#### Description of changes
Moves the smoketest script into the e2e test suite. Refactors the script to use the existing e2e suite.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/package.json
+++ b/packages/amplify-e2e-tests/package.json
@@ -21,7 +21,8 @@
     "general-config-e2e": "jest src/__tests__/general-config/general-config-headless-init.test.ts --verbose",
     "build-tests": "tsc --build tsconfig.tests.json",
     "setup-profile": "ts-node ./src/configure_tests.ts",
-    "clean-e2e-resources": "ts-node ./src/cleanup-e2e-resources.ts"
+    "clean-e2e-resources": "ts-node ./src/cleanup-e2e-resources.ts",
+    "smoketest": "DESTRUCTIVE=true jest --runInBand --testPathPattern='smoketest'"
   },
   "dependencies": {
     "@aws-amplify/amplify-category-auth": "3.0.1",
--- a/packages/amplify-e2e-tests/src/__tests__/smoketest.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/smoketest.test.ts
@@ -198,7 +198,10 @@ const createCommands = (amplify: Amplify, cliVersion: string): Command[] => [
   {
     description: `Install Amplify CLI v${cliVersion}`,
     run: async () => {
-      NPM.install(`@aws-amplify/cli@${cliVersion}`, true);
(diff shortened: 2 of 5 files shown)


packages/amplify-e2e-tests/package.json
```diff
@@ -21,7 +21,8 @@
     "general-config-e2e": "jest src/__tests__/general-config/general-config-headless-init.test.ts --verbose",
     "build-tests": "tsc --build tsconfig.tests.json",
     "setup-profile": "ts-node ./src/configure_tests.ts",
-    "clean-e2e-resources": "ts-node ./src/cleanup-e2e-resources.ts"
+    "clean-e2e-resources": "ts-node ./src/cleanup-e2e-resources.ts",
+    "smoketest": "jest --runInBand --testPathPattern='smoketest'"
```
sobolk, line 25: should we have 
`yarn smoketest rc` and `yarn smoketest latest` ?
I.e. make them npm install `@latest` and `@rc` respectively?

packages/amplify-e2e-tests/src/__tests__/smoketest.test.ts
```diff
@@ -0,0 +1,299 @@
...
+
+export type SmoketestArgs = {
+  projectDirectory: string;
+  cliVersion: string;
+  destructive: boolean;
+};
+
+function getArgs(): SmoketestArgs {
+  const { DESTRUCTIVE, CLI_VERSION, PROJECT_DIRECTORY } = process.env;
+  if (!CLI_VERSION) {
```
sobolk, line 34: I'd make CLI_VERSION an arg

sobolk (approved):
Should we add this to CCI as post release job ?

sobolk:
and... should we run smoke test as part of e2e tests to make sure it doesn't regress ?

packages/amplify-e2e-tests/package.json
sdstolworthy: I like that idea. I'm working on a separate PR to integrate w/ CI, I'll get a PR up next week if time allows.

packages/amplify-e2e-tests/src/__tests__/smoketest.test.ts
sdstolworthy: It was an arg previously, but I was having issues passing it in since the file is being run with jest. I didn't spend more than 10 minutes looking into it and just decided to switch to env vars. I can revisit this next week, as I personally preferred them as arguments.

sobolk: if passing args is hard then we should have `smoketest`, `smoketest-rc` , `smoketest-latest` targets. the latter two would set env var before calling jest.

sobolk (approved):

Amplifiyer (approved):

PR #12486, opened 2023-04-17
chore: reclassify common sdk errors

goldbez:
#### Description of changes
This PR adds try-catch blocks to SDK calls in push-resources.ts. These try-catch bocks call a function that checks if the thrown error matches a set of errors that we have pre-determined to be common user errors, and reclassifies them as such. In future PRs, this function can be used when adding try-catch blocks to SDK calls in other parts of the code.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-cli-core/src/errors/amplify-exception.ts
```diff
@@ -158,6 +158,7 @@ export type AmplifyErrorType =
   | 'ProjectNotFoundError'
   | 'ProjectNotInitializedError'
   | 'PushResourcesError'
+  | 'RateError'
```
Amplifiyer, line 161:
```suggestion
  | 'APIRateExceededError'
```

Amplifiyer (commented):
LGTM

jhockett (commented):
There's a lot of AWS-SDK calls that won't be checked with this approach. Is there a reason we can't do something in the exception handling logic itself rather than duplicate this call everywhere?

goldbez:
> There's a lot of AWS-SDK calls that won't be checked with this approach. Is there a reason we can't do something in the exception handling logic itself rather than duplicate this call everywhere?

The reason for not doing the error handling in a central location is that the stack track would be destroyed. This PR is only meant to cover the SDK calls in push-resources.ts. In later PRs, I will cover SDK calls in other places.

packages/amplify-provider-awscloudformation/src/__tests__/handle-common-sdk-errors.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      try {
+        throw new Error(
+          'User: ARN is not authorized to perform: ACTION:ACTION on resource: RESOURCE because no identity-based policy allows the ACTION:ACTION action',
+        );
+      } catch (error) {
+        throw handleCommonSdkError(error);
+      }
+    };
+
+    confirmCorrectError(toThrowError, 'Update the permissions.');
```
Amplifiyer, line 35:
```suggestion
    confirmErrorResolution(toThrowError, 'Update the permissions.');
```

Amplifiyer (approved):
nit: update the function name otherwise looks good.

jhockett (approved):

PR #12601, opened 2023-05-08
cb: enable basic e2e test coverage

kuhlmanp:
#### Description of changes
This adds several items to the CodeBuild setup including CodeCov updates and V8 supported coverage for e2e tests.

The CodeBuild updates include the function `_uploadCoverageLinux` in the shared-scripts.sh file. This setup utilizes the CodeCov Token secret, managed by SecretsManager and injected by the build.  It then gets the latest CodeCov uploader executable and verifies it via SHA. Once the verification is complete, the uploader API is called, automatically searching for coverage artifacts to upload. This update also includes an update to the codecov.yml file to conform to CodeCov validation.

Lastly, in this update we take advantage of the embedded ability of NodeJS to report coverage metrics: https://blog.npmjs.org/post/178487845610/rethinking-javascript-test-coverage.html. We then run the c8 tool to convert that coverage into something that both CodeCov and CodeBuild can consume. It also integrates these reports in CodeBuild directly.

Notes: currently, this mechanism is set to run tests from $TEST_SUITE and, in this PR, is set to run only the geo-add-b.test.ts file.

Manual build tests.

#### Checklist

- [x] PR description included


.circleci/local_publish_helpers.sh
```diff
@@ -153,7 +153,7 @@ function unsetSudoNpmRegistryUrl {
 }
 
 function changeNpmGlobalPath {
-    mkdir -p ~/.npm-global
+    mkdir -p ~/.npm-global/{bin,lib}
```
goldbez, line 156: nice bash work 😎

.circleci/local_publish_helpers.sh
```diff
@@ -269,14 +269,32 @@ function setAwsAccountCredentials {
...
 
 function runE2eTest {
+    _setupCoverage
     FAILED_TEST_REGEX_FILE="./amplify-e2e-reports/amplify-e2e-failed-test.txt"
 
     if [ -f  $FAILED_TEST_REGEX_FILE ]; then
         # read the content of failed tests
         failedTests=$(<$FAILED_TEST_REGEX_FILE)
-        yarn run e2e --forceExit --no-cache --maxWorkers=4 $TEST_SUITE -t "$failedTests"
+        NODE_V8_COVERAGE=$E2E_TEST_COVERAGE_DIR yarn run e2e --forceExit --no-cache --maxWorkers=4 $TEST_SUITE -t "$failedTests"
```
goldbez, line 278: Is there a reason to set this variable on the same line as calling "yarn run e2e"? If there is not a functional reason, I think this would be more legible on two lines.

codebuild_specs/run_coverage_linux.yml
```diff
@@ -0,0 +1,30 @@
+version: 0.2
+env:
+  shell: bash
+  variables:
+    CLI_REGION: us-east-1
+    E2E_TEST_COVERAGE_DIR: node_v8_coverage
+  secrets-manager:
+    CODECOV_TOKEN: arn:aws:secretsmanager:us-east-1:052112422291:secret:testing/codecov/token-OB8isC
```
goldbez, line 8: We should not have Secrets Manager ARNs on Github. This should be an env variable that is injected in

codebuild_specs/run_e2e_tests_linux.yml
```diff
@@ -3,7 +3,8 @@ env:
   shell: bash
   variables:
     CLI_REGION: us-east-1
-    TEST_SUITE: src/__tests__/auth_2a.test.ts|src/__tests__/auth_2b.test.ts|src/__tests__/auth_2d.test.ts|src/__tests__/auth_2f.test.ts
+    TEST_SUITE: src/__tests__/geo-add-b.test.ts
```
goldbez, line 6: I don't think you want to overwrite this line. It's important to run multiple tests here.

codebuild_specs/run_e2e_tests_linux.yml
```diff
@@ -26,19 +27,18 @@ phases:
...
-      - codebuild-breakpoint
   post_build:
     commands:
       - aws sts get-caller-identity
       - source ./shared-scripts.sh && _scanArtifacts
-
 artifacts:
   files:
-    - $CODEBUILD_SRC_DIR/packages/amplify-e2e-tests/amplify-e2e-reports/*
+    - '**/*'
```
goldbez, line 43: Why are you uploading artifacts like this instead of specifying specific files?

goldbez:
Also there are merge conflicts. Can you resolve those please?

.circleci/local_publish_helpers.sh
kuhlmanp: Setting it this way makes sure that it only exists as set for the duration of the following call. Since the V8 coverage is incredibly verbose, this minimizes the chances that something else gets run and produces coverage output.

codebuild_specs/run_coverage_linux.yml
```diff
@@ -0,0 +1,30 @@
...
+phases:
+  build:
+    commands:
+      # you can provide a codebuild source version to use old cache and skip all other jobs :)
+      - export CODEBUILD_SOURCE_VERSION=00f31ea25fc27aa865198398e5014ea4d6a126d9
```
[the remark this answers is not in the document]
kuhlmanp: Correct, this is another leftover from testing.

codebuild_specs/run_e2e_tests_linux.yml
kuhlmanp: Specific file names are unknown at the time of upload. Since all of these files are named with a GUID we have to specify the base directory (E2E_TEST_COVERAGE_DIR) and take all files in that directory. It should start clean at the beginning of each run and only have coverage files in it after a run is complete.

.circleci/local_publish_helpers.sh
goldbez: that makes sense. thanks for explaining

goldbez (approved):

shared-scripts.sh
```diff
@@ -261,7 +291,12 @@ function _runE2ETestsLinux {
     _loadTestAccountCredentials
     retry runE2eTest
 }
-
+function _unassumeTestAccountCredentials {
```
awsluja, line 294: no longer being used, not a huge deal since it is useful

awsluja (approved):
looks good, the other feedback can be fixed once you can experiment with full e2e

PR #12621, opened 2023-05-11
fix: don't buffer when extracting binary

sobolk:
#### Description of changes

Pipes the file instead of buffering in memory at binary installation.

E2E tests. I.e. `amplify_sudo_install_test`

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli-npm/binary.ts
```diff
@@ -153,27 +153,19 @@ export class Binary {
...
-      if (chunks.length) {
-        const data = Buffer.concat(chunks);
-        fs.writeFileSync(this.binaryPath, data, {
-          mode: 0o755,
-        });
-      }
-    });
     return extract;
   }
 }
```
sobolk, line 171: there's strong assumption in this code that TAR has only one file.

edwardfoyle (dismissed):

.circleci/config.base.yml
```diff
@@ -454,6 +454,10 @@ jobs:
             startLocalRegistry "$(pwd)/.circleci/verdaccio.yaml"
             setSudoNpmRegistryUrlToLocal
             changeSudoNpmGlobalPath
+            # limit memory for new processes to 1GB
+            # this is to make sure that install can work on small VMs
+            # i.e. not buffer content in memory while installing binary
+            ulimit -Sv 1000000
```
sobolk, line 460: **12.0.0 release**
image

**After applying fix**
image

packages/amplify-cli-npm/binary.ts
```diff
@@ -121,7 +121,10 @@ export class Binary {
     console.log(`Downloading release from ${getCompressedBinaryUrl()}`);
     try {
       const res = await axios({ url: getCompressedBinaryUrl(), responseType: 'stream' });
-      await pipeline(res.data, createGunzip(), this.extract());
+      // Dummy array to collect a promise from nested pipeline that extracts tar content to a file.
+      const extractPromiseCollector: Array<Promise<void>> = [];
+      await pipeline(res.data, createGunzip(), this.extract(extractPromiseCollector));
+      await extractPromiseCollector[0];
```
edwardfoyle, line 127: why only await the first promise and not Promise.allSettled?

sobolk: if we get here there's only one promise, but `Promise.all` would do too.

sobolk: Changed. I used `Promise.all` not `Promise.allSettled` . This should throw if extracting fails.

.circleci/config.base.yml
awsluja: nice

awsluja (approved):
looks good, only nit would be to add a note why we had to go this route (spawnSync was executing prematurely because tar.extract callbacks were not being awaited; this approach collects those nested callbacks and forces the process to await all) or something like that

edwardfoyle (dismissed):

awsluja (approved):

edwardfoyle (approved):

PR #12629, opened 2023-05-16
ci: Setup e2e scripts

awsluja:
#### Description of changes
This change allows us to trigger e2e workflows for codebuild, without the need for webhooks.
It also supports beta environments (cloud-e2e-beta), which is configurable via .env

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


.circleci/publish-codebuild.sh
```diff
@@ -1,4 +1,11 @@
 #!/bin/bash -e
```
goldbez, line 1: I'm wondering if it would be a good idea to move all of the codebuild scripts into the codebuild-specs directory, instead of having them mixed in with circle ci

goldbez (approved):

.circleci/publish-codebuild.sh
awsluja: I like this idea; seems like we are in a good spot to make a switch over; so I'll create a separate PR for that cleanup

sdstolworthy (approved):

scripts/cloud-e2e-beta.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# set exit on error to true
+set -e
```
sdstolworthy, line 4: This and the other e2e file are extremely similar. Can we extract these to a shared bash function and parameterize the values that are different?

sdstolworthy (changes requested):

kuhlmanp (dismissed):

sdstolworthy (approved):

kuhlmanp (approved):

Issue #12626, opened 2023-05-15
function mock errors are not stringified, printing `[object Object]`


josefaidt:
### How did you install the Amplify CLI?

pnpm

### If applicable, what version of Node.js are you using?

18

### Amplify CLI Version

12.0.0

### What operating system are you using?

mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

n/a

### Describe the bug

When mocking functions, if the function errors for any reason we are presented with a generic error message from CLI and the actual error as `[object Object]`

it appears to be going wrong here https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-util-mock/src/func/index.ts#L61

```
Ensuring latest function changes are built...
Starting execution...
🛑 sayhelloerror failed with the following error:
[object Object]
Finished execution.
```

### Expected behavior

error is printed

### Reproduction steps

1. `amplify init -y`
2. `amplify add function` > accept all defaults
3. add an error to the function template
	```js
	throw new Error('Hello from Lambda!')
	
	/**
	 * @type {import('@types/aws-lambda').APIGatewayProxyHandler}
	 */
	exports.handler = async (event) => {
	  console.log(`EVENT: ${JSON.stringify(event)}`)
	  return {
	    statusCode: 200,
	    //  Uncomment below to enable CORS requests
	    //  headers: {
	    //      "Access-Control-Allow-Origin": "*",
	    //      "Access-Control-Allow-Headers": "*"
	    //  },
	    body: JSON.stringify('Hello from Lambda!'),
	  }
	}
	```
4. `amplify mock function`
5. observe error

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Before submitting, please confirm:

- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

ykethan:
Was able to reproduce the issue. marking as bug.

PR #12636, opened 2023-05-16
feat: display error message instead of object

ykethan:
#### Description of changes

displays error message instead of object or stack trace 

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/12626

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-prompts/src/printer.ts
+++ b/packages/amplify-prompts/src/printer.ts
@@ -30,8 +30,16 @@ export class AmplifyPrinter implements Printer {
     this.writeLine(`${isHeadless ? '' : '⚠️ '}${chalk.yellow(line)}`);
   };
 
-  error = (line: string): void => {
+  // disable-eslint-next-line @typescript-eslint/no-explicit-any
+  error = (line: string, error?: any): void => {
     this.writeLine(`${isHeadless ? '' : '🛑 '}${chalk.red(line)}`);
+    if (error) {
+      if (error?.message) {
+        this.writeLine(`${chalk.red(error.message)}`);
+      } else {
+        this.writeLine(`${chalk.red(error)}`);
+      }
+    }
   };
 
   private writeSilenceableLine = (line?: string): void => {
@@ -59,7 +67,7 @@ export type Printer = {
   blankLine: () => void;
   success: (line: string) => void;
   warn: (line: string) => void;
-  error: (line: string) => void;
+  error: (line: string, error?: any) => void;
 };
 
 type Color = 'green' | 'blue' | 'yellow' | 'red' | 'reset';
--- a/packages/amplify-prompts/src/printer.ts
+++ b/packages/amplify-prompts/src/printer.ts
@@ -33,12 +33,9 @@ export class AmplifyPrinter implements Printer {
   // disable-eslint-next-line @typescript-eslint/no-explicit-any
   error = (line: string, error?: any): void => {
     this.writeLine(`${isHeadless ? '' : '🛑 '}${chalk.red(line)}`);
-    if (error) {
-      if (error?.message) {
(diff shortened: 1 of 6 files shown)


packages/amplify-util-mock/src/func/index.ts
```diff
@@ -58,7 +58,7 @@ export async function start(context: $TSContext): Promise<void> {
       printer.info(typeof result === 'undefined' ? '' : stringResult);
     } catch (err) {
       printer.error(`${resourceName} failed with the following error:`);
-      printer.info(err);
+      printer.info(err.message);
```
jhockett, line 61: Just to be safe to ensure no error is thrown in the catch block, please add an optional chaining operator:

```suggestion
      printer.info(err?.message);
```

jhockett (changes requested):

packages/amplify-util-mock/src/func/index.ts
```diff
@@ -58,7 +58,7 @@ export async function start(context: $TSContext): Promise<void> {
       printer.info(typeof result === 'undefined' ? '' : stringResult);
     } catch (err) {
       printer.error(`${resourceName} failed with the following error:`);
-      printer.info(err);
+      printer.info(err?.message);
```
sobolk, line 61: instead of doing this. could we modify functions in printer and make them accept `string | Error` ?
Then depending on input type extract apply desired strategy?

See https://github.com/aws-amplify/amplify-cli/blob/fd51b652c3fb6b24db7a400bb38bf5a35757efc2/packages/amplify-prompts/src/printer.ts#L11-L35 .

additionally. 
why do we `printer.error` first message and `printer.info` second line? Should this all be `printer.error` ?

sobolk: btw. if the answer to second concern is `yes` then we should only consider modifying `printer.error` to accept both string and error.

sobolk: @lazpavel @jhockett what do you think ?

sobolk: Then perhaps we need
`error = (line: string, error?: Error): void` on printer , i.e. add error as optional parameter and make sure it prints nicely together.

lazpavel: I like both suggestion, was thinking about info/error myself but didn't comment on it since it was an existing behavior. But in general we should keep errors to error console as someone can split the outputs

ykethan: updated the error according to the suggestions

jhockett: I missed this thread until now, but I talked with @ykethan and I think the suggestion makes sense

jhockett (approved):

packages/amplify-prompts/src/printer.ts
```diff
@@ -30,8 +30,16 @@ export class AmplifyPrinter implements Printer {
     this.writeLine(`${isHeadless ? '' : '⚠️ '}${chalk.yellow(line)}`);
   };
 
-  error = (line: string): void => {
+  // disable-eslint-next-line @typescript-eslint/no-explicit-any
+  error = (line: string, error?: any): void => {
     this.writeLine(`${isHeadless ? '' : '🛑 '}${chalk.red(line)}`);
+    if (error) {
```
lazpavel, line 36: to avoid nested if's can we do something like:
```
const errorMessage = error?.message ?? error;
if (errorMessage) {
  this.writeLine(`${chalk.red(errorMessage)}`);
}
```

lazpavel (approved):

sobolk (approved):

PR #12673, opened 2023-05-23
fix: warn about multiple schema files in admin modelgen task

phani-srikar:
#### Description of changes

If the customer is using multiple GraphQL schema files located at `<amplify/backend/api/<api_name>/schema>` folder instead of a single `schema.graphql` file, the current behavior is to fail softly by throwing the following error:
```
ENOENT: no such file or directory... schema.graphql
```
This error message occurs during the admin modelgen flow where we upload the Studio CMS artifacts to deployment S3 bucket. 
The current behavior is that we fail to update the schema, but generate and update the modelgen files in S3 bucket. When customer tries to open the Schema editor in Studio console, it fails to load.
With this change, we check and early return if we detect that there is no single schema file. In addition we display a warning message.

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/1074

Added unit test.
Tested from a JS app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/admin-modelgen.ts
+++ b/packages/amplify-provider-awscloudformation/src/admin-modelgen.ts
@@ -4,6 +4,7 @@ import _ from 'lodash';
 import * as path from 'path';
 import { S3 } from './aws-utils/aws-s3';
 import { ProviderName as providerName } from './constants';
+import { printer } from '@aws-amplify/amplify-prompts';
 
 /**
  * Generates DataStore Models for Admin UI CMS to consume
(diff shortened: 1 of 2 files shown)


packages/amplify-provider-awscloudformation/src/admin-modelgen.ts
```diff
@@ -26,6 +27,13 @@ export const adminModelgen = async (context: $TSContext, resources: $TSAny[]): P
     return;
   }
 
+  const localSchemaPath = path.join(pathManager.getResourceDirectoryPath(undefined, 'api', resourceName), 'schema.graphql');
+  // Early return with useful warning if the schema file does not exist
+  if (!fs.existsSync(localSchemaPath)) {
+    printer.warn(`Could not find the GraphQL schema file at "${localSchemaPath}". Amplify Studio's schema editor might not work as intended if you're using multiple schema files.`);
+    return;
+  }
```
sobolk, line 35: keep in mind that we generate these files regardless of customer using studio or not.
So for users who never opened Studio this warning might be confusing.

sobolk (approved):

sobolk (dismissed):

lazpavel (dismissed):

sobolk (approved):

lazpavel (approved):

PR #12710, opened 2023-05-31
chore: ask for consent for diagnose report

eddy2048:
#### Description of changes

Previously the CLI would ask permission for diagnose command on init but this PR moves the ask to the beginning of diagnose command. This is set on a per-project basis.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-core/src/init/initProjectHelper.ts
+++ b/packages/amplify-e2e-core/src/init/initProjectHelper.ts
@@ -6,6 +6,7 @@ import { v4 as uuid } from 'uuid';
 import { nspawn as spawn, getCLIPath, singleSelect } from '..';
 import { KEY_DOWN_ARROW, addCircleCITags, errorReportingTestHandler } from '../utils';
 import { amplifyRegions } from '../configure';
+import { mergeDeploymentSecrets } from '@aws-amplify/amplify-cli-core';
 
 const defaultSettings = {
   name: EOL,
@@ -56,7 +57,7 @@ export function initJSProjectWithProfile(cwd: string, settings?: Partial<typeof 
   if (mergedSettings?.name?.length > 20)
     console.warn('Project names should not be longer than 20 characters. This may cause tests to break.');
 
-  const chain = spawn(getCLIPath(s.testingWithLatestCodebase), cliArgs, {
+  const chain = spawn(getCLIPath(mergedSettings.testingWithLatestCodebase), cliArgs, {
     cwd,
     stripColors: true,
     env,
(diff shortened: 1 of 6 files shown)


packages/amplify-e2e-tests/src/__tests__/migration/api.connection.migration.test.ts
```diff
@@ -37,14 +37,19 @@ describe('amplify add api', () => {
       amplifyPushUpdate(
         projRoot,
         /Attempting to edit the global secondary index gsi-PostComments on the CommentTable table in the Comment stack.*/,
+        false,
+        false,
+        0,
+        false,
+        true,
```
evcodes, line 44: We should make a settings object within amplifyPushUpdate parameters.. I wrote this like this as a quick and dirty solution just to see if this would work.

0618:
E2E tests:
- https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli/22340

packages/amplify-cli/src/commands/diagnose.ts
```diff
@@ -87,9 +87,16 @@ const showLearnMore = (showOptOut: boolean): void => {
...
+    if (DebugConfig.Instance.promptSendReport()) {
+      const result = await prompter.yesOrNo('Help improve Amplify CLI by sharing non sensitive configurations on failures', false);
+      DebugConfig.Instance.setAndWriteShareProject(result);
+      if (result === false) {
+        return;
+      }
+      const diagnoseAction = await prompter.pick('What would you like to do?', choices);
+      if (diagnoseAction !== choices[0]) {
+        return;
+      }
```
sobolk, line 99: We should.

1. Double check this condition with https://github.com/aws-amplify/amplify-cli/blob/120e7feba8655f4404da756b0183c38f70ce3f95/packages/amplify-cli/src/commands/diagnose.ts#L35 . I.e. we must never ask this question in headless mode.
2. We should undo all the changes made to e2e/migration/other tests. They don't scale. Current implementation covers very narrow failure modes that are expected. All unexpected errors will now result in `nexpect` timing out on unexpected prompt should failure happen (transient or not). This will contribute to overall e2e test runtime.
3. Instead of that we should.
    1. Opt out existing e2e tests by default, i.e. make sure `DebugConfig.Instance.promptSendReport()` returns `false`, so this prompt is never asked there.
    2. Add test(s) and test utils dedicated to covering that particular scenario.

sobolk (changes requested):

packages/amplify-cli/src/commands/diagnose.ts
```diff
@@ -87,9 +87,16 @@ const showLearnMore = (showOptOut: boolean): void => {
 const zipSend = async (context: Context, skipPrompts: boolean, error: Error | undefined): Promise<void> => {
   const choices = ['Generate report', 'Nothing'];
   if (!skipPrompts) {
-    const diagnoseAction = await prompter.pick('What would you like to do?', choices);
-    if (diagnoseAction !== choices[0]) {
-      return;
+    if (DebugConfig.Instance.promptSendReport()) {
+      const result = await prompter.yesOrNo('Help improve Amplify CLI by sharing non sensitive configurations on failures', false);
```
0618, line 91: Looks like we've already got a prompt on line 37, why do we need to add this line to `zipSend` again? https://github.com/aws-amplify/amplify-cli/pull/12710/files#diff-71cabcaa3379af3b9458f6802b25c65f6c00768606d5ef7f8c010094d8720aa1R37

PR #12763, opened 2023-06-07
fix: update notification channel name lookup

lazpavel:
#### Description of changes
- when updating the notifications channel without passing the channel name as the command line argument, the customer is prompted to select the channel name, we incorrectly checked that the command line value is also set.

- manual testing
- e2e added
- e2e tests pass

#### Checklist
- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-category-notifications/src/commands/notifications/configure.ts
+++ b/packages/amplify-category-notifications/src/commands/notifications/configure.ts
@@ -26,14 +26,14 @@ export const alias = 'update';
  */
 export const run = async (context: $TSContext): Promise<$TSContext> => {
   const availableChannelViewNames = getAvailableChannelViewNames();
-  const channelName = context.parameters.first;
-  let channelViewName = channelName ? getChannelViewName(channelName) : undefined;
+  let channelViewName = context.parameters.first ? getChannelViewName(context.parameters.first) : undefined;
 
   if (!channelViewName || !availableChannelViewNames.includes(channelViewName)) {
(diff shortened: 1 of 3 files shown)


packages/amplify-category-notifications/src/commands/notifications/configure.ts
```diff
@@ -26,14 +26,14 @@ export const alias = 'update';
...
 
   if (!channelViewName || !availableChannelViewNames.includes(channelViewName)) {
     channelViewName = await prompter.pick('Choose the notification channel to configure', availableChannelViewNames);
   }
-  if (channelViewName && typeof channelName === 'string') {
-    const selectedChannel = getChannelNameFromView(channelViewName);
+
+  if (channelViewName) {
+    const channelName = getChannelNameFromView(channelViewName);
     let pinpointAppStatus = await pinpointHelper.ensurePinpointApp(context, undefined);
```
Amplifiyer, line 37: what's the type of `channelName` if not `string` and should that always be the same type for all channels?

Amplifiyer (commented):
nit: update the commit message and PR title to be more descriptive

packages/amplify-category-notifications/src/commands/notifications/configure.ts
lazpavel: the problem is that the channelName is not passed as argument, it needs to be obtained from the channelViewName prompted

Amplifiyer (approved):

jhockett (approved):

PR #12771, opened 2023-06-09
fix: adjust path finding for schemas in codebuild

goldbez:
#### Description of changes
There is a difference in the way that paths are resolved in CircleCI vs CodeBuild. This was causing many Windows tests to fail because the paths to schemas would not resolve correctly.

This PR also enables artifacts for Windows tests.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-core/src/categories/api.ts
+++ b/packages/amplify-e2e-core/src/categories/api.ts
@@ -22,11 +22,11 @@ import { multiSelect, singleSelect } from '../utils/selectors';
 import { selectRuntime, selectTemplate } from './lambda-function';
 import { modifiedApi } from './resources/modified-api-index';
 
-// const isWindowsPlatform = (): boolean => !!process?.platform?.startsWith('win');
+const isWindowsPlatform = (): boolean => !!process?.platform?.startsWith('win');
 
 export function getSchemaPath(schemaName: string): string {
   // This condition is to account for a difference in the use of __dirname and paths in CodeBuild Windows jobs
-  if (process.env.CODEBUILD_SRC_DIR /*&& isWindowsPlatform()*/) {
+  if (process.env.CODEBUILD_SRC_DIR && isWindowsPlatform()) {
     return path.join(process.env.CODEBUILD_SRC_DIR, 'packages', 'amplify-e2e-tests', 'schemas', schemaName);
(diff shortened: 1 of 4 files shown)


packages/amplify-e2e-core/src/categories/api.ts
```diff
@@ -23,6 +23,9 @@ import { selectRuntime, selectTemplate } from './lambda-function';
 import { modifiedApi } from './resources/modified-api-index';
 
 export function getSchemaPath(schemaName: string): string {
+  if (process.env.CODEBUILD_SRC_DIR) {
+    return path.join(process.env.CODEBUILD_SRC_DIR, 'packages', 'amplify-e2e-tests', 'schemas', schemaName);
+  }
   return path.join(__dirname, '..', '..', '..', 'amplify-e2e-tests', 'schemas', schemaName);
 }
```
sobolk, line 30: any idea why `__dirname` didn't work? `__dirname` should point to absolute path of the current file (i.e. api.js after compilation)

goldbez: I could not figure out why it did not work. I did verify from the .cast files that the tests are running using the amplify.exe binary.

sobolk: did it work on linux ? if this is just for windows please add comments in both places explaining that.
and perhaps add `ifWindows` condition to the mix.

packages/amplify-e2e-core/src/utils/index.ts
```diff
@@ -176,7 +177,9 @@ export const getFunctionSrcNode = (root: string, functionName: string, fileName
 };
 
 const getTestFileNamePath = (fileName: string): string =>
-  path.join(__dirname, '..', '..', '..', 'amplify-e2e-tests', 'functions', fileName);
+  process.env.CODEBUILD_SRC_DIR && isWindowsPlatform()
```
sobolk, line 180: nit: comment should go above the line with condition.

sobolk (approved):

awsluja (approved):
very nice

kuhlmanp (approved):

sobolk (approved):

PR #12774, opened 2023-06-09
fix: fix the handling of not found resources in the lambdas when deleting

danielleadams:
#### Description of changes
For the `hostedUIProviderLambda`, this runs deletes synchronously on the Cognito Identity Providers. For the `hostedUILambda`, when a User Pool Domain is attempting to be created, it will look for the correct Not Found error to send a "success" back to CFN.

Ran manually and following up with e2e tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


danielleadams:
The failing tests are fixed in other branches.

packages/amplify-category-auth/resources/auth-custom-resource/hostedUILambda.js
```diff
@@ -18,7 +18,7 @@ exports.handler = (event, context) => {
     .catch((err) => {
       console.log(err);
 
-      if (err.name === 'NotFoundException') {
+      if (err.name === 'NotFoundException' || err.message === 'No such domain or user pool exists.') {
```
sobolk, line 21: How does `err` look like when `No such domain or user pool exists.` is thrown?
Is there any other property like `code` that could be used instead of message in this condition?
I'm not sure if using message here is sustainable solution, not sure if service considers changing messaging a breaking change.

jhockett (approved):

packages/amplify-category-auth/resources/auth-custom-resource/hostedUIProviderLambda.js
```diff
@@ -13,22 +13,19 @@ exports.handler = (event, context) => {
...
-      } else {
-        const firstFailure = results.find((result) => result.status === 'rejected');
-        response.send(event, context, response.FAILED, firstFailure);
-      }
+    // Only 1 update can be sent at a time, so the SDK calls need to be run synchronously
+    hostedUIProviderMeta.forEach(async ({ ProviderName }) => {
+      await deleteIdentityProvider(ProviderName).catch((error) => {
+        if (!error?.code?.toString()?.match(/NotFoundException/)) {
+          response.send(event, context, response.FAILED, error);
+          return error;
```
sobolk, line 21: should this `throw error` to bubble it up?
i.e. does this prevent line 28 from executing?

packages/amplify-category-auth/resources/auth-custom-resource/hostedUILambda.js
sobolk: By looking at this https://stackoverflow.com/questions/54002030/how-to-fix-cognito-user-pool-domain-destruction-with-terraform-so-the-user-pool I seems that error should contain input values that were wrong, we could use that instead of the message.

danielleadams: I can add `err.name === 'InvalidParameterException'`

packages/amplify-category-auth/resources/auth-custom-resource/hostedUIProviderLambda.js
danielleadams: I think We can actually delete line 21

packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/auth-stack-builder/lambda-callout-migrations/__snapshots__/auth-cognito-stack-builder.test.ts.snap
```diff
@@ -0,0 +1,38 @@
...
+      const params = { ProviderName: providerName, UserPoolId: userPoolId };
+      return identity.deleteIdentityProvider(params).promise();
+    };
+
+    // Only 1 update can be sent at a time, so the SDK calls need to be run synchronously
+    hostedUIProviderMeta.forEach(async ({ ProviderName }) => {
+      await deleteIdentityProvider(ProviderName).catch((error) => {
+        if (!error?.code?.toString()?.match(/NotFoundException/)) {
+          response.send(event, context, response.FAILED, error);
+          return error;
```
sobolk, line 24: please update snapshot.

packages/amplify-category-auth/resources/auth-custom-resource/hostedUIProviderLambda.js
```diff
@@ -13,22 +13,19 @@ exports.handler = (event, context) => {
...
-
-    Promise.allSettled(providerPromises).then((results) => {
-      if (results.every(deleteSuccessOrNotFound)) {
-        response.send(event, context, response.SUCCESS);
-      } else {
-        const firstFailure = results.find((result) => result.status === 'rejected');
-        response.send(event, context, response.FAILED, firstFailure);
-      }
+    // Only 1 update can be sent at a time, so the SDK calls need to be run synchronously
+    hostedUIProviderMeta.forEach(async ({ ProviderName }) => {
```
sobolk, line 17: I just noticed that this is the same problem we discussed on the call.
The promise from callback won't be observed and code will proceed to line 28 regardless of result here...

sobolk (approved):

packages/amplify-category-auth/resources/auth-custom-resource/hostedUILambda.js
```diff
@@ -1,27 +1,26 @@
 const response = require('cfn-response');
 const aws = require('aws-sdk');
-const identity = new aws.CognitoIdentityServiceProvider();
+
+const { deleteUserPoolDomain } = new aws.CognitoIdentityServiceProvider();
 
 exports.handler = (event, context) => {
+  // Don't return promise, response.send() marks context as done internally
+  const ignoredPromise = handleEvent(event, context);
```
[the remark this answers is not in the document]
sobolk: feel free to won't fix this one.

PR #12789, opened 2023-06-13
fix: github release script

sobolk:
#### Description of changes

Adds `make_latest` flag to github release script

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/scripts/github-release.ts
+++ b/scripts/github-release.ts
@@ -13,6 +13,7 @@ const publishRelease = async (version: string, commit: string) => {
     method: 'PATCH',
     body: JSON.stringify({
       prerelease: false,
+      make_latest: 'true',
       target_commitish: commit,
     }),
   });


scripts/github-release.ts
```diff
@@ -13,6 +13,7 @@ const publishRelease = async (version: string, commit: string) => {
     method: 'PATCH',
     body: JSON.stringify({
       prerelease: false,
+      make_latest: 'true',
```
sobolk, line 16: The type is `string` see https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#update-a-release

edwardfoyle (commented):
Should we set this to 'legacy'? Seems like that would keep the behavior as it exists today

sobolk:
> Should we set this to 'legacy'? Seems like that would keep the behavior as it exists today

we could, but... don't we want to make it latest at this point anyway ? I'd rather choose non-legacy way of achieving same goal.

lazpavel (approved):

edwardfoyle (approved):

PR #12801, opened 2023-06-14
chore: refactor hostedUIProviderLambda.js to use async await

sobolk:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-category-auth/resources/auth-custom-resource/hostedUIProviderLambda.js
+++ b/packages/amplify-category-auth/resources/auth-custom-resource/hostedUIProviderLambda.js
@@ -4,33 +4,32 @@ const aws = require('aws-sdk');
 const identity = new aws.CognitoIdentityServiceProvider();
 
 exports.handler = (event, context) => {
+  // Don't return promise, response.send() marks context as done internally
+  const ignoredPromise = handleEvent(event, context);
+}
+
+async function handleEvent(event, context) {
   try {
     const userPoolId = event.ResourceProperties.userPoolId;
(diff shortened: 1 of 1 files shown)


packages/amplify-category-auth/resources/auth-custom-resource/hostedUIProviderLambda.js
```diff
@@ -4,33 +4,32 @@ const aws = require('aws-sdk');
...
-      } else {
-        const firstFailure = results.find((result) => result.status === 'rejected');
-        response.send(event, context, response.FAILED, firstFailure);
+    for (const { ProviderName } of hostedUIProviderMeta) {
+      try {
+        const params = { ProviderName, UserPoolId: userPoolId };
+        await identity.deleteIdentityProvider(params).promise();
+      } catch (e) {
+        if (!e?.code?.toString()?.match(/NotFoundException/)) {
+          response.send(event, context, response.FAILED, e);
```
jhockett, line 22: This line sends the error object directly, but line 33 wraps the error in an object. Intentional?

Additionally, if the next line throws anyways, can the `response.send` call happen in the outer catch block?

sobolk: line 33 is likely correct as it was  here before.

good catch with outer catch block - we don't want to call response.send twice.

jhockett (approved):

danielleadams (approved):

PR #12836, opened 2023-06-20
fix: make sure that AWS_REGION env variables match prescribed CLI_REGION in custom resource test

goldbez:
#### Description of changes
This test was failing because of an issue where the init step was being executed in a different region from the region specified by the CLI_REGION variable. This fix ensures that the init step happens in the same region.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-e2e-tests/src/__tests__/custom-resource-with-storage.test.ts
```diff
@@ -21,6 +21,12 @@ describe('adding custom resources test', () => {
   const projectName = 'cusres';
   let projRoot: string;
   const envName = 'dev';
+
+  beforeAll(async () => {
+    process.env.AWS_REGION = process.env.CLI_REGION;
+    process.env.AWS_DEFAULT_REGION = process.env.CLI_REGION;
+  });
```
sobolk, line 28: consider doing this globally for all tests.

sobolk (approved):

packages/amplify-e2e-tests/src/__tests__/custom-resource-with-storage.test.ts
sobolk: (as follow up ofc).

jhockett (approved):

sobolk (approved):

Issue #2087, opened 2019-08-19
Uploading the files of the web app to S3 should be in a different order


timoteialbu:
>**Note**: If your issue/bug is regarding the AWS Amplify Console service, please log it in the 
[Amplify Console GitHub Issue Tracker](https://github.com/aws-amplify/amplify-console/issues)

**Describe the bug**
We get the following error when deploying our app:
![image](https://user-images.githubusercontent.com/6344489/63282897-14574c00-c26d-11e9-8124-961775c44259.png)

This error happens for a brief 15-30 seconds, but its enough to give us less than 100% uptime.

The reason this happens is because of the order that files are uploaded in. Basically the `index.html` file is uploaded before the scripts that it requires are, so when a user refreshes the page in that upload time frame, the `index.html` file they receive references a script that is not in S3 yet and therefore it throws the error above.

I added a console log in here https://github.com/aws-amplify/amplify-cli/blob/master/packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js#L65 and was able to observe this upload order:

![image](https://user-images.githubusercontent.com/6344489/63283051-6ef0a800-c26d-11e9-8b4c-0902895dc94d.png)

**To Reproduce**
1. Upload the app.
2. This is a little harder to see if your internet is blazing fast, but if you can throttle it a little, then try to reload the page right after the `index.html` file is uploaded and you will see the above behavior.

**Expected behavior**
I expect the `index.html` to be uploaded last so that all the resources it needs are available to it. Its not about the order above, as `promise-sequential` which is a wrapper around `Promise.all` does not guarantee the order the promises are fulfilled in. So you probably need to upload everything else first and once that succeeds, upload the index.html file. I realize as I am typing this that maybe a non-react app while have a different entry point potentially, so maybe this is not the ideal solution for all use cases.

**Screenshots**
See above.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Firefox 
 - Version: 68.0.1

**Smartphone (please complete the following information):**
Did not test as this is not a browser/OS specific issue.

**Additional context**
This is a reactjs app.

Is this something I could work on in case its not a priority? I would be more than happy to contribute as we have been using `amplify-cli` extensively at my company for more than a year.

attilah:
[...]

PR #12852, opened 2023-06-22
fix: s3 upload index.html at last

0618:
#### Description of changes

This PR is to mitigate https://github.com/aws-amplify/amplify-cli/issues/2087 as proposed by @timoteialbu .

**Issue**: since `index.html` is not always uploaded to S3 at last, it sometimes causes file not found issue if a dependent file is uploaded after index.html

**Fix**: `await Promise.all` all other files except `index.html`, then upload `index.html` at the end.

**Note**: 
1. This PR is just a mitigation. To fundamentally resolve the issue, users need to enable some sort of cache or versionize.
2. I've checked some popular build tools and they all use `index.html`. The build tools I checked are: angular-cli@16,  react@18.2.0 + cra@5.0.1, react + next@13.4.7, react + vite@4.3.9, vue@3.3.4, vue + nuxt@3.5.2, vue + vite@4.3.9

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/2087

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-category-hosting/__tests__/lib/S3AndCloudFront/helpers/file-uploader.test.js
+++ b/packages/amplify-category-hosting/__tests__/lib/S3AndCloudFront/helpers/file-uploader.test.js
@@ -91,15 +91,6 @@ describe('cloudfront-manager', () => {
 
   beforeEach(() => {});
 
-  test('index.html files should be sorted to the end of the list', () => {
-    const sortedFiiles1 = fileUploader.exportForTesting.sortUploadFiles(['index.html', 'assets/index-87aefcd1.js', 'assets/index-d41dc7ea.css', 'assets/mapbox-gl-0f4a37a4.js']);
-    const sortedFiiles2 = fileUploader.exportForTesting.sortUploadFiles(['assets/index-87aefcd1.js', 'index.html', 'assets/index-d41dc7ea.css', 'assets/mapbox-gl-0f4a37a4.js', 'logo.png']);
-    const sortedFiiles3 = fileUploader.exportForTesting.sortUploadFiles(['assets/index-87aefcd1.js', 'index.html', 'assets/index-d41dc7ea.css', 'assets/index.html', 'logo.png']);
-    expect(sortedFiiles1).toEqual(['assets/index-87aefcd1.js', 'assets/index-d41dc7ea.css', 'assets/mapbox-gl-0f4a37a4.js', 'index.html']);
-    expect(sortedFiiles2).toEqual(['assets/index-87aefcd1.js', 'assets/index-d41dc7ea.css', 'assets/mapbox-gl-0f4a37a4.js', 'logo.png', 'index.html']);
-    expect(sortedFiiles3).toEqual(['assets/index-87aefcd1.js', 'assets/index-d41dc7ea.css', 'logo.png', 'index.html', 'assets/index.html']);
-  });
-
   test('run', async () => {
     await fileUploader.run(mockContext, 'mockDistributionFolder');
     expect(fileScanner.scan).toBeCalled();
--- a/packages/amplify-category-hosting/__tests__/lib/S3AndCloudFront/helpers/file-uploader.test.js
+++ b/packages/amplify-category-hosting/__tests__/lib/S3AndCloudFront/helpers/file-uploader.test.js
@@ -6,7 +6,9 @@ jest.mock('../../../../lib/S3AndCloudFront/helpers/file-scanner', () => {
     }),
   };
 });
-jest.mock('../../../../lib/S3AndCloudFront/helpers/upload-file', () => {return {uploadFile: jest.fn()}});
+jest.mock('../../../../lib/S3AndCloudFront/helpers/upload-file', () => {
+  return { uploadFile: jest.fn() };
+});
 
 const fs = require('fs-extra');
 const mime = require('mime-types');
@@ -98,11 +100,11 @@ describe('cloudfront-manager', () => {
     expect(fileScanner.scan).toBeCalled();
   });
 
-  test('uploadFileCalls should have the index.html at the end', () => { 
+  test('uploadFileCalls should have the index.html at the end', () => {
     expect(uploadFile).toHaveBeenCalled();
     const uploadFileCalls = uploadFile.mock.calls;
 
     expect(uploadFileCalls.length).toBe(5);
     expect(uploadFileCalls[4][3]).toBe('index.html');
-  })
+  });
 });
--- a/packages/amplify-category-hosting/__tests__/lib/S3AndCloudFront/helpers/file-uploader.test.js
+++ b/packages/amplify-category-hosting/__tests__/lib/S3AndCloudFront/helpers/file-uploader.test.js
@@ -91,6 +91,15 @@ describe('cloudfront-manager', () => {
 
   beforeEach(() => {});
 
+  test('index.html files should be sorted to the end of the list', () => {
+    const sortedFiiles1 = fileUploader.exportForTesting.sortUploadFiles(['index.html', 'assets/index-87aefcd1.js', 'assets/index-d41dc7ea.css', 'assets/mapbox-gl-0f4a37a4.js']);
+    const sortedFiiles2 = fileUploader.exportForTesting.sortUploadFiles(['assets/index-87aefcd1.js', 'index.html', 'assets/index-d41dc7ea.css', 'assets/mapbox-gl-0f4a37a4.js', 'logo.png']);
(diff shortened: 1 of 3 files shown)


packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
```diff
@@ -19,7 +19,13 @@ async function run(context, distributionDirPath) {
...
 
-  fileList.forEach(filePath => {
+
+  await Promise.all(fileList.map(async filePath => {
+    if(!filePath.includes('index.html')) {
+      return await uploadFileTasks.push(() => uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, hasCloudFront));
+  }}));
+
+  fileList.filter(filePath => filePath.includes('index.html')).forEach(filePath => {
     uploadFileTasks.push(() => uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, hasCloudFront));
```
sobolk, line 29: I suggest to refactor this into two sections for better readability. something like
```
const indexFile = ... select index from fileList...;
const otherFiles = ... select all other files from fileList....;

await Promise.all(...upload all files...);
await upload(index);

```

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
```diff
@@ -19,7 +19,13 @@ async function run(context, distributionDirPath) {
 
   const hasCloudFront = !!context?.exeInfo?.template?.Resources?.CloudFrontDistribution;
 
-  fileList.forEach(filePath => {
+
+  await Promise.all(fileList.map(async filePath => {
+    if(!filePath.includes('index.html')) {
+      return await uploadFileTasks.push(() => uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, hasCloudFront));
```
jhockett, line 25: nit: I don't think this `await` is needed since the Promise.all is being awaited

```suggestion
      return uploadFileTasks.push(() => uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, hasCloudFront));
```

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
0618: Done 3abfb5164c510f1697bc828246313a66f817646e

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
0618: Done 3abfb5164c510f1697bc828246313a66f817646e

lazpavel (approved):

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
```diff
@@ -19,13 +19,14 @@ async function run(context, distributionDirPath) {
...
+  const filesToUploadFirst = fileList.filter(filePath => !filePath.includes('index.html'));
+  const filesToUploadLast = fileList.filter(filePath => filePath.includes('index.html'));
 
-  await Promise.all(fileList.map(async filePath => {
-    if(!filePath.includes('index.html')) {
-      return await uploadFileTasks.push(() => uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, hasCloudFront));
-  }}));
+  await Promise.all(filesToUploadFirst.map(async filePath => {
+      return uploadFileTasks.push(() => uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, hasCloudFront));
+  }));
```
jhockett, line 27: I missed this the first time, but please move this inside the try/catch after the spinner starts so customers can see something is uploading

jhockett (changes requested):

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
0618: I think it doesn't matter for the user because `sequential` runs after the `spinner` and we have to push to the `uploadFileTasks` first

sobolk: I did not notice that `sequential` before and that we're accumulating deferred `() => uploadFile` callbacks. Sorry about that.

If actual upload happens in `sequential` and we'll never parallelize it then perhaps we should just make sure `fileList` has elements in right order.

sobolk: Consider just concatenating `filesToUploadFirst.concat(filesToUploadLast)` and feeding this to logic that was here before. We don't need two mappings nor Promise.all.

0618: That's a good point. I just noticed `sequential` when making the previous comment 😂 

Now I wonder if `sequential` is necessary. What's the purpose of `sequential`? Would it make the load too slow?

If we want to control the concurrency, would it be better to use [promise-pool](https://www.npmjs.com/package/@supercharge/promise-pool) instead?

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
```diff
@@ -34,6 +32,13 @@ async function run(context, distributionDirPath) {
   }
 }
 
+function sortUploadFiles(fileList) {
+  const filesToUploadLast = "index.html";
+  const sortFiles = (fileA, fileB) => fileA.includes(filesToUploadLast) ? 1 : fileB.includes(filesToUploadLast) ? -1 : 0;
```
0618, line 37: I'll add prettier in another PR

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
```diff
@@ -71,4 +76,7 @@ async function uploadFile(s3Client, hostingBucketName, distributionDirPath, file
 
 module.exports = {
   run,
+  exportForTesting: {
+    sortUploadFiles,
+  }
```
sobolk, line 81: we should not do this. I.e. don't open internals.
instead we should mock modules that are imported into uploader and assert order of `s3Client.upload` calls.

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
```diff
@@ -13,15 +13,13 @@
...
-  const uploadFileTasks = [];
   const s3Client = await getS3Client(context, 'update');
   const hostingBucketName = getHostingBucketName(context);
 
   const hasCloudFront = !!context?.exeInfo?.template?.Resources?.CloudFrontDistribution;
 
-  fileList.forEach(filePath => {
-    uploadFileTasks.push(() => uploadFile(s3Client, hostingBucketName, distributionDirPath, filePath, hasCloudFront));
-  });
+  uploadFileTasks = sortUploadFiles(fileList)
```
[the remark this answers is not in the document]
0618: good catch...

packages/amplify-category-hosting/lib/S3AndCloudFront/helpers/file-uploader.js
0618: Done e4ab4609c43863fc57d5f4d9e431c011a3e96927

packages/amplify-category-hosting/__tests__/lib/S3AndCloudFront/helpers/file-uploader.test.js
```diff
@@ -80,19 +84,27 @@ describe('cloudfront-manager', () => {
...
 
-  beforeAll(() => {
+  beforeAll(async () => {
     fs.createReadStream = jest.fn(() => {
       return {};
     });
     mime.lookup = jest.fn(() => {
       return 'text/plain';
     });
+    await fileUploader.run(mockContext, 'mockDistributionFolder');
```
sobolk, line 94: the "functionality under test" should not be in `beforeAll` block, that block is reserved for test setup.
Readability of the tests (i.e. right call in right block) should take precedence over duplication.

sobolk (commented):
Looks good. Please address remaining comment.

packages/amplify-category-hosting/__tests__/lib/S3AndCloudFront/helpers/file-uploader.test.js
0618: Done ada487f2515a0f9edef8eaeaa85a0e6c7c627024

sobolk (approved):

jhockett (approved):

PR #12857, opened 2023-06-23
ci: fix bug overusing parent account

awsluja:
#### Description of changes
USE_PARENT_ACCOUNT was set to 'false', which is considered truthy by this logic, which just checks if the variable is a non-empty string.
if [[ ! -z "$USE_PARENT_ACCOUNT" ]]; 

This lead to 80 tests running on the parent account, specifically on Windows.
Fixing this will allow those tests to pass more consistently.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


codebuild_specs/migration_tests_v8.yml
```diff
@@ -25,7 +25,7 @@ phases:
   build:
     commands:
       # you can provide a codebuild source version to use old cache and skip all other jobs :)
-      - export NODE_OPTIONS=--max-old-space-size=8096
+      # - export NODE_OPTIONS=--max-old-space-size=8096
```
goldbez, line 28: Why do we no longer need to set this option for these tests?

packages/amplify-e2e-core/src/categories/notifications.ts
```diff
@@ -43,7 +43,7 @@ export const addNotificationChannel = async (
   channel: string,
   hasAnalytics = false,
   hasAuth = false,
-  testingWithLatestCodebase = true,
+  testingWithLatestCodebase = false,
```
goldbez, line 46: Why is this set to false now?

goldbez (approved):

kuhlmanp (approved):

PR #12860, opened 2023-06-23
ci: fix windows test bugs

awsluja:
#### Description of changes
- Fix bug where too many tests were running on parent account
- Fix bug where tests were using "testingWithLatestCodebase" flag, when they shouldn't be (this is reserved for migration tests)
- Fix bug related to pathing for Geo files
- Fix bug related to code coverage causing failure on amplify-app tests / datastore modelgen tests
- other small bugs
image

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


goldbez (approved):

packages/amplify-e2e-tests/src/__tests__/function_14.test.ts
```diff
@@ -33,7 +33,7 @@ describe('amplify push function cases:', () => {
     const functionPath = pathManager.getResourceDirectoryPath(projRoot, AmplifyCategories.FUNCTION, funcName);
     const amplifyState = JSONUtilities.readJson<any>(path.join(functionPath, 'amplify.state'));
 
-    expect(amplifyState.scripts.build).toEqual('npm install --no-bin-links --production');
+    expect(amplifyState.scripts.build).toMatch('npm.* install --no-bin-links --production');
```
kuhlmanp, line 36: It probably won't ever come up, but this is an unsafe regex to be using. Better to do 'npm(?:\.exe) ...' if we're expecting a windows executable

packages/amplify-e2e-tests/src/__tests__/function_14.test.ts
```diff
@@ -45,7 +45,7 @@ describe('amplify push function cases:', () => {
     const functionPath = pathManager.getResourceDirectoryPath(projRoot, AmplifyCategories.FUNCTION, funcName);
     const amplifyState = JSONUtilities.readJson<any>(path.join(functionPath, 'amplify.state'));
 
-    expect(amplifyState.scripts.build).toEqual('yarn --no-bin-links --production');
+    expect(amplifyState.scripts.build).toMatch('yarn.* --no-bin-links --production');
```
kuhlmanp, line 48: See above comment about expecting npm

.circleci/local_publish_helpers.sh
```diff
@@ -236,7 +236,7 @@ function retry {
 
     resetAwsAccountCredentials
     TEST_SUITE=${TEST_SUITE:-"TestSuiteNotSet"}
-    aws cloudwatch put-metric-data --metric-name FlakyE2ETests --namespace amplify-cli-e2e-tests --unit Count --value $n --dimensions testFile=$TEST_SUITE
+    aws cloudwatch put-metric-data --metric-name FlakyE2ETests --namespace amplify-cli-e2e-tests --unit Count --value $n --dimensions testFile=$TEST_SUITE || true
```
sobolk, line 239: why are we swallowing error ?

.circleci/local_publish_helpers.sh
```diff
@@ -296,9 +296,19 @@ function runE2eTestCb {
...
+            NODE_V8_COVERAGE=$E2E_TEST_COVERAGE_DIR yarn e2e --forceExit --no-cache --maxWorkers=4 $TEST_SUITE -t "$failedTests"
+        fi
     else
-        NODE_V8_COVERAGE=$E2E_TEST_COVERAGE_DIR yarn e2e --forceExit --no-cache --maxWorkers=4 $TEST_SUITE
+        if [[ ! -z "$DISABLE_COVERAGE" ]]; then
+            echo Running WITHOUT coverage
+            yarn e2e --forceExit --no-cache --maxWorkers=4 $TEST_SUITE
+        else
+            NODE_V8_COVERAGE=$E2E_TEST_COVERAGE_DIR yarn e2e --forceExit --no-cache --maxWorkers=4 $TEST_SUITE
+        fi
```
sobolk, line 311: could we `export NODE_V8_COVERAGE=$E2E_TEST_COVERAGE_DIR` instead and refactor this if(disable coverage) outside to avoid this conditional Christmas tree

codebuild_specs/migration_tests_v10.yml
```diff
@@ -25,7 +25,7 @@ phases:
   build:
     commands:
       # you can provide a codebuild source version to use old cache and skip all other jobs :)
-      - export NODE_OPTIONS=--max-old-space-size=8096
+      # - export NODE_OPTIONS=--max-old-space-size=8096
```
sobolk, line 28: please remove commented out code.

packages/amplify-e2e-core/src/categories/notifications.ts
```diff
@@ -43,7 +43,7 @@ export const addNotificationChannel = async (
   channel: string,
   hasAnalytics = false,
   hasAuth = false,
-  testingWithLatestCodebase = true,
+  testingWithLatestCodebase = false,
```
sobolk, line 46: `false` seems correct (it's a pattern everywhere).
was `true` introduced by mistake? or was it always here?

goldbez (approved):

scripts/split-e2e-tests-codebuild.ts
```diff
@@ -240,6 +253,9 @@ const splitTestsV3 = (
       if (j.useParentAccount) {
         tmp.env.variables.USE_PARENT_ACCOUNT = 1;
       }
+      if (j.disableCoverage) {
+        tmp.env.variables.DISABLE_COVERAGE = 1;
+      }
```
sobolk, line 258: can we rename `j` and `tmp` to something meaningful?

.circleci/local_publish_helpers.sh
awsluja: Sometimes we reach a token expiration before this metric is collected; in those situations, we would like to ignore the error. Sometimes tests passes but token expires after test has completed, this causes unintentional "failures" which shouldn't be failures. A solution to token expiry is in progress

.circleci/local_publish_helpers.sh
awsluja: That could work; I image we'd need to set, and unset if we are exporting, as we only want this specific command to be tracked for coverage - I'll create a ticket for this

packages/amplify-e2e-core/src/categories/notifications.ts
awsluja: It must have been introduced by mistake, our windows image on circleci was forgiving & allowed this error to go unnoticed.

.circleci/local_publish_helpers.sh
sobolk: If this has to be localized then maybe not `export` but there should be a way to simplify this `if` tree here.

.circleci/local_publish_helpers.sh
sobolk: Please leave comment in the code explaining this.

scripts/split-e2e-tests-codebuild.ts
awsluja: updated

sobolk (dismissed):

goldbez (approved):

goldbez (approved):

sobolk (approved):

PR #12869, opened 2023-06-26
chore: prettier optimization

0618:
#### Description of changes

This PR does 3 things for prettier:
1. not ignore the some `package/*/lib/` so that prettier will work on those files
2. fix the prettierPath so that VS Code is able to auto-format on save
3. add prettier to husky so that it'll automatically prettify the changed code

Tested on local

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


.vscode/settings.json
```diff
@@ -20,6 +20,6 @@
     "**/.pnp.*": true
   },
   "typescript.enablePromptUseWorkspaceTsdk": true,
-  "prettier.prettierPath": "../node_modules/prettier",
```
0618, line 23: the 2nd change of this PR

.prettierignore
```diff
@@ -15,3 +15,7 @@ packages/amplify-e2e-tests/amplify-e2e-reports
 packages/amplify-environment-parameters/src/schemas/BackendParameters.schema.json
 packages/amplify-nodejs-function-runtime-provider/src/__tests__/utils/handlerWithSyntaxError.js
 packages/amplify-velocity-template/src/parse/velocity.yy
+
+# Exceptions:
+!packages/amplify-category-hosting/lib
+!packages/amplify-frontend-*/lib
```
0618, line 21: 1st change of this PR

package.json
```diff
@@ -174,5 +169,9 @@
     "nth-check": "^2.0.1",
     "undici": "^5.8.0",
     "pkg-fetch": "https://github.com/aws-amplify/pkg-fetch#ad4a21feb533d338bf951e7ba28cea7256aedeff"
+  },
+  "lint-staged": {
+    "*.js": "eslint --cache --fix",
+    "*.{js,css,md}": "prettier --write"
```
sobolk, line 175: this should also apply to .ts files.

sobolk (approved):

package.json
```diff
@@ -81,13 +82,6 @@
     "type": "git",
     "url": "https://github.com/aws-amplify/amplify-cli.git"
   },
-  "husky": {
-    "hooks": {
-      "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-push": "yarn build-tests-changed && yarn split-e2e-tests",
```
sobolk, line 87: we shouldn't drop these hooks.

sobolk (changes requested):

package.json
```diff
@@ -139,14 +133,15 @@
     "eslint-plugin-spellcheck": "^0.0.17",
     "execa": "^5.1.1",
     "glob": "^7.2.0",
-    "husky": "^3.1.0",
+    "husky": ">=6",
```
jhockett, line 136: Why upgrade husky in this PR?

Also, I think using `>=` is too permissive. We don't know that husky won't make breaking changes that will negatively affect our ability to run git hooks.

package.json
```diff
@@ -139,14 +133,15 @@
...
     "istanbul-lib-coverage": "^3.2.0",
     "istanbul-lib-report": "^3.0.0",
     "istanbul-reports": "^3.1.5",
     "jest": "^29.0.0",
     "jest-circus": "^29.0.0",
     "jest-junit": "^12.0.0",
     "js-yaml": "^4.0.0",
+    "lint-staged": ">=10",
```
jhockett, line 144: Why is this package needed? Same comment about the `>=` versioning as above.

package.json
0618: Oh....I just ran `npx mrm@2 lint-staged` ([code](https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged)) and didn't pay attention to the change 😂

package.json
0618: here's the doc https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged

package.json
0618: Good catch. Added them back 293ce82c0824dc4117c26fd96be8857f0352a238

package.json
jhockett: Can prettier be executed in the existing [pre-commit](https://github.com/aws-amplify/amplify-cli/blob/dev/package.json#L88) script? I tend to be conservative about adding dependencies and devDependencies.

package.json
0618: Good catch. It was automatically done by 

I just took the chance and upgraded it to the newest version 293ce82c0824dc4117c26fd96be8857f0352a238

package.json
```diff
@@ -69,7 +69,8 @@
     "verify-api-extract": "yarn extract-api && ./scripts/verify-extract-api.sh",
     "verify-commit": "yarn ts-node ./scripts/verify-commit.ts",
     "verify-yarn-lock": "./scripts/verify-yarn-lock.sh",
-    "workflow-results": "yarn ts-node ./scripts/cci-workflow-results.ts"
+    "workflow-results": "yarn ts-node ./scripts/cci-workflow-results.ts",
+    "postinstall": "husky install"
```
sobolk, line 73: please sort `scripts`

sobolk (dismissed):

package.json
0618: Yes, prettier can be done at pre-commit auto-magically. This way we don't have to wait until the PR checker fails. 

Also, the change in this PR will allow VS Code to format on save.

sobolk (dismissed):
huh. I didn't know we diverged from sorted scripts that much. Thanks for extra cleanup!

package.json
0618: Sure. Just sorted all the scripts. 

just FYI, there's a bug in VS Code, so the `-` can't be sorted correctly, but we can keep it as it is if we use a sorting tool on VS Code https://github.com/Tyriar/vscode-sort-lines/issues/11

sobolk (approved):

package.json
```diff
@@ -84,8 +85,7 @@
   "husky": {
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
-      "pre-push": "yarn build-tests-changed && yarn split-e2e-tests",
-      "pre-commit": "./scripts/package-json-check.sh && yarn verify-commit"
```
jhockett, line 88: `verify-commit` must not be removed

0618: and it's not, it's moved to the `.husky/pre-commit`

package.json
0618: We can't use that because 
1. `yarn prettier --write && yarn eslint --cache --fix` doesn't work directly in pre-commit
2. even though it would work, we only want to lint the staged files, which is much quicker

If not using `lint-staged`, we could write some script mentioned [here](https://prettier.io/docs/en/precommit.html#option-6-shell-script), but I think it could have unknown security issue as well. I would rather have Depend-bot warns me the issue.

jhockett: We have those scripts already defined in the root package.json. I should have said `yarn prettier-changes && yarn lint-fix`

0618: Right, but they run through the entire repo and they don't work at the pre-commit script. 

Maybe I missed something, did you test it works on your end?

jhockett: Those commands work on my end and don't run through the entire repo, but if they're not working for you then we perhaps the new package is the better approach.

.husky/.gitignore
```diff
@@ -0,0 +1 @@
+_
```
jhockett, line 1: What's the purpose of this file? Can the root `.gitignore` be used instead?

.husky/pre-commit
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
jhockett, line 4: Can we invoke with `yarn` instead of `npx`?

0618: I don't think so. `npx` runs the `bin`, but `yarn` runs the scripts in package.json

.husky/.gitignore
0618: there's a `_` folder generated under `.husky/`, which should be ignored.

This file is auto-generated by `npx mrm@2 lint-staged` https://prettier.io/docs/en/precommit.html#option-1-lint-stagedhttpsgithubcomokonetlint-staged

package.json
```diff
@@ -85,7 +85,7 @@
     "hooks": {
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
       "pre-push": "yarn build-tests-changed && yarn split-e2e-tests",
-      "pre-commit": "./scripts/package-json-check.sh && yarn verify-commit"
+      "pre-commit": "yarn prettier-changes && yarn lint-fix && ./scripts/package-json-check.sh && yarn verify-commit"
```
0618, line 88: use `yarn prettier-changes && yarn lint-fix` in `pre-commit` so that we don't need `lint-staged`

package.json
```diff
@@ -139,7 +139,7 @@
     "eslint-plugin-spellcheck": "^0.0.17",
     "execa": "^5.1.1",
     "glob": "^7.2.0",
-    "husky": "^3.1.0",
+    "husky": "^8.0.3",
```
0618, line 142: upgrade husky

package.json
```diff
@@ -11,12 +11,12 @@
     "build": "lerna run build",
     "check-version-conflicts:beta": "yarn ts-node ./scripts/check-version.ts beta",
     "check-version-conflicts:release": "yarn ts-node ./scripts/check-version.ts",
-    "cloud-e2e": "bash -c 'source ./scripts/cloud-e2e.sh && cloudE2E'",
-    "cloud-e2e-beta": "bash -c 'source ./scripts/cloud-e2e.sh && cloudE2EBeta'",
-    "cloud-e2e-local": "bash -c 'source ./scripts/cloud-e2e.sh && cloudE2ELocal'",
     "clean": "lerna run clean && yarn rimraf tsconfig.tsbuildinfo && lerna clean --yes && yarn rimraf node_modules",
+    "cloud-e2e-beta": "bash -c 'source ./scripts/cloud-e2e.sh && cloudE2EBeta'",
```
0618, line 15: scripts are sorted, but no changes

package.json
0618: That's a good point. I just updated the PR and removed `lint-staged` 🙌

jhockett (approved):

sobolk (approved):

package.json
0618: Note:  the pre-commit script didn't work on my end was because `lint-stage` changed my git hook path to `..husky`. 

```sh
$ git config core.hooksPath
> .husky
```

PR #12937, opened 2023-07-12
chore: use-peer deps for cdk/constructs on internal libs

alharris-at:
#### Description of changes
Flipping to peerDeps for most of our cdk and constructs dependencies (excluding top-level CLI, test packages, and deployed custom resource package) to reduce the burden of maintaining exact version parity (or worrying about duplicate versions) across all these packages. Bumps should just include changing top-level CLI target.

Related API Category change https://github.com/aws-amplify/amplify-category-api/pull/1664

#### Issue #, if available
N/A, but see ongoing release pain (I'm doing something similar for the api category).

Unit + E2E Tests

#### Checklist
- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-category-auth/package.json
+++ b/packages/amplify-category-auth/package.json
@@ -36,12 +36,10 @@
     "@aws-amplify/cli-extensibility-helper": "3.0.9",
     "amplify-headless-interface": "1.17.4",
     "amplify-util-headless-input": "1.9.13",
-    "aws-cdk-lib": "~2.68.0",
     "aws-sdk": "^2.1405.0",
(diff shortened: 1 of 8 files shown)


phani-srikar (approved):

sobolk (commented):
If we do this we should make top level package (`cli-internal`?) pin version to `~` or explicit.
So that bumping CDK is intentional action (other than yarn.lock).
We should also add check that asserts there's only one version of CDK in collective deps graph (by parsing `yarn why --recursive aws-cdk-lib` or something like that).

PR #12973, opened 2023-07-19
feat: generate components using graphql

awinberg-aws:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


Jshhhh (approved):

rtpascual (dismissed):

packages/amplify-util-uibuilder/src/__tests__/generateComponents.test.ts
```diff
@@ -32,18 +39,23 @@ jest.mock('../commands/utils/featureFlags', () => ({
   getTransformerVersion: jest.fn().mockReturnValue(2),
 }));
 
+const defaultStudioFeatureFlags = {
+  autoGenerateForms: 'true',
+  autoGenerateViews: 'true',
```
letsbelopez, line 44: do we need this anymore? Is views a part of feature flags?

awinberg-aws: I'll create a follow-up task in our backlog to remove all traces of Views in our codebase.

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
```diff
@@ -34,10 +37,66 @@ export const run = async (context: $TSContext, eventType: 'PostPush' | 'PostPull
       studioClient.isGraphQLSupported ? getAmplifyDataSchema(context) : Promise.resolve(undefined),
     ]);
 
-    const nothingWouldAutogenerate =
-      !dataSchema || !studioClient.metadata.autoGenerateForms || !studioClient.isGraphQLSupported || !studioClient.isDataStoreEnabled;
+    let canCodegenGraphqlComponents = false;
+    let apiConfiguration: ApiConfiguration = {
+      dataStoreConfig: {},
+    };
```
letsbelopez, line 43: This generateComponents function could be a lot cleaner with a helper function to get the api configuration.

packages/amplify-util-uibuilder/src/__tests__/generateComponents.test.ts
```diff
@@ -14,11 +15,17 @@ jest.mock('../commands/utils/featureFlags', () => ({
...
+}));
+
 const awsMock = aws as any;
 const utilsMock = utils as any;
+const isDataStoreEnabledMocked = isDataStoreEnabled as any;
+const getTransformerVersionMocked = getTransformerVersion as any;
+const getCodegenConfigMocked = getCodegenConfig as any;
 
-const isDataStoreEnabledMocked = jest.mocked(isDataStoreEnabled);
-const getTransformerVersionMocked = jest.mocked(getTransformerVersion);
```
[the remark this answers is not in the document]
awinberg-aws: Local tooling issue, I was getting errors that `jest.mocked()` does not exist and the entire test file would be failed. They're replaced with casts to `any` a couple lines higher, not sure why the diff ended up being removals and additions instead of as line updates.

packages/amplify-util-uibuilder/src/commands/generateComponents.ts
```diff
@@ -34,10 +37,66 @@ export const run = async (context: $TSContext, eventType: 'PostPush' | 'PostPull
...
+
+    if (!studioClient.isDataStoreEnabled && studioClient.metadata.isGraphQLEnabled) {
+      printer.debug('building graphql config');
+      // attempt to get api codegen info
+      const projectPath = context.exeInfo.localEnvInfo.projectPath;
+      const componentsPath = getUiBuilderComponentsPath(context);
+      let promptForUpdateCodegen = false;
+      const relativeToComponentsPath = (importPath: string) => {
```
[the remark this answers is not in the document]
awinberg-aws: good suggestion, pulling all the config generation out into a helper

letsbelopez (dismissed):
LGTM

rtpascual (dismissed):

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
```diff
@@ -0,0 +1,71 @@
...
+import { AmplifyStudioClient } from '../../clients';
+import { $TSContext } from '@aws-amplify/amplify-cli-core';
+import { getUiBuilderComponentsPath } from './getUiBuilderComponentsPath';
+import { getCodegenConfig } from 'amplify-codegen';
+import { ApiConfiguration } from 'aws-sdk/clients/amplifyuibuilder';
+import path from 'path';
+
+export function relativeToComponentsPath(importPath: string, context: $TSContext): string {
+  const componentsPath = getUiBuilderComponentsPath(context);
+  return path.relative(componentsPath, importPath).split(path.sep).join('/');
```
jhockett, line 11: Does this work on as expected on Windows?

I think it would be better to use `path.join` to combine the path pieces into a single string.

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
```diff
@@ -0,0 +1,71 @@
...
+      const isQueryingTooShallow = (codegenConfig.getQueryMaxDepth() || 0) < minQueryDepth;
+
+      if (studioClient.metadata.formFeatureFlags.isRelationshipSupported && isQueryingTooShallow) {
+        promptForUpdateCodegen = true;
+        printer.warn(`Forms with relationships require a maximum query depth of at least ${minQueryDepth}.`);
+      }
+      return apiConfiguration;
+    } catch {
+      promptForUpdateCodegen = true;
+      printer.warn(
```
jhockett, line 50: Why print a warning and not fail fast here?

packages/amplify-util-uibuilder/types/amplify-codegen.d.ts
```diff
@@ -0,0 +1,12 @@
+declare module 'amplify-codegen' {
```
jhockett, line 1: Why is this needed? Can this type be exported from the `amplify-codegen` package instead?

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
awinberg-aws: This works as expected and is actually because windows is the "exception" to the norm. These strings get dropped into jsx files as import paths, so the need to be built with `/` instead of `\` like we would get when running on windows

packages/amplify-util-uibuilder/types/amplify-codegen.d.ts
awinberg-aws: I asked the team owning the package and was told they do not plan to publish types. Hand-writing my own types was required due to our plugin having `"strict": true` in our tsconfig while other plugins referencing the library do not.

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
jhockett: That makes sense, thanks for explaining! Can you leave a comment explaining this in the code for future readers?

I'd prefer we use `path.posix.join` instead if it will work for all platforms, but it's not a merge blocker.

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
awinberg-aws: we chose to warn rather than fail fast as failing as an API configuration is not a hard requirement for every component a customer could be generating.

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
awinberg-aws: I like that suggestion, I'll make sure we try switching to `path.posix.join` and document why this is needed on the next update.

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
jhockett: I think either the warning's content is too alarming as it begins with "Error" or this is an inadequate mechanism to alert customers that a problem occurred they need to address.

Is there a limitation that's preventing us from determining if a component is expected from having an API configuration?

jhockett (dismissed):

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
```diff
@@ -0,0 +1,73 @@
...
+    }
+  }
+
+  return {
+    noApiConfig: {},
+  };
+}
+
+export function hasDataStoreConfiguration(apiConfiguration: ApiConfiguration): boolean {
+  return 'dataStoreConfig' in apiConfiguration && apiConfiguration.dataStoreConfig != undefined;
```
jhockett, line 68: Use of double equals

```suggestion
  return apiConfiguration?.dataStoreConfig !== undefined;
```

packages/amplify-util-uibuilder/src/commands/utils/getApiConfiguration.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export function hasGraphQLConfiguration(apiConfiguration: ApiConfiguration): boolean {
+  return 'graphQLConfig' in apiConfiguration && apiConfiguration.graphQLConfig != undefined;
```
jhockett, line 72: Same here

jhockett (approved):

rtpascual (approved):

sobolk (changes requested):
Please hold with merge for a moment. We lost static analysis in our pipelines.

sobolk:
> Please hold with merge for a moment. We lost static analysis in our pipelines.

Static analysis is restored in pipelines, but has not run on this PR. @awinberg-aws before merging I recommend to either validate PR manually by running `yarn lint-check` or merge latest `dev` into PR to refresh pipeline definition.

awinberg-aws:
@sobolk I'll rebase and re-push

jhockett (dismissed):

Milan-Shah (dismissed):

Milan-Shah (dismissed):

Amplifiyer (approved):

joebuono (approved):

Issue #12817, opened 2023-06-16
Parsing of Identity Pool Roles fails when role arn contains multiple '/' characters


geoff-mwi:
### How did you install the Amplify CLI?

yarn build from v12.1.0 tag

### If applicable, what version of Node.js are you using?

18.15.0

### Amplify CLI Version

12.0.3 (v12.1.0 tag)

### What operating system are you using?

Mac M2

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No manual changes made

### Describe the bug

I attempted and failed to run `amplify import auth` to import Cognito User Pool and Identity Pool settings from my AWS account.

When creating a Cognito Identity Pool, the v2 console guides you through creating Unauthenticated and Authenticated roles.

When creating the roles, the console will prepend `service-role/` to whatever IAM role name you specify.

For example, in this UI I am entering `my-role`:
![image](https://github.com/aws-amplify/amplify-cli/assets/136322535/8372465a-24a0-49f4-a106-116bfe0bc0eb)

And the subsequent role arn will be created as:
```
arn:aws:iam::<account-id>:role/service-role/my-role
```

I believe this probably differs from the original console UI which might not prepend `service-role/` to the arn.

When running `amplify import auth` the following code attempts to split the Identity Pool's Role arns into two parts and select the last part as the 'resourceName':
https://github.com/aws-amplify/amplify-cli/blob/bc5957a4d5fd2c036fde75d4da657ba733c83a22/packages/amplify-provider-awscloudformation/src/aws-utils/IdentityPoolService.ts#L107

When it attempts to parse `arn:aws:iam::<account-id>:role/service-role/my-role` it fails with an exception `Cannot parse arn: 'arn:aws:iam::<account-id>:role/service-role/my-role'.` due to there being more than two parts.

### Expected behavior

I expect `amplify import auth` to be compatible with the default Cognito Identity Pool console's created role resource names.

### Reproduction steps

1. Create a Cognito Identity Pool with reasonable defaults specified for Authenticated and Unauthenticated roles using the latest console version & ensuring that the role arn's are prepended with `service-role/`
2. Run `amplify import auth`
3. Receive exception `Cannot parse arn: <arn>`

### Additional information

Simply changing the amplify code to use the last part of a multi-part arn allows the `amplify import auth` command to succeed but subsequently breaks builds (as the build pipeline will use the official amplify-cli version with the bug in it).

### Before submitting, please confirm:

[...]

PR #12987, opened 2023-07-21
fix: parsing arn with multiple slashes when importing auth

jhockett:
#### Description of changes
Allow more than 1 forward slash in role arns for authenticated and unauthenticated roles when importing auth

#### Issue #, if available
Fixes https://github.com/aws-amplify/amplify-cli/issues/12817

Manually imported auth (Cognito User Pool and Identity Pool) with multiple slashes in the IAM role names for auth and unauth. Created test web app to confirm account creation and sign-in/sign-out.

These steps fail with 12.2.0 CLI, but succeed with PR changes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli-core/API.md
+++ b/packages/amplify-cli-core/API.md
@@ -8,6 +8,7 @@
 
 import Ajv from 'ajv';
 import { ApiKeyConfig } from '@aws-amplify/graphql-transformer-interfaces';
+import { ARN } from '@aws-sdk/util-arn-parser';
 import { BuildType } from '@aws-amplify/amplify-function-plugin-interface';
 import * as cdk from 'aws-cdk-lib';
 import { ChildProcess } from 'child_process';
@@ -1504,6 +1505,9 @@ export const packageManagers: Record<PackageManagerType, PackageManager>;
 
 // @public (undocumented)
 export type PackageManagerType = 'yarn' | 'npm' | 'pnpm' | 'custom';
+
+// @public (undocumented)
+export const parseArn: (arn: string) => ARN;
 
 // @public (undocumented)
 export function parseHelpCommands(input: $TSAny, commandsInfo: Array<CommandInfo>): {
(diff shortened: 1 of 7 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/IdentityPoolService.ts
```diff
@@ -102,9 +102,8 @@ export class IdentityPoolService implements IIdentityPoolService {
 
     if (arn) {
       const parts = arn.split('/');
-
-      if (parts.length === 2) {
-        resourceName = parts[1];
+      if (parts?.length >= 2) {
+        resourceName = parts.at(-1);
       }
```
sobolk, line 107: could we replace this logic with https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/modules/_aws_sdk_util_arn_parser.html ?

sobolk (approved):

packages/amplify-provider-awscloudformation/src/aws-utils/IdentityPoolService.ts
jhockett: Great suggestion! I'll create a wrapper function in `amplify-cli-core` so that the dependency on `@aws-sdk/utility-arn-parser` v3 can be isolated to one package for now. Thoughts?

sobolk: not sure if that's necessary. if we `^` version of that yarn should take care of duplicates. 
first check if this is solving our problem...

Amplifiyer (approved):

packages/amplify-cli-core/API.md
```diff
@@ -8,6 +8,7 @@
 
 import Ajv from 'ajv';
 import { ApiKeyConfig } from '@aws-amplify/graphql-transformer-interfaces';
+import { ARN } from '@aws-sdk/util-arn-parser';
```
sobolk, line 11: this means we're not really hiding this dependency.

sobolk (approved):

jhockett:
Closing as unnecessary because #13009 was merged to the hotfix branch

PR #13012, opened 2023-07-26
fix: upgrade proxy-agent to ^6.3.0 to remove vm2 deps

awsluja:
#### Description of changes
E2e run https://tiny.amazon.com/1gmqt246c/IsenLink

This removes the last of the vm2 dependencies by upgrading proxy-agent.
Proxy agent constructor has changed since v5, so there were updates as well.
See proxy-agent docs here: https://github.com/TooTallNate/proxy-agents/tree/main/packages/proxy-agent

Proxy-agent uses this module to determine which proxy to use, so we don't need to manually specify http/https anymore, see:
https://www.npmjs.com/package/proxy-from-env
https://www.npmjs.com/package/proxy-from-env#environment-variables

We do still need to conditionally check if proxy vars are set to decide if we want to initialize a new proxyagent.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


yarn.lock
```diff
@@ -31455,18 +31474,6 @@ node-pty@beta:
   languageName: node
   linkType: hard
 
-"vm2@npm:^3.9.3":
```
awsluja, line 31458: vm2 removal is here

packages/amplify-category-notifications/src/pinpoint-helper.ts
```diff
@@ -547,7 +547,7 @@ export const getPinpointClient = async (
   if (httpProxy) {
     aws.config.update({
       httpOptions: {
-        agent: proxyAgent(httpProxy),
+        agent: new ProxyAgent(),
```
sobolk, line 550: is it ok to drop `httpProxy` param here ?

it seems to be a pattern in the PR.

sobolk (approved):

packages/amplify-category-notifications/src/pinpoint-helper.ts
awsluja: yeah see description for more details; tldr it reads from process.env by default so we don't need to manually specify anymore

sobolk: if we read these variables anyway for the `if` I'd rather pass this explicitly.
unless they took away API to do this. In that case might be worth to add comment for future reader why `httpProxy` is not consumed.

awsluja: added notes

sobolk (approved):

shared-scripts.sh
```diff
@@ -125,6 +125,7 @@ function _verifyAPIExtract {
     # download [repo, .cache from s3]
     loadCache repo $CODEBUILD_SRC_DIR
     loadCache .cache $HOME/.cache
+    unset IS_AMPLIFY_CI
```
jhockett, line 128: Why is this included in this PR?

awsluja: short story: codebuild sets IS_AMPLIFY_CI globally, but circleci doesn't, so this achieves parity.
long story:
this PR changed the API slightly, which shined a light on something we didn't see before in codebuild, but happens all the time in circleci. That is, verify-api-extract actually downloads the amplify-cli & installs it due to a post install script for the amplify-cli-npm package. Downloading the CLI has no effect on the API step, but when because its running in codebuild CI - the IS_AMPLIFY_CI flag makes it try to download from s3 (pkg cli) using the binary hash. CircleCI does not set IS_AMPLIFY_CI for the verify-api-extract, so it succeeds there because it just downloads the production pkg cli. Download the pkg cli has no affect on verify-api-extract's result.

jhockett: Would you mind doing a quick one-off PR and merging it first? It will improve discoverability for the change and then this comment can become the PR description.

awsluja: https://github.com/aws-amplify/amplify-cli/pull/13019

jhockett (approved):

PR #13015, opened 2023-07-27
ci: update env names to match cleanup regex

awsluja:
#### Description of changes
Some bucket names were not being generated with a 'test' substring to match the cleanup regex.
This will allow those tests to produce buckets which can be cleaned up by the cleanup script.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
+++ b/packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
@@ -139,8 +139,7 @@ type AWSAccountInfo = {
 
 const PINPOINT_TEST_REGEX = /integtest/;
 const APPSYNC_TEST_REGEX = /integtest/;
-const BUCKET_TEST_REGEX =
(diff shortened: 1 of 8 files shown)


packages/amplify-e2e-tests/src/cleanup-e2e-resources.ts
```diff
@@ -139,7 +139,8 @@ type AWSAccountInfo = {
 
 const PINPOINT_TEST_REGEX = /integtest/;
 const APPSYNC_TEST_REGEX = /integtest/;
-const BUCKET_TEST_REGEX = /test/;
+const BUCKET_TEST_REGEX =
```
jhockett, line 142: This should be redundant once existing resources are cleaned up. Can you leave a comment explaining this is temporary and what it should be reverted to?

jhockett (approved):

sobolk (approved):
we should consider using resource tags instead of this naming convention at some point

PR #13027, opened 2023-07-28
test: enable more linter rules in tests

sobolk:
#### Description of changes

This PR:
- enables batch of linter rules, see changes in `.eslintrc.js` for the list.
- fixes violations in test codebase

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-e2e-tests/src/__tests__/tags.test.ts
+++ b/packages/amplify-e2e-tests/src/__tests__/tags.test.ts
@@ -58,7 +58,7 @@ describe('generated tags test', () => {
 function checkEquality(localTags: {}[], generatedTags: {}[]) {
   localTags.forEach((tagObj) => {
     const rootTag = generatedTags.find((obj) => obj['Key'] === tagObj['Key']);
-    if (tagObj['Key'] !== rootTag['Key']) return false;
+    return tagObj['Key'] === rootTag['Key'];
   });
 
   return true;
--- a/packages/amplify-util-mock/src/__tests__/utils/dynamo-db/util.test.ts
+++ b/packages/amplify-util-mock/src/__tests__/utils/dynamo-db/util.test.ts
@@ -219,6 +219,7 @@ describe('DynamoDB Utils', () => {
         };
         if (typeof callback === 'function') {
           callback(null, response);
+          return undefined;
         } else {
           return {
             promise: jest.fn().mockResolvedValue(response),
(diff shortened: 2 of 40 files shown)


packages/amplify-util-mock/src/__tests__/utils/dynamo-db/util.test.ts
```diff
@@ -219,6 +219,7 @@ describe('DynamoDB Utils', () => {
         };
         if (typeof callback === 'function') {
           callback(null, response);
+          return undefined;
         } else {
```
0618, line 223: nit: Why don't we return nothing or `null` instead?

0618 (approved):
Do we need to run e2e on this PR?

packages/amplify-util-mock/src/__tests__/utils/dynamo-db/util.test.ts
sobolk: `return undefined` preserves behavior.
It's not a goal of this PR to rework this test. Priority is to enable rules so new tests added are held to standard.

sobolk:
> Do we need to run e2e on this PR?
Yes, it's in progress. If we touch product logic OR e2e tests it's advised to validate change. Or face consequences of breaking `dev`.

packages/amplify-e2e-tests/src/__tests__/tags.test.ts
```diff
@@ -58,7 +58,7 @@ describe('generated tags test', () => {
 function checkEquality(localTags: {}[], generatedTags: {}[]) {
   localTags.forEach((tagObj) => {
     const rootTag = generatedTags.find((obj) => obj['Key'] === tagObj['Key']);
-    if (tagObj['Key'] !== rootTag['Key']) return false;
+    return tagObj['Key'] === rootTag['Key'];
```
jhockett, line 61: This function is fundamentally broken. The function always returns true, because forEach is itself a function.

I understand it's outside the scope of the PR, but we need to audit every test that uses this.

jhockett (approved):

edwardfoyle (approved):

packages/amplify-e2e-tests/src/__tests__/tags.test.ts
sobolk: Fortunatelly it's not a big blast radious.
image

PR #13040, opened 2023-07-31
build: upfront checkout

sobolk:
#### Description of changes

This PR adds a branch checkout step in build_linux step. This makes sure that every step in pipeline has some branch "checked out" instead of being in detached state and making lerna fail.

If we detect that we're in PR by examining refs or CB env variable (to support both manual and webhook trigger) we create temporary local branch.
Otherwise we expect that `BRANCH_NAME` variable has been provided upfront when starting workflow.

In action. Makes PR workflow happy.
image

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


.circleci/codebuild-checkout.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash -e
```
0618, line 1: I recently learned that if we add a `-v` or `-x` we should get all the command logged explicitly https://stackoverflow.com/questions/2853803/how-to-echo-shell-commands-as-they-are-executed

.circleci/codebuild-checkout.sh
```diff
@@ -0,0 +1,33 @@
...
+  echo "CODEBUILD_SOURCE_VERSION_REF=$CODEBUILD_SOURCE_VERSION_REF"
+fi
+
+# Codebuild doesn't checkout the branch by default
+if [[ "$CODEBUILD_WEBHOOK_TRIGGER" =~ ^pr/ || "$CODEBUILD_SOURCE_VERSION" =~ ^pr/ || "$CODEBUILD_SOURCE_VERSION_REF" =~ refs/pull/[0-9]+/head$ ]]; then
+  # If we're in PR workflow create temporary local branch.
+  # We detect if we're in PR by looking for pr/<number> pattern in code build env variables
+  # or by checking if commit is matching refs/pull/<number>/head.
+  echo "Creating temporary local branch for PR build"
+  TEMP_BRANCH_NAME=$(cat /proc/sys/kernel/random/uuid)
```
0618, line 22: do we need delete the `TEMP_BRANCH_NAME` somewhere?

0618 (approved):
LGTM

.circleci/codebuild-checkout.sh
sobolk: we do that sometimes when troubleshooting.

.circleci/codebuild-checkout.sh
sobolk: no. it's not pushed to github.

sobolk: I'm using uuid, not arbitrary string in case name is already taken by existing branch.
this branch exists only locally and is then dropped, i.e. not pushed to git.

jhockett (approved):

PR #13066, opened 2023-08-03
chore(cli): add share-project-config-on/off to configure

0618:
#### Description of changes

This PR is a part of the data consent re-work.
This PR 
- enables the user to toggle `shareProjectConfig` by runing `amplify configure share-project-config-on(off)`.
- update `amplify configure -h`
- add feedback once the user opt-in or opt-out
- fix typo `non sensitive` -> `non-sensitive`

|   | Before  | After  |
|---|---|---|
| configure -h  | image  | image  |
| init  | image  |  image image |

image

More upcoming changes for 
[ ] Show telemetry disclaimer (if not opted-out) on any first command which collecting telemetry data
[ ] Clean up old telemetry disclaimer in postinstall
[ ] Update docs

Ticket:
NA

Verified on local `amplify-dev`

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


0618:
CCI e2e: https://app.circleci.com/pipelines/github/aws-amplify/amplify-cli?branch=run-e2e/zhamujun/data-consent-configure

packages/amplify-cli/src/commands/configure.ts
```diff
@@ -15,12 +16,25 @@ export const run = async (context: Context) => {
...
+    context.print.success('Share Project Config has been turned off');
+    return;
+  }
+
+  if (context.parameters.options['share-project-config-on']) {
+    DebugConfig.Instance.setAndWriteShareProject(true);
+    context.print.success('Share Project Config has been turned on');
+    return;
+  }
+
```
0618, line 37: These lines are so spaghetti 😂  any suggestions how we can optimize it?

packages/amplify-cli/src/init-steps/s9-onSuccess.ts
```diff
@@ -65,6 +65,13 @@ export const onSuccess = async (context: $TSContext): Promise<void> => {
 
     await FeatureFlags.ensureDefaultFeatureFlags(true);
     const result = await prompter.yesOrNo('Help improve Amplify CLI by sharing non sensitive configurations on failures', false);
```
josefaidt, line 67:
```suggestion
    const result = await prompter.yesOrNo('Help improve Amplify CLI by sharing non-sensitive project configuration on failures', false);
```

packages/amplify-cli-core/src/help/commands-info.ts
```diff
@@ -59,7 +59,12 @@ export const commandsInfo: Array<CommandInfo> = [
     command: 'configure',
     commandDescription: 'Configure the CLI to work with your AWS profile',
     commandUsage: 'amplify configure [subcommand]',
-    commandFlags: [],
+    commandFlags: [
+      { short: '', long: 'usage-data-on', flagDescription: 'Turn on usage data sharing.' },
+      { short: '', long: 'usage-data-off', flagDescription: 'Turn off usage data sharing.' },
+      { short: '', long: 'share-project-config-on', flagDescription: 'Turn on non-sensitive configurations sharing on failures.' },
```
josefaidt, line 65:
```suggestion
      { short: '', long: 'share-project-config-on', flagDescription: 'Turn on non-sensitive project configuration sharing on failures.' },
```

packages/amplify-cli-core/src/help/commands-info.ts
```diff
@@ -59,7 +59,12 @@ export const commandsInfo: Array<CommandInfo> = [
...
+      { short: '', long: 'share-project-config-off', flagDescription: 'Turn off non-sensitive configurations sharing on failures.' },
```
josefaidt, line 66:
```suggestion
      { short: '', long: 'share-project-config-off', flagDescription: 'Turn off non-sensitive project configuration sharing on failures.' },
```

packages/amplify-cli-core/src/help/commands-info.ts
josefaidt: adding "project configuration" to help bridge the understanding with its flag "share-project-config-on/off"

0618: Cool. I'll do a global search to update it everywhere.

packages/amplify-cli/src/commands/configure.ts
sobolk: at least it's not spaghetti on christmas tree like in other places ...

sobolk (dismissed):
LGTM.
Feel free to merge when you get clean e2e run.

packages/amplify-cli/src/commands/configure.ts
```diff
@@ -15,12 +16,25 @@ export const run = async (context: Context) => {
     context.print.success('Usage Data has been turned off');
     return;
   }
-  if (context.parameters.options['usage-data-on']) {
+
+  if (context.parameters.options['usage-data-on ']) {
```
jhockett, line 20:
```suggestion
  if (context.parameters.options['usage-data-on']) {
```

packages/amplify-cli/src/commands/configure.ts
```diff
@@ -15,12 +16,25 @@ export const run = async (context: Context) => {
...
     write(context, { usageDataConfig: { isUsageTrackingEnabled: true } });
     context.print.success('Usage Data has been turned on');
     return;
   }
 
+  if (context.parameters.options['share-project-config-off']) {
+    DebugConfig.Instance.setAndWriteShareProject(false);
+    context.print.success('Share Project Config has been turned off');
```
jhockett, line 28: context.print operations are deprecated consider using `printer` from the `@aws-amplify/amplify-prompts` package, but not a merge blocker.

0618: Done 3a7220a335dc8aa1e4243982f973137ddd616726

packages/amplify-e2e-core/src/utils/nexpect.ts
```diff
@@ -753,7 +753,7 @@ export function nspawn(command: string | string[], params: string[] = [], option
   // For push operations in E2E we have to explicitly disable the Amplify Console App creation
   // as for the tests that need it, it is already enabled for init, setting the env var here
   // disables the post push check we have in the CLI.
-  if (params.length > 0 && params.find((param: string) => param.toLowerCase() === 'push')) {
+  if (params.length > 0 && params.some((param: string) => param.toLowerCase() === 'push')) {
```
0618, line 756: This change is not related to this PR...just came across it when reading the code

packages/amplify-e2e-core/src/configure/index.ts
```diff
@@ -41,42 +50,48 @@ const authenticationOptions = ['AWS profile', 'AWS access keys'];
...
-
-  return new Promise((resolve, reject) => {
-    const chain = spawn(getCLIPath(), ['configure'], { stripColors: true })
-      .wait('Sign in to your AWS administrator account:')
-      .wait('Press Enter to continue')
-      .sendCarriageReturn()
-      .wait('Specify the AWS Region');
-
-    singleSelect(chain, s.region, amplifyRegions);
+export function amplifyConfigure(cwd: string | null, settings: AmplifyConfiguration | null, flag?: CommandFlags): Promise<void> {
```
sobolk, line 53: If `settings` and `flag` are mutually exclusive (they are by looking at implementation) then please either change api to have union type (better) OR add validation to this function and throw if both are provided (not bad, but longer feedback loop than compiler error).

Yet other option is to split this function into two `amplifyConfigure` as is originally and `amplifyConfigureTelemetryFlags`.


```suggestion
export function amplifyConfigure(cwd: string | null, settings: AmplifyConfiguration | CommandFlags | null): Promise<void> {
```

packages/amplify-e2e-tests/src/__tests__/amplify-configure.test.ts
```diff
@@ -53,6 +73,8 @@ function testAmplifyConfigureValidation(): Promise<void> {
       .wait('You must enter a valid secretAccessKey')
       .sendLine(validMockAWSSecretAccessKey)
       .wait('Profile Name:')
+      // 'config-test' is a new profile name using defaultAWSSecretAccessKey and validMockAWSSecretAccessKey.
+      // Do NOT use TEST_PROFILE_NAME, otherwise, it will override the default profile.
```
sobolk, line 77:
```suggestion
      // Fake credentials used in this test are always invalid therefore
      // 'config-test' is a new profile name using defaultAWSSecretAccessKey and validMockAWSSecretAccessKey
      // to avoid overwriting profiles used in other tests.
      // Do NOT use TEST_PROFILE_NAME here, otherwise, it will override the default profile.
```

sobolk (commented):
LGTM.

packages/amplify-e2e-core/src/configure/index.ts
0618: They are not really "mutually exclusive". Maybe at this moment, but there are several scenarios we may need both
- if we add tests for subcommands without creating a new function (e.g. `amplifyConfigureProject`)
- if we add a new flag with requires an option

So I think It's more scalable to separate them into 2 different variables.

Ideally, it'd be even better if we could use a destracture IMO

sobolk: `settings` are not consumed if `flag` is passed at all. They are mutually exclusive effectively.
Besides scenarios in both logical branches in that if conditions are never going to be combined together. I.e. you won't ever set telemetry flags AND configure profiles.

```
if (flag) {
    return spawn(getCLIPath(), ['configure', `--${flag}`], { cwd, stripColors: true })
      .wait(commandFlagsReturnMessage[flag])
      .runAsync();
  } else {
    const allSettings = { ...defaultSettings, ...settings };
    const missingParam = MANDATORY_PARAMS.filter((p) => !Object.keys(allSettings).includes(p));
```

sobolk: > So I think It's more scalable to separate them into 2 different variables.

Define `scalable`

sobolk: This is classic SRP problem.

0618: I see. That's fair. Done 049cdcd7cec03d1d33f979861284b2884258417b

sobolk (approved):

joebuono (approved):

PR #13076, opened 2023-08-04
test: enforce no-unused-vars rule

sobolk:
#### Description of changes

**This PR:**
1. enables `@typescript-eslint/no-unused-vars` rule in tests
2. eliminates all unused symbols.

**Out of scope.**
This PR does not attempt to fix tests that does not have much sense. Just enable the rule for better future.

Running full E2E suite. Will merge if passes.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/utils/trigger-flow-auth-helper.test.js
+++ b/packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/utils/trigger-flow-auth-helper.test.js
@@ -1,9 +1,7 @@
 jest.mock('@aws-amplify/amplify-cli-core', () => {
   return {
     FeatureFlags: {
-      getBoolean: jest.fn().mockImplementation(() => {
-        return true;
-      }),
+      getBoolean: jest.fn().mockReturnValue(true),
     },
   };
 });
--- a/packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/utils/trigger-flow-auth-helper.test.js
+++ b/packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/utils/trigger-flow-auth-helper.test.js
@@ -1,7 +1,7 @@
 jest.mock('@aws-amplify/amplify-cli-core', () => {
   return {
     FeatureFlags: {
-      getBoolean: jest.fn().mockImplementation((name, defaultValue) => {
+      getBoolean: jest.fn().mockImplementation(() => {
         return true;
       }),
     },
--- a/packages/amplify-e2e-core/src/utils/nexpect.ts
+++ b/packages/amplify-e2e-core/src/utils/nexpect.ts
@@ -160,7 +160,7 @@ function chain(context: Context): ExecutionContext {
 
     wait(
       expectation: string | RegExp,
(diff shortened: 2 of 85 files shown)


joebuono (dismissed):

packages/amplify-category-auth/src/__tests__/provider-utils/awscloudformation/utils/trigger-flow-auth-helper.test.js
```diff
@@ -1,7 +1,7 @@
 jest.mock('@aws-amplify/amplify-cli-core', () => {
   return {
     FeatureFlags: {
-      getBoolean: jest.fn().mockImplementation((name, defaultValue) => {
+      getBoolean: jest.fn().mockImplementation(() => {
```
jhockett, line 4: nit: possibly out of scope, but could be simplified to
```ts
    getBoolean: jest.fn().mockReturnValue(true),
```

packages/amplify-e2e-core/src/utils/nexpect.ts
```diff
@@ -160,7 +160,7 @@ function chain(context: Context): ExecutionContext {
 
     wait(
       expectation: string | RegExp,
-      callback = (data: string) => {
+      callback: (data: string) => void = () => {
```
jhockett, line 163: Does this function parameter not need a default value?

packages/amplify-util-mock/src/__tests__/api/lambda-invoke.test.ts
```diff
@@ -53,6 +53,7 @@ describe('Invoke local lambda function', () => {
     expect(printer.info).toBeCalledWith(JSON.stringify(echoInput, undefined, 2));
     expect(printer.error).toBeCalledTimes(0);
     expect(printer.info).toBeCalledWith('Finished execution.');
+    expect(isBuilt).toBeTruthy();
```
jhockett, line 56: nit: although it doesn't matter in this context, I prefer to use `.toBe(true)` over `.toBeTruthy()` because it's more strict and avoids common errors when checking returned promises

```suggestion
    expect(isBuilt).toBe(true);
```

packages/amplify-e2e-core/src/utils/nexpect.ts
sobolk: this thing here is a default no-op implementation of https://github.com/aws-amplify/amplify-cli/blob/bf63580826133c56c56f86a59ecb918edfde8002/packages/amplify-e2e-core/src/utils/nexpect.ts#L75C20-L75C20

The code is correct. Parameter is ignored in both old and new implementation. But removing explicit parameter require typing `callback` parameter otherwise it's not matching `ExecutionContext` type.

jhockett (approved):

packages/amplify-e2e-core/src/utils/nexpect.ts
sobolk: This is legal in typescript.
```
let foo: (param: string) => void;
foo = () => {};
```

I.e. function must accept params from typing or less (less means it ignores inputs).

joebuono (approved):

PR #13092, opened 2023-08-08
ci: configure codeql to ignore e2e branches

jhockett:
#### Description of changes
Currently, starting e2e tests on a PR cause our CodeQL action to run again. This PR updates the action to ignore e2e branches.

https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/customizing-code-scanning#avoiding-unnecessary-scans-of-pull-requests

See checks below. E2e tests were intentionally cancelled.

#### Checklist

- [x] PR description included

### DIFF
--- a/.github/workflows/codeql-tests.yaml
+++ b/.github/workflows/codeql-tests.yaml
@@ -3,9 +3,7 @@ name: 'CodeQL Tests'
 
 on:
   push:
-    branches-ignore:
-      - run-e2e/**
-      - run-cb-e2e/**
+    branches: [dev]
   pull_request:
 
 jobs:
--- a/.github/workflows/codeql-tests.yaml
+++ b/.github/workflows/codeql-tests.yaml
@@ -3,6 +3,9 @@ name: 'CodeQL Tests'
 
 on:
   push:
+    branches-ignore:
+      - run-e2e/**
+      - run-cb-e2e/**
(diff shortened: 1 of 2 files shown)


sobolk (approved):

.github/workflows/codeql-tests.yaml
```diff
@@ -3,6 +3,9 @@ name: 'CodeQL Tests'
 
 on:
   push:
+    branches-ignore:
```
0618, line 6: Does that mean any other branch runs the tests? I wonder if we should only run the tests on specific branches (e.g. `dev`, `main`)?

jhockett: > Does that mean any other branch runs the tests? I wonder if we should only run the tests on specific branches (e.g. `dev`, `main`)?

Currently these checks run on all branches. We want these to run on arbitrary branches so PR code will be checked, but it's unnecessary to run them for e2e runs.

sobolk (approved):

.github/workflows/codeql-tests.yaml
jhockett: Good feedback, I updated the PR to include these suggestions.

0618 (approved):

PR #13115, opened 2023-08-14
test: isolate uibuilder test

sobolk:
#### Description of changes

uibuilder test calls `pkill -f react` which might affect other tests running in parallel if they start react apps.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


jhockett (approved):
Nice!

codebuild_specs/e2e_workflow_generated.yml
```diff
@@ -825,7 +826,16 @@ batch:
...
           TEST_SUITE: src/__tests__/auth_2e.test.ts
-          CLI_REGION: us-east-1
+          CLI_REGION: ap-southeast-1
+      depend-on:
+        - upb
+    - identifier: l_uibuilder
+      buildspec: codebuild_specs/run_e2e_tests_linux.yml
+      env:
+        variables:
+          compute-type: BUILD_GENERAL1_SMALL
```
awsluja, line 836: if i remember correctly we were running into out-of-memory issues with this test, so if we see it failing we might also need to put this job on a medium VM ; small vm only has 3gb and is fine for most tests; 
https://docs.aws.amazon.com/codebuild/latest/userguide/build-env-ref-compute-types.html

awsluja (approved):

codebuild_specs/e2e_workflow_generated.yml
sobolk: I kicked off a run. If that's the case I'll add option to customize compute per test.

sobolk: @awsluja looks like it runs on medium...

image

It seems that
```
      env:
        variables:
          compute-type: BUILD_GENERAL1_SMALL
```
is invalid yaml....
it should be 
```
      env:
        compute-type: BUILD_GENERAL1_SMALL
```

PR #13121, opened 2023-08-16
ci: reduce active wait on codebuild jobs

sobolk:
#### Description of changes

**Background**

We have introduced `_waitForJobs` mechanism to actively poll on CodeBuild jobs completion. I.e. almost all jobs in batch start eagerly and poll for their dependencies actively.

Why ?
It appears that Code Build executes batch graph in BFS manner. I.e. it waits for all level N jobs completion before starting N+1 level.
Let's say that we used `depend-on` directive instead of active poll. Then, for example, when `build_linux` step completes (level 1) all dependent jobs from level 2 start, including `test` and `publish_to_local_registry`.
Then `publish_to_local_registry` usually completes quickly, let's say 2 minutes. But all level 3 jobs will still wait for `test` to complete which takes ~10 minutes.

The workaround bypasses the problem, all jobs start eagerly and poll.

**The problem**

Polling burns codebuild api request quota and leads to throttling.

**Improvement**

To improve situation and eliminate some polling. We make all level 2+ jobs depend on `build_linux`.
Level 2 jobs won't poll. Level 3+ will keep polling.

Why do we have to depend on `build_linux` for Level 3+ jobs? If we don't they're treated as level 1 (started together with `build_linux`) and block level 2 jobs from starting.

**The exception**

We keep path from `build_linux` to `upb` through `publish_to_local_registry` and steps to build binaries hybrid.
I.e. in PR workflow, release workflows we gate them by `build_linux`.
But we keep polling to speed up path in E2E workflow while `build_windows` is still running.
I.e. to solve this problem:
image

1. PR workflow
2. E2E test workflow
3. Tagged release

image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/shared-scripts.sh
+++ b/shared-scripts.sh
@@ -502,6 +502,8 @@ function _amplifyConsoleIntegrationTests {
     retry yarn console-integration --no-cache --maxWorkers=4 --forceExit
 }
 function _integrationTest {
+    npm install -g ts-node
+
     echo "Restoring Cache"
     loadCache repo $CODEBUILD_SRC_DIR
 
--- a/codebuild_specs/base.yml
+++ b/codebuild_specs/base.yml
@@ -49,6 +49,8 @@ batch:
         - build_linux
     - identifier: verify_api_extract
       buildspec: codebuild_specs/verify_api_extract.yml
+      depend-on:
+        - build_linux
       env:
         compute-type: BUILD_GENERAL1_LARGE
       depend-on:
--- a/codebuild_specs/base.yml
+++ b/codebuild_specs/base.yml
@@ -49,8 +49,6 @@ batch:
         - build_linux
     - identifier: verify_api_extract
       buildspec: codebuild_specs/verify_api_extract.yml
-      depend-on:
-        - build_linux
       env:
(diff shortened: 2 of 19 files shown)


sobolk:
A new sequencing seems to be working in e2e pipe.
The build_linux -> ... -> upb -> tests speed is retained. Some leaf jobs start later, but that doesn't matter in big picture.
image

jhockett (approved):
LGTM with passing e2e

shared-scripts.sh
```diff
@@ -502,6 +502,8 @@ function _amplifyConsoleIntegrationTests {
     retry yarn console-integration --no-cache --maxWorkers=4 --forceExit
 }
 function _integrationTest {
+    npm install -g ts-node
```
awsluja, line 505: why this line

sobolk: Because `integration_test` started to fail after removing active wait there.

Active wait was installing that here https://github.com/aws-amplify/amplify-cli/blob/0fde28a08a3d82b371cbbf858f462648376a318a/shared-scripts.sh#L649-L650 .

image

awsluja (approved):
lgtm

PR #12596, opened 2023-05-06
fix(amplify-frontend-ios): encapsulate path option values with quotes

josefaidt:
#### Description of changes

adds quotes to `--path` option values to prevent breaking when absolute path to amplify project contains spaces

#### Issue #, if available

fixes #12593

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies
- [ ] [Pull request labels](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#labels) are added

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #13151, opened 2023-08-24
fix: allow whitespace characters in Xcode projects

sobolk:
#### Description of changes

This PR:
- Re-attempts fix from https://github.com/aws-amplify/amplify-cli/pull/12596 .
- Modifies "Xcode binding generating code", see https://github.com/aws-amplify/amplify-cli/blob/6c43a022754a196f549724336405e44bb2606c7a/packages/amplify-frontend-ios/package.json#L22
- Changes `execa` call to pass arguments as list. `execa` automatically escapes whitespace characters
- Adds sample xcode projects and integration test. These don't require creds or requests, just files on disk for the test.
- adds test sources to npm ignored (this was missed apparently, due to different dir convention in that package).

Added integration tests for `amplify-xcode` binary. These won't run in pipelines until we provide macOS compute.

Reproduced original whitespace issue before changing code and then.

Manually tested with both Xcode project with whitespace in path and without.

image

image

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-frontend-ios/tests/native-bindings-codegen.test.js
```diff
@@ -7,7 +7,7 @@ describe('amplify-xcode native bindings codegen', () => {
         name: 'option-name',
         kind: 'option',
       };
-      const expected = [[`if (params['${param.name}']) {`, `    command += \` --${param.name}=\${params['${param.name}']}\`;`, `  }`]];
+      const expected = [[`if (params['${param.name}']) {`, `    args.push(\`--${param.name}=\${params['${param.name}']}\`);`, `  }`]];
```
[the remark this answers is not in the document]
sobolk: There is. But it'll only work on macs. So the test is going to be skipped until we deliver proper compute.

Amplifiyer (approved):

packages/amplify-frontend-ios/tests/native-bindings-codegen.test.js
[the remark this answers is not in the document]
sobolk: Added test. It'll work on dev machines for now and later when we add macos.

packages/amplify-frontend-ios/tests/xcode.integration.test.js
```diff
@@ -0,0 +1,48 @@
...
+    beforeAll(async () => {
+      const source = path.resolve(__dirname, 'sample-xcode-project');
+      pathsToTest.push(source);
+      const destination = path.resolve(__dirname, 'sample xcode project');
+      pathsToTest.push(destination);
+      if (!existsSync(destination)) {
+        await fs.mkdir(destination);
+        await fs.cp(source, destination, {
+          recursive: true,
+        });
```
Amplifiyer, line 28: Do we not need to delete these files?

sobolk: we could.

sobolk: @Amplifiyer done.

Amplifiyer (approved):

packages/amplify-frontend-ios/native-bindings-codegen/templates/preamble.jst
```diff
@@ -22,5 +22,5 @@ const path = require('path');
 const { pathManager } = require('@aws-amplify/amplify-cli-core');
 
 const BINARY_PATH = 'resources/amplify-xcode';
-const PACKAGE_NAME = 'amplify-frontend-ios';
+const PACKAGE_NAME = '@aws-amplify/amplify-frontend-ios';
```
awsluja, line 25: why this change ? was it missed in the @ aws-amplify refactor?

awsluja (approved):

packages/amplify-frontend-ios/native-bindings-codegen/templates/preamble.jst
sobolk: No. it was hand-edited here https://github.com/aws-amplify/amplify-cli/blob/6c43a022754a196f549724336405e44bb2606c7a/packages/amplify-frontend-ios/lib/amplify-xcode.js#L25 when we added prefix.
But we didn't realize this code is generated, so this change fixes the generator.

PR #13152, opened 2023-08-24
ci: add build and unit test on macOS

sobolk:
#### Description of changes

This PR:
- Adds a github action that builds and runs tests on macOS

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/.github/workflows/build-test-mac.yml
+++ b/.github/workflows/build-test-mac.yml
@@ -1,0 +1,24 @@
+name: Build and Test on Mac
+
+on:
+  push:
+    branches: [dev]
+  pull_request:
+
+jobs:
+  build-and-test:
+    runs-on: macos-13
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+        with:
+          persist-credentials: false
+
+      - name: Install Dependencies
+        run: yarn install
+
+      - name: Build
+        run: yarn build
+
+      - name: Test
+        run: yarn test


packages/amplify-cli/src/domain/amplify-usageData/SerializableError.ts
```diff
@@ -70,7 +70,13 @@ const processPaths = (paths: string[]): string[] => {
...
   return result.map((r) => {
     if (r === longestString) {
-      return longestString.replace(path.join(...directoriesRemoved), '');
+      for (const directory of directoriesRemoved) {
+        if (r.startsWith(`/${directory}`)) {
+          r = r.replace(`/${directory}`, '');
+        } else {
+          break;
+        }
+      }
```
sobolk, line 79: Fixes 
image

packages/amplify-cli/src/domain/amplify-usageData/SerializableError.ts
```diff
@@ -70,7 +69,13 @@ const processPaths = (paths: string[]): string[] => {
 
   return result.map((r) => {
     if (r === longestString) {
-      return longestString.replace(path.join(...directoriesRemoved), '');
```
0618, line 73: The problem here seems to be `path.join(...directoriesRemoved)` missing a `/` in front...., so I guess it might work if we do 

```
return longestString.replace(`/${path.join(...directories removed)}`, '');
```

or

```
return longestString.replace(`/${directoriesRemoved.join("/")`, '');
```

what do you think?

sobolk: image

It's not that problem. But I think I know what's going on.
We're using `new Set` to store tokens, it may change order. Let me try fixing that.

sobolk: It also did not account for duplicates i.e. `/some/path/with/duplicate/duplicate/foo`

0618 (dismissed):

Amplifiyer:
Looks good, but I think the test from previous PR is failing on the new macos fleet. `fileNotFound: awsconfiguration.json not found.`

sobolk:
> Looks good, but I think the test from previous PR is failing on the new macos fleet. `fileNotFound: awsconfiguration.json not found.`

Yeah. I was expecting some bumps there. Will fix.

0618 (approved):

Amplifiyer (approved):

PR #13166, opened 2023-08-29
fix: add better error messaging when initializing an already initialized project

### LINKED ISSUE #11579: receiving `Invalid environment name: undefined` when running `amplify init -y` in an initialized project

pnpm

18

### Amplify CLI Version

10.5.1

mac

n/a

### Describe the bug

When running `amplify init -y` in an initialized project we receive `Invalid environment name: undefined`

### Expected behavior

We should receive a more helpful message such as

```
Amplify project ${appId} is already initialized for environment ${envName}. To create a new environment run `amplify add env`
```

### Reproduction steps

1. `amplify init -y`
2. `amplify init -y`

### Project Identifier

_No response_

### Log output

<details>

```


```

</details>

### Additional information

_No response_

- [X] I have removed any sensitive information from my code snippets and submission.

rtpascual:
#### Description of changes

Catch earlier and provide better error messaging when initializing an already initialized project with defaults (running `amplify init -y` for an already initialized project).

The error message changed when running `amplify init -y` the second time:
`Invalid environment name: undefined` -> `Amplify project <appId> is already initialized with environment <currentEnv>`

#### Issue #, if available

Fixes #11579

Added unit test for running `amplify init -y` for an initialized project and expecting the new error message.

Manually ran `amplify-dev init -y` twice, the second time gives the new error message (the same applies if initializing an app with a custom env name then running `amplify-dev init -y`):

Screenshot 2023-09-07 at 6 06 07 PM

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/__tests__/init-steps/s0-analyzeProject.test.ts
```diff
@@ -75,4 +76,35 @@ describe('analyzeProject', () => {
...
+    };
+    mockContext.exeInfo.inputParams = {
+      amplify: {},
+      yes: true,
+    };
+    mockContext.usageData = {
+      emitError: emitErrorMock,
+    };
+
+    jest.spyOn(fs, 'existsSync').mockReturnValue(true);
```
rtpascual, line 101: Mocked the bare minimum to replicate the situation where `amplify-dev init -y` was run the first time.

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -220,8 +220,11 @@ const getDefaultEnv = (context: $TSContext): string | undefined => {
       resolution: INVALID_ENV_NAME_MSG,
     });
   }
-
-  if (isNewProject(context) || !context.amplify.getAllEnvs().includes(defaultEnv)) {
+  if (
+    isNewProject(context) ||
+    !context.amplify.getAllEnvs().includes(defaultEnv) ||
+    (context.exeInfo.inputParams && context.exeInfo.inputParams.yes)
```
rtpascual, line 226: This condition was added so the error message would return the default environment name of `dev` instead of `undefined`

bombguy (approved):

packages/amplify-cli/src/init-steps/s0-analyzeProject.ts
```diff
@@ -241,8 +244,8 @@ const getEnvName = async (context: $TSContext): Promise<string> => {
     });
   } else if (context.exeInfo.inputParams && context.exeInfo.inputParams.yes) {
     throw new AmplifyError('ProjectInitError', {
-      message: `Invalid environment name: ${context.exeInfo.inputParams.amplify.envName}`,
-      resolution: INVALID_ENV_NAME_MSG,
+      message: `Amplify project is already initialized with environment name: ${getDefaultEnv(context)}`,
+      resolution: `To create a new environment run \`amplify add env\``,
     });
   }
```
sobolk, line 250: Couple of things:

I'm struggling to understand how do we end up in this branch if we run `amplify init -y` in already initialized project.
And therefore I'm not sure if just changing this error is safe. I.e. I'm not sure if we're not missing some other case where new error doesn't make sense. Running full e2e suite may reveal where problem could be (but it's not guaranteed to find anything). Use `yarn cloud-e2e` to trigger tests.
The caller of this code `analyzeProject` seem to set `isNewProject` flag. We could potentially use it to make sure that we throw the new error in already existing project.


Another thing is. If we want to say that `Amplify project is already initialized with environment name` we shouldn't use "default env". But rather read what env is in project files. Look for `stateManager.getCurrentEnvName` to get this value.

Apart from unit testing could you check if there's any e2e test that covers this scenario (i.e. init -y in already initialized project?). If there isn't it might be worth adding one.

rtpascual: Running `yarn cloud-e2e` didn't find anything but makes sense that we might be missing another case for this error so having the error separate and earlier would be best.

Good catch on "default env", in the case when running `amplify init` with a different environment name then `amplify init -y` we would want to get the current environment. This would also work in a multi-environment app as well.

I'll look into existing e2e tests and add one if there aren't any existing test for this scenario.

packages/amplify-e2e-tests/src/__tests__/init-special-case.test.ts
```diff
@@ -72,4 +74,12 @@ describe('amplify init', () => {
       getParameters(projectRoot, category, authResourceName);
     }).not.toThrow();
   });
+
+  it('should fail if running amplify init -y on already initialized project', async () => {
```
sobolk, line 78: Please work with @bombguy and @0618 to get this included in codebuild manifests and tested in e2e test run.

sobolk (commented):
LGTM.

packages/amplify-e2e-tests/src/__tests__/init-special-case.test.ts
sobolk: Well. Skip the codebuild manifest part, since it's addition to already existing test file. But please obtain e2e run where this test passes.

packages/amplify-e2e-tests/src/__tests__/init-special-case.test.ts
```diff
@@ -72,4 +73,11 @@ describe('amplify init', () => {
...
+    const envName = 'devtest';
+    const resourceName = 'twoInitDefaultTest';
+    await initJSProjectWithProfile(projectRoot, { disableAmplifyAppCreation: false, name: resourceName, envName });
+    await expect(initNonInteractive(projectRoot)).rejects.toThrowError('Process exited with non zero exit code 1');
```
bombguy, line 81: Can you not just run `initJSProjectWithProfile` twice and expect it to throw?

packages/amplify-e2e-core/src/init/initProjectHelper.ts
```diff
@@ -516,3 +516,7 @@ export function amplifyStatus(cwd: string, expectedStatus: string, testingWithLa
 export function initHeadless(cwd: string, envName: string, appId: string): Promise<void> {
   return spawn(getCLIPath(), ['init', '--yes', '--envName', envName, '--appId', appId], { cwd, stripColors: true }).runAsync();
 }
+
+export function initNonInteractive(cwd: string): Promise<void> {
```
bombguy, line 520: can we just re-use the helper function ^? You should be able to just make the 2nd & 3rd parameter optional.

bombguy (dismissed):

packages/amplify-e2e-core/src/init/initProjectHelper.ts
sobolk: this is good feedback. let's do this.

packages/amplify-e2e-tests/src/__tests__/init-special-case.test.ts
sobolk: `initJSProjectWithProfile` is interactive.

rtpascual: The `Invalid environment name: undefined` error only occurs for non-interactive init with `amplify init -y` where envName isn't passed in as an option or through the input prompt

bombguy: pretty weird edge case. But thanks for the clarification

packages/amplify-e2e-core/src/init/initProjectHelper.ts
rtpascual: addressed this in the latest commit

bombguy (approved):
LGTM

Issue #13001, opened 2023-07-25
'amplify add function' seems to be broken


ginobean:
### How did you install the Amplify CLI?

npm

### If applicable, what version of Node.js are you using?

16.19.1

### Amplify CLI Version

12.2.0

### What operating system are you using?

Ubuntu

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

I used 'amplify add function' to add  a lambda, with the default 'Hello World' js option.

Then, I tried to run 'amplify function build', and got this error:

```
Packaging lambda function failed with the error 
Command failed with exit code 1: yarn install
Usage Error: The nearest package directory (/home/gino/local/amplify-app/amplify/backend/function/foobar/src) doesn't seem to be part of the project declared in /home/gino/local/amplify-app.

```

BTW, as a side note, in trying to resolve this issue, I noticed that the package.json, for the lambda function, was in the src subdirectory for the function. Out of curiosity, I tried moving package.json to the parent of src directory, e.g. 'mv package.json ..' . And 'amplify add function' no longer errored out. Not sure if this is the correct resolution.. 

I believe 'amplify push' also yields the exact same error message as 'amplify add function'. And moving the package.json, for the lambda function, from the src subdir to the parent of the src directory, 'mv package.json ..' also seemed to enable 'amplify push' to work correctly.

### Expected behavior

Expection is that the function would be built correctly.

### Reproduction steps

See the bug description for the reproducible steps.

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Before submitting, please confirm:

- [X] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [X] I have removed any sensitive information from my code snippets and submission.

josefaidt:
Hey @ginobean :wave: thanks for raising this! Out of curiosity what version of yarn are you using? Is there a workspaces setting that is picking up this directory?

ginobean:
yarn 3.5.0
Could not find any workspaces setting that is picking up the lambda related directory.

josefaidt:
Hey @ginobean thanks for clarifying! I'm not able to reproduce with the latest yarn berry release and `amplify add function`. Would you mind sharing the full command execution output? I'm curious if there is some setting or prompt that is triggering this

josefaidt:
Hey @ginobean I was able to reproduce this by running `amplify build`

```
➜  amplify build function
[...]

PR #13197, opened 2023-09-07
fix: yarn modern install workspace

0618:
#### Description of changes

**Issue**: Modern Yarn (2,3) doesn't create yarn.lock when running `yarn install` if it can find a yarn.lock in it's parent directory. So we can see the folloing error:

https://github.com/aws-amplify/amplify-cli/issues/13001

image

**Fix**: Add yarn.lock to category resources directories if there's no yarn.lock

image

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


Amplifiyer:
Can't we use this new https://yarnpkg.com/configuration/yarnrc#nmHoistingLimits i.e. creating `.yarnrc.yml` for yarn3 workspaces instead?

Also we should have automated tests for this change.

0618:
> Can't we use this new https://yarnpkg.com/configuration/yarnrc#nmHoistingLimits i.e. creating `.yarnrc.yml` for yarn3 workspaces instead?
> 
> Also we should have automated tests for this change.

That's a good point. Thanks for pointing it out. 

However, it only works if we modify the .yarnrc.yml in the root directory. It doesn't work if we add  .yarnrc.yml in the resource' directory :(

sobolk (commented):
We should add tests.

shared-scripts.sh
```diff
@@ -643,6 +640,19 @@ function _buildTestsStandalone {
     yarn build-tests
 }
 
+function _buildTestsStandaloneYarn3 {
```
0618, line 643: This is the only change of this file. Others are from prettier 😂

packages/amplify-cli-core/src/utils/packageManager.ts
```diff
@@ -43,9 +44,31 @@ class YarnPackageManager implements PackageManager {
...
+      if (fs.existsSync(`${resourceDir}/${this.lockFile}`)) {
+        console.log(`${resourceDir}/${this.lockFile} already exists.`);
+      } else {
+        fs.writeFile(`${resourceDir}/${this.lockFile}`, '', (err) => {
+          if (err) {
+            console.error(`Error creating ${resourceDir}/${this.lockFile}`, err);
+          } else {
+            console.log(`${resourceDir}/${this.lockFile} created successfully.`);
+          }
+        });
```
Amplifiyer, line 65: Shouldn't we be using sync methods or promisified ones instead of callbacks? `execa.sync` after will execute right away whether file was written or not.

codebuild_specs/build_tests_standalone-yarn-3.yml
```diff
@@ -0,0 +1,11 @@
+version: 0.2
+env:
+  shell: bash
+phases:
+  build:
+    commands:
+      - echo "Build Tests Standalone"
+      - source ./shared-scripts.sh && _buildTestsStandaloneYarn3
```
Amplifiyer, line 8: Can you add a comment on how this is testing the issue fixed in this PR?

0618: I think the name of the test is self-explanatory because the PR fix the issue for yarn3 and the test just verify if yarn3 works.

Amplifiyer: my confusion stems from the fact that the test is asserting that we can build our own tests using yarn3(not even running them) while the fix seems to be for customers to be able to build their functions using yarn3. So I might be missing something here on how these two are related.

packages/amplify-cli-core/src/utils/packageManager.ts
0618: Done! 1df95f1d4136e841a07181ab13e2b454421c4030

codebuild_specs/build_tests_standalone-yarn-3.yml
0618: Good point. I re-did the test de785772eb2e6da7c2554153e46b0ac43d30a328

Amplifiyer (dismissed):

sobolk (dismissed):

0618:
failed e2e tests are: l_diagnose_hooks_a_mock_api and l_function_3a_init_special_case_http_migration, which are the same as the `dev` branch. They are not from this PR.

sobolk (approved):

Amplifiyer (approved):

codebuild_specs/e2e_workflow_generated.yml
```diff
@@ -530,210 +530,202 @@ batch:
           TEST_SUITE: src/__tests__/auth_5g.test.ts|src/__tests__/auth/admin-api.test.ts|src/__tests__/auth/hosted-ui.test.ts
       depend-on:
         - upb
-    - identifier: l_user_groups_s3_access_user_groups_build_function
```
0618, line 533: ran `yarn split-e2e-tests-codebuild`

Issue #12575, opened 2023-05-03
Populate custom resources CDK stacks with region and account


mattiarossi:
### Amplify CLI Version

11.1.1

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

When using a custom CDK deployment and trying to reference existing infrastructure, it is expected that the stack refrences a default account and region.
Usually it is possible to setup these values using the documented env variables CDK_DEFAULT_REGION/CDK_DEFAULT_ACCOUNT or AWS_DEFAULT_REGION/AWS_DEFAULT_ACCOUNT
but amplify does not honor these.

A possible solution was suggested in #9360 and a pr provided and an attempt to merge it was done in the v7.6.12 release but it has been reverted and this functionality is still missing/behaviour has reverted to functionality prior to v7.6.12

### Expected behavior

being able to create a reference object to an existing resource like :
```
import * as cdk from '@aws-cdk/core';
import * as AmplifyHelpers from '@aws-amplify/cli-extensibility-helper';
import { AmplifyDependentResourcesAttributes } from '../../types/amplify-dependent-resources-ref';
import * as ec2 from '@aws-cdk/aws-ec2';

export class cdkStack extends cdk.Stack {
  constructor(scope: cdk.Construct, id: string, props?: cdk.StackProps, amplifyResourceProps?: AmplifyHelpers.AmplifyResourceProps) {
    super(scope, id, props)
    /* Do not remove - Amplify CLI automatically injects the current deployment environment in this input parameter */
    new cdk.CfnParameter(this, 'env', {
      type: 'String',
      description: 'Current Amplify CLI env name',
    });
    /* AWS CDK code goes here - learn more: https://docs.aws.amazon.com/cdk/latest/guide/home.html */
    
    const vpc = ec2.Vpc.fromLookup(this, "VPC", {
      isDefault: true,
    });

    // creation of resources inside the VPC
  }
}
```

and not getting the 'missing env' error:

```
⠙ Building custom resources🛑 There was an error building the custom resources
🛑 Error: Cannot retrieve value from context provider vpc-provider since account/region are not specified at the stack level. Configure "env" with an account and region when you define your stack.See https://docs.aws.amazon.com/cdk/latest/guide/environments.html for more details.
[...]

PR #13316, opened 2023-10-09
fix: populate custom resources CDK stacks with region and account (#12575)

bombguy:
#### Description of changes
Custom resource should populate with region and account from CDK default variables.

https://docs.aws.amazon.com/cdk/v2/guide/environments.html

#### Issue #, if available
[#12575](https://github.com/aws-amplify/amplify-cli/issues/12575)

Manual testing, unit test & e2e tests added

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)

### DIFF
--- a/packages/amplify-category-custom/src/utils/generate-cfn-from-cdk.ts
+++ b/packages/amplify-category-custom/src/utils/generate-cfn-from-cdk.ts
@@ -2,11 +2,24 @@ import * as cdk from 'aws-cdk-lib';
 import { JSONUtilities, pathManager } from '@aws-amplify/amplify-cli-core';
 import * as path from 'path';
 import { categoryName } from './constants';
+import { StackProps } from '@aws-cdk/core';
 
 export type AmplifyResourceProps = {
   category: string;
   resourceName: string;
 };
+
+export function getCDKProps(): StackProps {
+  if (process.env.CDK_DEFAULT_ACCOUNT || process.env.CDK_DEFAULT_REGION) {
+    return {
+      env: {
+        account: process.env.CDK_DEFAULT_ACCOUNT,
+        region: process.env.CDK_DEFAULT_REGION,
+      },
+    };
+  }
+  return {};
+}
 
 export async function generateCloudFormationFromCDK(resourceName: string) {
   const targetDir = pathManager.getResourceDirectoryPath(undefined, categoryName, resourceName);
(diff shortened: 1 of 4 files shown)


packages/amplify-category-custom/src/utils/generate-cfn-from-cdk.ts
```diff
@@ -2,19 +2,32 @@ import * as cdk from 'aws-cdk-lib';
...
 import { categoryName } from './constants';
+import { StackProps } from 'aws-cdk-lib';
 
 export type AmplifyResourceProps = {
   category: string;
   resourceName: string;
 };
 
+export function getCDKProps(): StackProps {
+  if (process.env.CDK_DEFAULT_ACCOUNT || process.env.CDK_DEFAULT_REGION) {
```
rtpascual, line 13: What is the expected behavior if only one of the envs was set (account but no region and region but no account)?

bombguy: Good question! This environment object is passed to the cdk to build the stack.

According to the cdk docs, if either of those values are not defined, the cdk will substitute it with special strings
https://docs.aws.amazon.com/cdk/v2/guide/environments.html

packages/amplify-category-custom/src/utils/generate-cfn-from-cdk.ts
```diff
@@ -2,19 +2,32 @@ import * as cdk from 'aws-cdk-lib';
...
+    return {
```
0618, line 14: nit: should we save some key-stroles by using destructure?

0618 (approved):

rtpascual (approved):

packages/amplify-e2e-tests/src/__tests__/custom_resources.test.ts
```diff
@@ -130,4 +131,23 @@ describe('adding custom resources test', () => {
...
+
+    fs.copyFileSync(srcCustomResourceFilePath, destCustomResourceFilePath);
+
+    // no aws account & region set, should throw error,
+    await expect(amplifyPushAuth(projRoot)).rejects.toThrowError();
+
+    // should pickup to default cdk env vars and build cfn stack.
+    process.env.CDK_DEFAULT_ACCOUNT = 'some_id';
+    process.env.CDK_DEFAULT_REGION = process.env.CLI_REGION;
+    await amplifyPushAuth(projRoot);
```
sobolk, line 151: this lacks assertion.
we have developed the following mechanism in override tests to assert that injected values actually work
https://github.com/aws-amplify/amplify-cli/blob/d60e505391b4f8fc58c2cc4d352698a6157f1b59/packages/amplify-e2e-tests/overrides/override-auth.ts#L15-L21
Can we please use similar pattern in `barebone-cdk-stack.ts` ?

PR #13510, opened 2023-12-28
fix: s3 bucket name randomizer

0618:
#### Description of changes

**Issue**: canary run `amplify init ....` sometimes have "The specified bucket does not exist" error.

**Hypothesis**: the S3 bucket name is not random enough, so the bucket could be accidentally deleted by another task.

**Fix**: Make the bucket name more random

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/amplify-service-manager.ts
+++ b/packages/amplify-provider-awscloudformation/src/amplify-service-manager.ts
@@ -17,7 +17,7 @@ export async function init(amplifyServiceParams) {
 
   let amplifyAppId;
   let verifiedStackName = stackName;
-  let deploymentBucketName = `${stackName}-deployment`;
+  let deploymentBucketName = `${stackName}-deploy`;
 
   const amplifyClient = await getConfiguredAmplifyClient(context, awsConfigInfo);
   if (!amplifyClient) {
--- a/packages/amplify-provider-awscloudformation/src/amplify-service-manager.ts
+++ b/packages/amplify-provider-awscloudformation/src/amplify-service-manager.ts
@@ -17,7 +17,7 @@ export async function init(amplifyServiceParams) {
 
   let amplifyAppId;
   let verifiedStackName = stackName;
-  let deploymentBucketName = `${stackName}-deploy`;
+  let deploymentBucketName = `${stackName}-deployment`;
 
   const amplifyClient = await getConfiguredAmplifyClient(context, awsConfigInfo);
   if (!amplifyClient) {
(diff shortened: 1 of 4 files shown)


packages/amplify-provider-awscloudformation/src/amplify-service-manager.ts
```diff
@@ -17,7 +17,7 @@ export async function init(amplifyServiceParams) {
 
   let amplifyAppId;
   let verifiedStackName = stackName;
-  let deploymentBucketName = `${stackName}-deployment`;
+  let deploymentBucketName = `${stackName}-deploy`;
```
sobolk, line 20: I'm not sure this is safe. I.e. there might be plenty of logic that rely on deployment bucket name, and studio might be reading from it too.

packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -56,8 +55,7 @@ export const run = async (context: $TSContext): Promise<void> => {
     context.exeInfo ??= { inputParams: {}, localEnvInfo: {} as unknown as LocalEnvInfo };
     const { projectName } = context.exeInfo.projectConfig;
     const initTemplateFilePath = path.join(__dirname, '..', 'resources', 'rootStackTemplate.json');
-    /* eslint-disable-next-line spellcheck/spell-checker */
-    const timeStamp = process.env.CIRCLECI ? uuid().substring(0, 5) : `${moment().format('Hmmss')}`;
+    const timeStamp = uuid().substring(0, 6);
```
sobolk, line 58: which command uses this code ?
one potential risk here is that stack name will be one character longer and some existing projects may start hitting stack name length validation for emepheral environments.

0618: that's for `amplify init` command.

I checked the S3 bucket name limit is 63 charactors, so I add 1 character here and removed 4 charactors from `development` 😂

packages/amplify-provider-awscloudformation/src/amplify-service-manager.ts
0618: 🤯 

I wish there're some test to confirm that. If not, we'll have to keep `deployment` and think about other ways to keep the bucket name short

sobolk: Unfortunately this might require digging to confirm :/

packages/amplify-provider-awscloudformation/src/initializer.ts
sobolk: I think we should just `const timeStamp = uuid().substring(0, 5);` . UUID should be better than timestamp.

sobolk (approved):

packages/amplify-provider-awscloudformation/src/initializer.ts
0618: Done

packages/amplify-provider-awscloudformation/src/amplify-service-manager.ts
0618: revert

rtpascual (approved):

PR #13512, opened 2024-01-02
fix: limit query parameter parsing in auth admin api

sobolk:
#### Description of changes

This PR fixes an error where `limit` in Auth Admin API templates would be passed incorrectly to Cognito SDK as `string` whereas SDK expects `number`.

I.e. this error:

```
SerializationException: STRING_VALUE can not be converted to an Integer
    at throwDefaultError (/var/runtime/node_modules/@aws-sdk/smithy-client/dist-cjs/default-error-handler.js:8:22)
    at deserializeAws_json1_1ListUsersInGroupCommandError (/var/runtime/node_modules/@aws-sdk/client-cognito-identity-provider/dist-cjs/protocols/Aws_json1_1.js:5121:51)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /var/runtime/node_modules/@aws-sdk/middleware-serde/dist-cjs/deserializerMiddleware.js:7:24
    at async /var/runtime/node_modules/@aws-sdk/middleware-signing/dist-cjs/middleware.js:13:20
    at async StandardRetryStrategy.retry (/var/runtime/node_modules/@aws-sdk/middleware-retry/dist-cjs/StandardRetryStrategy.js:51:46)
    at async /var/runtime/node_modules/@aws-sdk/middleware-logger/dist-cjs/loggerMiddleware.js:6:22
    at async listUsersInGroup (/var/task/cognitoActions.js:231:20)
    at async /var/task/app.js:229:18 {
```

Added e2e tests that are calling Auth Admin API

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


edwardfoyle (approved):

codebuild_specs/e2e_workflow_generated.yml
```diff
@@ -145,7 +136,7 @@ batch:
       env:
         variables:
           TEST_SUITE: src/__tests__/diagnose.test.ts|src/__tests__/hooks-a.test.ts|src/__tests__/mock-api.test.ts
-          CLI_REGION: us-west-2
+          CLI_REGION: us-east-2
```
0618, line 139: why changing region?

codebuild_specs/e2e_workflow_base.yml
```diff
@@ -7,7 +7,6 @@ env:
     AWS_REGION: us-east-1
     CDK_DEFAULT_REGION: us-east-1
     CLI_REGION: us-east-1
-    TEST_SUITE: src/__tests__/auth_2b.test.ts
```
sobolk, line 10: Changes in this file are to find some space as code build started to throw:
```
[Container] 2024/01/02 23:22:43.265930 Phase complete: DOWNLOAD_SOURCE State: FAILED
--
104 | [Container] 2024/01/02 23:22:43.265948 Phase context status code: YAML_FILE_ERROR Message: batch yaml definition exceeds size limit
```

Changes made:
- removed global line with env var (this is not used as every job defines it).
- removed compute type medium from jobs as global defaults define it above.

packages/amplify-e2e-core/src/utils/auth-utils.ts
```diff
@@ -42,12 +42,8 @@ export async function setupUser(
   }
 }
 
-export async function addUserToGroup(
-  cognitoClient: CognitoIdentityServiceProvider,
-  userPoolId: string,
-  username: string,
-  groupName?: string,
-): Promise<void> {
+export async function addUserToGroup(userPoolId: string, username: string, groupName: string, region?: string): Promise<void> {
```
0618, line 45: Where do we use `addUserToGroup` in the repo? I somehow couldn't find it 😂

sobolk: In new test. I just changed signature to be easier to use since nothing else uses it.

codebuild_specs/e2e_workflow_generated.yml
sobolk: That's default behavior of `yarn split-e2e-tests-codebuild` . (which we sometimes revert manually, but in this case it doesn't make much sense since tests got reshuffled a bit more than usually).

sobolk: I.e. this place https://github.com/aws-amplify/amplify-cli/blob/9034f865172fbc754a37c00783490ea972e02410/scripts/split-e2e-tests-codebuild.ts#L157

codebuild_specs/wait_for_ids.json
```diff
@@ -1,22 +1,20 @@
 [
-  "l_S3server_api_8_function_8",
   "l_amplify_app",
```
0618, line 2: what are the changes in this file?

0618: It's generated from https://github.com/aws-amplify/amplify-cli/blob/9034f865172fbc754a37c00783490ea972e02410/scripts/split-e2e-tests-codebuild.ts#L367-L368

Thanks @sobolk for pointing that out.

packages/amplify-category-auth/resources/adminAuth/admin-auth-app.js
```diff
@@ -178,14 +175,11 @@ app.get('/listUsers', async (req, res, next) => {
...
-    if (req.query.token) {
-      response = await listGroups(req.query.limit || 25, req.query.token);
-    } else if (req.query.limit) {
-      response = await listGroups((Limit = req.query.limit));
-    } else {
-      response = await listGroups();
+    let limit = 25;
+    if (req.query.limit) {
+      limit = parseInt(req.query.limit);
     }
```
0618, line 181: nit: Would is be _slightly_ easier to read?
```suggestion
      const limit = req.query.limit ? parseInt(req.query.limit) : 25;
```

 (and the rest)

packages/amplify-e2e-tests/src/__tests__/auth/admin-api.test.ts
```diff
@@ -0,0 +1,177 @@
...
+    const userPoolId = getUserPoolId(projRoot);
+    await configureAmplify(projRoot);
+
+    for (let i = 0; i < 5; i++) {
+      const username = `testUser${i}`;
+      const password = `Password12#${i}`;
+      await setupUser(userPoolId, username, password, 'Users', adminLambdaRegion);
+    }
+
+    for (let i = 0; i < 5; i++) {
```
0618, line 57: Why don't we only use one loop to set both user and admin?

packages/amplify-e2e-tests/src/__tests__/auth/admin-api.test.ts
```diff
@@ -0,0 +1,177 @@
...
+        limit: '1',
+        username: 'testAdminUser1',
+      },
+    };
+    const response = await callAdminLambda(request);
+    expect(response.statusCode).toBe(200);
+    const responseBody = JSON.parse(response.body);
+    expect(responseBody.Groups.length).toBe(1);
+  });
+});
```
0618, line 177: Do we need to test a failure case? e.g. the error in the description

0618 (approved):
LGTM. Thanks for adding the test. 

Just left some non-blocking nits and questions.

packages/amplify-e2e-tests/src/__tests__/auth/admin-api.test.ts
sobolk: Before applying fix (changes in `packages/amplify-category-auth/resources/adminAuth/admin-auth-app.js` file) the tests that explicitly specify `limit` would fail with 500 error. (which I locally verified before applying the fix).
After fixing the implementation (adding `parseInt`) they're now passing.

packages/amplify-e2e-tests/src/__tests__/auth/admin-api.test.ts
sobolk: it might have spared a line or two but it would obscure the intent a bit, i.e. that we want to create two (rather independent) groups of users. so I'd rather keep this as is.

0618 (approved):

packages/amplify-e2e-tests/src/__tests__/auth/admin-api.test.ts
0618: What if we create a function `setUpUserGroup`? 

Then,

```ts
await setupUserGroup({groupName: "Users", toAddUserToGroup: false});
await setupUserGroup({groupName: "Admins", toAddUserToGroup: true})
```

Would it be easier to read?

sobolk: Rather 

```
await setupUsers({ groups: ['Admins', 'Users'], numberOfUsers: 5, namePrefix: 'testAdminUser'} );
await setupUsers({ groups: ['Users'], numberOfUsers: 5, namePrefix: 'testUser'} );
```

But it'd do that when we get second test file that needs this kind of logic, i.e. when we'd need to repeat this pattern. We have couple of places

edwardfoyle (approved):

PR #13710, opened 2024-04-12
chore(cli): upgrade api plugin dep version

sundersc:
Upgrade API dependencies after the latest API amplify-cateogory-api 5.11.0 release.

- All the E2E tests pass.


packages/amplify-cli/package.json
```diff
@@ -66,7 +66,7 @@
     "@aws-amplify/amplify-util-mock": "5.9.4",
     "@aws-amplify/amplify-util-uibuilder": "1.14.3",
     "@aws-cdk/cloudformation-diff": "~2.68.0",
-    "amplify-codegen": "4.7.3",
+    "amplify-codegen": "4.8.0",
```
AaronZyLee, line 69: nit: `amplify-codegen` is bumped here but not in the `yarn.lock` as its version was fixed in the root package json resolution. 
https://github.com/aws-amplify/amplify-cli/blob/dev/package.json#L178-L179

AaronZyLee (approved):

packages/amplify-cli/package.json
sundersc: We do not have any codegen changes that needs to be shipped for Gen1 CLI, let's take care of it when we need to ship new codegen.

sobolk (approved):

PR #13714, opened 2024-04-15
chore: update tar dependency

sobolk:
#### Description of changes

Bumps `tar` dependency.

Note: the bump is via override as lerna depends on it and doesn't have a fix in 6.x line.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -175,6 +175,7 @@
     "semver": "^7.5.4",
     "word-wrap": "^1.2.4",
     "amplify-codegen": "4.7.3",
-    "@aws-amplify/appsync-modelgen-plugin": "2.8.1"
+    "@aws-amplify/appsync-modelgen-plugin": "2.8.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -175,6 +175,7 @@
     "semver": "^7.5.4",
     "word-wrap": "^1.2.4",
     "amplify-codegen": "4.7.3",
-    "@aws-amplify/appsync-modelgen-plugin": "2.8.1"
+    "@aws-amplify/appsync-modelgen-plugin": "2.8.1",
+    "tar": "^6.2.1"
```
sobolk, line 179: Via override because `lerna` depends in it and doesn't have fix in `6.x` line (would require MV bump otherwise).

edwardfoyle (approved):

PR #13718, opened 2024-04-16
fix: reduce memory pressure while uploading files

sobolk:
#### Description of changes

This PR is a partial revert of https://github.com/aws-amplify/amplify-cli/pull/13493/files which caused https://github.com/aws-amplify/amplify-cli/issues/13681 .

I.e. using `s3.upload` API for all uploads puts a memory pressure due to https://github.com/aws/aws-sdk-js/issues/2552 .

This PR changes file uploading logic to buffer all files below 5mb in memory. This partially restoring previous behavior but avoiding full revert to `s3.putObject` API call which doesn't have retries.

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/13681

PR checks and E2E tests.

Manual verification on `t2.nano` with sample test project where `ENOMEM` has been reproduced. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.ts
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.ts
@@ -134,6 +134,8 @@ export class S3 {
         // Previous implementation used s3.putObject for small uploads, but it didn't have retries, see https://github.com/aws-amplify/amplify-cli/pull/13493.
         // On the other hand uploading small streams leads to memory leak, see https://github.com/aws/aws-sdk-js/issues/2552.
         // Therefore, buffering small files ourselves seems to be middle ground between memory leak and loosing retries.
+        // Buffering small files brings back balance between leaking and non-leaking uploads that is matching
+        // the ratio from before https://github.com/aws-amplify/amplify-cli/pull/13493.
         augmentedS3Params.Body = await consumers.buffer(augmentedS3Params.Body);
       }
       uploadTask = this.s3.upload(augmentedS3Params);
--- a/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.ts
+++ b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.ts
@@ -10,6 +10,8 @@ import { $TSAny, $TSContext, AmplifyError, AmplifyFault, stateManager } from '@a
 
 import _ from 'lodash';
 
+import fs from 'fs-extra';
+import { buffer } from 'node:stream/consumers';
 import ora from 'ora';
 import { ListObjectVersionsOutput, ListObjectVersionsRequest, ObjectIdentifier } from 'aws-sdk/clients/s3';
 import { pagedAWSCall } from './paged-call';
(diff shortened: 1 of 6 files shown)


packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.ts
```diff
@@ -10,13 +10,15 @@ import { $TSAny, $TSContext, AmplifyError, AmplifyFault, stateManager } from '@a
...
 
+import fs from 'fs-extra';
 import ora from 'ora';
 import { ListObjectVersionsOutput, ListObjectVersionsRequest, ObjectIdentifier } from 'aws-sdk/clients/s3';
 import { pagedAWSCall } from './paged-call';
 import { loadConfiguration } from '../configuration-manager';
 import aws from './aws';
 
 const providerName = require('../constants').ProviderName;
+const consumers = require('stream/consumers');
```
sobolk, line 21: I was unable to bend TS compiler to accept any form of `import`, so I switched to `require` to workaround it.

It seems that bumping TS compiler to `5.x` would be required to make it right. (which is out of scope here).
The other alternative would be to not use this Node api and write own consumer which is not ideal either.

packages/amplify-provider-awscloudformation/src/download-api-models.ts
```diff
@@ -50,6 +50,9 @@ const extractAPIModel = async (context: $TSContext, resource: $TSObject, framewo
 
   fs.ensureDirSync(tempDir);
 
+  // This has been working fine before bumping node types to 18.x
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
sobolk, line 55: I had to suppress compiler here after bumping node typings...
The alternative would be 5 logical branches to handle union type of `data.body` .

Since this was working fine, I suggest we fix this later as a follow up.

0618: Could you add the comment in the PR to the code? (same for zip-util)

Reading the commend alone in the code doesn't make much sense.

packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.ts
```diff
@@ -126,6 +128,14 @@ export class S3 {
         spinner.start('Uploading files.');
       }
       logger('uploadFile.s3.upload', [others])();
+      const minChunkSize = 5 * 1024 * 1024; // 5 MB
+      if (augmentedS3Params.Body instanceof fs.ReadStream && fs.statSync(augmentedS3Params.Body.path).size <= minChunkSize) {
+        // Buffer small files to avoid memory leak.
```
0618, line 133: Why buffering small files help? Shouldn't we buffer big files?

sobolk: This is because before this change https://github.com/aws-amplify/amplify-cli/pull/13493 all files that were less than 5mb were uploaded via `s3.putObject` API. 
I.e. population of files smaller than 5mb was NOT running into memory leak.
The memory leak only happens if `Body` is a stream, so buffering small files brings back former balance between uploads that leak and not leak.
(keep in mind that large files always leaked, but there wasn't a lot of them, at least not enough to cause current problem we have now).

packages/amplify-provider-awscloudformation/src/download-api-models.ts
sobolk: done.

edwardfoyle (approved):
Change LGTM. Did you also verify that uploading on a slow network still works as expected? See verification description in the previous PR: https://github.com/aws-amplify/amplify-cli/pull/13493#issue-2049608122

0618 (approved):

sobolk:
> Change LGTM. Did you also verify that uploading on a slow network still works as expected? See verification description in the previous PR: [#13493 (comment)](https://github.com/aws-amplify/amplify-cli/pull/13493#issue-2049608122)

I played a bit with `3g` profile, seems to be working.

PR #13827, opened 2024-06-19
fix(cli): add feature flag to control subscription field redaction behavior

sundersc:
Add a new FF for data plugin to allow customers to opt-in for a subscriptions field redaction behavior.

- If the feature flag is disabled (when set to `false`), relational fields will be over-redacted in mutations response and subscribed clients won't receive this data in payload.
- If the feature flag is enabled (when set to `true`), subscriptions will inherit the auth rules of the model for the relational fields. It means that the relational fields won't be over-redacted on mutations.


sobolk (approved):

awsluja (approved):

dpilch (approved):

DarylBeattie:
2 questions here:

1. How do we use this feature flag?
2. Does it work for the update mutations as well as the subscriptions?

PR #13856, opened 2024-07-18
feat(cli): initial migration merge

sdstolworthy:
#### Description of changes

Initial commit migration tooling.

Unit tests validate consistent code generation. An initial e2e test does precursory validations on correctness of the tool, but does not attempt to deploy changes. Manual testing has been used to validate that generated code deploys as expected.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/package.json
```diff
@@ -58,6 +58,7 @@
     "@aws-amplify/amplify-frontend-ios": "3.7.6",
     "@aws-amplify/amplify-frontend-javascript": "3.10.17",
     "@aws-amplify/amplify-go-function-template-provider": "1.4.8",
+    "@aws-amplify/amplify-migration": "^0.0.1",
```
sobolk, line 61: Consider naming this `@aws-amplify/amplify-gen1-gen2-migration` or something more specific.

Why?
1. There are still folks on very old Gen1 CLI who want to move to latest gen1 and may hope this is what they're looking for.
2. Some day there might be Gen 3.

packages/amplify-gen1-codegen-auth-adapter/API.md
```diff
@@ -0,0 +1,21 @@
+## API Report File for "@aws-amplify/amplify-gen1-codegen-auth-adapter"
+
+> Do not edit this file. It is a report generated by [API Extractor](https://api-extractor.com/).
+
+```ts
+
+import { AuthDefinition } from '@aws-amplify/amplify-gen2-codegen';
+import { UserPoolType } from '@aws-sdk/client-cognito-identity-provider';
```
sobolk, line 8: Didn't we decide to not use SDK types in APIs ?

sobolk (approved):
Since we're merging this as is please consider addressing comments in future PRs.

packages/amplify-gen1-codegen-auth-adapter/API.md
sdstolworthy: We discussed not using them in the codegen package so that the codegen package can stay independent. The purpose of the `amplify-gen1-codegen-{auth,storage}` packages are to reduce descriptions of AWS resources down to the inputs required for codegen. This allows codegen to specify its own types and API, while creating a simple way to adhere to that API using common SDK outputs. 

I broke out these separate adapter packages so that directly leveraging SDK types would be more feasible.

packages/amplify-cli/package.json
sdstolworthy: Ack. I have this addressed locally and will include the change in the follow-up PR so it can be properly reviewed.

packages/amplify-gen1-codegen-auth-adapter/API.md
sobolk: The idea isn't bad.

However it has two risks:
1. It's easier to enforce single rule, i.e. "no sdk types" rather than "in this package sdk types are ok". Perhaps this script could be enhanced to enforce it https://github.com/aws-amplify/amplify-cli/blob/dev/scripts/verify-extract-api.sh (but it's some effort to get it right that we might want to avoid).
2. If top level codegen ever re-exports/uses these types they'll transitively get exposed there.


Perhaps middle ground is to continue with this approach but have a release blocker to inspect this later.

PR #13858, opened 2024-07-18
fix(amplify-cli-core): use build script for overrides

brianlenz:
Updated the TypeScript compilation of overrides so that it doesn't require `node_modules/.bin/tsc`. Instead, it simply relies on the `build` script to execute `tsc`. This is more flexible and can support alternative setups w/ hoisting (e.g. via Yarn workspaces).

This is an override corollary fix to #11854, which is for custom resources.

#### Description of changes
Remove hard-coded dependency on `node_modules/.bin/tsc` for overrides to instead use the `build` script from `package.json`, which is more flexible.

Ran `yarn test` and all pre-commit hooks without issue.
Tested `amplify build --debug` and `amplify push` locally. The commands were failing on Amplify CLI `12.12.4`, and it succeeds with these changes in place (via `amplify-dev`).

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


brianlenz:
@josefaidt sorry to pester, but any chance this can get some attention? Would love to be able to stop using custom builds of the CLI! Thanks! 🙏

sobolk (approved):

brianlenz:
Thanks, @sobolk! If you want to add feature parity, there is #11854 which has a nearly identical fix for custom-resources.

sobolk:
> Thanks, @sobolk! If you want to add feature parity, there is #11854 which has a nearly identical fix for custom-resources.

I did see that. The #11854 is a bit out of date, we'll need to rebase it or open new PR.

rtpascual (approved):

PR #13757, opened 2024-05-02
chore: upgrade cdk to 2.129.0

phani-srikar:
#### Description of changes

Upgrading the CDK library version to `2.129.0` which vends the upgraded Node 18 runtimes for the Lambda functions. This is necessary because:
> We are ending support for Node.js 16 in Lambda on June 12, 2024. This follows Node.js 16 End-Of-Life (EOL) reached on September 11, 2023 [1]

This version is chosen so to be the target based on our testing and ease to upgrade without running into any complex breaking changes.

Notable changes:
- Had to update the AWS SDK dependency in the custom pipeline awaiter.
- Had to bump the package size which I believe is a natural increase because of the CDK library.

#### Description of how you validated changes
- We first upgraded the CDK dependency in the API plugin repository and published a tag `acdk-upgrade-2-129`. [Link to branch](https://github.com/aws-amplify/amplify-category-api/tree/tagged-release/acdk-upgrade-2-129)
- We've upgraded the CDK dependency in CLI repo in the `guest/data/upgrade-cdk-2.129.0` feature branch. 
- We've created a `tagged-release-without-e2e-tests/acdk-upgrade-2-129` branch in CLI that consumes the upgraded API plugin via  `acdk-upgrade-2-129` tag and then published a CLI tag also with the same name.
- We've run full E2E on `tagged-release-without-e2e-tests/acdk-upgrade-2-129` branch in CLI. [Link to run](https://tiny.amazon.com/1g3b4wxmd/IsenLink)
- We've then consumed the `acdk-upgrade-2-129` tag of upgraded CLI in API repo and run full E2E suite there. [Link to run](https://tiny.amazon.com/la5clms/IsenLink).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [ ] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)
- [ ] Relevant documentation is changed or added (and PR referenced)
- [ ] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies
- [ ] [Pull request labels](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#labels) are added

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #13872, opened 2024-07-31
fix: update awaiter s3 key

dpilch:
#### Description of changes

With the CDK version upgrade the runtime version of the custom resource lambda changed from 16 to 18. With the runtime change the lambda code needed be changed to account for breaking changes between 16 and 18. However, the S3 key for the zip file containing the lambda code was not changed. During the CFN deploy if the S3 key for the lambda code is unchanged the lambda code will not be updated. 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html

> Changes to a deployment package in Amazon S3 or a container image in ECR are not detected automatically during stack updates. To update the function code, change the object key or version in the template.

This resulted in the runtime version being upgraded to 18, but the code was not updated (and stayed with the runtime version 16 compatible implementation). The lambda would then throw errors when executed.

This change updates the lambda code s3 key to append the runtime version. This will cause the next deploy to update the lambda code because the s3 key has changed.

Changes in API repo: https://github.com/aws-amplify/amplify-category-api/pull/2732

#### Issue #, if available

https://github.com/aws-amplify/amplify-category-api/issues/2730

Manual testing. This issue only presents itself when upgrading from an earlier version of the CLI with an existing awaiter function deployed. We do not have an E2E testing framework that allows multiple versions of the CLI to be used. Adding this testing would delay the release of this fix. The tests can be added at a later time.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


dpilch:
This will also need release from API category.

packages/amplify-provider-awscloudformation/src/resource-package/resource-packager.ts
```diff
@@ -57,7 +57,7 @@ export abstract class ResourcePackager {
 
   constructor(context: $TSContext, deployType: ResourceDeployType) {
     this.context = context;
-    this.elasticContainerZipFiles = ['custom-resource-pipeline-awaiter.zip', 'codepipeline-action-buildspec-generator-lambda.zip'];
+    this.elasticContainerZipFiles = ['custom-resource-pipeline-awaiter-18.zip', 'codepipeline-action-buildspec-generator-lambda.zip'];
```
sdstolworthy, line 60: see: https://github.com/aws-amplify/amplify-category-api/pull/2732/files#r1699028311.

What is the number 18 referencing? What would cause this to change in the future? Should this be parameterized?

palpatim (commented):
LGTM

phani-srikar (dismissed):

phani-srikar (approved):

packages/amplify-provider-awscloudformation/resources/custom-resource-pipeline-awaiter-18.zip
sobolk, line 1: I guess there are changes in zip's content right? 
Is there any chance to see the diff?

dpilch: The contents changed in this PR: https://github.com/aws-amplify/amplify-cli/pull/13757. This is just changing the name (to change the S3 Key).

sobolk (approved):

awsluja (approved):

sobolk (approved):

packages/amplify-provider-awscloudformation/src/resource-package/resource-packager.ts
awsluja: looks like a ref to node18

PR #13970, opened 2024-10-09
Add Gen2 migration template generation command

abhi7cr:
#### Description of changes

Add generate templates command to `to-gen-2` top level migrate command. Usage:

`npx @aws-amplify/migrate --to-gen-2 generate-templates --from gen1RootStackName --to gen2RootStackName`.

TODO: Finalize command messages with @josefaidt. 

local testing, unit tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


bzsurbhi (approved):

packages/amplify-migration/API.md
```diff
@@ -10,6 +10,9 @@ export type AuthCliInputs = Record<string, unknown>;
 // @public (undocumented)
 export function execute(): Promise<void>;
 
+// @public (undocumented)
+export function generateTemplates(fromStack: string, toStack: string): Promise<void>;
```
sobolk, line 14: does this have to be public ?

abhi7cr: It was generated when I ran the api extract command. It needs to be exported from index file to be consumed by the command handler within the same package (amplify-migration). Did it mark it as public since its exported from an index file ?

sobolk: if this is needed only within same package it should be taken out from public surface.

abhi7cr: yeah I think it was because of the `index` file and exporting from it since we can look at it in the `extractApi` method:
`const hasEntryPoint = fs.pathExistsSync(path.join(packagePath, 'lib', 'index.js'));`
I renamed the index file, deleted `API.md`, and ran `yarn extract-api` and it didn't generate the API.md file :)

sobolk: One thing to consider is to leave index.ts in place as entry point (so that package is valid from module loader perspective, I recall some tools like jest may complain about absent entry point).

But just have `index.ts` empty.

sobolk: btw. if this is top level CLI/executable, it probably shouldn't have public API.

abhi7cr: oh got it. Yeah it is a top level CLI executable, so makes sense to not have a public API. `jest` has been kind so far. I can add a dummy `index.ts` if it starts complaining.

sobolk (approved):

PR #13984, opened 2024-10-21
feat: templategen command e2e integration tests

Sanayshah2:
#### Description of changes
Added e2e integration test for templategen command which runs as a part of the `yarn e2e-migration` command

Manually checked and verified the deployed stack resources, assertion statements
#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-migration-e2e/src/templategen.ts
+++ b/packages/amplify-migration-e2e/src/templategen.ts
@@ -1,0 +1,175 @@
+import execa from 'execa';
+import path from 'node:path';
+import * as fs from 'fs-extra';
+import { getNpxPath, retry } from '@aws-amplify/amplify-e2e-core';
+import { runGen2SandboxCommand } from '.';
+
+export function runTemplategenCommand(cwd: string, gen1StackName: string, gen2StackName: string) {
+  const parentDir = path.resolve(cwd, '..');
+  const processResult = execa.sync(
+    getNpxPath(),
+    ['--prefix', parentDir, '@aws-amplify/migrate', 'to-gen-2', 'generate-templates', '--from', gen1StackName, '--to', gen2StackName],
+    {
+      cwd,
+      env: { ...process.env, npm_config_user_agent: 'npm' },
+      encoding: 'utf-8',
+    },
+  );
+
+  if (processResult.exitCode !== 0) {
+    throw new Error(`Templategen command exit code: ${processResult.exitCode}, message: ${processResult.stderr}`);
+  }
+}
+
+function uncommentBucketNameLineFromBackendFile(projRoot: string) {
+  const backendFilePath = path.join(projRoot, 'amplify', 'backend.ts');
+  const backendFileContent = fs.readFileSync(backendFilePath, 'utf8');
+  const regex = /^\s*\/\/\s*(s3Bucket\.bucketName)/m;
+  const updatedBackendFileContent = backendFileContent.replace(regex, '$1');
+  fs.writeFileSync(backendFilePath, updatedBackendFileContent);
+}
+
+function setEnvVariable(name: string, value: string) {
+  process.env[name] = value;
+}
+
+function extractContent(readmeContent: string, startRegex: string, endRegex: string) {
+  const pattern = new RegExp(`${startRegex}([\\s\\S]*?)${endRegex}`, 'i');
+  const match = readmeContent.match(pattern);
+
+  if (match && match[1]) {
+    return match[1].trim();
+  }
+  throw new Error('README file parsing failed to get the stack refactor commands');
+}
+
+function extractCommands(readmeContent: string) {
+  const pattern = /```([\s\S]*?)```/g;
+  const matches = readmeContent.matchAll(pattern);
+  const commands = [];
+
+  for (const match of matches) {
+    if (match[1]) {
+      commands.push(match[1].trim());
+    }
+  }
+  if (commands.length === 0) {
+    throw new Error('README file parsing failed to get the stack refactor commands');
+  }
+  return commands;
+}
+
+function getCommandsFromReadme(readmeContent: string) {
+  const step1Content = extractContent(readmeContent, '### STEP 1', '#### Rollback step');
+  const step2Content = extractContent(readmeContent, '### STEP 2', '#### Rollback step');
+  const step3Content = extractContent(readmeContent, '### STEP 3', '#### Rollback step');
+  const step1Commands = extractCommands(step1Content);
+  const step2commands = extractCommands(step2Content);
+  const step3Commands = extractCommands(step3Content);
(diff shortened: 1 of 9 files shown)


packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,175 @@
...
+
+function uncommentBucketNameLineFromBackendFile(projRoot: string) {
+  const backendFilePath = path.join(projRoot, 'amplify', 'backend.ts');
+  const backendFileContent = fs.readFileSync(backendFilePath, 'utf8');
+  const regex = /^\s*\/\/\s*(s3Bucket\.bucketName)/m;
+  const updatedBackendFileContent = backendFileContent.replace(regex, '$1');
+  fs.writeFileSync(backendFilePath, updatedBackendFileContent);
+}
+
+function setEnvVariable(name: string, value: string) {
```
hoangnbn, line 32: delete these envVars after test run

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,175 @@
...
+async function assertRefactorStepCompletion(command: string) {
+  const processResult = JSON.parse(await executeCommand(command));
+  return processResult.Status;
+}
+
+async function executeStep1(cwd: string, commands: string[]) {
+  await executeCommand(commands[0], cwd);
+  await retry(
+    () => assertStepCompletion(commands[1]),
+    (status) => status.includes('COMPLETE') && !status.includes('IN_PROGRESS'),
```
abhi7cr, line 108: nit: move `IN_PROGRESS` to a constant.

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,175 @@
...
+    {
+      times: 50,
+      delayMS: 1000, // 1 second
+      timeoutMS: 1000 * 60 * 5, // 5 minutes
+    },
+    (status) => status.includes('FAILED'),
+  );
+}
+
+export async function stackRefactor(projRoot: string, category: string, bucketName: string) {
```
abhi7cr, line 161: nit: make `category` param a type to constrain the possibilities:  `export RefactorCategory = 'auth' | 'storage'`

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,175 @@
...
+      delayMS: 1000, // 1 second
+      timeoutMS: 1000 * 60 * 5, // 5 minutes
+    },
+    (status) => status.includes('FAILED'),
+  );
+}
+
+async function executeStep3(cwd: string, commands: string[], bucketName: string) {
+  setEnvVariable('BUCKET_NAME', bucketName);
+  await executeCommand(commands[1], cwd);
```
abhi7cr, line 134: can we name the commands instead of accessing by indices for easier understanding: e.g. `const [uploadSourceTemplate , uploadDestinationTemplate, ...]` ?

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,175 @@
...
+  const regex = /^\s*\/\/\s*(s3Bucket\.bucketName)/m;
+  const updatedBackendFileContent = backendFileContent.replace(regex, '$1');
+  fs.writeFileSync(backendFilePath, updatedBackendFileContent);
+}
+
+function setEnvVariable(name: string, value: string) {
+  process.env[name] = value;
+}
+
+function extractContent(readmeContent: string, startRegex: string, endRegex: string) {
```
abhi7cr, line 36: nice!

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,175 @@
...
+  const pattern = new RegExp(`${startRegex}([\\s\\S]*?)${endRegex}`, 'i');
+  const match = readmeContent.match(pattern);
+
+  if (match && match[1]) {
+    return match[1].trim();
+  }
+  throw new Error('README file parsing failed to get the stack refactor commands');
+}
+
+function extractCommands(readmeContent: string) {
```
abhi7cr, line 46: nice!

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,175 @@
...
+      commands.push(match[1].trim());
+    }
+  }
+  if (commands.length === 0) {
+    throw new Error('README file parsing failed to get the stack refactor commands');
+  }
+  return commands;
+}
+
+function getCommandsFromReadme(readmeContent: string) {
```
hoangnbn, line 62: I think the README parsing logic is difficult to follow here. Unless a code reviewer knows a logic beforehand, it is not trivial to know what command 1,2,3 are. Some options here:
1. Black box : Parse the README and get the list of commands, execute them one by one and doesn't care what it is. the test only cares about the end state of the stacks (i.e stack in UPDATE_COMPLETE) and resources are moved/renamed to the right stacks. (Note: You can use stack state to know if update/refactor operation is completed or not, doesn't have to get refactor object or changeset)
2. White-box: parse the README and assume knowledge on the migration logic, like command 1 : upload to s3, command 2: update stack, command 3: refactor, and so on.
```
const stackMigrationHandler = StackMigrationHandler.fromTemplateGenReadme(<readme path>)
stackMigrationHandler.uploadToS3()
stackMigrationHandler.updateStack()
stackMigrationHandler.refactorStacks()
......
```

I would suggest the black box testing for simplicity.

abhi7cr: That's a great idea. We can do both kinds of testing based on what we are testing. For the happy path, we can do black box testing since we only care about the end state. We already have `getCommandsFromReadme`, we can just execute them one by one.  
For the rollback scenario, we need to extract Rollback specific commands. So that can be its own test.

packages/amplify-migration-e2e/src/templategen.ts
Sanayshah2: If we follow the Black Box Testing approach, this approach is irrelevant then?

abhi7cr: Yup!

packages/amplify-migration-e2e/src/templategen.ts
Sanayshah2: Sounds good, will keep the logic for the `getCommandsFromReadme` and executing commands one by one logic as it is. Should the commands be stored in a variable or accessing them by indices should be okay for this approach?

abhi7cr: yeah in the black box mode, we don't need the names. Iterating them one by one suffice

PR #13993, opened 2024-10-23
feat: templategen command e2e integration tests

Sanayshah2:
#### Description of changes
Added e2e integration test for templategen command which runs as a part of the `yarn e2e-migration` command

Manually checked and verified the deployed stack resources, assertion statements
#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-migration-e2e/src/__tests__/migration_codegen_e2e.test.ts
+++ b/packages/amplify-migration-e2e/src/__tests__/migration_codegen_e2e.test.ts
@@ -1,6 +1,6 @@
 import path from 'node:path';
 import assert from 'node:assert';
-import { createNewProjectDir, npmInstall } from '@aws-amplify/amplify-e2e-core';
+import { createNewProjectDir, generateRandomShortId, npmInstall } from '@aws-amplify/amplify-e2e-core';
 import { createGen2Renderer } from '@aws-amplify/amplify-gen2-codegen';
 import { copyFunctionFile } from '../function_utils';
 import { copyGen1Schema } from '../api_utils';
@@ -46,7 +46,7 @@ void describe('Codegen E2E tests', () => {
     beforeEach(async () => {
       const baseDir = process.env.INIT_CWD ?? process.cwd();
       projRoot = await createNewProjectDir('codegen_e2e_flow_test', path.join(baseDir, '..', '..'));
-      projName = `test${Math.floor(Math.random() * 1000000)}`;
+      projName = `test${generateRandomShortId()}`;
     });
 
     afterEach(async () => {
@@ -60,10 +60,7 @@ void describe('Codegen E2E tests', () => {
       await runCodegenCommand(projRoot);
       await copyFunctionFile(projRoot, 'function', gen1FunctionName);
       await copyGen1Schema(projRoot, projName);
-
-      // TODO: replace below line with correct package version
       await updatePackageDependency(projRoot, '@aws-amplify/backend');
-
       await npmInstall(projRoot);
       const gen2StackName = await runGen2SandboxCommand(projRoot);
       await assertAuthResource(projRoot, gen1UserPoolId, gen1ClientIds, gen1IdentityPoolId, gen1Region);
@@ -80,10 +77,7 @@ void describe('Codegen E2E tests', () => {
       await runCodegenCommand(projRoot);
       await copyFunctionFile(projRoot, 'auth', gen1FunctionName);
       await removeErrorThrowsFromAuthResourceFile(projRoot);
-
-      // TODO: replace below line with correct package version
       await updatePackageDependency(projRoot, '@aws-amplify/backend');
-
       await npmInstall(projRoot);
       await toggleSandboxSecrets(projRoot, 'set');
       const gen2StackName = await runGen2SandboxCommand(projRoot);
@@ -98,10 +92,7 @@ void describe('Codegen E2E tests', () => {
         projRoot,
       );
       await runCodegenCommand(projRoot);
-
-      // TODO: replace below line with correct package version
       await updatePackageDependency(projRoot, '@aws-amplify/backend');
-
       await npmInstall(projRoot);
       await runGen2SandboxCommand(projRoot);
       await assertAuthResource(projRoot, gen1UserPoolId, gen1ClientIds, gen1IdentityPoolId, gen1Region);
--- a/packages/amplify-migration-e2e/src/__tests__/migration_codegen_e2e.test.ts
+++ b/packages/amplify-migration-e2e/src/__tests__/migration_codegen_e2e.test.ts
@@ -60,8 +60,10 @@ void describe('Codegen E2E tests', () => {
       await runCodegenCommand(projRoot);
       await copyFunctionFile(projRoot, 'function', gen1FunctionName);
       await copyGen1Schema(projRoot, projName);
+
       // TODO: replace below line with correct package version
-      await updatePackageDependency(projRoot, '@aws-amplify/backend', '0.0.0-test-20241003180022');
+      await updatePackageDependency(projRoot, '@aws-amplify/backend');
+
       await npmInstall(projRoot);
       const gen2StackName = await runGen2SandboxCommand(projRoot);
       await assertAuthResource(projRoot, gen1UserPoolId, gen1ClientIds, gen1IdentityPoolId, gen1Region);
@@ -78,8 +80,10 @@ void describe('Codegen E2E tests', () => {
       await runCodegenCommand(projRoot);
       await copyFunctionFile(projRoot, 'auth', gen1FunctionName);
       await removeErrorThrowsFromAuthResourceFile(projRoot);
+
       // TODO: replace below line with correct package version
-      await updatePackageDependency(projRoot, '@aws-amplify/backend', '0.0.0-test-20241003180022');
+      await updatePackageDependency(projRoot, '@aws-amplify/backend');
+
       await npmInstall(projRoot);
       await toggleSandboxSecrets(projRoot, 'set');
       const gen2StackName = await runGen2SandboxCommand(projRoot);
@@ -94,8 +98,10 @@ void describe('Codegen E2E tests', () => {
         projRoot,
       );
       await runCodegenCommand(projRoot);
+
       // TODO: replace below line with correct package version
-      await updatePackageDependency(projRoot, '@aws-amplify/backend', '0.0.0-test-20241003180022');
+      await updatePackageDependency(projRoot, '@aws-amplify/backend');
+
       await npmInstall(projRoot);
       await runGen2SandboxCommand(projRoot);
       await assertAuthResource(projRoot, gen1UserPoolId, gen1ClientIds, gen1IdentityPoolId, gen1Region);
--- a/packages/amplify-migration-e2e/src/__tests__/migration_templategen_e2e.test.ts
+++ b/packages/amplify-migration-e2e/src/__tests__/migration_templategen_e2e.test.ts
@@ -43,7 +43,7 @@ void describe('Templategen E2E tests', () => {
       const gen2StackName = await runGen2SandboxCommand(projRoot);
       assert(gen2StackName);
       await runTemplategenCommand(projRoot, gen1StackName, gen2StackName);
-      // await stackRefactor(projRoot, 'auth', bucketName);
+      await stackRefactor(projRoot, 'auth', bucketName);
       await stackRefactor(projRoot, 'storage', bucketName);
     });
   });
--- a/packages/amplify-migration-e2e/src/__tests__/migration_templategen_e2e.test.ts
+++ b/packages/amplify-migration-e2e/src/__tests__/migration_templategen_e2e.test.ts
@@ -1,6 +1,6 @@
 import path from 'node:path';
 import assert from 'node:assert';
-import { createNewProjectDir, npmInstall, deleteS3Bucket } from '@aws-amplify/amplify-e2e-core';
+import { createNewProjectDir, npmInstall, deleteS3Bucket, generateRandomShortId } from '@aws-amplify/amplify-e2e-core';
 import { assertDefaultGen1Setup } from '../assertions';
 import { setupAndPushDefaultGen1Project, runCodegenCommand, runGen2SandboxCommand, cleanupProjects } from '..';
 import { copyFunctionFile } from '../function_utils';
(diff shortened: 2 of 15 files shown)


packages/amplify-migration-e2e/src/migrationReadmeParser.ts
```diff
@@ -0,0 +1,35 @@
...
+      commands.push(match[1].trim());
+    }
+  }
+  if (commands.length === 0) {
+    throw new Error('README file parsing failed to get the stack refactor commands');
+  }
+  return commands;
+}
+
+export function getCommandsFromReadme(readmeContent: string) {
```
Sanayshah2, line 27: Referencing the comment from the previous PR: https://github.com/aws-amplify/amplify-cli/pull/13984#discussion_r1809384187

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,143 @@
...
+    () => assertStepCompletion(commands[1]),
+    (status) => status.includes(STATUS_COMPLETE) && !status.includes(STATUS_IN_PROGRESS),
+    RETRY_CONFIG,
+    (status) => status.includes(STATUS_FAILED),
+  );
+}
+
+async function executeStep3(cwd: string, commands: string[], bucketName: string) {
+  toggleEnvVariable('BUCKET_NAME', 'SET', bucketName);
+  await executeCommand(commands[1], cwd);
```
Sanayshah2, line 93: eferencing the comment from the previous PR: https://github.com/aws-amplify/amplify-cli/pull/13984#discussion_r1809356785

packages/amplify-migration-e2e/src/gen1ResourceDetailsFetcher.ts
```diff
@@ -0,0 +1,42 @@
...
+import { getResourceDetails } from './sdk_calls';
+
+async function getGen1AuthResourceDetails(projRoot: string) {
+  const gen1ProjRoot = path.join(projRoot, '.amplify', 'migration');
+  const gen1Meta = getProjectMeta(gen1ProjRoot);
+  const gen1Region = gen1Meta.providers.awscloudformation.Region;
+  const { gen1UserPoolId } = await assertUserPool(gen1Meta, gen1Region);
+  const { gen1IdentityPoolId } = await assertIdentityPool(gen1Meta, gen1Region);
+  const { gen1ClientIds } = await assertUserPoolClients(gen1Meta, gen1Region);
+  const [gen1ClientIdWeb, gen1ClientId] = gen1ClientIds;
```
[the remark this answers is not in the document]
Sanayshah2: This variable will be used later, there is a bug which does creates only one UserPoolClient instead of 2. Once, that bug is resolved, the second UserPoolClient will also be used for assertions.

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,143 @@
...
+  }
+}
+
+async function executeCreateStackRefactorCallCommand(command: string, cwd: string) {
+  const processResult = JSON.parse(await executeCommand(command, cwd));
+  const stackRefactorId = processResult.StackRefactorId;
+  return stackRefactorId;
+}
+
+async function executeStep1(cwd: string, commands: string[]) {
```
hoangnbn, line 71: what is the commands[0] and [1] in step1? To make the code self documented, assert that the commands[x] is the command you expected. i.e assert(command[x] has prefix `npx amplify....`).
Similarly for step 2 and step 3 pls.

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,143 @@
...
+  if (category === 'storage') await uncommentS3BucketLineFromBackendFile(projRoot);
+
+  await runGen2SandboxCommand(projRoot);
+
+  toggleEnvVariable('BUCKET_NAME', 'DELETE');
+  toggleEnvVariable('STACK_REFACTOR_ID', 'DELETE');
+
+  const { gen2ResourceIds, gen2ResourceDetails } = await getGen2ResourceDetails(projRoot, category);
+  assert.deepEqual(gen1ResourceIds, gen2ResourceIds);
+  assert.deepEqual(gen1ResourceDetails, gen2ResourceDetails);
```
hoangnbn, line 142: add assertion that these gen1ResourceIds no longer exists in gen1 stacks.

packages/amplify-migration-e2e/src/migrationReadmeParser.ts
hoangnbn: nit: give some sample of the readme fie, like the step 1 portion, would be helpful to visualize the file format.

packages/amplify-migration-e2e/src/templategen.ts
Sanayshah2: The `gen2ResourceIds` are fetched from the `Outputs` after the final `npx ampx sandbox` which is the final step for stack refactor. I believe the resources can only be present in one stack and CFN doesn't allow it to be in two stacks. So, I think the assertion that gen1ResourceIds no longer exists in gen1 stacks gets handled there since we are making sure the ids are fetched from the outputs and not reusing `gen2ResourceIds `. Also, to do this assertion, I would have to make a separate call to get the nested stack name and other call for assertion.

packages/amplify-migration-e2e/src/templategen.ts
Sanayshah2: These assertions are handled in the unit tests which checks if every command is generated correctly and matches the expected string.

hoangnbn: I see, yet I still think we need to add these assertion here for below reasons:
1. Self documented code. Otherwise, it is difficult to review, maintain, and extend in the future.
2. Easy to identify bugs. Specifically, if some command changes, unit test will adapt in the same CR. However, this e2e may (or may not) fail at random places which would take time to debug.
3. Unit test vs e2e serve different purposes, yet may share some assertions. It is best to decouple them and avoid making assumptions. E2e should be from end-user perspective, and it, by itself, should give us confident on the UX.

packages/amplify-migration-e2e/src/templategen.ts
hoangnbn: fair point!

packages/amplify-migration-e2e/src/templategen.ts
Sanayshah2: But I believe, every step waits for the status to be complete and if the command doesn't get parsed correctly, the test would fail at that specific step which is reflected in the stackStatus. We are waiting for that to be complete and only then move forward, so we would know for sure which specific command failed. 

Also, there are definitely some changes in the steps once the other teams resolve the bugs, so every time there is a small change, the e2e tests will every time have to be changed.

Also, if I assert every step having some prefix, won't it defeat the Black Box Testing purpose? 

This line asserts the exact command in unit tests - 
https://github.com/aws-amplify/amplify-cli/blob/bc4a6627db22f008ade914edc1809d0240403ff4/packages/amplify-migration-template-gen/src/migration-readme-generator.test.ts#L67

packages/amplify-migration-e2e/src/gen2ResourceDetailsFetcher.ts
```diff
@@ -0,0 +1,37 @@
...
+  const gen2BucketName = gen2Meta.storage.bucket_name;
+  const gen2ResourceIds = [gen2BucketName];
+  const gen2ResourceDetails = await getResourceDetails('AWS::S3::Bucket', gen2BucketName, gen2Region);
+  return { gen2ResourceIds, gen2ResourceDetails };
+}
+
+export async function getGen2ResourceDetails(projRoot: string, category: RefactorCategory) {
+  if (category === 'auth') {
+    return await getGen2AuthResourceDetails(projRoot);
+  } else {
```
abhi7cr, line 34: nit: explicitly check for `storage` if we ever expand to other categories.

packages/amplify-migration-e2e/src/templategen.ts
abhi7cr, line 93: do we need to remove this line (`commands[1]`) since we are running the 3rd command below (`commands[2]`)

packages/amplify-migration-e2e/src/templategen.ts
```diff
@@ -0,0 +1,143 @@
...
+async function executeCreateStackRefactorCallCommand(command: string, cwd: string) {
+  const processResult = JSON.parse(await executeCommand(command, cwd));
+  const stackRefactorId = processResult.StackRefactorId;
+  return stackRefactorId;
+}
+
+async function executeStep1(cwd: string, commands: string[]) {
+  await executeCommand(commands[0], cwd);
+  await retry(
+    () => assertStepCompletion(commands[1]),
```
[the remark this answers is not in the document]
Sanayshah2: `commands[0]` is the command for updating the stack in step 1 and `commands[1]` is the describe stacks command to get the `stackStatus`. So, that command is executed for asserting the status complete.

abhi7cr (approved):

packages/amplify-migration-e2e/src/__tests__/migration_templategen_e2e.test.ts
```diff
@@ -0,0 +1,50 @@
...
+  void describe('Full Migration Templategen Flow', () => {
+    let projRoot: string;
+    let projName: string;
+    let bucketName: string;
+
+    beforeEach(async () => {
+      const baseDir = process.env.INIT_CWD ?? process.cwd();
+      projRoot = await createNewProjectDir('templategen_e2e_flow_test', path.join(baseDir, '..', '..'));
+      projName = `test${Math.floor(Math.random() * 1000000)}`;
+      bucketName = `testbucket${Math.floor(Math.random() * 1000000)}`;
```
sobolk, line 22: we have `generateRandomShortId` somewhere in `amplify-e2e-core` for this kind of thing.

packages/amplify-migration-e2e/src/__tests__/migration_templategen_e2e.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    void it('should init a project & add auth, function, storage, api with defaults & perform full migration templategen flow', async () => {
+      await setupAndPushDefaultGen1Project(projRoot, projName);
+      const { gen1StackName, gen1FunctionName, gen1Region } = await assertDefaultGen1Setup(projRoot);
+      await createS3Bucket(bucketName, gen1Region);
+      assert(gen1StackName);
+      await runCodegenCommand(projRoot);
+      await copyFunctionFile(projRoot, 'function', gen1FunctionName);
+      await copyGen1Schema(projRoot, projName);
+
+      // TODO: replace below line with correct package version
```
sobolk, line 39: is this comment still relevant ?

packages/amplify-migration-e2e/src/__tests__/migration_codegen_e2e.test.ts
```diff
@@ -78,8 +80,10 @@ void describe('Codegen E2E tests', () => {
       await runCodegenCommand(projRoot);
       await copyFunctionFile(projRoot, 'auth', gen1FunctionName);
       await removeErrorThrowsFromAuthResourceFile(projRoot);
+
       // TODO: replace below line with correct package version
-      await updatePackageDependency(projRoot, '@aws-amplify/backend', '0.0.0-test-20241003180022');
+      await updatePackageDependency(projRoot, '@aws-amplify/backend');
```
sobolk, line 85: are todo comments relevant still ?

packages/amplify-migration-e2e/src/envVariables.ts
```diff
@@ -0,0 +1,9 @@
+type EnvVariableAction = 'SET' | 'DELETE';
+
+export function toggleEnvVariable(name: string, option: EnvVariableAction, value?: string) {
+  if (option === 'SET') {
+    process.env[name] = value;
+  } else if (option === 'DELETE') {
+    delete process.env[name];
+  }
+}
```
sobolk, line 9: TBH. this doesn't seem to be providing lot of simplification over calling `process.env` directly. It's just changing syntax of how we set and unset variable without much value added.

If we want to build something that takes care of scope of env vars I'd suggest to do something like this.

```
function runWithEnvVarialbe(name: string, value: string, callable: ()=> void) => {
 process.env[name] = value;
 try {
   callable();
 } finally {
   delete process.env[name]
 }

}
```

packages/amplify-migration-e2e/src/gen1ResourceDetailsFetcher.ts
```diff
@@ -0,0 +1,43 @@
...
+
+async function getGen1AuthResourceDetails(projRoot: string) {
+  const gen1ProjRoot = path.join(projRoot, '.amplify', 'migration');
+  const gen1Meta = getProjectMeta(gen1ProjRoot);
+  const gen1Region = gen1Meta.providers.awscloudformation.Region;
+  const { gen1UserPoolId } = await assertUserPool(gen1Meta, gen1Region);
+  const { gen1IdentityPoolId } = await assertIdentityPool(gen1Meta, gen1Region);
+  const { gen1ClientIds } = await assertUserPoolClients(gen1Meta, gen1Region);
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [gen1ClientIdWeb, gen1ClientId] = gen1ClientIds;
```
sobolk, line 15: Please don't do this.

If this will be used earlier then this should be added later in follow up PR.
Otherwise if follow up never happens we're left with debt.

packages/amplify-migration-e2e/src/envVariables.ts
Sanayshah2: Sure, but I think even this wouldn't quite work, because there are a few steps which involve setting an env var, and then based on those steps running correctly, the stdout of those commands, gives another envvar which need to be set. So, if the case where I had an env var used for specific steps and then other steps use other vars, this would work well. But for this use case, I think simply going with `setEnvVariable` and `deleteEnvVariable` functions would be simple to read and understand.

sobolk (approved):

packages/amplify-migration-e2e/src/gen1ResourceDetailsFetcher.ts
Sanayshah2: Noted. I'll keep this in mind. [b28e101](https://github.com/aws-amplify/amplify-cli/pull/13993/commits/b28e1014fa504a64624b444c916fde099102071e)

packages/amplify-migration-e2e/src/__tests__/migration_templategen_e2e.test.ts
Sanayshah2: Irrelevant and removed - [b28e101](https://github.com/aws-amplify/amplify-cli/pull/13993/commits/b28e1014fa504a64624b444c916fde099102071e)

packages/amplify-migration-e2e/src/__tests__/migration_codegen_e2e.test.ts
Sanayshah2: Irrelevant and removed - [b28e101](https://github.com/aws-amplify/amplify-cli/pull/13993/commits/b28e1014fa504a64624b444c916fde099102071e)

packages/amplify-migration-e2e/src/__tests__/migration_templategen_e2e.test.ts
Sanayshah2: Thanks for pointing this out! [b28e101](https://github.com/aws-amplify/amplify-cli/pull/13993/commits/b28e1014fa504a64624b444c916fde099102071e)

PR #14016, opened 2024-11-18
fix: added user pool client codegen

bzsurbhi:
#### Description of changes
- Added codegen for userPoolClient configuration
- Added retain policy for identity pool, user pool and s3 bucket
- Fixed a bug in s3 encryption algorithm codegen

Added unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-gen2-codegen/src/backend/synthesizer.ts
+++ b/packages/amplify-gen2-codegen/src/backend/synthesizer.ts
@@ -178,33 +178,36 @@ export class BackendSynthesizer {
   }
 
   private createUserPoolClientAssignment(userPoolClient: UserPoolClientType, imports: any[]) {
-    const userPoolAtrributesMap = new Map<string, string>();
-    userPoolAtrributesMap.set('ClientName', 'userPoolClientName');
-    userPoolAtrributesMap.set('ClientSecret', 'generateSecret');
-    userPoolAtrributesMap.set('ReadAttributes', 'readAttributes');
-    userPoolAtrributesMap.set('WriteAttributes', 'writeAttributes');
-    userPoolAtrributesMap.set('RefreshTokenValidity', 'refreshTokenValidity');
-    userPoolAtrributesMap.set('AccessTokenValidity', 'accessTokenValidity');
-    userPoolAtrributesMap.set('IdTokenValidity', 'idTokenValidity');
-    userPoolAtrributesMap.set('RefreshToken', 'refreshToken');
-    userPoolAtrributesMap.set('AccessToken', 'accessToken');
-    userPoolAtrributesMap.set('IdToken', 'idToken');
-    userPoolAtrributesMap.set('AllowedOAuthScopes', 'scopes');
-    userPoolAtrributesMap.set('CallbackURLs', 'callbackUrls');
-    userPoolAtrributesMap.set('LogoutURLs', 'logoutUrls');
-    userPoolAtrributesMap.set('DefaultRedirectURI', 'defaultRedirectUri');
-    userPoolAtrributesMap.set('AllowedOAuthFlowsUserPoolClient', 'disableOAuth');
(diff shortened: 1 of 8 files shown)


packages/amplify-gen2-codegen/src/backend/synthesizer.ts
```diff
@@ -115,6 +122,291 @@ export class BackendSynthesizer {
...
+
+      if (identifiers.length > 0) {
+        imports.push(this.createImportStatement(identifiers, 'aws-cdk-lib/aws-cognito'));
+      }
+    }
+
+    return test;
+  }
+
+  private createNestedObjectExpression(object: Record<string, any>, map: Map<string, string>): ts.ObjectLiteralExpression {
```
abhi7cr, line 229: Nice work on making this function generic enough for nested objects. 
Not blocking this PR since Cognito APIs are wildly different from their CDK counterparts. I was thinking for the next iteration on this if we can pass a callback function that can be run inside this function to provide custom behavior. That way it can become a generic function that can be used across categories :)

abhi7cr, line 229: nit: rename `map` to something like `gen2PropertyMap`

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
```diff
@@ -115,6 +122,291 @@ export class BackendSynthesizer {
...
+        factory.createIdentifier('RemovalPolicy.RETAIN'),
+        factory.createObjectLiteralExpression(
+          [factory.createPropertyAssignment(factory.createIdentifier('applyToUpdateReplacePolicy'), factory.createTrue())],
+          false,
+        ),
+      ],
+    );
+  }
+
+  private createUserPoolAssignment(userPoolClient: UserPoolClientType, imports: any[]) {
```
abhi7cr, line 180: `userPoolClientAssignment` ?

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
```diff
@@ -288,6 +582,12 @@ export class BackendSynthesizer {
       );
     }
 
+    if (renderArgs.auth?.userPoolClient) {
```
abhi7cr, line 585: nit: Can we add a comment stating why we need to codegen this additional user pool client ? Something like:
`Since Gen2 only supports 1 user pool client by default, we need to add CDK overrides for the additional user pool client from Gen1`

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
bzsurbhi: Done

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
bzsurbhi: Done

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
bzsurbhi: Sure, that's a good idea. I'll update that in the next CR

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
```diff
@@ -115,6 +122,291 @@ export class BackendSynthesizer {
...
+        factory.createObjectLiteralExpression(
+          [factory.createPropertyAssignment(factory.createIdentifier('applyToUpdateReplacePolicy'), factory.createTrue())],
+          false,
+        ),
+      ],
+    );
+  }
+
+  private createUserPoolAssignment(userPoolClient: UserPoolClientType, imports: any[]) {
+    const userPoolAtrributesMap = new Map<string, string>();
```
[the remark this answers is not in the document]
bzsurbhi: Done

abhi7cr (approved):

PR #14023, opened 2024-11-21
feat: dotnet8 support for linux image

awsluja:
#### Description of changes
Updates Linux image build to use dotnet8.
Tested build using local pipeline.

image

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -147,11 +147,11 @@ FROM tools AS runtimes
 
 #****************     .NET-CORE     *******************************************************
 
-ENV DOTNET_6_SDK_VERSION="6.0.405"
+ENV DOTNET_8_SDK_VERSION="8.0.11"
 ENV DOTNET_ROOT="/root/.dotnet"
 
 # Add .NET Core 6 Global Tools install folder to PATH
-RUN  /usr/local/bin/dotnet-install.sh -v $DOTNET_6_SDK_VERSION \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -147,11 +147,11 @@ FROM tools AS runtimes
 
 #****************     .NET-CORE     *******************************************************
 
-ENV DOTNET_6_SDK_VERSION="6.0.405"
+ENV DOTNET_8_SDK_VERSION="8.0.404"
```
sobolk, line 150: there should be a way instal both 6 and 8 if we want backwards compatibility for some time. (think hotfixes, rollbacks in main repo).

awsluja: currently I am planning to be able to just create this image for testing (since the old images are not affected), ie:
- enable way to create test image (to expedite full-e2e testing and iteration, I will use the images via overrides settings for codebuild batch until that is stable)
- as a follow up, I can submit a PR to introduce image with multiple version of dotnet (and test with that as well using overrides); if that goes well, we can use this as final image, but currently I am not able to do much testing in bulk and want to have a standalone v8 to set a baseline

sobolk (approved):

awsluja:
closing for now, there is a way to build an image from a branch without merging to /latest

PR #14026, opened 2024-11-25
feat: lambda dotnet 8 support

awsluja:
#### Description of changes
Re: https://aws.amazon.com/blogs/compute/introducing-the-net-8-runtime-for-aws-lambda/
Migrate Lambda functions to .Net 8

Related: https://github.com/aws-amplify/amplify-cli/pull/14023

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/codebuild_specs/run_e2e_tests_windows.yml
+++ b/codebuild_specs/run_e2e_tests_windows.yml
@@ -25,7 +25,6 @@ env:
 phases:
   build:
     commands:
-      - choco install -fy jq
       - git config --global core.longpaths true
       - bash ./codebuild_specs/scripts-windows/load-e2e-cache.sh
       - bash ./codebuild_specs/scripts-windows/rename-packaged-cli.sh
--- a/codebuild_specs/scripts-windows/shared-scripts-windows.sh
+++ b/codebuild_specs/scripts-windows/shared-scripts-windows.sh
@@ -117,8 +117,8 @@ function _install_packaged_cli_win {
     echo Move to CLI Binary to already existing PATH
     # This is a Hack to make sure the Amplify CLI is in the PATH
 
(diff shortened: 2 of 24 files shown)


codebuild_specs/run_e2e_tests_windows.yml
```diff
@@ -25,7 +25,6 @@ env:
 phases:
   build:
     commands:
-      - choco install -fy jq
```
awsluja, line 28: we no longer need this, the image was updated to include jq by default

codebuild_specs/scripts-windows/shared-scripts-windows.sh
```diff
@@ -98,9 +98,6 @@ function _loadTestAccountCredentials {
 function _lsOut {
     ls ..
     ls ~
-    ls $HOME
```
awsluja, line 101: The reason for removing calls to $HOME is they are no longer defined in newer images, and we don't really use this anyway.
Updated to use specific path which is set to C:/Users/ContainerAdministrator

packages/amplify-e2e-tests/src/__tests__/import_s3_3.test.ts
```diff
@@ -38,9 +36,6 @@ describe('headless s3 import', () => {
     const shortId = getShortId();
     bucketNameToImport = `${bucketPrefix}${shortId}`;
 
-    const credentials = new aws.SharedIniFileCredentials({ profile: profileName });
-    aws.config.credentials = credentials;
```
awsluja, line 42: this was the only test with this still left in, and is not needed since the aws creds for the test acc are already available. Perhaps this was left in from local testing or before the configure_tests was added

sobolk (approved):

rtpascual (approved):

PR #14074, opened 2025-01-16
fix: choose current env, update gitignore for gen2, choose function name from output prop

abhi7cr:
#### Description of changes

QA bug fixes/improvements:
- Choose current environment instead of listing all and asking customer to select one
- Update .gitignore to gen2 specific files
- Look for `Cognito` service name for grabbing the resource name since `Cognito-UserPool-Groups` can also co-exist
- Derive function names from `output` key instead of the resource name since in the case of triggers, they are different
- Remove `demandCommands` to surface top level handler errors

Local testing, unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-migration/src/app_functions_definition_fetcher.ts
+++ b/packages/amplify-migration/src/app_functions_definition_fetcher.ts
@@ -2,7 +2,7 @@ import assert from 'node:assert';
 import { FunctionDefinition } from '@aws-amplify/amplify-gen2-codegen';
 import { getFunctionDefinition } from '@aws-amplify/amplify-gen1-codegen-function-adapter';
 import { BackendEnvironmentResolver } from './backend_environment_selector';
-import { LambdaClient, GetFunctionCommand } from '@aws-sdk/client-lambda';
+import { GetFunctionCommand, LambdaClient } from '@aws-sdk/client-lambda';
 import { StateManager } from '@aws-amplify/amplify-cli-core';
 
 interface AuthConfig {
@@ -72,10 +72,9 @@ export class AppFunctionsDefinitionFetcher {
     });
 
     const getFunctionPromises = Object.keys(functions).map((key) => {
-      const functionName = key;
       return this.lambdaClient.send(
         new GetFunctionCommand({
-          FunctionName: functionName + '-' + backendEnvironment?.environmentName,
+          FunctionName: meta.function[key].output.Name,
         }),
       );
     });
--- a/packages/amplify-migration/src/backend_environment_selector.ts
+++ b/packages/amplify-migration/src/backend_environment_selector.ts
@@ -1,13 +1,7 @@
 import assert from 'node:assert';
 import { stdin as input, stdout as output } from 'node:process';
 import readline from 'node:readline/promises';
-import {
-  AmplifyClient,
-  BackendEnvironment,
-  GetBackendEnvironmentCommand,
-  GetBackendEnvironmentCommandOutput,
-  ListBackendEnvironmentsCommand,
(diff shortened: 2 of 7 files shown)


packages/amplify-migration/src/app_functions_definition_fetcher.ts
```diff
@@ -72,10 +72,9 @@ export class AppFunctionsDefinitionFetcher {
     });
 
     const getFunctionPromises = Object.keys(functions).map((key) => {
-      const functionName = key;
       return this.lambdaClient.send(
         new GetFunctionCommand({
-          FunctionName: functionName + '-' + backendEnvironment?.environmentName,
+          FunctionName: meta.function[key].output.Name,
```
Sanayshah2, line 77: Do we want to assert this before or is it safe to assume this will always be present?

packages/amplify-migration/src/backend_environment_selector.ts
```diff
@@ -1,37 +1,22 @@
...
+    const { backendEnvironment } = await this.amplifyClient.send(
+      new GetBackendEnvironmentCommand({
+        appId: this.appId,
+        environmentName: envInfo.envName,
+      }),
+    );
+    assert(backendEnvironment, 'No backend environment found');
+    this.selectedEnvironment = backendEnvironment;
+    return this.selectedEnvironment;
   };
```
Sanayshah2, line 21: Question: Why are we doing this change of not giving an option to the customer to select the backend env?

packages/amplify-migration/src/command-handlers.ts
```diff
@@ -199,21 +200,26 @@ const unsupportedCategories = (): Map<string, string> => {
 };
 
 export async function execute() {
-  const amplifyClient = new AmplifyClient();
+  let backendEnvironment: BackendEnvironment | undefined;
+  let backendEnvironmentResolver: BackendEnvironmentResolver | null = null;
```
Sanayshah2, line 204: Doesn't the assertion take care of `await backendEnvironmentResolver.selectBackendEnvironment()` returning null? Do we need this initialization?

packages/amplify-migration/src/app_functions_definition_fetcher.ts
abhi7cr: yeah it would always be present since in Gen1 they should have deployed the code to cloud :)

packages/amplify-migration/src/backend_environment_selector.ts
abhi7cr: Good question! It just makes it easier for the customer since they are already in the context of an environment.

packages/amplify-migration/src/command-handlers.ts
abhi7cr: This is needed since we moved the following call inside a callback function:

```
  await printer.indicateProgress(`Inspecting Amplify app ${appId} with current backend`, async () => {
    const amplifyClient = new AmplifyClient();
    backendEnvironmentResolver = new BackendEnvironmentResolver(appId, amplifyClient);
    backendEnvironment = await backendEnvironmentResolver.selectBackendEnvironment();
  });
```

Sanayshah2: Ahh got it, missed that.

packages/amplify-migration/src/backend_environment_selector.ts
Sanayshah2: Sounds good!

Sanayshah2 (approved):

PR #14093, opened 2025-02-01
fix(amplify-cli-core): use build script properly for overrides

brianlenz:
Updated the TypeScript compilation of overrides so that it doesn't require `node_modules/.bin/tsc`. Instead, it simply relies on the `build` script to execute `tsc`. This is more flexible and can support alternative setups w/ hoisting (e.g. via Yarn workspaces).

This is an override corollary fix to #11854, which is for custom resources.

This is an improvement over my previous PR in #13858 in that it works with any package manager by ensuring the `--project` and `tsconfig.json` files are passed through to the `tsc` script. The previous implementation didn't work with `npm` because it doesn't pass through additional args like `yarn` does. The fix was easy: simply separate the build run with `--` so that the remaining args are treated as positional for the `tsc` script.

I've confirmed the fix works with both `yarn` and `npm` 💪

#### Description of changes
Remove hard-coded dependency on `node_modules/.bin/tsc` for overrides to instead use the `build` script from `package.json`, which is more flexible.

Ran `yarn test` and all pre-commit hooks without issue.
Tested `amplify build --debug` and `amplify push` locally _with both `yarn` and `npm`_. The commands were failing with `yarn` prior to these changes, and it succeeds with these changes in place (via `amplify-dev`).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli-core/src/overrides-manager/override-skeleton-generator.ts
+++ b/packages/amplify-cli-core/src/overrides-manager/override-skeleton-generator.ts
@@ -93,17 +93,8 @@ export const buildOverrideDir = async (cwd: string, destDirPath: string): Promis
     const tsConfigSampleFilePath = path.join(__dirname, '..', '..', 'resources', 'overrides-resource', 'tsconfig.resource.json');
     fs.writeFileSync(tsConfigDestFilePath, fs.readFileSync(tsConfigSampleFilePath));
 
-    // get locally installed tsc executable
-
-    const localTscExecutablePath = path.join(cwd, 'node_modules', '.bin', 'tsc');
-
-    if (!fs.existsSync(localTscExecutablePath)) {
-      throw new AmplifyError('MissingOverridesInstallationRequirementsError', {
-        message: 'TypeScript executable not found.',
-        resolution: 'Please add it as a dev-dependency in the package.json file for this resource.',
-      });
-    }
-    execa.sync(localTscExecutablePath, [`--project`, `${tsConfigDestFilePath}`], {
+    // ensure the --project and specific tsconfig.json file get passed through to tsc by using --
+    execa.sync(packageManager.executable, [`run`, `build`, '--', `--project`, `${tsConfigDestFilePath}`], {
       cwd: tsConfigDir,
       stdio: 'pipe',
       encoding: 'utf-8',


brianlenz:
@sobolk it's working now with `npm`! Just a slight tweak to my last PR needed (using `--` to pass through the args to `tsc`). Tested and confirmed to be working.

It's worth noting that #11854 does _not_ have the same issue because it doesn't pass a custom `--project` arg. As such, #11854 should be safe to merge as-is for consistency.

If there's anything else I can help with, please let me know!

Thank you 🙏

sobolk:
> @sobolk it's working now with `npm`! Just a slight tweak to my last PR needed (using `--` to pass through the args to `tsc`). Tested and confirmed to be working.
> 
> It's worth noting that #11854 does _not_ have the same issue because it doesn't pass a custom `--project` arg. As such, #11854 should be safe to merge as-is for consistency.
> 
> If there's anything else I can help with, please let me know!
> 
> Thank you 🙏

@brianlenz I have slightly different solution in mind for this. Where we do not have dependency on `build` script but we do use package manager to locate `tsc`.

Package managers have a way to run executables (`.bin` scripts) from npm packages.
For example:
```
npx tsc <args> // note different first token here
yarn tsc <args>
pnpm tsc <args>
```

Therefore what I would suggest is to 
1. Add `packageManager.runner` to https://github.com/aws-amplify/amplify-cli/blob/4d368b3d587616b7cba3164b701c370429c66ae2/packages/amplify-cli-core/src/utils/packageManager.ts#L21 .
2. Have sub-classes provide `npx`, `yarn` or `pnpm` for that new property.
3. Use `packageManager.runner` when calling `tsc` (see examples above).
4. Test solution with npm yarn and pnpm.

brianlenz:
@sobolk thanks for sticking with me and the recommendations here 🙏 I like your suggestions, so I'll take a crack at it!

brianlenz:
@sobolk done! Check it out and let me know if you think it needs further tweaking 🙏  Tested with `npm`, `yarn`, and `pnpm`. Works great and seems like a much cleaner solution ✨

sobolk:
> @sobolk done! Check it out and let me know if you think it needs further tweaking 🙏 Tested with `npm`, `yarn`, and `pnpm`. Works great and seems like a much cleaner solution ✨

I'll test it on my end and I'll get back to you.

sobolk (approved):

rtpascual (approved):

PR #14108, opened 2025-02-13
fix: added dynamic reference to the env name in backend.ts and fixed a couple of bugs

bzsurbhi:
#### Description of changes

- During `prepare` phase, add an environment variable to customers Amplify Hosting build for their gen 1 backend name - https://app.asana.com/0/1206146466379945/1209214806262350
- Throw new error in function resource.ts - https://app.asana.com/0/1209297236712871/1209207544854343
- Fix the identity provider type for LoginWithAmazon - https://app.asana.com/0/1209297236712871/1209313621506855
- Fix the storage access groups

Unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-gen1-codegen-storage-adapter/src/storage_access.test.ts
+++ b/packages/amplify-gen1-codegen-storage-adapter/src/storage_access.test.ts
@@ -43,6 +43,13 @@ void describe('getStorageAccess', () => {
           assert.deepEqual(access?.groups?.[group].sort(), gen2[group].sort());
         }
       });
+
+      void it('returns empty group permissions', () => {
+        const input = getCLIInput();
+        input.groupAccess = {};
+        const access = getStorageAccess(input);
+        assert.deepEqual(access?.groups, undefined);
+      });
     }
   });
   void describe('auth and unauth', () => {
--- a/packages/amplify-gen1-codegen-storage-adapter/src/storage_access.test.ts
+++ b/packages/amplify-gen1-codegen-storage-adapter/src/storage_access.test.ts
@@ -48,6 +48,7 @@ void describe('getStorageAccess', () => {
         const input = getCLIInput();
         input.groupAccess = {};
         const access = getStorageAccess(input);
+        assert.notEqual(access, undefined);
         assert.deepEqual(access?.groups, undefined);
       });
     }
--- a/packages/amplify-gen2-codegen/src/backend/synthesizer.ts
+++ b/packages/amplify-gen2-codegen/src/backend/synthesizer.ts
@@ -48,6 +48,8 @@ export interface BackendRenderParameters {
   unsupportedCategories?: Map<string, string>;
 }
 
+const amplifyGen1EnvName = 'AMPLIFY_GEN_1_ENV_NAME';
+
 export class BackendSynthesizer {
   private importDurationFlag = false;
   private oAuthFlag = false;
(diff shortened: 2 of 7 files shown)


packages/amplify-gen2-codegen/src/backend/synthesizer.ts
```diff
@@ -443,6 +448,17 @@ export class BackendSynthesizer {
     ]);
   }
 
+  private createTemplateLiteralExpression(id1: string, id2: string, templateHead: string) {
```
abhi7cr, line 451: nit: not blocking, but can we add a comment showing how the statement will look after codegen ?

packages/amplify-gen2-codegen/src/function/source_builder.ts
```diff
@@ -24,13 +24,12 @@ export function renderFunctions(definition: FunctionDefinition, appId?: string,
...
   groupsComment.push(
-    factory.createJSDocComment(
-      factory.createNodeArray([
-        factory.createJSDocText(
-          `Source code for this function can be found in your Amplify Gen 1 Directory.\nSee .amplify/migration/amplify/backend/function/${definition.resourceName}/src \n`,
-        ),
-      ]),
-    ),
+    factory.createCallExpression(factory.createIdentifier('throw new Error'), undefined, [
+      // eslint-disable-next-line spellcheck/spell-checker
```
abhi7cr, line 28: why is the spell checker disabled? Is it because we removed the new line char?

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
```diff
@@ -443,6 +448,17 @@ export class BackendSynthesizer {
...
+    return factory.createExpressionStatement(
+      factory.createAssignment(
+        factory.createPropertyAccessExpression(factory.createIdentifier(id1), factory.createIdentifier(id2)),
+        factory.createTemplateExpression(factory.createTemplateHead(templateHead), [
+          factory.createTemplateSpan(factory.createIdentifier('AMPLIFY_GEN_1_ENV_NAME'), factory.createTemplateTail('')),
```
abhi7cr, line 456: nit: move it to a constant?

packages/amplify-gen1-codegen-storage-adapter/src/storage_access.test.ts
```diff
@@ -43,6 +43,13 @@ void describe('getStorageAccess', () => {
           assert.deepEqual(access?.groups?.[group].sort(), gen2[group].sort());
         }
       });
+
+      void it('returns empty group permissions', () => {
+        const input = getCLIInput();
+        input.groupAccess = {};
+        const access = getStorageAccess(input);
+        assert.deepEqual(access?.groups, undefined);
```
abhi7cr, line 51: nit: can we add an assert for making sure `access` is not undefined?

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
abhi7cr: nit: since its specific to setting the branch env variable, can we rename this to something like `createTemplateLiteralExpressionWithBranchEnv` ?

bzsurbhi: We can make it generic in future, if needed, just by extracting out the 'AMPLIFY_GEN_1_ENV_NAME' identifier. That's why kept it generic

abhi7cr: ah got it, makes sense to keep it generic. In that case, should we just combine the env with user pool name and pass it together in the same param from the caller ? That way, we can let consumers pass what they want.

bzsurbhi: No, we cannot pass it together with the userPoolName
// factory.createTemplateHead(templateHead) - this part creates the userPoolName (anything before ${AMPLIFY_GEN_1_ENV_NAME})
// factory.createTemplateSpan(factory.createIdentifier(amplifyGen1EnvName) - this creates ${AMPLIFY_GEN_1_ENV_NAME}
// factory.createTemplateTail('') - this creates anything after ${AMPLIFY_GEN_1_ENV_NAME}

abhi7cr (approved):

PR #14123, opened 2025-02-24
chore: copy schema in data/resource.ts, add error message if env variable is not set

bzsurbhi:
#### Description of changes
- Copy schema from schema.graphql file to data/resource.ts file. In case of split schema, merge the schemas and paste it in data/resource.ts https://app.asana.com/0/1209417997044572/1209380394918346
- Throw an error message in backend.ts if the env variable AMPLIFY_GEN_1_ENV_NAME is not set by the customer
- Strip the -${env} suffix from strings set on name properties in defineThing - https://app.asana.com/0/1209417997044572/1209390080804200

Unit tests

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-gen2-codegen/src/backend/synthesizer.ts
```diff
@@ -557,13 +558,19 @@ export class BackendSynthesizer {
       factory.createVariableDeclarationList([backendVariable], ts.NodeFlags.Const),
     );
 
+    let envVariableIdentifier = 'process.env.AMPLIFY_GEN_1_ENV_NAME';
+    if (ci.isCI && !process.env.AMPLIFY_GEN_1_ENV_NAME) {
```
abhi7cr, line 562: should the CI check be part of the codegen so that it can be evaluated at runtime in `backend.ts` file ?

packages/amplify-migration/src/data_definition_fetcher.ts
```diff
@@ -1,11 +1,63 @@
...
+
+      assert(apiName);
+
+      const rootDir = pathManager.findProjectRoot();
+      assert(rootDir);
+      const apiPath = path.join(rootDir, 'amplify', 'backend', 'api', apiName);
+
+      // Check for schema folder first
+      const schemaFolderPath = path.join(apiPath, 'schema');
+      if (fs.existsSync(schemaFolderPath) && fs.statSync(schemaFolderPath).isDirectory()) {
```
abhi7cr, line 36: nit: can we use `node:fs/promises` api instead of `fs-extra` ?

packages/amplify-gen2-codegen/package.json
```diff
@@ -45,6 +45,7 @@
     "@aws-sdk/client-lambda": "^3.651.1",
     "@aws-sdk/client-s3": "^3.651.1",
     "aws-cdk-lib": "^2.127.0",
+    "ci-info": "^3.8.0",
```
abhi7cr, line 48: We don't need this dependency now right ?

abhi7cr: LGTM overall. Just had one minor comment about removing `ci-info` build time dependency since we moved it to runtime.

abhi7cr (approved):

PR #14145, opened 2025-03-21
Gen2 migrations QA fixes

abhi7cr:
#### Description of changes

- Skip `PreTokenGenerationConfig` key from Cognito during `prepare`
- Add new line to separate import, body and export statements in codegen
- Wrap `throw new Error` call expression in a statement. This adds a semicolon at the end
- Add `@types/node` to fix `process.env` type errors
- Resolve CFN parameters in Gen2 stack during revert
- Surface errors during  Gen2 codegen in `prepare` command
- Add support for `--debug` flag for `prepare` command to output the stack trace

local testing with `prepare`, `execute` and `revert` commands

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-gen2-codegen/src/backend/synthesizer.ts
+++ b/packages/amplify-gen2-codegen/src/backend/synthesizer.ts
@@ -855,7 +855,7 @@ export class BackendSynthesizer {
     }
 
     return factory.createNodeArray(
-      [...imports, newLineIdentifier, ...errors, newLineIdentifier, backendStatement, newLineIdentifier, ...nodes],
+      [...imports, newLineIdentifier, ...errors, newLineIdentifier, backendStatement, ...nodes],
       true,
     );
   }
--- a/packages/amplify-gen2-codegen/src/backend/synthesizer.ts
+++ b/packages/amplify-gen2-codegen/src/backend/synthesizer.ts
@@ -14,6 +14,7 @@ import { BucketAccelerateStatus, BucketVersioningStatus } from '@aws-sdk/client-
 import { AccessPatterns, ServerSideEncryptionConfiguration } from '../storage/source_builder.js';
 import { ExplicitAuthFlowsType, OAuthFlowType, UserPoolClientType } from '@aws-sdk/client-cognito-identity-provider';
 import assert from 'assert';
+import { newLineIdentifier } from '../ts_factory_utils';
 
 const factory = ts.factory;
 export interface BackendRenderParameters {
@@ -570,10 +571,8 @@ export class BackendSynthesizer {
     }
 
     if (renderArgs.function) {
-      const functionIdentifiers: Identifier[] = [];
       const functionNameCategories = renderArgs.function.functionNamesAndCategories;
       for (const [functionName, category] of functionNameCategories) {
-        functionIdentifiers.push(factory.createIdentifier(functionName));
         const functionProperty = factory.createShorthandPropertyAssignment(factory.createIdentifier(functionName));
         defineBackendProperties.push(functionProperty);
         imports.push(this.createImportStatement([factory.createIdentifier(functionName)], `./${category}/${functionName}/resource`));
@@ -855,6 +854,9 @@ export class BackendSynthesizer {
       nodes.push(tagAssignment);
     }
 
-    return factory.createNodeArray([...imports, ...errors, backendStatement, ...nodes], true);
+    return factory.createNodeArray(
+      [...imports, newLineIdentifier, ...errors, newLineIdentifier, backendStatement, newLineIdentifier, ...nodes],
+      true,
+    );
   }
 }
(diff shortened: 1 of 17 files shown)


packages/amplify-gen2-codegen/src/index.ts
```diff
@@ -142,9 +143,7 @@ export const createGen2Renderer = ({
           new TypescriptNodeArrayRenderer(
             async () => renderFunctions(func),
             (content) => {
-              return fileWriter(content, path.join(dirPath, 'resource.ts'))
-                .then(() => fileWriter('', path.join(dirPath, 'handler.ts')))
-                .catch(console.error);
```
abhi7cr, line 147: This was in the wrong place. It was intended to catch codegen errors but this is while writing code to the files.

packages/amplify-migration/src/error_handler.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const message = 'Unauthorized';
+    const error = new Error('Unauthorized');
+    expect.assertions(4);
+    try {
+      await errorHandler(message, error, true);
+    } catch (e) {
+      expect(e).toBeInstanceOf(Error);
+      expect(printSpy).toBeCalledTimes(2);
+      expect(printSpy).toHaveBeenNthCalledWith(1, message);
+      expect(printSpy.mock.calls[1][0]).toContain('at Promise');
```
abhi7cr, line 69: `at <Method call>` is used  to detect stack trace.

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
```diff
@@ -855,6 +854,6 @@ export class BackendSynthesizer {
       nodes.push(tagAssignment);
     }
 
-    return factory.createNodeArray([...imports, ...errors, backendStatement, ...nodes], true);
+    return factory.createNodeArray([...imports, newLineIdentifier, ...errors, newLineIdentifier, backendStatement, ...nodes], true);
```
surbhibansal, line 857: Nice! We should do it in category specific resource.ts files too (in next CR)

packages/amplify-migration/src/error_handler.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    try {
+      await errorHandler(message, error);
+    } catch (e) {
+      expect(e).toBeInstanceOf(Error);
+      expect(printSpy).toBeCalledTimes(1);
+      expect(printSpy).toBeCalledWith(message);
+    }
+  });
+
+  it('should print error with debug flag', async () => {
```
surbhibansal, line 58: nit: debug `true` right?

packages/amplify-migration/src/error_handler.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  beforeAll(() => {
+    const parser = createMainParser(version);
+    errorHandler = generateCommandFailureHandler(parser);
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('should print error without debug flag', async () => {
```
surbhibansal, line 44: with debug false?

packages/amplify-gen2-codegen/src/backend/synthesizer.ts
surbhibansal: nvm, see you have it already

packages/amplify-gen2-codegen/src/function/source_builder.ts
```diff
@@ -39,11 +39,13 @@ export function renderFunctions(definition: FunctionDefinition, appId?: string,
   namedImports['@aws-amplify/backend'].add('defineFunction');
 
   postImportStatements.push(
-    factory.createCallExpression(factory.createIdentifier('throw new Error'), undefined, [
-      factory.createStringLiteral(
-        `Source code for this function can be found in your Amplify Gen 1 Directory. See .amplify/migration/amplify/backend/function/${definition.resourceName}/src`,
-      ),
-    ]),
+    factory.createExpressionStatement(
```
surbhibansal, line 42: We should also wrap other `throw new Error` in createExpressionStatement(). We've some instances in backend.ts and also auth/resource.ts

packages/amplify-migration/src/error_handler.test.ts
abhi7cr: Good question. Its not a boolean flag, the presence of `--debug` indicates its true and absence is false.

packages/amplify-gen2-codegen/src/function/source_builder.ts
abhi7cr: Good call out. Will do it other places too.

surbhibansal (approved):

PR #14153, opened 2025-04-01
fix: upgrade most packages that use `rimraf` and `glob`

ShadowCat567:
#### Description of changes

Currently when you install `@amplify/cli`, this message appears:
```
npm warn deprecated inflight@1.0.6: This module is not supported, and leaks memory. Do not use it. Check out lru-cache if you want a good and tested way to coalesce async requests by a key value, which is much more comprehensive and powerful.
npm warn deprecated rimraf@3.0.2: Rimraf versions prior to v4 are no longer supported
npm warn deprecated glob@7.2.3: Glob versions prior to v9 are no longer supported
```
This PR is an attempt to upgrade as many packages as possible that use the deprecated version of these dependencies as well as upgrade the versions of `rimraf` and `glob` that are used in our repository (`inflight` is a dependency of versions of `glob` before v9, upgrading to v9 or higher resolves the `inflight` problem).
Many packages were out of date and using deprecated versions of `rimraf` and `glob` that had memory leaks. This PR fixes the vast majority of the outdated packages.
Some notes about `glob` and `rimraf`:
- `glob` and `rimraf` not longer have default exports, since we were using `rimraf.sync` and `glob.sync`, I switched us to their named export equivalents
- apparently `-f/-r/-rf` has never done anything, it seems people would add them because `rm` takes these arguments

Moved some of the packages that cannot be updated to devDependencies. 
Most remaining outdated packages that have not been moved to devDependencies come from data repositories, these will be addressed in a separate PR.

Ensured that the project still builds and the tests still work. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


Amplifiyer:
Do we need this PR?

ShadowCat567:
> Do we need this PR?

Additionally if this PR is deemed unnecessary, is it worth the effort to do similar updates in the Gen2 CLI and Data repositories?

package.json
```diff
@@ -51,8 +51,8 @@
     "tagged-release-without-e2e-local": "bash -c 'source ./scripts/cloud-release.sh && TaggedRCLocal'",
     "tagged-release-without-e2e-beta": "bash -c 'source ./scripts/cloud-release.sh && TaggedRCBeta'",
     "tagged-release-without-e2e-prod": "bash -c 'source ./scripts/cloud-release.sh && TaggedRCProd'",
-    "rm-aa-dev-link": "rimraf -f \".bin/amplify-app-dev\"",
-    "rm-dev-link": "rimraf -f \".bin/amplify-dev\"",
+    "rm-aa-dev-link": "rimraf \".bin/amplify-app-dev\"",
+    "rm-dev-link": "rimraf \".bin/amplify-dev\"",
```
ShadowCat567, line 55: https://github.com/isaacs/rimraf/issues/253

packages/amplify-cli/src/initialize-env.ts
```diff
@@ -91,7 +91,8 @@ export const initializeEnv = async (
     const initializationTasks: (() => Promise<$TSAny>)[] = [];
     const providerPushTasks: (() => Promise<$TSAny>)[] = [];
 
-    for (const provider of context.exeInfo?.projectConfig?.providers) {
+    // eslint-disable-next-line no-unsafe-optional-chaining
+    for (const provider of context.exeInfo!.projectConfig?.providers) {
```
Amplifiyer, line 95: Is the force coercion of `exeInfo` intentional here?

packages/amplify-e2e-tests/src/__tests__/custom_policies_function.test.ts
```diff
@@ -53,7 +53,7 @@ it('should init and deploy storage DynamoDB + Lambda trigger, attach custom poli
   );
 
   const meta = getProjectMeta(projRoot);
-  const { Region: region } = meta?.providers?.awscloudformation;
+  const { Region: region } = meta!.providers!.awscloudformation;
```
Amplifiyer, line 56: similarly here, is the coercion intentional?

Amplifiyer (commented):
Looks good overall.

packages/amplify-cli/src/initialize-env.ts
ShadowCat567: The coercions of `exeinfo` (and a few other things) came from lint errors I was experiencing while working on this PR. I can revert them and see if the lint errors still occur, if they do would it be preferable to add an ignore to them?

Amplifiyer: I'd recommend defensive approach here. Though I'm not sure why eslint would complaint about multiple optional chaining here.

Amplifiyer (approved):

sobolk (approved):

PR #14156, opened 2025-04-02
fix: codegen custom resources and added function scheduling

bzsurbhi:
#### Description of changes

**Codegen custom resource**

Copy custom/ folder from gen1 -> gen2
Update cdk-stack.ts file to remove @aws-amplify/cli-extensibility-helper dependency
-- Check for existence of AmplifyHelpers.addResourceDependency and throw an error if found (errorMessage to be updated)
-- Replace AmplifyHelpers.getProjectInfo() with {envName: 'envName', projectName: 'projectName'}
-- Replace AmplifyHelpers.AmplifyResourceProps with {category: 'custom', resourceName: resource}
-- Remove the import statement import * as AmplifyHelpers from '@aws-amplify/cli-extensibility-helper';

**Add scheduling in functions**
#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-gen2-codegen/src/function/source_builder.ts
+++ b/packages/amplify-gen2-codegen/src/function/source_builder.ts
@@ -159,7 +159,9 @@ export function createFunctionDefinition(
   if (definition?.schedule) {
     const rawScheduleExpression = definition.schedule;
     let scheduleExpression: string | undefined;
-    const scheduleValue = rawScheduleExpression.match(/\(([^)]+)\)/)?.[1];
+    const startIndex = rawScheduleExpression.indexOf('(') + 1;
+    const endIndex = rawScheduleExpression.lastIndexOf(')');
+    const scheduleValue = startIndex > 0 && endIndex > startIndex ? rawScheduleExpression.slice(startIndex, endIndex) : undefined;
     if (rawScheduleExpression?.startsWith('rate(')) {
       // Convert rate expression to a more readable format
       const rateValue = scheduleValue;
--- a/packages/amplify-gen2-codegen/src/function/source_builder.ts
+++ b/packages/amplify-gen2-codegen/src/function/source_builder.ts
@@ -175,7 +175,7 @@ export function createFunctionDefinition(
           day: 'd',
           days: 'd',
         };
-        scheduleExpression = `every ${value}${unitMap[unit] || unit}`;
+        scheduleExpression = `every ${value}${unitMap[unit]}`;
       }
     } else if (rawScheduleExpression?.startsWith('cron(')) {
       // Extract the cron expression as-is
--- a/packages/amplify-gen2-codegen/src/function/source_builder.ts
+++ b/packages/amplify-gen2-codegen/src/function/source_builder.ts
@@ -12,6 +12,7 @@ export interface FunctionDefinition {
   environment?: EnvironmentResponse;
   runtime?: Runtime | string;
   resourceName?: string;
+  schedule?: string;
 }
 
 const factory = ts.factory;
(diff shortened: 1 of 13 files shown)


packages/amplify-migration/src/command-handler.test.ts
```diff
@@ -1,15 +1,38 @@
...
 jest.mock('node:fs/promises', () => ({
   access: jest.fn(),
   readFile: jest.fn(),
   writeFile: jest.fn(),
+  mkdir: jest.fn(),
+  cp: jest.fn(),
 }));
 
-jest.mock('@aws-amplify/amplify-cli-core');
+// jest.mock('@aws-amplify/amplify-cli-core');
```
abhi7cr, line 16: nit: remove comment

packages/amplify-migration/src/command-handlers.ts
```diff
@@ -282,6 +286,102 @@ async function updateGitIgnoreForGen2() {
...
+    });
+
+    const sourceTypesPath = path.join(amplifyGen1BackendDir, TYPES_DIR);
+    const destinationTypesPath = path.join(amplifyGen2Dir, TYPES_DIR);
+    await fs.mkdir(destinationTypesPath, { recursive: true });
+    await fs.cp(sourceTypesPath, destinationTypesPath, { recursive: true });
+
+    await updateCdkStackFile(customResources, destinationCustomResourcePath, rootDir);
+  } else {
+    console.log(`Custom resource category does not exist in the meta`);
```
abhi7cr, line 324: I don't think we should log this if we are skipping it ?

packages/amplify-migration/src/command-handlers.ts
```diff
@@ -282,6 +286,102 @@ async function updateGitIgnoreForGen2() {
...
+        /AmplifyHelpers\.AmplifyResourceProps/g,
+        `{category: 'custom', resourceName: '${resource}' }`,
+      );
+
+      // Remove the import statement for AmplifyHelpers
+      cdkStackContent = cdkStackContent.replace(amplifyHelpersImport, '');
+
+      await fs.writeFile(cdkStackFilePath, cdkStackContent, { encoding: 'utf-8' });
+    } catch (error) {
+      console.error(`Error updating ${cdkStackFilePath}: ${error.message}`);
```
abhi7cr, line 360: we should throw error here else it might be in an inconsistent state.

packages/amplify-gen2-codegen/src/function/source_builder.ts
```diff
@@ -155,5 +156,34 @@
     defineFunctionProperties.push(createParameter('runtime', factory.createNumericLiteral(nodeRuntime)));
   }
 
+  if (definition?.schedule) {
+    const rawScheduleExpression = definition.schedule;
+    let scheduleExpression: string | undefined;
+    const scheduleValue = rawScheduleExpression.match(/\(([^)]+)\)/)?.[1];
```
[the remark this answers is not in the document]
abhi7cr: will it be faster to have separate regex for rate and cron ?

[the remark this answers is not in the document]
bzsurbhi: The format for rate schedule - `rate(5 minutes)`
The format for cron schedule - `cron(12 20 * 1 ? *)`
The regex is just getting the values between the parentheses in both formats. I think it's ok to keep the common regex. The error is thrown in case the input data is very large which won't be the case here. I can try to simplify the regex anyways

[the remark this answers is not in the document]
abhi7cr: ah makes sense. Thanks!

abhi7cr (approved):

packages/amplify-gen2-codegen/src/function/source_builder.ts
```diff
@@ -155,5 +156,36 @@ export function createFunctionDefinition(
...
+        const [value, unit] = rateValue.split(' ');
+        const unitMap: Record<string, string> = {
+          minute: 'm',
+          minutes: 'm',
+          hour: 'h',
+          hours: 'h',
+          day: 'd',
+          days: 'd',
+        };
+        scheduleExpression = `every ${value}${unitMap[unit] || unit}`;
```
abhi7cr, line 178: nit: Will there be a case when `unitMap[unit]` would be `undefined` and fallback to `unit` ?

bzsurbhi: Probably not, I'll remove it

abhi7cr (approved):

PR #14199, opened 2025-05-06
Dismantling global config object from SDK V2

ShadowCat567:
#### Description of changes

This is in preparation of SDK V2 to V3 migrations changes (nothing has actually be migrated yet). 
SDK V2 reaches end of life on September 8, 2025 and we must upgrade SDK to V3 before that happens: https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-javascript-v2/
SDK V2 has a global configuration object that can be used to store a configuration which is automatically loaded into all of the service objects that are created unless the service object specifies an alternate configuration when it is created. 
This global configuration object does not exist in SDK V3 so this PR works to move all the work that used to be done by the global config object to each of the individual service objects. 

The threshold for the size of all the binaries has also temporarily been raised. 

Global Config must remain around for some Aurora DB utilities on the Data side, can be completely removed when the PR on the data side is merged and released.
Sibling PR in the data repo handles dismantling of the global config that data inherited from CLI: https://github.com/aws-amplify/amplify-category-api/pull/3310

Did a combination of manual testing (ensured that proxies and credentials with Access Keys and Profiles works as well as add, push, and delete commands) and updating the current tests to ensure that they could work without the global sdk config.
Ensured that SSO credentials workaround still works: 
image

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-category-notifications/src/auth-helper.ts
```diff
@@ -150,8 +152,16 @@ const getIamClient = async (context: $TSContext, action: string | undefined): Pr
   const providerPlugins = context.amplify.getProviderPlugins(context);
   // eslint-disable-next-line import/no-dynamic-require, global-require, @typescript-eslint/no-var-requires
   const provider = require(providerPlugins[providerName]);
-  const aws = await provider.getConfiguredAWSClient(context, AmplifyCategories.NOTIFICATIONS, action);
-  return new aws.IAM();
+  let httpAgent = undefined;
+  const httpProxy = process.env.HTTP_PROXY || process.env.HTTPS_PROXY;
+  if (httpProxy) {
+    httpAgent = new ProxyAgent();
+  }
```
sobolk: Should this be inside `getConfiguredAWSClient` ?
If proxy is configured I would expect that we want to add it to config for all clients.

packages/amplify-category-notifications/src/pinpoint-helper.ts
```diff
@@ -543,17 +538,14 @@ export const getPinpointClient = async (
...
+  let httpAgent = undefined;
 
   // HTTP_PROXY & HTTPS_PROXY env vars are read automatically by ProxyAgent, but we check to see if they are set before using the proxy
   if (httpProxy) {
-    aws.config.update({
-      httpOptions: {
-        agent: new ProxyAgent(),
-      },
-    });
+    httpAgent = new ProxyAgent();
```
sobolk, line 545: same here

packages/amplify-console-hosting/src/utils/client-factory.js
```diff
@@ -1,13 +1,14 @@
 const constants = require('../constants/plugin-constants');
+const AWS = require('aws-sdk');
 
 async function getAmplifyClient(context) {
-  const AWS = await getAWSClient(context);
-  return new AWS.Amplify();
+  const config = await getAWSClient(context);
```
sobolk: It seems we should rename `getAWSClient` to `getAWSClientConfig`

packages/amplify-container-hosting/src/lib/ElasticContainer/file-uploader.js
```diff
@@ -1,12 +1,13 @@
...
 const mime = require('mime-types');
 const constants = require('../constants');
+const S3 = require('aws-sdk/clients/s3');
 
 export async function getS3Client(context, action) {
   const providerPlugins = context.amplify.getProviderPlugins(context);
   const provider = require(providerPlugins[constants.providerName]);
-  const aws = await provider.getConfiguredAWSClient(context, constants.CategoryName, action);
-  return new aws.S3();
+  const config = await provider.getConfiguredAWSClient(context, constants.CategoryName, action);
```
sobolk: similarly here `provider.getAWSClientConfig`.

packages/amplify-provider-awscloudformation/API.md
```diff
@@ -50,13 +50,11 @@ export const getLocationRegionMapping: () => $TSObject;
 // @public (undocumented)
 export const getLocationSupportedRegion: (region: string) => string;
 
-// Warning: (ae-forgotten-export) The symbol "AwsSecrets" needs to be exported by the entry point index.d.ts
+// Warning: (ae-forgotten-export) The symbol "AwsSdkConfig" needs to be exported by the entry point index.d.ts
```
sobolk: might be good to address this warning while we're touching this code.

packages/amplify-container-hosting/src/lib/ElasticContainer/file-uploader.js
ShadowCat567: renaming this to `provider.getConfiguredAWSClientConfig` since I found `provider.getAWSClientConfig` a little confusing when used in combination with `getAWSClientConfig` in `amplify-console-hosting/.../client-factory.js`

packages/amplify-category-notifications/src/auth-helper.ts
ShadowCat567: taking another look at what is going on in `getConfiguredAWSClient`, looks like I already moved proxy stuff into there and I should be able to remove this

packages/amplify-category-notifications/src/pinpoint-helper.ts
ShadowCat567: what's happening here is not quite the same as `amplify-category-notifications/.../auth-helper.ts` since creds are coming from `getConfiguredCredentials`, which seems to just return region

sobolk (approved):

PR #14236, opened 2025-07-09
refactor: e2es and anything else that remains

ShadowCat567:
#### Description of changes

**Migrating:**
 - amplify-e2e-core
 - amplify-e2e-tests
 - amplify-category-hosting
 - amplfiy-container-hosting
 - any remaining infrastructure that uses AWS SDK V2
 
**Removed 2 files:**
- `packages/amplify-e2e-tests/functions/dynamodb-scan-v2.js` -- tested the Scan Command in DynamoDB in AWS SDK V2. This is not being used by any files, removed it
- `packages/amplify-provider-awscloudformation/src/aws-utils/aws.js` -- this used to be the heart of the AWS SDK V2 configuration in the `amplify-provider-awscloudformation` package. Since everything has been migrated to V2, nothing uses this any more. 

**Changes to CodeBuild Clean up Script:**
Removed functionality to check whether a Codebuild job is currently running before deleting its resources. The overhead of getting information about Codebuild jobs to see whether they were completed or not was consuming too much memory ever since the migration to V3. 
The Codebuild spec for our e2e tests already ensures that the clean up resources job does not run until all the other jobs have been completed and the cleanup script already skips cleaning up stacks that were created in the last 2 hours, so I think we are safe to remove this functionality for better memory usage. 

Had to downgrade pinpoint-client back to an older version since the newest versions seem to break some of our notifications e2e tests. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-e2e-tests/src/cleanup-codebuild-resources.ts
ShadowCat567, line 1: if the changes in the last commit don't work, revert it but keep the garbage collection thing it added and up the node space memory

packages/amplify-e2e-tests/src/__tests__/import_s3_3.test.ts
```diff
@@ -65,14 +66,14 @@ describe('headless s3 import', () => {
   });
 
   afterAll(async () => {
-    // Delete bucket
-    const s3 = new aws.S3();
+    // Delete bucket- in SDK V3, buckets are very picky about the region their client is in
```
svidgen, line 69: They sure are. 😒

packages/amplify-e2e-tests/src/cleanup-codebuild-resources.ts
```diff
@@ -1060,27 +1081,36 @@ const cleanupAccount = async (account: AWSAccountInfo, accountIndex: number, fil
   await deleteResources(account, accountIndex, staleResources);
   await deleteOrphanedOidcProviders(account);
   console.log(`[ACCOUNT ${accountIndex}] Cleanup done!`);
+
+  if (global.gc) {
+    global.gc();
+  }
```
svidgen: Gross. This solves the memory issue? Is the memory issue a time bomb? Does this just help a little when account resource cleanup gets kind of "backed up?" Or, are we bumping up against memory limits on every cleanup run? ... Any hints as to why GC isn't being engaged sooner on its own? What are the heavy things that make this required!?

svidgen (dismissed):

packages/amplify-e2e-tests/src/cleanup-codebuild-resources.ts
ShadowCat567: This may not be required anymore since I removed the check to see if a Codebuild job was is running before trying to delete a stack, since that apparently took up a lot of memory. But right after I migrated the cleanup script to V3, I immediately started hitting memory limits every it ran (you can check some of the failing test runs in this PR)

PR #14238, opened 2025-07-15
refactor: provider-cloudformation and amplify-category-storage

ShadowCat567:
#### Description of changes

**Migrating:**
- amplify-category-storage
- amplify-util-import (completing migration)
- amplify-provider-awscloudformation
- amplify-console-hosting

Changes the way credentials are handled to use the V3-style object instead of the V2-style one.
Updates `@types/node` to at least Node 20 in all of the places where outdated versions of Node were being used for `@types/node` -- the entire repo was upgraded to use Node 22 several months ago, not sure why these were overlooked then.

**Additional Testing- Proxies**
Tested an app with storage, auth, a lambda function, REST API, and graphQL API to ensure that all of them deploy and update properly through a proxy and that updates don't happen when a proxy is turned off (if they are expecting to go through a proxy). Deployment worked as expected. 

**Additional Testing - Iterative Deployments**
Tested making updates that triggered iterative deployments (added and removed indexes and primary keys) and made an update that allow me to test iterative rollback. It does not seem like the migration introduced new issues related to iterative deployments/rollbacks. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -365,104 +380,104 @@ class CloudFormation {
...
-                  self.readStackEvents(stackName);
-
-                  const cfnCompleteStatus = 'stackUpdateComplete';
-                  if (updateErr) {
-                    if (self.pollForEvents) {
-                      clearTimeout(self.pollForEvents);
-                    }
-                    return reject(updateErr);
-                  }
-                  cfnModel.waitFor(cfnCompleteStatus, cfnStackCheckParams, async (completeErr) => {
```
ShadowCat567, line 421: need some additional feedback on this section

packages/amplify-e2e-core/src/init/amplifyPull.ts
```diff
@@ -47,6 +47,8 @@ export const amplifyPull = (
     args.push(providerJson);
   }
 
+  args.push('--debug');
+
```
sobolk: Here and in other places.
I'm assuming this will get removed in final version of PR.
In general we should be running tests without it. To make sure that command without it works as expected (there's always small risk that debug flag might alter logic - if that happens it would be unwanted regression).

For test debugging the alternative is to set env var mentioned here https://github.com/aws-amplify/amplify-cli/blob/4ce6d0c0dc1ec5daad855a0f67855883daba91fd/packages/amplify-prompts/src/flags.ts#L4 .
in the workflow definition or codebuild build. (there's probably no automation/runbook for it yet, but should be possible).

packages/amplify-provider-awscloudformation/src/aws-utils/aws-lambda.ts
```diff
@@ -70,7 +79,7 @@ export class Lambda {
     }
 
     try {
-      await Promise.all(deletionPromises);
+      await Promise.all(deletionPromises.map((fn) => fn()));
```
sobolk: is this `.map` necessary?

packages/amplify-provider-awscloudformation/src/initialize-env.ts
```diff
@@ -13,6 +14,14 @@ import { generateDependentResourcesType } from '@aws-amplify/amplify-category-cu
...
 const logger = fileLogger('initialize-env');
 
+const streamToString = async (stream: Readable) => {
+  const chunks = [];
+  for await (const chunk of stream) {
+    chunks.push(Buffer.from(chunk));
+  }
+  return Buffer.concat(chunks).toString('utf-8');
+};
+
```
sobolk: There is `import { text } from 'node:stream/consumers';` that could be used for this.

packages/amplify-e2e-core/src/init/amplifyPull.ts
ShadowCat567: yep, if you see any `--debug` or `console.log` statements, those are relics from testing

packages/amplify-provider-awscloudformation/src/aws-utils/aws-lambda.ts
ShadowCat567: maybe not, I'll check

packages/amplify-provider-awscloudformation/src/initialize-env.ts
ShadowCat567: `'node:stream/consumers'` only works with node versions 16+, we have some packages that have `@types/node` set to 12, which seem to be blocking us from using this

sobolk: you can try upgrading `@types/node`. The runtimes we're using now are 20+.

packages/amplify-provider-awscloudformation/src/initialize-env.ts
```diff
@@ -3,16 +3,26 @@
...
 import { buildOverridesEnabledResources } from './build-override-enabled-resources';
 import { S3BackendZipFileName } from './constants';
 import { fileLogger } from './utils/aws-logger';
 import { downloadZip, extractZip } from './zip-util';
 import { generateDependentResourcesType } from '@aws-amplify/amplify-category-custom';
+import { text } from 'node:stream/consumers';
 
 const logger = fileLogger('initialize-env');
 
+const streamToString = async (stream: Readable) => {
```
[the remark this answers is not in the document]
ShadowCat567: will remove this

packages/amplify-util-uibuilder/package.json
```diff
@@ -14,7 +14,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@aws-amplify/amplify-category-api": "^5.15.0",
+    "@aws-amplify/amplify-category-api": "5.15.1-sdk-0-ssmSend.0",
```
svidgen: Tagged release still required here?

ShadowCat567: Not for much longer -> https://github.com/aws-amplify/amplify-cli/pull/14263
Once that is merged, I'll also go through and resolve the conflicts

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -76,44 +93,48 @@ class CloudFormation {
...
     const { context } = this;
-    const cfnCompleteStatus = 'stackCreateComplete';
     const cfnStackCheckParams = {
       StackName: cfnParentStackParams.StackName,
     };
     const self = this;
     self.eventStartTime = new Date();
 
-    return new Promise((resolve, reject) => {
+    return new Promise(async (resolve, reject) => {
```
svidgen: This bit here looks fishy. Can we skip the extra `new Promise` layer?

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
```diff
@@ -302,38 +323,31 @@ class CloudFormation {
...
+      if (e && e.name === 'Throttling') {
+        return [];
+      }
+      throw e;
+    }
   }
 
-  getStackParameters(stackName) {
-    return this.cfn
-      .describeStack({ StackName: stackName })
```
svidgen, line 329: This isn't even a valid v2 call ... is it? How was this working.

packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
```diff
@@ -68,39 +76,40 @@ export class SSMClientWrapper {
...
-        }
-      });
+    try {
+      await this.ssmClient.send(
+        new DeleteParameterCommand({
+          Name: secretName,
+        }),
+      );
+    } catch (err) {
+      if (err.code !== 'ParameterNotFound') {
```
svidgen: Would be curious to know if all these error cases have been exercised via tests, since some of the error hand handling changes in v3. Not necessarily this specific one, but throughout.

Any idea what coverage of these catches is like?

packages/amplify-provider-awscloudformation/src/__tests__/amplify-service-manager.test.ts
```diff
@@ -39,16 +43,11 @@ describe('init', () => {
...
-        }),
-      }),
-    } as unknown as Amplify;
-    getConfiguredAmplifyClientMock.mockClear();
-    getConfiguredAmplifyClientMock.mockResolvedValue(amplifyClientGeneralConfigStub);
+    mockAmplifyClient.on(CreateAppCommand).rejectsOnce({
+      name: 'ConfigError',
+      message: 'Missing Region in Config',
+    });
+    getConfiguredAmplifyClientMock.mockResolvedValue(mockAmplifyClient as unknown as AmplifyClient);
```
svidgen, line 50: `as unknown as X` is always a treat. 😅

svidgen (approved):
A couple oddities I wouldn't mind seeing some attention on. But, I'm OK to ship as-is if the tests pass.

packages/amplify-provider-awscloudformation/src/__tests__/amplify-service-manager.test.ts
ShadowCat567: alas, this is the way that I found to get the mock clients to get recognized as their real counterparts

packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
ShadowCat567: looks like I forgot to flip this to `err.name`...also I don't think there is unit testing for this file (this is surprisingly common in Gen1 and for the most part I did not update test coverage since the migration itself is so time-consuming to begin with) but this might have coverage in e2es...not sure

packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js
ShadowCat567: so this line would turn into:
```
return async () => { ...
```
The `Promise` seems to be here for the `reject/resolve` logic that is implemented here to mimic what was happening when this used v2.

svidgen: No, if I'm reading it correctly:

1. [`createResourceStack(cfnParentStackParams)`](https://github.com/aws-amplify/amplify-cli/blob/89f6baccefa7318ab2e49d069808ffed5a62f803/packages/amplify-provider-awscloudformation/src/aws-utils/aws-cfn.js#L93) becomes `async createResourceStack(cfnParentStackParams)`
2. `return new Promise(async (resolve, reject) => {` and its closing `})` are removed.
3. `resolve(x)` turns into `return x`
4. `reject(x)` turns `throw x`

Only reasons I can think *not* to simplify it down like this are:

1. We record a reference to the promise we're returning somewhere for later something-ing. (But, I don't see us doing that.)
2. I read it all incorrectly and it's just not that simple.
3. "I don't wanna. I'm tired of working on this." (Which I can sympathize with.)

ShadowCat567: I can do this as a separate commit, if e2es are not happy, I'll revert it and we keep this promise thing (since I'm going in and adjusting some errors any way)

svidgen (approved):

PR #14271, opened 2025-10-06
refactor: amplify-category-notifications aws sdk v3 migration

dgandhi62:
* refactor: start notifications category migration

* refactor: migrate pinpoint-helper to sdk v3

* refactor: migrate channel handlers to sdk v3

* refactor: migrate auth-helper to sdk v3

* refactor: update test mocks for sdk v3

* refactor: migrate cli logout command to sdk v3

* refactor: migrate e2e-core pinpoint utils to sdk v3

* fix: update package dependencies

* fix: correct client import names

* fix: yarn lock updates

- [x] PR description included
- [ ] `yarn test` passes


packages/amplify-category-notifications/src/__tests__/channel-apns.test.ts
```diff
@@ -65,21 +65,11 @@ describe('channel-APNS', () => {
   const mockCertificateConfig = {};
 
   const mockPinpointClient = {
-    updateApnsChannel: jest.fn().mockReturnValue({
```
ShadowCat567, line 68: For all of these tests, see if you can use `'aws-sdk-client-mock'` like this test file does: https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-provider-awscloudformation/src/__tests__/amplify-service-manager.test.ts
Since just checking that send was called is not 1:1 with what the previous test was checking.

packages/amplify-category-notifications/src/pinpoint-helper.ts
```diff
@@ -476,24 +477,25 @@ const deleteApp = async (context: $TSContext, pinpointAppId: string): Promise<$T
...
-    });
-  });
+  try {
+    const data = await pinpointClient.send(new DeleteAppCommand(params));
+    spinner.succeed(`Successfully deleted Pinpoint project: ${data.ApplicationResponse?.Name || params.ApplicationId}`);
+    if (data.ApplicationResponse) {
+      return data.ApplicationResponse;
+    }
+    return { Id: params.ApplicationId };
+    // Removed region property assignment due to changes in sdk v3 ApplicationResponse interface
```
ShadowCat567: you don't need to add this in the actual code but can you link to what you are referring to as a response to this comment? I want to understand why the last else statement was dropped for this review

packages/amplify-cli/src/commands/logout.ts
ShadowCat567, line 1: why are changes to the `amplify-cli` package required for the Pinpoint changes?

ShadowCat567 (commented):
Left some suggestions for improvements and questions

packages/amplify-category-notifications/src/pinpoint-helper.ts
dgandhi62: The last 'else' statement was in the case of an error not being present. I am currently implementing the same functionality in the "try" block. Therefore, if there is no error, only the try block would implement. If an error exists, the catch handles the other two conditions. 

My comment referes to this line. I wasn't able to find matching region property in sdk3 - https://github.com/aws-amplify/amplify-cli/pull/14271/files#diff-6a4a421e3edb17c2b530c8caf0e61f97a41e99f622584ab91c0d3d236a76d982L492

ShadowCat567: ok, sounds good

packages/amplify-cli/src/commands/logout.ts
dgandhi62: It's not. I think I mis-understood one part of the flow. I was trying to understand how the cli dependencies work and this was one change I thought was needed for the errors to subside. I looked closer now and I don't see a connection, removing it for a separate pr.

ShadowCat567: these changes will be coming in one of my PRs

packages/amplify-category-notifications/src/__tests__/channel-apns.test.ts
```diff
@@ -14,6 +17,8 @@ jest.mock('../apns-cert-config');
 jest.mock('@aws-amplify/amplify-prompts');
 const prompterMock = prompter as jest.Mocked<typeof prompter>;
 
+const mockPinpointClient = mockClient(PinpointClient as any);
```
ShadowCat567: why are you casting to `any` here and in all the other places when working with commands? 
If you have a lot of lint failures, all of these any casts are probably why, it is generally discouraged to use `any` in typescript since it defeats the purpose of typescript, and you should not need to do all these casts to `any` since the types should be compatible. 
I am leaving this comment here, but take a look at the other test files and try to get rid of the `any` casts

packages/amplify-category-notifications/src/__tests__/channel-apns.test.ts
```diff
@@ -64,49 +69,22 @@ describe('channel-APNS', () => {
...
-    }),
-  };
-
   const mockContext: $TSContext = {
     exeInfo: {
       serviceMeta: {
         output: mockServiceOutput,
       },
-      pinpointClient: mockPinpointClient,
+      pinpointClient: mockPinpointClient as any,
```
ShadowCat567: you should be able to switch this to `mockPinpointClient as unknown as PinpointClient` so you do not have to cast this to `any` - same with the other test files

packages/amplify-category-notifications/src/__tests__/channel-fcm.test.ts
```diff
@@ -37,11 +42,11 @@ const getError = async <TError>(call: () => unknown): Promise<TError> => {
 };
 const channelName = 'FCM';
 
-const mockPinpointResponseData = (status: boolean, action: ChannelAction): IChannelAPIResponse => ({
+const mockPinpointResponseData = (status: boolean, action: ChannelAction, output: any): IChannelAPIResponse => ({
```
ShadowCat567, line 45: why is `output` getting added here as a parameter?

dgandhi62: All the tests call this function in which we are returning actual data - https://github.com/aws-amplify/amplify-cli/blob/a03b6bf80939b822f755c99fd2389b350002588a/packages/amplify-category-notifications/src/channel-fcm.ts#L69

This was causing an error since output was hardcoded as undefined before. For it to exist as a normal parameter, I would either need to redeclare it above or pass it as a parameter. Passing as a parameter felt safer than changing location of code

packages/amplify-category-notifications/src/__tests__/channel-fcm.test.ts
```diff
@@ -51,110 +56,97 @@ const mockPinpointResponseData = (status: boolean, action: ChannelAction): IChan
...
-    const errCert: AmplifyFault = await getError(async () => channelFCM.enable(context as unknown as $TSContext, 'successMessage'));
-    expect(context.exeInfo.pinpointClient.updateGcmChannel).toBeCalled();
+    const context = mockContext(mockServiceOutput);
+    const errCert: AmplifyFault = await getError(async () => channelFCM.enable(context, 'successMessage'));
+    expect(mockPinpointClient).toHaveReceivedCommand(UpdateGcmChannelCommand);
     expect(errCert?.downstreamException?.message).toContain(mockPinpointResponseErr.message);
   });
 
   test('disable', async () => {
-    const data = await channelFCM.disable(mockContextReject(mockServiceOutput, mockPinpointClient));
```
ShadowCat567, line 150: Is this an instance of the original test being weird? Since it seems like the point of this test is to mock a successful instance of disabling a notification channel, yet for some reason this was initially mocked using reject instead of resolve

packages/amplify-category-notifications/src/pinpoint-helper.ts
```diff
@@ -476,24 +477,28 @@ const deleteApp = async (context: $TSContext, pinpointAppId: string): Promise<$T
...
+  try {
+    const data = await pinpointClient.send(new DeleteAppCommand(params));
+    spinner.succeed(`Successfully deleted Pinpoint project: ${data.ApplicationResponse?.Name || params.ApplicationId}`);
+    if (data.ApplicationResponse) {
+      return {
+        ...data.ApplicationResponse,
+        Region: pinpointClient.config.region as any,
+      };
+    }
+    return { Id: params.ApplicationId };
```
ShadowCat567, line 489: why was this if statement with `data.ApplicationResponse` added?

scripts/split-e2e-tests-codebuild.ts
```diff
@@ -36,6 +40,7 @@ const RUN_SOLO = [
   'src/__tests__/import_s3_2b.test.ts',
   'src/__tests__/import_s3_2c.test.ts',
   'src/__tests__/import_s3_3.test.ts',
+  'src/__tests__/notifications-analytics-compatibility-in-app-2.test.ts',
```
ShadowCat567: why have all of these tests been added to run solo? Also if you are going to alter this file, you should run `yarn split-e2e-tests-codebuild` so the changes are reflected in Codebuild

packages/amplify-category-notifications/src/__tests__/channel-fcm.test.ts
dgandhi62: So mockContextReject doesnt really reject anything. It just creates a context object for whatever you pass into it. The intended idea for its usage in general was to pass a failure client into it. I think the previous naming was weird

packages/amplify-category-notifications/src/pinpoint-helper.ts
dgandhi62: ApplicationResponse can hold undefined. Previously, we were not checking this and this could throw errors in runtime (which I suspect was one of the errors I was facing). To prevent errors being thrown, I added a check. 

https://github.com/aws/aws-sdk-js-v3/blob/eb2bf70de8531f4e0337ea38fc8b63569c9335c3/clients/client-pinpoint/src/models/models_0.ts#L6236

scripts/split-e2e-tests-codebuild.ts
dgandhi62: I added them to help mitigate the "worker killed during runtime" issue. Essentially they were memory overloading when run in parallel so I tried running them individually

And interesting. I did not know this

ShadowCat567 (approved):
Looks good, nice job!
Remember to run PR testing on this

ShadowCat567 (approved):

PR #14272, opened 2025-10-06
refactor: uibuilder, function, cli, console-integ, env params

ShadowCat567:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-util-uibuilder/src/__tests__/syncAmplifyUiComponents.test.ts
```diff
@@ -158,29 +179,10 @@ describe('should sync amplify ui builder components', () => {
               style: {},
             },
           ],
-        })),
-      })),
-      exportViews: jest.fn(() => ({
```
ShadowCat567, line 163: `ExportViews` seems to only have existed in tests. There is no `ExportViewsCommand` in aws sdk v3, we should not be losing any code functionality by dropping this.

iliapolo: > There is no ExportViewsCommand in aws sdk v3

A little suspicious - my guess is it never existed in v2 as well, and we just mistakenly added it here to a test mock. Can you double check?

ShadowCat567: yep, according to the AWS SDK v2 API reference `exportViews` never existed: https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/AmplifyUIBuilder.html
Not sure why it was added here to begin with

PR #14280, opened 2025-10-10
fix: update throttling retry

ShadowCat567:
#### Description of changes

AWS SDK V3 no longer has `retryable` as a property on the errors that are thrown when throttling happens. This lead to the custom retry configuration we had set up in `describeStacks` not properly activating, which caused an increase in throttling errors with customers who had large/complicated applications. 

Created a script that ran a large number (1000) of `describeStacks` operations and observed how it ran with and without the custom retry configuration set up. 

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


svidgen (approved):
Makes sense to me. Thanks for digging on this!

PR #14284, opened 2025-10-17
fix: cli throws an error when `thirdPartyAuth` is enabled, but `authProvider` is not included

otaviomacedo:
The function that loads the `awscloudformation` provider assumes that the `authProviders` property is present whenever `thirdPartyAuth` is present. This object comes from a user controlled file (`parameter.json` in the application resource dir), so there is no guarantee that this is true.

Default to an empty array if it's not present.


ShadowCat567 (approved):
Considering the previous cli maintainers used this `includes('accounts.google.com')` previously, I'm not too concerned about these CodeQL errors.

PR #14303, opened 2025-10-28
feat(gen2-migration): cfn drift detection

9pace:
#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/index.ts
+++ b/packages/amplify-provider-awscloudformation/src/index.ts
@@ -65,8 +65,6 @@ export {
 } from './utils/ssm-utils/env-parameter-ssm-helpers';
 export { AwsSdkConfig } from './utils/auth-types';
 
-// Drift detection exports have been moved to amplify-cli package
-
 function init(context) {
   return initializer.run(context);
 }


packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+    verbose: context.parameters?.options?.verbose || false,
+    format: context.parameters?.options?.format || 'summary',
+    'output-file': context.parameters?.options?.['output-file'],
+  };
+
+  const detector = new AmplifyDriftDetector(context);
+  const exitCode = await detector.detect(options);
+
+  if (exitCode !== 0) {
+    process.exitCode = exitCode;
```
iliapolo, line 39: Prefer throwing an error (`AmplifyError`?) to make it clearer to the user the command failed

package.json
```diff
@@ -161,6 +161,7 @@
   },
   "packageManager": "yarn@3.5.0",
   "resolutions": {
+    "@aws-sdk/client-cloudformation": "^3.624.0",
```
iliapolo, line 164: Why is this necessary?

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+ * Amplify drift detection command
+ * Based on AWS CDK CLI drift implementation
+ */
 
+import { $TSContext, AmplifyError } from '@aws-amplify/amplify-cli-core';
+import { printer } from '@aws-amplify/amplify-prompts';
+import chalk from 'chalk';
+import { detectStackDriftRecursive, type DriftDisplayFormat } from './drift-detection';
+import { CloudFormationService, AmplifyConfigService, FileService, DriftFormatter } from './drift-detection/services';
+import type { CloudFormationClient } from '@aws-sdk/client-cloudformation';
```
iliapolo, line 11: Unused

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
-import { $TSContext } from '@aws-amplify/amplify-cli-core';
+/**
+ * Amplify drift detection command
+ * Based on AWS CDK CLI drift implementation
+ */
 
+import { $TSContext, AmplifyError } from '@aws-amplify/amplify-cli-core';
+import { printer } from '@aws-amplify/amplify-prompts';
+import chalk from 'chalk';
+import { detectStackDriftRecursive, type DriftDisplayFormat } from './drift-detection';
```
iliapolo, line 9: `DriftDisplayFormat` unused

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+    // 1. Validate Amplify project
+    this.configService.validateAmplifyProject();
+
+    // 2. Get stack name and project info
+    const stackName = this.configService.getRootStackName();
+    const projectName = this.configService.extractProjectName(stackName);
+
+    // Display initial status
+    printer.info('');
+    printer.info(chalk.cyan.bold(`Started Drift Detection for Project: ${projectName}`));
```
iliapolo, line 75: We should also say which environment we are checking

packages/amplify-cli/src/commands/drift-detection/services/cloudformation-service.ts
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * Service for CloudFormation operations
+ */
+export class CloudFormationService {
+  /**
+   * Get CloudFormation client
+   * Creates a new client each time to ensure fresh credentials
+   */
+  public async getClient(context: $TSContext): Promise<CloudFormationClient> {
+    const { loadConfiguration } = require('@aws-amplify/amplify-provider-awscloudformation');
```
iliapolo, line 19: Why a dynamic require instead of a regular import? is this an established pattern?

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+
+    // 3. Get CloudFormation client
+    const cfn = await this.cfnService.getClient(this.context);
+
+    // 4. Validate stack exists
+    if (!(await this.cfnService.validateStackExists(cfn, stackName))) {
+      printer.error(chalk.red('Stack not found'));
```
iliapolo, line 82: Given we immediately throw - is this log adding anything?

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+        } else if (msg.includes('Checking drift for nested stack:')) {
+          // Show nested stack checking without indentation
+          const nestedStackName = msg.replace('Checking drift for nested stack:', '').trim();
+          printer.info(chalk.gray(`Checking drift for nested stack: ${chalk.yellow(nestedStackName)}`));
+        } else if (!msg.includes('Checking drift for stack:')) {
+          printer.info(msg);
+        }
+      },
+      debug: (msg: string) => {
+        if (options.verbose) printer.info(msg);
```
iliapolo, line 148: Why not `printer.debug(msg)`?

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+  /**
+   * Create print object for drift detection output
+   */
+  private createPrintObject(options: DriftOptions) {
+    return {
+      info: (msg: string) => {
+        // Parse and format messages based on their content
+        if (msg.includes('Found') && msg.includes('nested')) {
+          // Extract indentation, count, and type
+          const indent = msg.match(/^(\s*)/)?.[1] || '';
```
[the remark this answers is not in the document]
iliapolo: Copilot is right.

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+          const count = msg.match(/Found (\d+)/)?.[1] || '0';
+          const level = msg.match(/level (\d+)/)?.[1];
```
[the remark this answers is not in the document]
iliapolo: Same here

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+    const hasDrift = output.totalDrifted > 0;
+    return hasDrift ? 1 : 0;
+  }
+
+  /**
+   * Create print object for drift detection output
+   */
+  private createPrintObject(options: DriftOptions) {
+    return {
+      info: (msg: string) => {
```
iliapolo, line 129: Seems like this is just adding color to specific parts of the message - any reason not to create those parts with color from the get go? i.e just use `chalk` when you're creating the message?

packages/amplify-cli/src/commands/drift.ts
```diff
@@ -1,13 +1,191 @@
...
+    // 5. Detect drift recursively (including nested stacks)
+    const print = this.createPrintObject(options);
+    const combinedResults = await detectStackDriftRecursive(cfn, stackName, print);
+
+    printer.info(chalk.green('Drift detection completed'));
+    printer.info('');
+
+    // 6. Handle no results
+    if (!combinedResults.rootStackDrifts.StackResourceDrifts) {
+      printer.warn(`${stackName}: No drift results available`);
```
iliapolo, line 101: Why not throw an error? is this expected?

packages/amplify-cli/src/commands/gen2-migration/_validations.ts
```diff
@@ -7,9 +7,9 @@ import { STATEFUL_RESOURCES } from './stateful-resources';
 export class AmplifyGen2MigrationValidations {
   constructor(private readonly context: $TSContext) {}
 
-  public async validateDrift(): Promise<void> {
-    return new AmplifyDriftDetector(this.context).detect();
-  }
+  // public async validateDrift(): Promise<void> {
```
iliapolo, line 10: Why comment out?

packages/amplify-cli/src/commands/drift-detection/services/amplify-config-service.ts
```diff
@@ -0,0 +1,69 @@
...
+  /**
+   * Extract category from logical ID
+   */
+  public extractCategory(logicalId: string): string {
+    const idLower = logicalId.toLowerCase();
+    if (idLower.includes('auth')) return 'auth';
+    if (idLower.includes('storage')) return 'storage';
+    if (idLower.includes('function')) return 'function';
+    if (idLower.includes('api')) return 'api';
+    if (idLower.includes('hosting')) return 'hosting';
```
iliapolo, line 65: is there really a stack for hosting?

packages/amplify-cli/src/commands/drift-detection/services/amplify-config-service.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (idLower.includes('analytics')) return 'analytics';
+    return 'other';
```
iliapolo, line 67: Should throw, this is unexpected.

packages/amplify-cli/src/commands/drift-detection/detect-stack-drift.ts
```diff
@@ -0,0 +1,313 @@
...
+        notCheckedResources.map((r) => `  - ${r.LogicalResourceId} (${r.ResourceType})`).join('\n'),
+    );
+  }
+
+  // Log warning for resources with UNKNOWN status (actual problems)
+  const unknownResources = driftResults.StackResourceDrifts?.filter((drift) => drift.StackResourceDriftStatus === 'UNKNOWN');
+
+  if (unknownResources && unknownResources.length > 0 && print?.debug) {
+    print.debug(
+      'WARNING: Drift detection failed for some resources. This may be due to insufficient permissions or throttling:\n' +
```
iliapolo, line 101: If this isn't expected - lets throw. In the migration context, I want to be extract protective. I would agree that for normal drift detection we can be more permissive. We can provide an option later on to toggle "faileOnUnknown" or something.

packages/amplify-cli/src/commands/drift-detection/services/drift-formatter.ts
```diff
@@ -0,0 +1,705 @@
...
+    // Resources in sync
+    dashboard += this.formatDashboardLine(
+      'Resources in Sync',
+      chalk.green(this.summary.totalInSync),
+      DISPLAY_CONSTANTS.LABEL_RESOURCES_SYNC,
+    );
+
+    // Unchecked resources
+    dashboard += this.formatDashboardLine(
+      'Unchecked Resources',
```
iliapolo, line 275: Lets print the unchecked resources so the user can see they are less critical (it would be bad if we cant check a stateful resource).

packages/amplify-cli/src/commands/drift-detection/detect-stack-drift.ts
```diff
@@ -0,0 +1,313 @@
...
+      if (print?.info) {
+        print.info(`Checking drift for nested stack: ${nestedStack.LogicalResourceId}`);
+      }
+
+      // Extract stack name from PhysicalResourceId
+      // Handle both ARN format and direct stack names
+      let nestedStackName = nestedStack.PhysicalResourceId;
+
+      // ARN format: arn:aws:cloudformation:region:account:stack/stack-name/id
+      if (nestedStackName.startsWith('arn:aws:cloudformation:')) {
```
iliapolo, line 230: Can this be false? when?

packages/amplify-cli/src/commands/drift-detection/detect-stack-drift.ts
```diff
@@ -0,0 +1,313 @@
...
+            if (resourcePart && resourcePart.startsWith('stack/')) {
+              // Extract stack name from stack/stack-name/id
+              const stackParts = resourcePart.split('/');
+              if (stackParts.length >= 2) {
+                nestedStackName = stackParts[1];
+              }
+            }
+          }
+        } catch (e) {
+          // If parsing fails, log and use the original value
```
iliapolo, line 246: If we are good with both a name and an ARN - why choose over the other?

packages/amplify-cli/src/commands/drift-detection/detect-stack-drift.ts
```diff
@@ -0,0 +1,313 @@
...
+          logger.logInfo({
+            message: `Failed to parse ARN for nested stack ${nestedStack.LogicalResourceId}: ${nestedStackName}. Using original value.`,
+          });
+        }
+      }
+
+      // Validate the extracted name
+      if (!nestedStackName || nestedStackName === nestedStack.PhysicalResourceId) {
+        logger.logInfo({
```
iliapolo, line 255: Why validate if we are just logging?

packages/amplify-cli/src/commands/drift-detection/detect-stack-drift.ts
```diff
@@ -0,0 +1,313 @@
...
+        } catch (e) {
+          // If parsing fails, log and use the original value
+          logger.logInfo({
+            message: `Failed to parse ARN for nested stack ${nestedStack.LogicalResourceId}: ${nestedStackName}. Using original value.`,
+          });
+        }
+      }
+
+      // Validate the extracted name
+      if (!nestedStackName || nestedStackName === nestedStack.PhysicalResourceId) {
```
iliapolo, line 254: Why not just use `nestedStack.PhysicalResourceId` and call it a day? what scenario are you trying to protect against?

packages/amplify-cli/src/commands/drift-detection/detect-stack-drift.ts
```diff
@@ -0,0 +1,313 @@
...
+        print.warning(`Failed to check drift for nested stack ${nestedStack.LogicalResourceId}: ${error.message}`);
+      }
+      logger.logError({
+        message: `detectStackDriftRecursive.nestedStack.error: ${nestedStack.LogicalResourceId}`,
+        error: error,
+      });
+    }
+  }
+
+  logger.logInfo({
```
iliapolo, line 297: When are these being printed to screen? with `--verbose`?

packages/amplify-cli/src/commands/drift-detection/detect-stack-drift.ts
```diff
@@ -0,0 +1,313 @@
...
+      nestedResults.nestedStackPhysicalIds.forEach((value, key) => {
+        const fullKey = `${nestedStack.LogicalResourceId}/${key}`;
+        nestedStackPhysicalIds.set(fullKey, value);
+      });
+
+      logger.logInfo({
+        message: `detectStackDriftRecursive.nestedStack: ${nestedStack.LogicalResourceId}, ${nestedResults.rootStackDrifts.StackResourceDrifts?.length} direct resources, ${nestedResults.nestedStackDrifts.size} sub-stacks`,
+      });
+    } catch (error: any) {
+      // Log error but continue checking other nested stacks
```
iliapolo, line 286: Yeah but again in the context of migration I would prefer this to fail - to be extra safe.

iliapolo (approved):

PR #14311, opened 2025-10-30
chore: update directory structure and file names

dgandhi62:
Changes include:

1) Renaming amplify-gen2-migration-codegen-dg directory to codegen-generate
2) Removing src from migration_head
3) Rename migration_head to codegen-head
4) Remove codegen/cli directory

The src directory needs to remain for the codegen-generate folder since there would be an additional tsconfig file added for codegen in the future. This is to enforce strict:false behavior in tsconfig as it was in the old tool. Currently, amplify-cli's tsconfig has been modified to make the option strict:false from strict:true (which needs to change back before merging to dev)


kaizencc (commented):
thanks! i'm still confused by these parts of the structure:

- `src` folder --> why is it necessary to have another `src` folder? the folder structure is now `amplify-cli/src/comands/codegen-generate/src`. `src` typically is a top-level folder that delineates where all the source code lives so it's confusing to me to see it again. 
- `migration_head` folder -> what does this mean? why is this the head of the migration?
- `cll` folder -> are we releasing `generate` as a CLI separate from the `migrate` tool?
- `npm_package` folder -> are we releasing `generate` as a separate npm package?
- `test_utils` folder -> ok, but where are the tests? I can only see `auth/index.test.ts`. in the structure of the `gen2-migration` branch, i see `src/__tests__/gen2-migration` with some tests from Sai and Sanjana. i _think_ that's where all the unit tests belong, and if so, that's where `test_utils` belong too if they are truly just utilities for tests

PR #14320, opened 2025-11-03
feat(gen2-migration): support kinesis analytics migration

kaizencc:
This PR supports kinesis analytics migration. Kinesis is a category that exists in Amplify Gen1 but is not supported in Amplify Gen2. 

The general idea is to grab the CloudFormation nested stack that includes the Kinesis Stream definition, and send that to a new dependency, `cdk-from-cfn`, which will spit out CDK L1 code that can live in conjunction with Amplify Gen2.

The following limitations still exist and must be solved before this PR fully supports Kinesis Analytics:

- [x] Need to use the stacks defined by `cdk-from-cfn` in `backend.ts`. We will need to pass in the parameters.
- [x] CFN conditions in the CFN template need to be resolved in a preprocessing step prior to `cdk-from-cfn`. We do this  for the supported categories https://github.com/aws-amplify/amplify-cli/blob/migrations/packages/amplify-migration-template-gen/src/resolvers/cfn-condition-resolver.ts.

To suport other unsupported categories, `cdk-from-cfn` will need to support custom resources.

### Update 1:

This update builds on the initial PR by implementing the analytics (Kinesis) codegen pipeline for Gen1 to Gen2 migration. Previously, the migration generated the CDK stack file via cdk-from-cfn and now also generates the resource.ts wrapper and integrates it into backend.ts.

### Changes:
1. Type Safety & Interfaces:

- cdk-from-cfn.ts : Added KinesisAnalyticsDefinition and AnalyticsCodegenResult interfaces, removing all any types for proper type safety

- app_analytics_definition_fetcher.ts : Fixed return type to use typed Record<string, KinesisAnalyticsDefinition>

2. Code Generation:

- generators/analytics/index.ts : Created new generator with renderAnalytics() function that generates resource.ts with proper Backend type and nested stack instantiation using TypeScript AST

- migration-pipeline.ts : Updated to use TypescriptNodeArrayRenderer for generating analytics/resource.ts alongside the stack file

3. Backend Integration:

- synthesizer.ts : Updated to add analytics import ( import { analytics } from './analytics/resource') and analytics(backend) call in the generated backend.ts

The generated resource.ts creates a NestedStack under the Amplify backend root stack using backend.createStack('analytics'), passing auth role names from backend.auth.resources. All parameter values (kinesisStreamName, authPolicyName, etc.) are derived from the resource name or runtime backend references.

### Update 2:

This update resolves conditions in the CFN template in a preprocessing step prior to cdk-from-cfn. cdk-from-cfn generates local variables for CFN conditions (ex., `const shouldNotCreateEnvResources = props.env! === 'NONE'`). This causes issues when migrating analytics resources from Gen1 to Gen2.

The `env` parameter must be renamed to `amplify-env` because `env` is a reserved property in CDK's StackProps that CDK uses for AWS account/region configuration. However, the generated condition code still references `props.env`.

Additionally, Gen2 doesn't use CloudFormation parameters for environment. The generated condition variables reference parameters that don't exist in the Gen2 props interface, causing TypeScript type errors.

To address this, we preprocess Gen1 templates to resolve CFN conditions before passing to cdk-from-cfn, following the same pattern used for supported categories. We fetch deployed stack parameters via CloudFormation DescribeStacks API, use the existing CFNConditionResolver to evaluate conditions with actual values, and remove the Conditions block after resolution.

cdk-from-cfn never sees conditions, so no local variables are generated. The output contains only concrete resources matching the deployed state.

### Changes:

- `cdk-from-cfn.ts`: Added `getAnalyticsStackParameters()` and condition resolution in `preTransmute()`
- `migration-pipeline.ts`: Added `rootStackName` and `cfnClient` to render options
- `command-handlers.ts`: Pass CloudFormationClient and rootStackName to pipeline

Note: Per discussion, Backend\<any> is used because the alternatives are more complex without providing additional safety. Using a generic \<T extends DefineBackendProps> causes TypeScript to lose type information about backend.auth.resources - it can't infer the specific authenticatedUserIamRole property and errors with "Property does not exist on type 'ReturnType<T[string]["getInstance"]>["resources"]'". Importing the backend instance directly (typeof backend) creates a circular dependency since backend.ts already imports this analytics module. While we could define a custom interface with the exact auth structure, that duplicates type information and doesn't use the Backend\<T> type as requested. Since the actual backend type is validated at the call site in backend.ts, Backend\<any> is the pragmatic solution. This could be further investigated.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


sai-ray:
This update resolves conditions in the CFN template in a preprocessing step prior to cdk-from-cfn. cdk-from-cfn generates local variables for CFN conditions (ex., `const shouldNotCreateEnvResources = props.env! === 'NONE'`). This causes issues when migrating analytics resources from Gen1 to Gen2.

The `env` parameter must be renamed to `amplify-env` because `env` is a reserved property in CDK's StackProps that CDK uses for AWS account/region configuration. However, the generated condition code still references `props.env`.

Additionally, Gen2 doesn't use CloudFormation parameters for environment. The generated condition variables reference parameters that don't exist in the Gen2 props interface, causing TypeScript type errors.


To address this, we preprocess Gen1 templates to resolve CFN conditions before passing to cdk-from-cfn, following the same pattern used for supported categories. We fetch deployed stack parameters via CloudFormation DescribeStacks API, use the existing CFNConditionResolver to evaluate conditions with actual values, and remove the Conditions block after resolution.

cdk-from-cfn never sees conditions, so no local variables are generated. The output contains only concrete resources matching the deployed state.

### Changes:

- `cdk-from-cfn.ts`: Added `getAnalyticsStackParameters()` and condition resolution in `preTransmute()`
- `migration-pipeline.ts`: Added `rootStackName` and `cfnClient` to render options
- `command-handlers.ts`: Pass CloudFormationClient and rootStackName to pipeline

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
```diff
@@ -44,10 +45,13 @@ interface CodegenCommandParameters {
   logger: Logger;
   outputDirectory: string;
   backendEnvironmentName: string | undefined;
+  rootStackName: string | undefined;
```
dgandhi62, line 48: Unless I'm missing something, rootStackName is already a property in the abstract class for the `generate` command. Shouldn't it already be accessible?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
```diff
@@ -239,6 +254,16 @@ const unsupportedCategories = (): Map<string, string> => {
           }
         });
       }
+    } else if (category === 'analytics') {
+      const analytics = meta?.analytics ?? {};
+      Object.keys(analytics).forEach((analytic) => {
+        const analyticObj = analytics[analytic];
+        if (analyticObj.service === 'Pinpoint') {
+          const analyticsDocLink = unsupportedCategories.get(analyticsKey);
```
dgandhi62, line 263: Why is pinPoint unsupported?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/unsupported/cdk-from-cfn.ts
```diff
@@ -0,0 +1,154 @@
...
+      // All Fn::If references have been resolved, and resources with unmet conditions have been removed
+      delete resolved.Conditions;
+      return resolved;
+    }
+
+    return template;
+  }
+}
+
+async function getS3ObjectContent(s3Url: string): Promise<CFNTemplate> {
```
dgandhi62: There should be methods like this in supported/s3. If not, can we add this there and import it here?

dgandhi62: I misread - this is extracting cfn templates from s3. Could the function name be more explicit about this?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/unsupported/cdk-from-cfn.ts
```diff
@@ -0,0 +1,154 @@
...
+  const url = new URL(s3Url);
```
dgandhi62, line 154: I think the url format seems to be assumed below. Do we know if this format is universally true?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
```diff
@@ -0,0 +1,177 @@
...
+ * import { Backend } from '@aws-amplify/backend';
+ *
+ * export const analytics = (backend: Backend<any>) => {
+ *   const analyticsStack = backend.createStack('analytics');
+ *   new stackClassName(analyticsStack, 'resourceName', {
+ *     kinesisStreamName: 'resourceName',
+ *     kinesisStreamShardCount: 1,
+ *     authPolicyName: 'resourceName-auth-policy',
+ *     unauthPolicyName: 'resourceName-unauth-policy',
+ *     authRoleName: backend.auth.resources.authenticatedUserIamRole.roleName,
```
dgandhi62, line 35: Is Auth being present a necessity for analytics / kinesis?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  // Create the new NestedStack instantiation with props
+  const newStackExpression = factory.createExpressionStatement(
+    factory.createNewExpression(factory.createIdentifier(stackClassName), undefined, [
+      factory.createIdentifier('analyticsStack'),
+      factory.createStringLiteral(resourceName),
+      factory.createObjectLiteralExpression(
+        [
+          factory.createPropertyAssignment(factory.createIdentifier('kinesisStreamName'), factory.createStringLiteral(resourceName)),
+          factory.createPropertyAssignment(factory.createIdentifier('kinesisStreamShardCount'), factory.createNumericLiteral(1)),
```
dgandhi62: Why is this hardcoded to 1?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
```diff
@@ -0,0 +1,177 @@
...
+          factory.createPropertyAssignment(factory.createIdentifier('unauthRoleName'), createUnauthRoleAccess()),
+          factory.createPropertyAssignment(factory.createIdentifier('amplifyEnv'), createEnvExpression()),
+        ],
+        true,
+      ),
+    ]),
+  );
+
+  // Create the arrow function: export const analytics = (backend: Backend<any>) => { ... }
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any -- Using 'any' for generated code to avoid complex type inference
```
dgandhi62, line 151: The point of typescript is for strict type checking. Can you explain why `any` is needed here?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
```diff
@@ -0,0 +1,177 @@
...
+ * Generated output:
+ * ```typescript
+ * import { stackClassName } from './stackFileName';
+ * import { Backend } from '@aws-amplify/backend';
+ *
+ * export const analytics = (backend: Backend<any>) => {
+ *   const analyticsStack = backend.createStack('analytics');
+ *   new stackClassName(analyticsStack, 'resourceName', {
+ *     kinesisStreamName: 'resourceName',
+ *     kinesisStreamShardCount: 1,
```
dgandhi62: Can you link me to the source of these properties? I was unable to find the right documentation page for some reason and just wanted to crosscheck

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
sai-ray: Yes, rootStackName is in the abstract class, but generateGen2Code() is a standalone function, not a class method, so it can't access class properties directly. It needs the value passed through the CodegenCommandParameters interface hence I added it here. Please correct me if I'm wrong!

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
sai-ray: Pinpoint is being sunsetted, so it's not a priority over other categories right now.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/unsupported/cdk-from-cfn.ts
sai-ray: Makes sense, I have updated the function name.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/unsupported/cdk-from-cfn.ts
sai-ray: From what I have seen, Amplify Gen1 uses path-style S3 URLs for template storage, so this format should be consistent. But this may or not be universally true so I have also added virtual-hosted style support for safety.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
sai-ray: Yes, Kinesis analytics in Gen1 uses Cognito auth/unauth IAM roles to control access to the stream, so Auth is required.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
sai-ray: These properties come from the Gen1 Kinesis CloudFormation template that gets deployed when users add analytics. You can find it [here](https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-category-analytics/src/provider-utils/awscloudformation/cloudformation-templates/kinesis-cloudformation-template.json).

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
sai-ray: Good catch. While adding analytics the default is 1, but customers can always change it manually. Updated to fetch the value dynamically from the deployed stack

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
sai-ray: Per discussion, Backend\<any> is used because the alternatives are more complex without providing additional safety. Using a generic \<T extends DefineBackendProps> causes TypeScript to lose type information about backend.auth.resources - it can't infer the specific authenticatedUserIamRole property and errors with "Property does not exist on type 'ReturnType<T[string]["getInstance"]>["resources"]'". Importing the backend instance directly (typeof backend) creates a circular dependency since backend.ts already imports this analytics module. While we could define a custom interface with the exact auth structure, that duplicates type information and doesn't use the Backend\<T> type as requested. Since the actual backend type is validated at the call site in backend.ts, Backend\<any> is the pragmatic solution. This could be further investigated.

dgandhi62: Can we add this to the pr message to keep track?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/unsupported/cdk-from-cfn.ts
dgandhi62: Okay

dgandhi62:
Look good! Can we update the pr message to add all the updated features / support and solve the merge conflicts? I'll approve it after

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/analytics/index.ts
sai-ray: Added all the updates to the PR message plus added this point as a note!

dgandhi62 (approved):

PR #14323, opened 2025-11-05
fix: use lowercase OAuth scopes to fix google login failures

dgandhi62:
This fix is based on the issue where incorrect scopes caused failed online Google OAuth login

- Change OAuth scope type from uppercase to lowercase format
- Fix test cases to use lowercase scopes

Source of truth - https://docs.aws.amazon.com/cognito/latest/developerguide/cognito-user-pools-define-resource-servers.html

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/auth/index.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/auth/index.ts
@@ -772,7 +772,7 @@ function createSecretErrorStatements(secretVariables: string[]): ts.Node[] {
     factory.createCallExpression(factory.createIdentifier('throw new Error'), undefined, [
       // eslint-disable-next-line spellcheck/spell-checker
       factory.createStringLiteral(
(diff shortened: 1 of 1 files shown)


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/auth/index.ts
```diff
@@ -768,7 +770,10 @@ const createUserAttributeAssignments = (
 function createSecretErrorStatements(secretVariables: string[]): ts.Node[] {
   return secretVariables.map((secret) =>
     factory.createCallExpression(factory.createIdentifier('throw new Error'), undefined, [
-      factory.createStringLiteral(`Secrets need to be reset, use \`npx ampx sandbox secret set ${secret}\` to set the value`),
+      // eslint-disable-next-line spellcheck/spell-checker
+      factory.createStringLiteral(
+        `Secrets need to be reset, use \`npx ampx sandbox secret set ${secret}\` to set the value. Do not deply unless this is done`,
```
iankhou:
```suggestion
        `Secrets need to be reset, use \`npx ampx sandbox secret set ${secret}\` to set the value. Do not deploy unless this is done`,
```

iankhou (approved):

PR #14326, opened 2025-11-07
feat(gen2-migration): add deployment status validation for lock command

sai-ray:
#### Description of changes

Adds deployment status validation to the gen2-migration lock command to prevent locking when the CloudFormation stack is not in `UPDATE_COMPLETE` or `CREATE_COMPLETE`.

Added UX improvement with printer.success() to display validation success status.

- Manual verification against real Amplify Gen1 projects.
- All tests pass successfully with yarn test.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


yarn.lock
iankhou, line 1: There are no dependency changes for your code, so we should not modify the lockfile in this PR.

sai-ray: The lock.ts file has an import added to the code, wouldn't that warrant changes in the yarn.lock file? The Build and Test CI check is failing without it.

sai-ray: You were correct, removed the yarn.lock changes. The earlier CI failure was due to version mismatches between the base branch's lockfile and what my local yarn install resolved. After merging the latest gen2-migration (which has updated dependency versions), everything is in sync and CI passes. Please do correct me if I am wrong.

PR #14327, opened 2025-11-07
feat: replace env variable with gen2 branch name

sanjanaravikumar-az:
Initial problem: Previously the migrated code contained AMPLIFY_GEN_1_ENV_NAME = process.env.AMPLIFY_ENV which should be changed to branchName = process.env.AWS_BRANCH. Also any env references would be cdk.Fn.ref('env') which does not suit the gen2 format. 

There are patterns like cdk.Fn.ref('env') or AmplifyHelpers.getProjectInfo().envName which need to be transformed to branchName, and we also need to remove the import related to this. Also before, when we migrate the gen1 app to gen2, you get this line of code at the top which is like AMPLIFY_GEN_1_ENV_NAME = process.env.AMPLIFY_ENV which we replace with branchName = process.env.AWS_BRANCH.

The code scans the gen1 app for env variables and references and replaces them with the proper gen2 branch names and references. This is because gen2 apps only use branch names and have no concept of environments.

Changes:
- Add `AmplifyHelperTransformer` to handle standalone `cdk.Fn.ref('env')` calls
- Update `CodeTransformer` to replace `cdk.Fn.ref('env')` with `branchName` variable
- Update command handlers to generate `const branchName = process.env.AWS_BRANCH ?? "sandbox"`. The AWS_BRANCH is used to get the branch name of the current build.
- Keep environment variable setting as `AMPLIFY_GEN_1_ENV_NAME` for compatibility

Generated CDK stack files will now use:
- `const branchName = process.env.AWS_BRANCH ?? "sandbox";`
- All `cdk.Fn.ref('env')` replaced with `branchName`
- All `AmplifyHelpers.getProjectInfo().envName` replaced with `branchName`

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
@@ -31,6 +31,8 @@ import { AppFunctionsDefinitionFetcher } from './app_functions_definition_fetche
 import { printer } from './printer';
 import { format } from './format';
 import ora from 'ora';
+import execa from 'execa';
+import { Logger } from '../../../../gen2-migration';
 import * as ts from 'typescript';
 import { AmplifyHelperTransformer } from '../../../codegen-custom-resources/transformer/amplify-helper-transformer';
 
@@ -444,8 +446,7 @@ const hasUncommentedDependency = (fileContent: string, matchString: string) => {
 
   return false;
 };
-
-export async function prepare() {
+export async function prepare(logger: Logger) {
   const appId = resolveAppId();
   const amplifyClient = new AmplifyClient();
   const backendEnvironmentResolver = new BackendEnvironmentResolver(appId, amplifyClient);


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/functions/index.ts
```diff
@@ -58,7 +58,7 @@ export function renderFunctions(definition: FunctionDefinition, appId?: string,
       amplifyGen1EnvName,
       undefined,
       undefined,
-      factory.createIdentifier('process.env.AMPLIFY_GEN_1_ENV_NAME ?? "sandbox"'),
+      factory.createIdentifier('process.env.AWS_BRANCH ?? "sandbox"'),
```
iankhou, line 61: I had to check the docs to understand this, because on the surface, it just looks like a rename. You mention this change in your PR description, but it's always helpful to also indicate why, and provide a reference.

https://docs.aws.amazon.com/amplify/latest/userguide/environment-variables.html#amplify-console-environment-variables

yarn.lock
iankhou, line 1: Don't modify this unless you're modifying dependencies in package.json.

packages/amplify-cli/src/__tests__/commands/gen2-migration/codegen-custom-resources/transformer/ast-code-transformer.test.ts
sanjanaravikumar-az, line 1: this file is unnecessary because i've added basically the same functionality in amplify-helper-transformer.ts, and it is more efficient.

packages/amplify-cli/src/__tests__/commands/gen2-migration/codegen-custom-resources/transformer/code-transformer.test.ts
sanjanaravikumar-az, line 1: this is removed for pretty much the same reason as packages/amplify-cli/src/__tests__/commands/gen2-migration/codegen-custom-resources/transformer/ast-code-transformer.test.ts. its duplicate code that isnt needed anymore.

yarn.lock
sanjanaravikumar-az: got it, it wasn't supposed to be there originally, i've deleted it now.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/functions/index.ts
sanjanaravikumar-az: noted, i'll add this in the PR description so its more clear.

sanjanaravikumar-az:
@kaizencc  I've tried to add a bit more context for the reason behind this change, and I've updated the project board as well as per your suggestion, so hopefully i was able to provide a little more background. I've also added inline comments for the reasoning behind deleting the files I removed, and updated the yarn.lock file so there are no changes to it. 

> * the PR description is missing a "why" or a link to an issue where the reviewer can gain context on the motivation behind the PR. 
> * the PR description "changes" section (or, alternatively, you could review your own PR and add relevant inline comments) does not include deletion of `code-transformers.ts` and `ast-code-transformer.ts` and why that is necessary
> * `yarn.lock` file should not change if there was no update to dependencies in this PR
> the main thing is providing necessary context for a smooth review though. currently any reviewer of your PR needs to have context supplied elsewhere and I just don't have that context yet

packages/amplify-cli/src/commands/gen2-migration/codegen-custom-resources/transformer/amplify-helper-transformer.ts
```diff
@@ -0,0 +1,125 @@
...
+        }
+        return ts.visitNode(node, visit);
+      };
+    };
+
+    const result = ts.transform(sourceFile, [transformer]);
+    return result.transformed[0] as ts.SourceFile;
+  }
+
+  static addRequiredImports(sourceFile: ts.SourceFile): ts.SourceFile {
```
iliapolo: Maybe ?

```suggestion
  static addBranchNameVariable(sourceFile: ts.SourceFile): ts.SourceFile {
```

packages/amplify-cli/src/commands/gen2-migration/codegen-custom-resources/transformer/amplify-helper-transformer.ts
```diff
@@ -0,0 +1,125 @@
...
+    );
+
+    const filteredStatements = sourceFile.statements.filter(
+      (stmt) =>
+        !(
+          ts.isImportDeclaration(stmt) &&
+          ts.isStringLiteral(stmt.moduleSpecifier) &&
+          stmt.moduleSpecifier.text === '@aws-amplify/cli-extensibility-helper'
+        ),
+    );
```
iliapolo: Check if we can get rid of this because these imports are already being removed in the `transform` method

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/functions/index.ts
```diff
@@ -19,7 +19,7 @@ export interface FunctionDefinition {
 
 const factory = ts.factory;
 
-const amplifyGen1EnvName = 'AMPLIFY_GEN_1_ENV_NAME';
+const amplifyGen1EnvName = 'branchName';
```
iliapolo:
```suggestion
const gen2BranchNameVariableName = 'branchName';
```

PR #14337, opened 2025-11-12
feat(gen2-migration): analytics category refactor

iankhou:
#### Description of changes

Enables refactor of the `analytics` category (for Kinesis).

Requires a Gen1 application with analytics configured using Kinesis, and a Gen2 application that deploys an analytics CFn stack that matches the Gen1 structure.

Create a Gen1 app with analytics, deploy the backend.

Create a Gen2 app with analytics, deploy the backend. This is more difficult because we don't have codegen for it, so we need to align the CFn resource configuration with the Gen1 app, within the analytics stack we create though CDK.

Screenshot 2025-11-13 at 15 44 55

_The Kinesis stream was successfully moved from a Gen1 stack to a Gen2 stack._

#### Checklist

- [X] PR description included
- [X] `yarn test` passes


packages/amplify-cli/src/commands/gen2-migration/refactor/generators/template-generator.ts
```diff
@@ -48,18 +49,21 @@ const AUTH_RESOURCES_TO_REFACTOR = [
...
   [CFN_AUTH_TYPE.IdentityPoolRoleAttachment.valueOf(), 'IdentityPoolRoleMap'],
   [CFN_AUTH_TYPE.UserPoolDomain.valueOf(), 'UserPoolDomain'],
   [CFN_S3_TYPE.Bucket.valueOf(), 'S3Bucket'],
+  [CFN_ANALYTICS_TYPE.Stream.valueOf(), 'KinesisStream'],
 ]);
 const LOGICAL_IDS_TO_REMOVE_FOR_REVERT_MAP = new Map<CATEGORY, CFN_RESOURCE_TYPES[]>([
   ['auth', AUTH_RESOURCES_TO_REFACTOR],
   ['auth-user-pool-group', AUTH_USER_POOL_GROUP_RESOURCES_TO_REFACTOR],
   ['storage', [CFN_S3_TYPE.Bucket]],
+  ['analytics', ANALYTICS_RESOURCES_TO_REFACTOR],
```
iankhou, line 69: Used for rollback, not used yet.

packages/amplify-cli/src/commands/gen2-migration/refactor/generators/template-generator.ts
```diff
@@ -33,7 +34,7 @@ import ora from 'ora';
 const CFN_RESOURCE_STACK_TYPE = 'AWS::CloudFormation::Stack';
 const GEN2_AMPLIFY_AUTH_LOGICAL_ID_PREFIX = 'amplifyAuth';
 
-const CATEGORIES: CATEGORY[] = ['auth', 'storage'];
+const CATEGORIES: CATEGORY[] = ['auth', 'storage', 'analytics'];
```
iankhou, line 37: Strings in this array are used to check against the prefix of category stacks in both Gen1 and Gen2. In Gen1, the analytics stack always starts with "analyticsamplify", but in Gen2, it can be any name. Codegen will need to handle the naming accordingly.

packages/amplify-cli/src/commands/gen2-migration/refactor/types.ts
```diff
@@ -101,6 +101,10 @@ export enum CFN_S3_TYPE {
   Bucket = 'AWS::S3::Bucket',
 }
 
+export enum CFN_ANALYTICS_TYPE {
+  Stream = 'AWS::Kinesis::Stream',
```
iankhou, line 111: This is the only stateful resource for Kinesis-based analytics. The rest are IAM roles and policies.

packages/amplify-cli/src/commands/gen2-migration/refactor/resolvers/cfn-output-resolver.ts
```diff
@@ -161,6 +161,11 @@ class CfnOutputResolver {
             return {
               Arn: `arn:aws:lambda:${this.region}:${this.accountId}:function:${resourceIdentifier}`,
             };
+          case 'AWS::Kinesis::Stream':
+            return {
+              // output is already in ARN format
+              Arn: resourceIdentifier,
+            };
           default:
```
iankhou, line 188: Should probably default to resourceIdentifier instead of `undefined`... but we will take a deeper look at this later.

packages/amplify-cli/src/commands/gen2-migration/refactor/types.ts
```diff
@@ -70,6 +70,7 @@ export enum NON_CUSTOM_RESOURCE_CATEGORY {
   AUTH = 'auth',
   STORAGE = 'storage',
   AUTH_USER_POOL_GROUP = 'auth-user-pool-group',
+  ANALYTICS = 'analytics',
```
iankhou, line 73: Included this here so that the custom resource flow is not triggered.

packages/amplify-cli/src/commands/gen2-migration/refactor/resolvers/cfn-output-resolver.ts
```diff
@@ -161,6 +161,11 @@ class CfnOutputResolver {
             return {
               Arn: `arn:aws:lambda:${this.region}:${this.accountId}:function:${resourceIdentifier}`,
             };
+          case 'AWS::Kinesis::Stream':
+            return {
+              // output is already in ARN format
+              Arn: resourceIdentifier,
```
9pace, line 186: This assumes resourceIdentifier is already an ARN, but getResourceAttribute is called from two paths.
Path 1 is via outputs - where resourceIdentifier is the full ARN 
Path 2 is via `tryReplaceLogicalResourceRefWithPhysicalId` - where resourceIdentifier is the physical resource ID (stream name for kinesis according to [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-kinesis-stream.html))

This may be okay if the Kinesis template ALWAYS exposes the ARN in stack outputs, in which case path 2 is never reached for Kinesis. 
For comparison, case `AWS::IAM::Role`: handles both cases defensively. Is the Kinesis ARN guaranteed to always be in outputs?

The standard `kinesis-cloudformation-template.json` has 
```
"Outputs": {
  "kinesisStreamArn": {
    "Value": { "Fn::GetAtt": ["KinesisStream", "Arn"] }
  },
```

So from this I conclude yes and that the code will resolve fine as path 2 is never reached. But I wanted to outline my concerns and confirm my understanding.

packages/amplify-cli/src/commands/gen2-migration/refactor/resolvers/cfn-output-resolver.ts
9pace: if it resolves to undefined we then resolve it to `stackResourcePhysicalId`

This works for some resource types like sns topic, but not others. 

`Line 105`
```
if (resourceArn) {
  stackTemplateResourcesString = stackTemplateResourcesString.replaceAll(fnGetAttRegExpPerLogicalId, `"${resourceArn.Arn}"`);
} else {
  stackTemplateResourcesString = stackTemplateResourcesString.replaceAll(
    fnGetAttRegExpPerLogicalId,
    `"${stackResourcePhysicalId}"`,
  );
```

But the proposed change (to resourceIdentifier) doesnt do anything different because resourceIdentifier (the parameter name in getResourceAttribute) IS stackResourcePhysicalId. (except for some special handling for sqs)

```
const stackResourcePhysicalId = stackResourceWithMatchingLogicalId.PhysicalResourceId;
```
```
// Few resources like SQS have their physical ids as their HTTP URLs. We need to construct the arn manually in such cases.
const resourceId = stackResourcePhysicalId.startsWith('http') ? stackResourcePhysicalId.split('/')[2] : stackResourcePhysicalId;
```

packages/amplify-cli/src/commands/gen2-migration/refactor/types.ts
9pace, line 79:
```suggestion
  | NON_CUSTOM_RESOURCE_CATEGORY.AUTH_USER_POOL_GROUP
  | NON_CUSTOM_RESOURCE_CATEGORY.ANALYTICS

```

9pace:
```suggestion
export type CFN_CATEGORY_TYPE = CFN_AUTH_TYPE | CFN_S3_TYPE | CFN_ANALYTICS_TYPE | CFN_IAM_TYPE | string;
```

9pace (dismissed):

packages/amplify-cli/src/commands/gen2-migration/refactor/resolvers/cfn-output-resolver.ts
```diff
@@ -94,6 +94,21 @@ class CfnOutputResolver {
           const fnGetAttRegExpPerLogicalId = new RegExp(`{"${GET_ATT}":\\["${groups.LogicalResourceId}","(?<AttributeName>\\w+)"]}`, 'g');
           const stackResourcePhysicalId = stackResourceWithMatchingLogicalId.PhysicalResourceId;
           assert(stackResourcePhysicalId);
+
+          // Kinesis streams require their ARN to be exposed in CloudFormation outputs.
+          // The physical resource ID for Kinesis streams is the stream name, not the ARN.
+          if (
+            stackResourceWithMatchingLogicalId.ResourceType === 'AWS::Kinesis::Stream' &&
+            groups.AttributeName === 'Arn' &&
+            !stackResourcePhysicalId.startsWith('arn:aws:kinesis')
```
9pace, line 103: This condition (the check that its not an arn) will always be true for Kinesis streams.
Technically redundant but included for safety.

packages/amplify-cli/src/commands/gen2-migration/refactor/resolvers/cfn-output-resolver.ts
```diff
@@ -94,6 +94,21 @@ class CfnOutputResolver {
...
+          ) {
+            throw new Error(
```
9pace, line 105: Since most of the gen2-migration refactor code uses Error, and AmplifyError is only used in _validations.ts. 

maybe this should change in the future.

iliapolo (approved):

PR #14349, opened 2025-11-18
fix(gen2-migration): excluding deployment bucket validation for decommission command

sai-ray:
#### Description of changes

This PR fixes the stateful resources validation for the amplify gen2-migration decommission command. The decommission step expects the environment to have an S3 deployment bucket containing root stack data, which should be removed during decommission. However, the stateful resources validation was correctly identifying this deployment bucket as a stateful resource and throwing an error, blocking the decommission process.

To resolve this, the validateStatefulResources method now accepts an optional excludeDeploymentBucket parameter (default: false). When set to true, the validation logic retrieves the deployment bucket name from the meta configuration and excludes it from stateful resource checks. The decommission command passes true for this parameter, allowing it to proceed while still protecting actual user data resources.

This approach ensures the validation logic itself owns the exclusion decision, rather than manipulating input data. Other migration commands that call validateStatefulResources without the parameter (or with false) continue to validate all stateful resources including the deployment bucket as expected.

Changes:
- Modified validateStatefulResources in _validations.ts to accept excludeDeploymentBucket parameter (default: false)
- Updated decommission.ts to pass true for deployment bucket exclusion
- Added unit tests to verify deployment bucket exclusion behavior

- Manually verified using live amplify gen1 environments.
- Added unit test scenarios to cover the changes.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/commands/gen2-migration/decommission.ts
```diff
@@ -17,10 +17,28 @@ import { invokeDeleteEnvParamsFromService } from '../../extensions/amplify-helpe
...
+        if (
+          change.Type === 'Resource' &&
+          change.ResourceChange?.ResourceType === 'AWS::S3::Bucket' &&
+          change.ResourceChange?.PhysicalResourceId === deploymentBucketName
+        ) {
+          return false;
+        }
+        return true;
+      });
+    }
```
iankhou: Consider the intent of this change. We want to exclude the deployment bucket from the decommission step, due to a validation preventing us from proceeding.

That means the validation itself is not correct. Instead of modifying the input to the validation, consider modifying the validation itself (`validateStatefulResources`).

iliapolo (approved):

PR #14383, opened 2025-12-08
feat(gen2-migration): add naming to codegenerated function resource

dgandhi62:
This pr adds the `name` property to the `defineFunction()` construct in the codegenerated code. This prevents cloudformation from inferring the name by default to `handler.ts`.

Furthermore, since Graphql schemas might reference function by name, we need to assign lambda function names to maintain a standard. This pr refactors the code to generate Lambda function names via escape hatches in backend.ts.

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/backend/synthesizer.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/backend/synthesizer.ts
@@ -1023,6 +1023,52 @@ export class BackendSynthesizer {
       nodes.push(userPoolDomainRemovalStatementCommented);
     }
 
+    // Function name escape hatch - set function names with branch suffix
+    if (renderArgs.function) {
+      const branchNameStatement = factory.createVariableStatement(
+        [],
+        factory.createVariableDeclarationList(
+          [
+            factory.createVariableDeclaration(
+              'branchName',
+              undefined,
+              undefined,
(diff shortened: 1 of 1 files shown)


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/functions/index.ts
```diff
@@ -157,10 +165,13 @@ export function createFunctionDefinition(
...
-  // Set the entry point to the standard Gen 2 handler file
-  defineFunctionProperties.push(createParameter('entry', factory.createStringLiteral('./handler.ts')));
-
-  // Remove name parameter - should not be included
+  if (definition?.entry) {
+    defineFunctionProperties.push(createParameter('entry', factory.createStringLiteral('./handler.ts')));
+  }
+  if (definition?.name) {
+    const funcNameAssignment = createTemplateLiteral(`${definition.resourceName}-`, gen2BranchNameVariableName, '');
+    defineFunctionProperties.push(createParameter('name', funcNameAssignment));
```
iliapolo, line 173: Should be in `backend.ts` in the form of:

```ts
backend.amplifytodogen1e23e5d28.resources.cfnResources.cfnFunction.functionName = `amplifytodogen1e23e5d28-${branchName}`;
```

dgandhi62: Incorporated in the latest pr

iliapolo (approved):

PR #14384, opened 2025-12-08
refactor(gen2-migration): storage bucket name comment generation in Gen2 migration

dgandhi62:
Moved storage bucket name comment generation from the `storage/resource.ts` to `backend.ts`

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/backend/synthesizer.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/backend/synthesizer.ts
@@ -925,7 +925,7 @@ export class BackendSynthesizer {
       ts.addSyntheticLeadingComment(
         bucketNameComment2,
         ts.SyntaxKind.SingleLineCommentTrivia,
-        ` name: '${renderArgs.storage.bucketName}',`,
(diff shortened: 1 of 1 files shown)


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/backend/synthesizer.ts
```diff
@@ -917,6 +917,18 @@ export class BackendSynthesizer {
...
       nodes.push(cfnStorageVariableStatement);
+
+      // Add comments as synthetic leading comments
+      const bucketNameComment1 = factory.createEmptyStatement();
+      ts.addSyntheticLeadingComment(bucketNameComment1, ts.SyntaxKind.SingleLineCommentTrivia, ` Use this bucket name post refactor`, true);
+      const bucketNameComment2 = factory.createEmptyStatement();
+      ts.addSyntheticLeadingComment(
+        bucketNameComment2,
+        ts.SyntaxKind.SingleLineCommentTrivia,
+        ` name: '${renderArgs.storage.bucketName}',`,
```
iliapolo: this should be:

```ts
s3Bucket.bucketName = ...
```

dgandhi62: Updated!

PR #14392, opened 2025-12-10
feat: implementation plan prompt

sanjanaravikumar-az:
Pass a prompt like "Create implementation task list to migrate custom resources from AWS Amplify Gen1 to Gen2." along with implementation_plan_prompt.md as context to Q. It should be able to generate a task list to follow to complete the task. 

### DIFF
--- a/implementation_plan_prompt_v3.md
+++ b/implementation_plan_prompt_v3.md
@@ -5,8 +5,9 @@ You are a Senior Software Development Engineer that is excellent at creating tas
 ## Step 1: Analyze Feature Type
 
 First, determine the feature type:
-- **MIGRATION**: Feature exists in both Gen1 and Gen2 (custom resources, auth, API, storage)
-- **STANDALONE**: Feature that works across versions or is tooling-related (drift detection, rollback)
+- **MIGRATION**: Transform Gen1 code to Gen2 syntax (creates new resources)
+- **REFACTOR**: Update Gen2 code to reuse existing Gen1 resources (preserves resources)
+- **STANDALONE**: Migration tooling (drift detection, rollback, validation)
 
 ## Your Job (Conditional)
 
@@ -18,6 +19,15 @@ First, determine the feature type:
 5. **Find Differences** - Identify what changed between Gen1 and Gen2
 6. **List All Changes** - Document every difference that needs to be implemented
 7. **Create Task List** - Break down changes into atomic, sequential tasks
+
+**If REFACTOR Feature:**
+1. Identify Gen1 deployed resources (resource types, ARNs, names, logical IDs)
+2. Read Gen2 documentation for resource import/reference patterns
+3. **Find resource preservation strategies** (CloudFormation imports, CDK from* methods, logical ID preservation)
+4. **Identify stateful vs stateless resources** (which MUST be reused vs can be recreated)
+5. **Document resource mapping** (Gen1 resource → Gen2 reference method)
+6. **Find resource identifiers** (how to get ARNs, names, IDs from deployed Gen1 stack)
+7. **Create Task List** - Break down resource reuse into atomic, sequential tasks
 
 **If STANDALONE Feature:**
 1. Read current documentation for [Feature]
@@ -49,7 +59,7 @@ First, determine the feature type:
 ```markdown
 # Implementation Plan: [Feature Name]
 
-## Feature Type: [MIGRATION | NEW_FEATURE | STANDALONE]
+## Feature Type: [MIGRATION | REFACTOR | STANDALONE]
 
 ## What I Found in Documentation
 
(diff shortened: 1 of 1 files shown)


implementation_plan_prompt_v3.md
```diff
@@ -0,0 +1,178 @@
+# Implementation Plan Generator
+
+You are a Senior Software Development Engineer that is excellent at creating task lists, and you are tasked with creating a simple, actionable task list. When creating the task list, **you MUST follow the prompt line by line**. You create task lists for implementing features in Amplify. Your output is a simple, actionable task list in the form of implementation_plan.md.
```
iliapolo: Will this work? so it creates feature specific file names.

```suggestion
You are a Senior Software Development Engineer that is excellent at creating task lists, and you are tasked with creating a simple, actionable task list. When creating the task list, **you MUST follow the prompt line by line**. You create task lists for implementing features in Amplify. Your output is a simple, actionable task list in the form of [Feature]-implementation_plan.md.
```

implementation_plan_prompt_v3.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Step 1: Analyze Feature Type
+
+First, determine the feature type:
+- **MIGRATION**: Feature exists in both Gen1 and Gen2 (custom resources, auth, API, storage)
+- **NEW_FEATURE**: Feature only exists in Gen2 (new capabilities)
```
iliapolo: What would be an example of that?

implementation_plan_prompt_v3.md
```diff
@@ -0,0 +1,178 @@
...
+- **STANDALONE**: Feature that works across versions or is tooling-related (drift detection, CLI tools, dev utilities)
```
iliapolo: `CLI tools` and `dev utilities` feel out of place to me, Gen2 is authored in a separate repo and doesn't have the same CLI...

I think its better to avoid vague examples to keep Q focused on what we need.

implementation_plan_prompt_v3.md
```diff
@@ -0,0 +1,178 @@
...
+
+## Your Job (Conditional)
+
+**If MIGRATION Feature:**
+1. Read Gen1 documentation - find ALL ways to implement [Feature]
+2. Read Gen2 documentation - find ALL ways to implement [Feature] 
+3. **Extract example code from both docs**
+4. **Perform line-by-line comparison of examples**
```
iliapolo, line 18: This sounds relevant only to features that can be expressed in code in Gen1 (like overrides, custom resources, ...) - but many don't. How can we make this applicable to, for example, `refactor --rollback`?

implementation_plan_prompt_v3.md
```diff
@@ -0,0 +1,178 @@
...
+## Your Job (Conditional)
+
+**If MIGRATION Feature:**
+1. Read Gen1 documentation - find ALL ways to implement [Feature]
+2. Read Gen2 documentation - find ALL ways to implement [Feature] 
+3. **Extract example code from both docs**
+4. **Perform line-by-line comparison of examples** 
+5. **Find Differences** - Identify what changed between Gen1 and Gen2
+6. **List All Changes** - Document every difference that needs to be implemented
+7. **Create Task List** - Break down changes into atomic, sequential tasks
```
iliapolo, line 21: Why does it have to be sequential?

implementation_plan_prompt_v3.md
```diff
@@ -0,0 +1,178 @@
...
+# Implementation Plan: [Feature Name]
+
+## Feature Type: [MIGRATION | NEW_FEATURE | STANDALONE]
+
+## What I Found in Documentation
+
+**[IF MIGRATION Feature]:**
+### Gen1 Implementation Methods
+1. **Method 1**: [Name] - [Brief description]
+   - File path: `[exact path from docs]`
```
iliapolo, line 70: Will not always be relevant. Can we instead ask it to provide an "example of how to use/configure the feature in Gen1"?

implementation_plan_prompt_v3.md
```diff
@@ -0,0 +1,178 @@
...
+- **Integration changes**: How it connects to existing code
+- **Dependencies**: What external libraries or services are needed
+- **Configuration**: What settings or options need to be supported
+
+## Task List
+
+### Sequence Logically
+- Start with foundational changes (file structure, dependencies)
+- Then core implementation
+- Then integration/testing
```
iliapolo: Skip integration tests for now, we don't have a good example of this yet.

implementation_plan_prompt_v3.md
sanjanaravikumar-az: something like drift detection, or rollback. There were limited capabilities in gen1 so they can be considered "newer" features. If the wording is vague or confusing I can change it

implementation_plan_prompt_v3.md
sanjanaravikumar-az: got it, then i can remove the new feature and just keep the standalone, without CLI tools and dev utilities

implementation_plan_prompt_v3.md
sanjanaravikumar-az: yes, so the overrides and custom resources would be considered as a migration feature, along with the other supported categories. Rollback would be considered under a standalone feature. I'm thinking refactor might need some more specific instructions so I'll add a section that handles refactor as a separate category.

implementation_plan_prompt_v3.md
sanjanaravikumar-az: Q is better at handling sequential tasks than parallel, it performs better if it has a checklist type of instruction set so it can cross them off and move on to the next one. If it's parallel, it kind of tends to miss out on following through with the task. The sequential approach is just to make sure it stays focused on one task at a time.

implementation_plan_prompt_v3.md
sanjanaravikumar-az: yes, that would be helpful, i'll add this to the prompt

implementation_plan_prompt_v3.md
sanjanaravikumar-az: got it

implementation_plan_prompt_v3.md
sanjanaravikumar-az: made the change

implementation_plan_prompt_v3.md
sanjanaravikumar-az: made change

implementation_plan_prompt_v3.md
sanjanaravikumar-az: made change

implementation_plan_prompt_v3.md
sanjanaravikumar-az: added refactor portion

PR #14393, opened 2025-12-10
feat(gen2-migration): add additional auth provider escape hatches

dgandhi62:
This PR adds support for migrating additional authentication providers from Amplify Gen1 to Gen2 during the migration process. The implementation fetches additional auth providers directly from the AWS AppSync API and generates appropriate CDK escape hatches in the migrated Gen2 backend.

Source of truth - https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_appsync.CfnGraphQLApi.AdditionalAuthenticationProviderProperty.html

### Migration Flow
1. Extract AppSync API ID from `amplify-meta.json`
2. Fetch live auth provider configuration from AWS AppSync API
3. Generate typed `AdditionalAuthProvider` objects
4. Create CDK escape hatches in generated `backend.ts`

### Gen1 Configuration
AppSync API with additional Cognito User Pool auth provider configured in AWS console.

### Generated Gen2 Code
```typescript
// amplify/backend.ts
const backend = defineBackend({ auth, data });

// Additional auth providers escape hatch
const cfnGraphQLApi = backend.data.resources.cfnResources.cfnGraphQLApi;
cfnGraphQLApi.additionalAuthenticationProviders = [
  {
    authenticationType: 'AMAZON_COGNITO_USER_POOLS',
    userPoolConfig: {
      userPoolId: backend.auth.resources.userPool.userPoolId,
      awsRegion: 'us-east-1'
    }
  }
];

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts
@@ -10,6 +10,7 @@ import { BackendEnvironmentResolver } from './backend_environment_selector';
 import { BackendDownloader } from './backend_downloader';
 import { pathManager } from '@aws-amplify/amplify-cli-core';
 import { fileOrDirectoryExists } from './directory_exists';
+import { AppSyncClient, GetGraphqlApiCommand } from '@aws-sdk/client-appsync';
 
 /**
  * Fetches and processes data definitions from Amplify Gen1 projects for migration to Gen2.
@@ -112,7 +113,6 @@ export class DataDefinitionFetcher {
    */
   private getAdditionalAuthProvidersFromConsole = async (apiId: string): Promise<AdditionalAuthProvider[]> => {
     try {
-      const { AppSyncClient, GetGraphqlApiCommand } = require('@aws-sdk/client-appsync');
       const client = new AppSyncClient({});
       const response = await client.send(new GetGraphqlApiCommand({ apiId }));
 
@@ -125,8 +125,7 @@ export class DataDefinitionFetcher {
         })) || []
       );
     } catch (error) {
-      console.warn('Failed to fetch additional auth providers from AWS:', error.message);
-      return [];
+      throw new Error(`Failed to fetch additional auth providers from AWS: ${error.message}`);
     }
   };
 

### CODE REVIEW

File: packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts [hunk 1]
```diff
@@ -2,6 +2,26 @@ import ts, { ObjectLiteralElementLike, ObjectLiteralExpression } from 'typescrip
...
+  openIdConnectConfig?: {
+    authTtl?: number;
+    clientId?: string;
+    iatTtl?: number;
+    issuer: string;
+  };
+  userPoolConfig?: {
+    appIdClientRegex?: string;
+    awsRegion?: string;
+    userPoolId: string;
```
iliapolo: This should be populated with a reference to the `backend.auth...` and not the hardcoded gen1 id.

iliapolo (changes requested):

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
dgandhi62: Updated

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts
```diff
@@ -105,6 +107,29 @@ export class DataDefinitionFetcher {
     }
   };
 
+  /**
+   * Fetches additional authentication providers from AWS AppSync API.
+   */
+  private getAdditionalAuthProvidersFromConsole = async (apiId: string): Promise<AdditionalAuthProvider[]> => {
+    try {
+      const { AppSyncClient, GetGraphqlApiCommand } = require('@aws-sdk/client-appsync');
```
9pace: why not import at top with 
'import { AppSyncClient, GetGraphqlApiCommand } from '@aws-sdk/client-appsync';'
static import preferred over dynamic

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts
```diff
@@ -105,6 +107,29 @@ export class DataDefinitionFetcher {
...
+      return (
+        response.graphqlApi?.additionalAuthenticationProviders?.map((provider: any) => ({
+          authenticationType: provider.authenticationType,
+          ...(provider.lambdaAuthorizerConfig && { lambdaAuthorizerConfig: provider.lambdaAuthorizerConfig }),
+          ...(provider.openIDConnectConfig && { openIdConnectConfig: provider.openIDConnectConfig }),
+          ...(provider.userPoolConfig && { userPoolConfig: provider.userPoolConfig }),
+        })) || []
+      );
+    } catch (error) {
+      console.warn('Failed to fetch additional auth providers from AWS:', error.message);
```
9pace: silent error catching, shouldnt we throw error here to indicate failure?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/backend/synthesizer.ts
```diff
@@ -607,6 +609,83 @@ export class BackendSynthesizer {
...
+            ),
+          );
+        }
+
+        // Replace hardcoded userPoolId with backend.auth reference
+        if (provider.userPoolConfig.userPoolId) {
+          userPoolConfigProps.push(
+            factory.createPropertyAssignment(
+              factory.createIdentifier('userPoolId'),
+              this.createPropertyAccessExpression(factory.createIdentifier('backend'), 'auth.resources.userPool.userPoolId'),
```
9pace, line 651: conditional logic for existence of backend.auth needed

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts
dgandhi62: Trying to follow modularity. Do we have a known project preference on this?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts
dgandhi62: Could you elaborate on how this categorizes as silent error catching? I'm printing a warning to the console. 

The design decision I'm making is that `generate` should work regardless of additional auth mode escape hatching

9pace: Is this the design choice we are making elsewhere, I thought if we can't generate accurately we throw an error. If we are in the code block then supposedly we have additional auth mode. But if intended to just skip then thats fine.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts
9pace: Not from what I have seen i think that its usually best practice to static import due to stronger typing and dependency mappings, technically small detail and i think its okay to do like this but I think i have seen tsc compiler complain about require style imports before.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts
dgandhi62: @iliapolo any inputs? I think this is a non-essential feature and that migration shouldn't be blocked over its functionality (until we have a full understanding of how gen2 handles string schema inputs)

iliapolo: @dgandhi62 I think its better to throw an error rather than risking generating an incomplete app, especially is the error is really unexpected. From a user perspective, wouldn't it be easier to solve this error instead of an unknown build/deploy time error?

Its true that if a user doesn't happen to need additional auths they wouldn't mind this failure, but thats an assumption on the app structure that I don't think the code should be making.

iliapolo: In general - fail early, unless there is a viable logical scenario where an error would occur

dgandhi62: Changed

PR #14394, opened 2025-12-10
feat(gen2-migration): add logging configuration support for api category

dgandhi62:
Adds support for migrating AppSync logging configuration from Gen1 to Gen2 `defineData()`. Previously, logging settings were lost during migration.

Reference - https://github.com/aws-amplify/amplify-backend/blob/ffdc9dc5deea61bef70848c5efdf67ef3ee13ad0/packages/backend-data/src/types.ts#L151

https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-appsync-graphqlapi-logconfig.html

### Example
**Gen1 AppSync API with logging:**
- Field resolver log level: `ERROR`
- Exclude verbose content: `true`

**Generated Gen2 code:**
```typescript
  schema,
  logging: {
    fieldLogLevel: "error",
    excludeVerboseContent: true
  }
});

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
@@ -315,10 +315,8 @@ export const generateDataSource = async (dataDefinition?: DataDefinition): Promi
   // Add logging configuration if available
   if (dataDefinition?.logging) {
     if (dataDefinition.logging === true) {
-      // Simple boolean logging configuration
       dataRenderProperties.push(factory.createPropertyAssignment('logging', factory.createTrue()));
     } else if (typeof dataDefinition.logging === 'object') {
-      // DataLogConfig object
       const loggingConfig = dataDefinition.logging;
       const loggingProperties: ObjectLiteralElementLike[] = [];
 

### CODE REVIEW

File: packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/data_definition_fetcher.ts [hunk 1]
```diff
@@ -129,6 +129,28 @@ export class DataDefinitionFetcher {
     }
   };
 
+  /**
+   * Fetches logging configuration from AWS AppSync API.
+   */
+  private getLoggingConfigFromConsole = async (apiId: string) => {
+    try {
+      const { AppSyncClient, GetGraphqlApiCommand } = require('@aws-sdk/client-appsync');
```
9pace: Not required, import already at top of file.
```suggestion
```

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
```diff
@@ -312,6 +312,33 @@ export const generateDataSource = async (dataDefinition?: DataDefinition): Promi
...
+    if ('excludeVerboseContent' in loggingConfig && loggingConfig.excludeVerboseContent !== undefined) {
+      loggingProperties.push(
+        factory.createPropertyAssignment(
+          'excludeVerboseContent',
+          loggingConfig.excludeVerboseContent ? factory.createTrue() : factory.createFalse(),
+        ),
+      );
+    }
+
+    if ('retention' in loggingConfig && loggingConfig.retention !== undefined) {
```
9pace: from what i can tell there is no retention property in loggingConfig from appysnc
https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-properties-appsync-graphqlapi-logconfig.html

what case is this for?

9pace: also missing cloudwatch logs role arn property?

9pace (changes requested):

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
dgandhi62: What you're looking at is the cloudformation property. Amplify has apis on top of it. Will update the reference link in the pr message

Source of truth - https://github.com/aws-amplify/amplify-backend/blob/ffdc9dc5deea61bef70848c5efdf67ef3ee13ad0/packages/backend-data/src/types.ts#L151

9pace: I see thank you for your clarification

9pace (approved):

PR #14397, opened 2025-12-11
feat(gen2-migration): copy all function source files during Gen 2 migration

dgandhi62:
Currently, Gen 2 migration only copies the main `index.js` file from Gen 1 functions, leaving behind utility files, configurations, and dependencies that the handler imports. This breaks migrated functions that rely on additional source files.

### Solution
- Copy entire `src/` directory from Gen 1 functions
- Preserve original file extensions (no forced `.ts` conversion)
- Skip package management files (`package.json`, lock files)
- Rename main handler while keeping extension (`index.js` → `handler.js`)
- Fail fast with clear error when source files are missing

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/functions/index.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/functions/index.ts
@@ -165,7 +165,7 @@ export function createFunctionDefinition(
 ) {
   const defineFunctionProperties: ObjectLiteralElementLike[] = [];
 
-  // Always set entry point - Lambda functions require one
+  // Fallback to index.js if there is no entry
   const entryPoint = definition?.entry || './index.js';
   defineFunctionProperties.push(createParameter('entry', factory.createStringLiteral(entryPoint)));
   if (definition?.name) {


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
```diff
@@ -106,6 +106,46 @@ export interface Gen2RenderingOptions {
...
+    for (const entry of srcEntries) {
+      if (entry.isFile()) {
+        const file = path.relative(gen1SrcDir, path.join(entry.parentPath || entry.path, entry.name));
+        const fileName = path.basename(file);
+        const skipFiles = ['package.json', 'package-lock.json', 'yarn.lock', 'pnpm-lock.yaml'];
+
+        if (!skipFiles.includes(fileName)) {
+          const srcPath = path.join(gen1SrcDir, file);
+          const content = await fs.readFile(srcPath, 'utf-8');
+          const destFile = file.startsWith('index.') ? `handler${path.extname(file)}` : file;
```
iliapolo: Why rename? It may override a utility file called `handler.js`. Lets preserve `index.js` and configure `handler: index.js` on the function resource.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
```diff
@@ -106,6 +106,46 @@ export interface Gen2RenderingOptions {
...
+  resourceName: string,
+  destDir: string,
+  fileWriter: (content: string, path: string) => Promise<void>,
+): Promise<void> => {
+  try {
+    const gen1SrcDir = path.join('amplify', 'backend', 'function', resourceName, 'src');
+    const srcEntries = await fs.readdir(gen1SrcDir, { recursive: true, withFileTypes: true });
+
+    for (const entry of srcEntries) {
+      if (entry.isFile()) {
```
iliapolo, line 125: What if its a directory that contains necessary utility files?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
```diff
@@ -106,6 +106,46 @@ export interface Gen2RenderingOptions {
...
+        const file = path.relative(gen1SrcDir, path.join(entry.parentPath || entry.path, entry.name));
```
iliapolo: How can `entry.parentPath` be `undefined`?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
dgandhi62: `recursive: true` handles this. `readdir` finds files inside directories as separate entries. `mkdir` preserves the directory structure

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
dgandhi62: Changed! I was unsure what parentPath defaults to when used at the root but I checked and its an empty string

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
dgandhi62: This makes sense. Updated code

PR #14399, opened 2025-12-12
feat(gen2-migration): merge Gen 1 function dependencies into Gen 2 package.json during migration

dgandhi62:
This PR enables functionality to automatically extract and merge dependencies from all Gen 1 Lambda function `package.json` files into the generated Gen 2 project `package.json` 

**_NOTE:_** In the case where we have two lambda functions which rely on a conflicting and incompatible version dependencies, the lambda requiring the older dependency may break. This is due to the inherent limitations introduced in gen2 by unifying dependencies to one package.json.

### Example
**Before:** Gen 2 migration would lose function dependencies
```json
// Gen 1 function package.json
{
  "dependencies": {
    "lodash": "^4.17.0",
    "axios": "^1.0.0"
  }
}

// Generated Gen 2 package.json (missing function deps)
{
  "name": "my-gen2-app",
  "dependencies": {},
  "devDependencies": {
    "@aws-amplify/backend": "^1.18.0"
  }
}
```

**After:**
```json
// Generated Gen 2 package.json (includes function deps)
{
  "name": "my-gen2-app", 
  "dependencies": {
    "lodash": "^4.17.0",
    "axios": "^1.0.0"
  },
  "devDependencies": {
    "@aws-amplify/backend": "^1.18.0"
  }
}
```

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
@@ -129,7 +129,7 @@ const extractGen1FunctionDependencies = async (
 };
 
 /**
- * Merges dependencies from all Gen 1 functions
+ * Merges dependencies from all Gen 1 functions, selecting highest version for conflicts
  * @param functions - Array of function definitions
  * @returns Combined dependencies and devDependencies
  */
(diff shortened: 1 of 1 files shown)


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/core/migration-pipeline.ts
```diff
@@ -106,6 +106,50 @@ export interface Gen2RenderingOptions {
...
+const mergeAllFunctionDependencies = async (
+  functions: FunctionDefinition[],
+): Promise<{ dependencies: Record<string, string>; devDependencies: Record<string, string> }> => {
+  const functionDeps: Record<string, string> = {};
+  const functionDevDeps: Record<string, string> = {};
+
+  for (const func of functions) {
+    if (func.resourceName) {
+      const deps = await extractGen1FunctionDependencies(func.resourceName);
+      Object.assign(functionDeps, deps.dependencies || {});
```
9pace: Is it possible that two different lambdas will have the same package dependency with different version? If so then here it would only choose the 'latest' dependency version in array of functions. Could this break in that scenario?

dgandhi62: Agreed, updated

9pace (approved):

PR #14406, opened 2025-12-23
docs(gen2-migration): cloudformation custom resource implementation plan

sanjanaravikumar-az:
This is a task list output from using implemetation-plan-prompt.md. The task list extends the existing Gen1→Gen2 migration tool to support raw CloudFormation custom resources (customCloudformation service type) in addition to the already-supported CDK custom resources. 
After explicitly suggesting the plan uses CfnInclude, this final implementation plan was made.


9pace (changes requested):
Nice. Could you make a new directory (could be like docs/implementation-plans) just so that we don't have the docs scattered around in the root?

9pace (approved):

PR #14408, opened 2025-12-23
feat(gen2-migration): handling cdk custom resource dependencies

sanjanaravikumar-az:
### Overview
This PR implements comprehensive support for migrating Amplify Gen1 custom resource dependencies to Gen2 during the migration process. Previously, custom resources with AmplifyHelpers.addResourceDependency calls would fail migration with an error message directing users to manual migration documentation.

### Why Resource Dependencies Need Special Handling in Gen1 vs Gen2:

**Gen1 Dependency Model:**

- Uses CloudFormation-based dependency injection via AmplifyHelpers.addResourceDependency()
- Dependencies are resolved at deployment time through CloudFormation stack references
- Resources access other resources through CloudFormation outputs and parameters
- Example: amplifyResources.auth.userPool.userPoolId resolves to CloudFormation stack outputs

**Gen2 Dependency Model:**

- Uses TypeScript constructor injection with direct object references
- Dependencies are resolved at build time through the backend definition
- Resources receive dependency objects as constructor parameters
- Example: auth.resources.userPool.userPoolId directly accesses the backend auth construct

**Key Changes**

1. Dependency Extraction (extractResourceDependencies)

- Parses Gen1 custom resource files to identify AmplifyHelpers.addResourceDependency calls
- Extracts category dependencies (auth, storage, api, function) from dependency arrays
- Maps Gen1 categories to Gen2 backend property names

2. AST Transformation (AmplifyHelperTransformer)

- Dependency Tracking: Identifies and removes AmplifyHelpers.addResourceDependency variable statements
- Constructor Transformation: Adds resource dependency parameters to custom resource constructor
- Property Access Mapping: Transforms Gen1 resource references to Gen2 equivalents:
- amplifyResources.auth.userPool.userPoolId → auth.resources.userPool.userPoolId
- amplifyResources.storage.bucket.bucketName → storage.resources.bucket.bucketName
- amplifyResources.function.myFunc.functionArn → functions.myFunc.resources.lambda.functionArn

3. Backend Registration (BackendUpdater)

- Generates TypeScript imports for custom resources
- Creates instantiation statements with proper dependency injection
- Maps Gen1 categories to Gen2 backend properties (function→functions, api→data)

Gen1:

```
// Custom resource constructor
constructor(scope: Construct, id: string, props: AmplifyResourceProps) {
  super(scope, id, props);
  
  const dependencies = AmplifyHelpers.addResourceDependency(this, props.category, props.resourceName, [
    { category: "auth", resourceName: "userPool" },
    { category: "storage", resourceName: "s3Bucket" }
  ]);
  
  // Access resources through CloudFormation outputs
  const userPoolId = amplifyResources.auth.userPool.userPoolId;
}
```

Gen2:

```
// Custom resource constructor with dependency injection. This is in resource.ts.
constructor(scope: Construct, id: string, auth: any, storage: any) {
  super(scope, id);
  
  // Direct access to backend resources
  const userPoolId = auth.resources.userPool.userPoolId;
}

// Backend registration with dependencies. This is in backend.ts.
new myCustomResource(backend.createStack("myCustomResource"), "myCustomResource", backend.auth, backend.storage);
```


packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
```diff
@@ -104,6 +168,32 @@ export class AmplifyHelperTransformer {
                 return ts.factory.createStringLiteral('custom');
               }
             }
+
+            // Transform property access like amplifyResources.storage.bucket.bucketName
+            // Only transform the outermost property access (when this node is not part of a larger chain)
+            if (!ts.isPropertyAccessExpression(node.parent)) {
```
dgandhi62, line 180: Should this equation be the other way around? Not sure I fully follow

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
```diff
@@ -104,6 +168,32 @@ export class AmplifyHelperTransformer {
...
+              const fullAccess = AmplifyHelperTransformer.getPropertyAccessChain(node);
+              const parts = fullAccess.split('.');
+              // Match pattern: amplifyResources.category.resourceName.property (4+ parts)
+              if (parts.length >= 4 && parts[0].includes('amplifyResources')) {
+                const gen1Category = parts[1];
+                const gen2Category = AmplifyHelperTransformer.CATEGORY_MAP[gen1Category] || gen1Category;
```
dgandhi62, line 186: What about unsupported categories? Would this condition not break?

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
```diff
@@ -104,6 +168,32 @@ export class AmplifyHelperTransformer {
...
+                const resourceName = parts[2];
+                const gen1Attribute = parts[3];
+                const mappedAttribute = AmplifyHelperTransformer.ATTRIBUTE_MAP[gen1Category]?.[gen1Attribute];
```
dgandhi62, line 189: Same question here. What if the custom resource is dealing with an unsupported category?

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
```diff
@@ -104,6 +168,32 @@ export class AmplifyHelperTransformer {
...
+
+                // Functions need resource name preserved: functions.myFunc.resources.lambda.functionArn
+                if (gen1Category === 'function') {
+                  return AmplifyHelperTransformer.createPropertyAccessFromString(
+                    `${gen2Category}.${resourceName}.resources.${gen2Property}`,
+                  );
+                }
+
+                // Other categories: auth.resources.userPool.userPoolId
+                return AmplifyHelperTransformer.createPropertyAccessFromString(`${gen2Category}.resources.${gen2Property}`);
```
dgandhi62, line 200: Does this also handle unsupported?

packages/amplify-cli/src/commands/gen2-migration/codegen-custom-resources/transformer/amplify-helper-transformer.ts
```diff
@@ -1,11 +1,45 @@
...
+      UserPoolId: 'userPool.userPoolId',
+      UserPoolArn: 'userPool.userPoolArn',
+      IdentityPoolId: 'identityPool.identityPoolId',
+      AppClientID: 'userPoolClient.userPoolClientId',
+      AppClientIDWeb: 'userPoolClient.userPoolClientId',
+    },
+    api: {
+      GraphQLAPIIdOutput: 'cfnResources.cfnGraphqlApi.attrApiId',
+      GraphQLAPIEndpointOutput: 'cfnResources.cfnGraphqlApi.attrGraphQlUrl',
+      GraphQLAPIKeyOutput: 'cfnResources.cfnGraphqlApi.attrApiKey',
```
dgandhi62: Can you link me to where this property exists in cdk?

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
sanjanaravikumar-az: So we want to only transform the outermost (complete) property access chain, not intermediate nodes. When TypeScript AST visits amplifyResources.storage.bucket.bucketName, it visits each nested PropertyAccessExpression node from innermost to outermost:
```
 amplifyResources.storage
``` 
(parent is PropertyAccessExpression)
```
amplifyResources.storage.bucket
``` 
(parent is PropertyAccessExpression)
```
amplifyResources.storage.bucket.bucketName 
```
(parent is NOT PropertyAccessExpression - it's the statement/expression using it)

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
sanjanaravikumar-az: Right now the unsupported categories won't break, they'll pass through unchanged. If the category isn't in the map, it'll have the gen1 name, and the attribute map also uses optional chaining, so it will fall back to the gen1 property path

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
sanjanaravikumar-az: Yes, it would be the same as above, the fallbacks prevent breaking but there wouldn't be any transformation, the unsupported categories would have the original gen1 property path.

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
sanjanaravikumar-az: An unsupported category like amplifyResources.notifications.myResource.myAttr would transform to notifications.resources.myAttr, so it won't necessarily break, but there is a risk of the generated path not being valid in gen2.

packages/amplify-cli/src/commands/gen2-migration/codegen-custom-resources/transformer/amplify-helper-transformer.ts
sanjanaravikumar-az: https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_appsync.CfnGraphQLApi.html#attrgraphqlurl
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_appsync.CfnGraphQLApi.html#attrapiid
https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_appsync.CfnApiKey.html#attrapikey
I just made a small change to the last property

dgandhi62: attrApiKey seems like a property of CfnApiKey not CfnGraphqlApi. Why are we using the latter? Just trying to understand

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
dgandhi62: That was what I meant by breaking -> do we know if the path for gen2 properties is necessarily`${gen2Category}.${resourceName}.resources.${gen2Property}` where `gen2Category` is the same as `gen1Category`. Is it possible for there to be differences in naming between unsupported gen1 and gen2

packages/amplify-cli/src/commands/gen2-migration/codegen-custom-resources/transformer/amplify-helper-transformer.ts
dgandhi62: Ah I see this is resolved

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
```diff
@@ -21,7 +21,8 @@ export class AmplifyHelperTransformer {
     api: {
       GraphQLAPIIdOutput: 'cfnResources.cfnGraphqlApi.attrApiId',
       GraphQLAPIEndpointOutput: 'cfnResources.cfnGraphqlApi.attrGraphQlUrl',
-      GraphQLAPIKeyOutput: 'cfnResources.cfnGraphqlApi.attrApiKey',
+      // Note: API key is on CfnApiKey construct, not CfnGraphQLApi
+      GraphQLAPIKeyOutput: 'cfnResources.cfnApiKey.Default.attrApiKey',
```
dgandhi62, line 25: Why is there a default here?

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
dgandhi62: Understood

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
sanjanaravikumar-az: it's a general assumption, because that's the pattern all resources follow, but if we think there might be a mismatch then we can try to figure out a different way to handle unsupported categories

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
sanjanaravikumar-az: In Gen2's Amplify Data construct, cfnResources.cfnApiKey returns a Record<string, CfnApiKey> (a map of API keys by name), not a single CfnApiKey directly. The key "Default" is the name Amplify assigns to the default API key when you enable API key auth.

sanjanaravikumar-az:
> How were these changes tested? If manually tested, can you include the steps to reproduce manual verificatio

Yes, I manually tested it:

1. Set up an Amplify test app with various resource dependency patterns in cdk-stack.ts using Gen1 syntax (e.g., amplifyResources.auth.userPool.userPoolId)
2. Executed the lock and generate migration commands
3. Compared the generated output line-by-line against expected Gen2 syntax, validated against official Gen2 documentation and codebase examples

All transformations produced the correct Gen2 paths.

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
dgandhi62: I suggest creating a mapping for unsupported categories. Can collaborate with @sai-ray on this

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
dgandhi62: Lets put a pin in this and note this assumption explicitly in the pr message. Maybe we can have a small heading "assumptions" to keep it organized

packages/amplify-cli/src/commands/gen2-migration/generate/custom-resources/transformer/amplify-helper-transformer.ts
dgandhi62: Can we sync on this? I'm not sure I follow. cfnResources.cfnApiKey is a cdk construct. How is this related to amplify data?

PR #14410, opened 2025-12-24
feat(gen2-migration): detecting service type to distinguish between cdk and cfn custom resources

sanjanaravikumar-az:
### Overview
This PR adds service type detection for custom resources during Gen1 to Gen2 migration. It introduces the ability to distinguish between CDK-based (customCDK) and CloudFormation-based (customCloudformation) custom resources.

**Key Changes**

1. Custom Resource Service Type Detection (command-handlers.ts)

- Added ```CustomResourceServiceType``` type to distinguish between customCDK and customCloudformation resources
- Added ```getCustomResourcesWithType()``` function that reads from amplify-meta.json and returns a map of resource names to their service types
- This enables different handling paths for CDK vs CloudFormation custom resources during migration

2. API Key Attribute Mapping Fix (amplify-helper-transformer.ts)

- Reverted ```GraphQLAPIKeyOutput``` mapping back to ```cfnResources.cfnGraphqlApi.attrApiKey```
- Removed incorrect comment about CfnApiKey construct


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
```diff
@@ -322,6 +322,21 @@
   await fs.writeFile(`${cwd}/.gitignore`, newGitIgnore, { encoding: 'utf-8' });
 }
 
+type CustomResourceServiceType = 'customCDK' | 'customCloudformation';
+
+const getCustomResourcesWithType = (): Map<string, CustomResourceServiceType> => {
```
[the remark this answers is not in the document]
iankhou: Can you address this as either true positive or false positive?

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
```diff
@@ -492,6 +504,50 @@ const hasUncommentedDependency = (fileContent: string, matchString: string) => {
...
+        // Extract category values from dependency objects
+        const categoryRegex = /category:\s*['"]([^'"]+)['"]/g;
+        let categoryMatch;
+
+        while ((categoryMatch = categoryRegex.exec(dependenciesArray)) !== null) {
+          const category = categoryMatch[1];
+          if (!dependencies.includes(category)) {
+            dependencies.push(category);
+          }
+        }
```
iankhou, line 535: For this kind of logic I would generally like to see some kind of test, since regex is not verifiable by reading it (at least not by me). A unit test may be sufficient for now.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/codegen-head/command-handlers.ts
```diff
@@ -418,13 +436,7 @@ export async function updateCdkStackFile(customResources: string[], destinationC
...
       let cdkStackContent = await fs.readFile(cdkStackFilePath, { encoding: 'utf-8' });
 
-      // Check for existence of AmplifyHelpers.addResourceDependency and throw an error if found
-      if (hasUncommentedDependency(cdkStackContent, 'AmplifyHelpers.addResourceDependency')) {
-        cdkStackContent = cdkStackContent.replace(
-          /export class/,
-          `throw new Error('Follow https://docs.amplify.aws/react/start/migrate-to-gen2/ to update the resource dependency');\n\nexport class`,
-        );
-      }
+      // AmplifyHelpers.addResourceDependency is now supported by the transformer
```
iankhou: :)

iankhou (changes requested):
Looks good, but difficult for me to verify functionality by looking at it. How was it tested?

We are making a lot of changes to transformations here. I think it would be good to have unit tests that cover functionality introduced by the new parameter `resourceDependencies`.

PR #14413, opened 2025-12-29
fix(gen2-migration): function entry is not configured correctly

iliapolo:
#### Description of changes

> _AI Assisted with Kiro_

Currently, functions are being generated like so:

```ts
export const quotegenerator = defineFunction({
    entry: "index.handler",
    ...
});
```

However, `index.handler` is not a valid value for the `entry` property. Deployment causes the following error:

```console
5:53:39 PM [ERROR] [BackendBuildError] Unable to deploy due to CDK Assembly Error
  ∟ Caused by: [AssemblyError] Assembly builder failed
    ∟ Caused by: [NodeJSFunctionConstructInitializationError] Failed to instantiate nodejs function construct
      ∟ Caused by: [ValidationError] Only JavaScript or TypeScript entry files are supported.
```

Instead, it should point to the handler file, i.e `./index.js`. 

```ts
export const quotegenerator = defineFunction({
    entry: "./index.js",
    ...
});
```

Where `index.js` is the file we copied over from Gen1. This PR adds a function that derives the correct value of `entry` property from the function's raw `Handler` property.

Added unit tess

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/adapters/functions/index.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/adapters/functions/index.ts
@@ -17,6 +17,27 @@ export type AmplifyMetaWithFunction = {
   function: Record<string, AmplifyMetaFunction>;
 };
 
+/**
+ * Extracts the file path from an AWS Lambda handler string.
+ * Converts handler strings like "index.handler" or "src/handler.myFunction"
+ * to file paths like "./index.js" or "./src/handler.js".
+ *
+ * @param handler - The AWS Lambda handler string (e.g., "index.handler", "src/handler.myFunction")
+ * @returns The file path with .js extension (e.g., "./index.js", "./src/handler.js")
+ */
+const extractFilePathFromHandler = (handler: string): string => {
+  // Split on the last dot to separate file path from function name
+  const lastDotIndex = handler.lastIndexOf('.');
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/adapters/functions/index.ts
```diff
@@ -17,6 +17,27 @@ export type AmplifyMetaWithFunction = {
...
+ *
+ * @param handler - The AWS Lambda handler string (e.g., "index.handler", "src/handler.myFunction")
+ * @returns The file path with .js extension (e.g., "./index.js", "./src/handler.js")
+ */
+const extractFilePathFromHandler = (handler: string): string => {
+  // Split on the last dot to separate file path from function name
+  const lastDotIndex = handler.lastIndexOf('.');
+  if (lastDotIndex === -1) {
+    // No dot found, treat the whole string as the file path
+    return `./${handler}.js`;
```
dgandhi62, line 33: This makes sense

We are still assuming .js here however. I suppose the end goal is also to dynamically detect the extension based on the coding language?

iliapolo: Yeah. Once we introduce new languages this will likely change.

PR #14419, opened 2025-12-30
fix(gen2-migration): main.tsx in app4 imports the wrong file

dgandhi62:
Currently, app4 is importing a file that is only created in gen2. It needs to import a gen1 config file. 

Both aws_exports or amplifyconfiguration.json can be used. They contain the same information in different formats

### DIFF
--- a/amplify-migration-apps/app-4/src/main.tsx
+++ b/amplify-migration-apps/app-4/src/main.tsx
@@ -4,8 +4,8 @@ import './index.css';
 import App from './App.tsx';
 
 import { Amplify } from 'aws-amplify';
-import amplifyconfig from '../amplify_outputs.json';
-Amplify.configure(amplifyconfig);
+import awsExports from './aws-exports.js';
+Amplify.configure(awsExports);
 
 createRoot(document.getElementById('root')!).render(
   <StrictMode>


amplify-migration-apps/app-4/src/main.tsx
```diff
@@ -4,8 +4,8 @@ import './index.css';
 import App from './App.tsx';
 
 import { Amplify } from 'aws-amplify';
-import amplifyconfig from '../amplify_outputs.json';
-Amplify.configure(amplifyconfig);
+import awsExports from './aws-exports.js';
```
iliapolo, line 7: Why this one and not `amplifyconfiguration.json` like we normally do? Was that intentional?

dgandhi62: From my understanding, both files contain the same information just in different formats. They can be used interchangeably. amplifyconfiguration was introduced later by amplify to accomodate the configs in json.

dgandhi62:
> @dgandhi62 please rephrase the PR title to explain **what the bug is** - not just that **a bug exists**. The title should be a concise summary of the PR body - where you can (and should) go into more detail.

Apologies - will do. I was going through the list of changes to make and didn't spend a lot of time on this one. I'll keep that in mind

amplify-migration-apps/app-4/src/main.tsx
jiayiwang7: wait but TS can't infer types from JS directly without additional tsconfig right? importing JS without a declaration file will cause a build error

PR #14422, opened 2025-12-30
fix(gen2-migration): `cfnIdentityPool` constant is created for no reason

iliapolo:
#### Description of changes

Currently we generate the following statement:

```ts
const cfnIdentityPool = backend.auth.resources.cfnResources.cfnIdentityPool;
```

But the constant itself is not being used. This PR removes the constant if its not needed.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/backend/synthesizer.test.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/backend/synthesizer.test.ts
@@ -11,6 +11,54 @@ describe('BackendSynthesizer', () => {
     synthesizer = new BackendSynthesizer();
   });
 
+  describe('identity pool overrides', () => {
+    it('allowUnauthenticatedIdentities is escape hatched if guest login is false', () => {
+      const renderArgs: BackendRenderParameters = {
+        auth: {
+          importFrom: './auth/resource',
+          guestLogin: false,
+        },
+      };
+
+      const result = synthesizer.render(renderArgs);
+      const source = printNodeArray(result);
+      expect(source).toMatchInlineSnapshot(`
+        "import { auth } from "./auth/resource";
+        import { defineBackend } from "@aws-amplify/backend";
+
+
+        const backend = defineBackend({
+            auth
+        });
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/backend/synthesizer.ts
```diff
@@ -908,14 +908,14 @@ export class BackendSynthesizer {
     }
 
     // Identity pool overrides
-    if (renderArgs.auth?.guestLogin === false || (renderArgs.auth?.identityPoolName && !renderArgs?.auth?.referenceAuth)) {
-      const cfnIdentityPoolVariableStatement = this.createVariableStatement(
-        this.createVariableDeclaration('cfnIdentityPool', 'auth.resources.cfnResources.cfnIdentityPool'),
+    if (renderArgs.auth?.guestLogin === false && renderArgs.auth?.identityPoolName && !renderArgs?.auth?.referenceAuth) {
```
iliapolo: Looking at the code, the constant is only used if `renderArgs.auth?.guestLogin === false`, which means that this entire code block should only execute if both conditions are met.

dgandhi62: I'm not sure changing the condition is necessarily correct. Its possible I might be missing logic from the case where only the second part is true. Let me recheck this feature logic and I'll update you

dgandhi62: This check for `identityPoolName` is redundant. This is likely leftover logic from when we were actually generating the identity pool names in case they had been customized. We should just remove the second check altogether

dgandhi62 (changes requested):
We can remove the additional check of identityPoolNames for this

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/backend/synthesizer.ts
iliapolo: Done

PR #14435, opened 2026-01-05
fix(gen2-migration): functions are generated with incorrect API access env variables

iliapolo:
#### Description of changes

If you configure your function to have access to your GraphQL API, we currently generate:

```ts
const function = defineFunction({
  ...,
  env: { API_PRODUCTCATALOG_GRAPHQLAPIENDPOINTOUTPUT: 'https://<gen1-appsync-endpoint>', ... }
})
```

But the Gen2 function needs to point to the new Gen2 AppSync API. So really, it should be:

```ts
const function = defineFunction({
  ...,
  env: { API_PRODUCTCATALOG_GRAPHQLAPIENDPOINTOUTPUT: backend.data.graphqlEndpoint, ... }
})
```

We don't have access to the `backend` instance in this part of the code so the fix would have to be different. In the meantime, we remove these env variables since they are wrong anywa.

Unit test + Manual integ test.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/adapters/functions/index.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/adapters/functions/index.ts
@@ -2,14 +2,6 @@ import { FunctionDefinition } from '../../core/migration-pipeline';
 import { FunctionConfiguration } from '@aws-sdk/client-lambda';
 import assert from 'node:assert';
 
-const ENV_VARIABLES_TO_REMOVE = [
-  // we remove these because their value points to the Gen1 appsync API.
-  // we need to point to the new one.
-  'GRAPHQLAPIKEYOUTPUT',
-  'GRAPHQLAPIENDPOINTOUTPUT',
-  'GRAPHQLAPIIDOUTPUT',
-];
-
 export type AmplifyMetaFunction = {
   service: string;
   providerPlugin: 'awscloudformation';


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/adapters/functions/index.ts
```diff
@@ -52,6 +60,19 @@ export const getFunctionDefinition = (
     funcDef.name = configuration?.FunctionName;
     funcDef.timeoutSeconds = configuration?.Timeout;
     funcDef.memoryMB = configuration?.MemorySize;
+
+    // we remove these because their value points to the Gen1 appsync API.
+    // the correct value needs to come from `backend.data` attributes, which we don't have access to here
+    // since `backend` is configured in a different file. we can't import that file because it would create
+    // a circular import. instead, we need to generate some code in `backend.ts` (TODO)
+    for (const envSuffix of ['GRAPHQLAPIKEYOUTPUT', 'GRAPHQLAPIENDPOINTOUTPUT', 'GRAPHQLAPIIDOUTPUT']) {
```
iliapolo, line 60: These are known suffixes, as hinted to [here](https://github.com/aws-amplify/amplify-cli/blob/gen2-migration/packages/amplify-nodejs-function-template-provider/resources/lambda/appsync-request/index.js.ejs#L9).

dgandhi62: Noted

PR #14436, opened 2026-01-06
fix(gen2-migration): code structure update to data codegen

sanjanaravikumar-az:
This PR is to make some changes in the code related to graphql api codegen. Alot of functionalities were handled in one single function:
```
generateDataSource()
```
This function contained code for handling various responsibilities:
- Schema generation
- Table mapping generation
- Auth mode generation
- Logging generation
- Final assembly

The changes are basically just creating a specific function for each of these sections, so the code is modularized and easier to maintain. Testing was done by running the existing unit test, and 9/9 test cases passed. I also manually tested with app1(product-catalog).

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
@@ -29,16 +29,26 @@ export interface AdditionalAuthProvider {
  */
 interface Gen1AuthorizationConfig {
   defaultAuthentication?: {
-    authenticationType?: string;
+    authenticationType?: 'API_KEY' | 'AMAZON_COGNITO_USER_POOLS' | 'AWS_IAM' | 'OPENID_CONNECT' | 'AWS_LAMBDA';
     apiKeyConfig?: {
+      description?: string;
       apiKeyExpirationDays?: number;
+      apiKeyExpirationDate?: Date;
     };
-    lambdaAuthorizerConfig?: {
-      ttlSeconds?: number;
+    userPoolConfig?: {
+      userPoolId: string;
     };
     openIDConnectConfig?: {
-      issuer?: string;
+      name?: string;
+      issuerUrl?: string;
       clientId?: string;
+      iatTTL?: number;
+      authTTL?: number;
+    };
+    lambdaAuthorizerConfig?: {
+      lambdaFunction?: string;
+      lambdaArn?: string;
+      ttlSeconds?: number;
     };
   };
 }
@@ -303,15 +313,15 @@ const generateLambdaAuthMode = (lambdaConfig: { ttlSeconds?: number } | undefine
  * @param oidcConfig - OpenID Connect configuration from Gen1
  * @returns ObjectLiteralElementLike for oidcAuthorizationMode, or undefined
  */
-const generateOidcAuthMode = (oidcConfig: { issuer?: string; clientId?: string } | undefined): ObjectLiteralElementLike | undefined => {
+const generateOidcAuthMode = (oidcConfig: { issuerUrl?: string; clientId?: string } | undefined): ObjectLiteralElementLike | undefined => {
   if (!oidcConfig) {
     return undefined;
   }
 
   const oidcProps: ObjectLiteralElementLike[] = [];
 
-  if (oidcConfig.issuer) {
-    oidcProps.push(factory.createPropertyAssignment('oidcIssuerUrl', factory.createStringLiteral(oidcConfig.issuer)));
+  if (oidcConfig.issuerUrl) {
+    oidcProps.push(factory.createPropertyAssignment('oidcIssuerUrl', factory.createStringLiteral(oidcConfig.issuerUrl)));
   }
   if (oidcConfig.clientId) {
     oidcProps.push(factory.createPropertyAssignment('clientId', factory.createStringLiteral(oidcConfig.clientId)));
@@ -357,6 +367,10 @@ const generateAuthModesProperty = (authorizationModes: AuthorizationModes | unde
         break;
       case 'OPENID_CONNECT':
         modeSpecificConfig = generateOidcAuthMode(gen1AuthModes.defaultAuthentication?.openIDConnectConfig);
+        break;
+      case 'AWS_IAM':
+      case 'AMAZON_COGNITO_USER_POOLS':
+        // These auth modes don't require additional configuration
         break;
     }
 
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.test.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.test.ts
@@ -1,6 +1,6 @@
 import assert from 'node:assert';
 import { printNodeArray } from '../../test_utils/ts_node_printer';
-import { generateDataSource } from './index';
+import { generateDataSource, DataDefinition } from './index';
 
 describe('Data Category code generation', () => {
   beforeEach(() => {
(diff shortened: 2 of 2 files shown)


packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
```diff
@@ -221,161 +487,26 @@ export const generateDataSource = async (gen1Env: string, dataDefinition?: DataD
...
-      }
-    }
+  // Add logging configuration property if available
+  const loggingProperty = generateLoggingProperty(dataDefinition?.logging);
+  if (loggingProperty) {
+    dataRenderProperties.push(loggingProperty);
   }
 
-  // Add schema reference to the data configuration
-  if (dataDefinition?.schema) {
```
9pace, line 376: Why remove this conditional?
If dataDefinition.schema is undefined, this will add a reference to an undefined schema variable in the generated code.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
```diff
@@ -132,15 +165,275 @@ export type DataDefinition = {
...
+    factory.createIdentifier('`' + processedSchema + '`'),
+  );
+
+  const schemaStatementAssignment = factory.createVariableStatement(
+    [],
+    factory.createVariableDeclarationList([schemaVariableDeclaration], ts.NodeFlags.Const),
+  );
+  statements.push(schemaStatementAssignment);
+
+  return { statements, processedSchema };
```
9pace: I believe schemaResult.processedSchema is never accessed it is safe to not return this (and also remove the return type) as it only gets used internally.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
9pace: This type is from gen2s backend-data. To be consistent we should change the type to be `Gen1AuthorizationConfig` which would eliminate the unsafe casting that gets done on line 340:
`const gen1AuthModes = authorizationModes as Gen1AuthorizationConfig;`

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
```diff
@@ -132,15 +165,275 @@ export type DataDefinition = {
...
+ *
+ * @param authorizationModes - Authorization configuration from Gen1
+ * @returns ObjectLiteralElementLike for authorizationModes property, or undefined if no auth config
+ */
+const generateAuthModesProperty = (authorizationModes: AuthorizationModes | undefined): ObjectLiteralElementLike | undefined => {
+  if (!authorizationModes) {
+    return undefined;
+  }
+
+  const gen1AuthModes = authorizationModes as Gen1AuthorizationConfig;
```
9pace: Unsafe casting can be eliminated with the other comment on line 57, since you added the better fitting `Gen1AuthorizationConfig` type.

9pace (changes requested):

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
sanjanaravikumar-az: True, i added it back.

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
sanjanaravikumar-az: yes, i made a few changes, this shouldn't be a problem anymore

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
sanjanaravikumar-az: made the change

packages/amplify-cli/src/commands/gen2-migration/codegen-generate/src/generators/data/index.ts
sanjanaravikumar-az: done!

9pace (commented):
I think something weird happened when you force pushed to solve merge conflicts since I see the merge conflict markers in the actual code. same thing as what codeQL is complaining about.

sanjanaravikumar-az:
> I think something weird happened when you force pushed to solve merge conflicts since I see the merge conflict markers in the actual code. same thing as what codeQL is complaining about.

yeah i dont know how that happened, but fixed it

packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
```diff
@@ -23,8 +24,49 @@ export interface AdditionalAuthProvider {
   };
 }
 
+/**
+ * Gen1 authorization configuration structure from amplify-meta.json
+ */
+interface Gen1AuthorizationConfig {
```
9pace: From what i see in amplify-meta.json Authconfig looks like: 

```
"authConfig": {
  "additionalAuthenticationProviders": [],
  "defaultAuthentication": {
    "apiKeyConfig": {
      "apiKeyExpirationDays": 7
    },
    "authenticationType": "API_KEY"
  }
},
```

This config is generated from amplify-category-api. 

in `graphql-api-providers.ts` we export `AppSyncAuthConfiguration`  Which is the correct type. 

it looks like this:
```
export type AppSyncAuthConfiguration = {
  defaultAuthentication: AppSyncAuthConfigurationEntry;
  additionalAuthenticationProviders: Array<AppSyncAuthConfigurationEntry>;
};

export type AppSyncAuthConfigurationEntry =
  | AppSyncAuthConfigurationUserPoolEntry
  | AppSyncAuthConfigurationAPIKeyEntry
  | AppSyncAuthConfigurationIAMEntry
  | AppSyncAuthConfigurationOIDCEntry
  | AppSyncAuthConfigurationLambdaEntry;

export type AppSyncAuthConfigurationAPIKeyEntry = {
  authenticationType: 'API_KEY';
  apiKeyConfig?: ApiKeyConfig;
};
export type AppSyncAuthConfigurationUserPoolEntry = {
  authenticationType: 'AMAZON_COGNITO_USER_POOLS';
  userPoolConfig?: UserPoolConfig;
};
export type AppSyncAuthConfigurationIAMEntry = {
  authenticationType: 'AWS_IAM';
};

export type AppSyncAuthConfigurationOIDCEntry = {
  authenticationType: 'OPENID_CONNECT';
  openIDConnectConfig?: OpenIDConnectConfig;
};

...
```

This is basically a manual reconstruction of this type, which I think is lacking the additionalAuthenticationProviders key. To be more complete we can instead:

`import { AppSyncAuthConfiguration } from '@aws-amplify/graphql-transformer-interfaces';`
and add dep to package.json.

and then we can use AppSyncAuthConfiguration for authmodes instead 
`  authorizationModes?: AppSyncAuthConfiguration; `

9pace, line 30: From what i see in amplify-meta.json Authconfig looks like: 

```
"authConfig": {
  "additionalAuthenticationProviders": [],
  "defaultAuthentication": {
    "apiKeyConfig": {
      "apiKeyExpirationDays": 7
    },
    "authenticationType": "API_KEY"
  }
},
```

This config is generated from amplify-category-api. 

in `graphql-api-providers.ts` we export `AppSyncAuthConfiguration`  Which is the correct type. 

it looks like this:
```
export type AppSyncAuthConfiguration = {
  defaultAuthentication: AppSyncAuthConfigurationEntry;
  additionalAuthenticationProviders: Array<AppSyncAuthConfigurationEntry>;
};

export type AppSyncAuthConfigurationEntry =
  | AppSyncAuthConfigurationUserPoolEntry
  | AppSyncAuthConfigurationAPIKeyEntry
  | AppSyncAuthConfigurationIAMEntry
  | AppSyncAuthConfigurationOIDCEntry
  | AppSyncAuthConfigurationLambdaEntry;

export type AppSyncAuthConfigurationAPIKeyEntry = {
  authenticationType: 'API_KEY';
  apiKeyConfig?: ApiKeyConfig;
};
export type AppSyncAuthConfigurationUserPoolEntry = {
  authenticationType: 'AMAZON_COGNITO_USER_POOLS';
  userPoolConfig?: UserPoolConfig;
};
export type AppSyncAuthConfigurationIAMEntry = {
  authenticationType: 'AWS_IAM';
};

export type AppSyncAuthConfigurationOIDCEntry = {
  authenticationType: 'OPENID_CONNECT';
  openIDConnectConfig?: OpenIDConnectConfig;
};

...
```

This is basically a manual reconstruction of this type, which I think is lacking the additionalAuthenticationProviders key. To be more complete we can instead:

`import { AppSyncAuthConfiguration } from '@aws-amplify/graphql-transformer-interfaces';`
and add dep to package.json.

and then we can use AppSyncAuthConfiguration for authmodes instead 
`  authorizationModes?: AppSyncAuthConfiguration; `

packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
```diff
@@ -221,158 +494,24 @@ export const generateDataSource = async (gen1Env: string, dataDefinition?: DataD
...
-      if (loggingConfig.retention !== undefined) {
-        loggingProperties.push(factory.createPropertyAssignment('retention', factory.createStringLiteral(loggingConfig.retention)));
-      }
-
-      if (loggingProperties.length > 0) {
-        dataRenderProperties.push(factory.createPropertyAssignment('logging', factory.createObjectLiteralExpression(loggingProperties)));
-      }
-    }
+  // Add logging configuration property if available
+  const loggingProperty = generateLoggingProperty(dataDefinition?.logging);
```
9pace, line 509: A little weird to me to call the function first on something you are not sure is populated and then after check if it returned undefined, Instead of checking if the property is populated and then calling if it is.

Maybe this is just style I am not used to? 
It works either way, so just preference.

PR #14455, opened 2026-01-09
feat(gen2-migration): lambda access to api

dgandhi62:
This pr implements CloudFormation template analysis to detect precise GraphQL permissions (query, mutation, subscription) and generate Gen2 permission grants.

Our scanner generates something like this (just an example) -
```
Function A: { hasQuery: false, hasMutation: false, hasSubscription: false } → FILTERED OUT
Function B: { hasQuery: true, hasMutation: false, hasSubscription: false }  → INCLUDED
Function C: { hasQuery: true, hasMutation: true, hasSubscription: true }    → INCLUDED

Result: Only Functions B & C get grant calls in backend.ts
```
This lets us generate cdk calls in the backend.ts
Source of truth - https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_appsync.GraphqlApi.html

### Example 
Input
testfunctionaccess-cloudformation-template.json
```json
{
              "Effect": "Allow",
              "Action": [
                "appsync:GraphQL"
              ],
              "Resource": [
                {
                  "Fn::Join": [
                    "",
                    [
                      "arn:aws:appsync:",
                      {
                        "Ref": "AWS::Region"
                      },
                      ":",
                      {
                        "Ref": "AWS::AccountId"
                      },
                      ":apis/",
                      {
                        "Ref": "apidiscussionsGraphQLAPIIdOutput"
                      },
                      "/types/Query/*"
                    ]
                  ]
                },
                {
                  "Fn::Join": [
                    "",
                    [
                      "arn:aws:appsync:",
                      {
                        "Ref": "AWS::Region"
                      },
                      ":",
                      {
                        "Ref": "AWS::AccountId"
                      },
                      ":apis/",
                      {
                        "Ref": "apidiscussionsGraphQLAPIIdOutput"
                      },
                      "/types/Mutation/*"
                    ]
                  ]
                }
              ]
            },
```

Output
```
// testfunctionaccess: Mutation + Subscription only
backend.data.resources.graphqlApi.grantMutation(backend.notifierFunction.resources.lambda);
backend.data.resources.graphqlApi.grantSubscription(backend.notifierFunction.resources.lambda);

```


iliapolo (changes requested):
Looks like this PR is not in the right state as it has unexpected code that relies on incorrect IAM action names (e.g `appsync:GraphQLQuery`). Please rework it.

PR #14471, opened 2026-01-15
chore(gen2-migration): fitness tracker app and various fixes

iliapolo:
Apologies in advance for the large and wide PR. There are several small issues I ran into and fixed during testing this app.

##### CLI Changes

- Fix a bug with `preSignup` auth trigger - it didn't map correctly to Cognito triggers.
- Added deletion of `package-lock.json` and `node_modules` as part of generate to create a more reliable lock file.
- Comment out fetching of rest apis to prevent code generating them - the generated code needs more work.
- Remove another reliance of the local `amplify-meta.json` file I found.
- Use pagination to fetch AppSync APIs - this is an attempt to fix the non existing dynamo mappings we are seeing from time to time. Theory is we were only fetching the first page and thus not finding the expected API. 

#### App Changes

- Removed app-5. It didn't add to our test coverage apart from stress testing a large schema which we will re-add later on.
- Renamed app-4 to fitness-tracker.
- Added preSignup auth trigger that does email domain validation.
- Simplify rest api (and frontend) by providing a single route that write to a dynamo model (e.g `Meal`). This model is then queried by the frontend.
- The app was still using project-boards terminology in the code because it was based on it. Renamed to fitness tracker terms.

Manual deployment and migration of fitness tracker app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.test.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.test.ts
@@ -11,7 +11,8 @@ describe('Data Category code generation', () => {
     jest.restoreAllMocks();
   });
   it('generates the correct import', async () => {
-    const source = printNodeArray(await generateDataSource('main', { schema: 'type Test { id: ID! }' }));
+    const tableMappings = { Todo: 'my-todo-mapping' };
+    const source = printNodeArray(await generateDataSource('main', { tableMappings, schema: 'type Test { id: ID! }' }));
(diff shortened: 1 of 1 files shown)


amplify-migration-apps/fitness-tracker/src/main.tsx
```diff
@@ -4,7 +4,7 @@ import './index.css';
 import App from './App.tsx';
 
 import { Amplify } from 'aws-amplify';
-import awsExports from './aws-exports.js';
+import awsExports from './amplifyconfiguration.json';
```
iliapolo:
```suggestion
import awsExports from './aws-exports.js';
```

PR #14475, opened 2026-01-15
docs(gen2-migration): add ddb refactor instructions for gen2-migration guide

9pace:
Add instructions based on ddb refactor commit changes.

### DIFF
--- a/GEN2_MIGRATION_GUIDE.md
+++ b/GEN2_MIGRATION_GUIDE.md
@@ -379,6 +379,8 @@ If your application contains an S3 bucket as part of the storage category, edit 
 > This is required in order to sync your local bucket name with the deployed template. 
 Otherwise, pushing changes to the `gen2-main` branch will result in a bucket replacement.
 
+Commit the changes:
+
 ```console
 git add .
 git commit -m "fix: reuse gen1 storage bucket"
@@ -396,6 +398,8 @@ If your application contains a DynamoDB table as part of the storage category, e
 
 > This is required in order to sync your local table name with the deployed template. 
 Otherwise, pushing changes to the `gen2-main` branch will result in a table replacement.
+
+Commit the changes:
 
 ```console
 git add .


GEN2_MIGRATION_GUIDE.md
```diff
@@ -379,11 +379,34 @@ If your application contains an S3 bucket as part of the storage category, edit
 > This is required in order to sync your local bucket name with the deployed template. 
 Otherwise, pushing changes to the `gen2-main` branch will result in a bucket replacement.
 
-And push the changes:
```
sai-ray, line 382: I think we should add this back or something similar.

GEN2_MIGRATION_GUIDE.md
```diff
@@ -379,11 +379,34 @@ If your application contains an S3 bucket as part of the storage category, edit
...
+
+```diff
+- new Table(storageStack, "myTable", { partitionKey: ... });
+- // Add this property to the Table above post refactor: tableName: 'my-table-main'
++ new Table(storageStack, "myTable", { tableName: 'my-table-main', partitionKey: ... });
+```
+
+> This is required in order to sync your local table name with the deployed template. 
+Otherwise, pushing changes to the `gen2-main` branch will result in a table replacement.
+
```
sai-ray, line 401: Same here

GEN2_MIGRATION_GUIDE.md
9pace: I separated this out into a new step #### 4.3 Deploy
Thats because if you have both a s3 and ddb, you need to make both changes before the push

GEN2_MIGRATION_GUIDE.md
9pace: See other comment

GEN2_MIGRATION_GUIDE.md
sai-ray: Yeah makes sense. How about "commit the changes:"?

PR #14497, opened 2026-01-20
feat(gen2-migration): add test scripts to discussions app

sanjanaravikumar-az:
This script provides frontend-backend testing for the discussions app's Amplify backend.

What it tests:

- GraphQL queries (listTopics, getTopic, listPosts, getPost, listComments, getComment, fetchUserActivity)
- GraphQL mutations (CRUD operations for Topics, Posts, and Comments)
- User activity tracking via Lambda function

How it works:

- Runs public queries first to verify data retrieval (Topics, Posts, Comments)
- Tests filtered queries (posts by topic, comments by post)
- Authenticates via Cognito user pool
- Performs full CRUD cycle on Topics (create → update → delete)
- Performs full CRUD cycle on Posts linked to Topics (create → update → delete)
- Performs full CRUD cycle on Comments linked to Posts (create → update → delete)
- Tests user activity tracking for the authenticated user
- Cleans up test data in reverse order (comments → posts → topics) and signs out

Users need to update TEST_USER credentials before running authenticated tests.

### DIFF
--- a/amplify-migration-apps/discussions/test-utils.ts
+++ b/amplify-migration-apps/discussions/test-utils.ts
@@ -263,12 +263,12 @@ export function createQueryTests(queries: GraphQLQueries) {
 // Mutation Test Functions
 // ============================================================
 export function createMutationTests(mutations: GraphQLMutations, getCurrentUserId: () => string | null) {
-  async function testCreateTopic(discussionId: string = 'tech'): Promise<string | null> {
+  async function testCreateTopic(): Promise<string | null> {
     console.log('\n🆕 Testing createTopic...');
     const publicClient = generateClient({ authMode: 'apiKey' });
 
     const topicName = `Test Topic ${Date.now()}`;
-    const content = `${discussionId}:${topicName}`;
+    const content = `tech:${topicName}`;
 
     const result = await publicClient.graphql({
       query: mutations.createTopic,
@@ -563,7 +563,7 @@ async function runTopicMutationTests(
   console.log('='.repeat(60));
 
   // Create topic
-  const topicId = await runTest('createTopic', () => mutationTests.testCreateTopic('tech'));
+  const topicId = await runTest('createTopic', () => mutationTests.testCreateTopic());
   if (!topicId) {
     console.log('❌ Failed to create topic, skipping remaining topic tests');
     return null;


amplify-migration-apps/discussions/gen2-test-script.ts
```diff
@@ -0,0 +1,579 @@
...
+// Polyfill crypto for Node.js environment (required for Amplify Auth)
+import { webcrypto } from 'crypto';
+if (typeof globalThis.crypto === 'undefined') {
+  (globalThis as any).crypto = webcrypto;
+}
+
+import { generateClient } from 'aws-amplify/api';
+import { Amplify } from 'aws-amplify';
+import { signIn, signOut, getCurrentUser } from 'aws-amplify/auth';
+import amplifyconfig from './src/amplify_outputs.json';
```
9pace, line 15: This file is the same as the other test script, except for this line.
Can we consolidate the test script into one single file, that takes this as some parameter. 

then each gen1 and gen2 file can look something like

```
// gen1-test-script.ts
import amplifyconfig from './src/amplifyconfiguration.json';
import { runDiscussionsTests } from './test-runner'; // or whatever you want to call it


const TEST_USER = {
  username: 'YOUR_USERNAME_HERE',
  password: 'YOUR_PASSWORD_HERE',
}; 
^ this can go in the generic file too or here, it may make sense that we want a different user per app.


then run the generic shared test:
runDiscussionsTests({
  amplifyConfig: amplifyconfig,
  testUser: TEST_USER,
});
Or something along those lines
```

amplify-migration-apps/discussions/gen2-test-script.ts
```diff
@@ -0,0 +1,579 @@
...
+// ============================================================
+// CONFIGURATION - Update with your test user credentials
+// ============================================================
+const TEST_USER = {
+  username: 'YOUR_USERNAME_HERE', // Phone number format
+  password: 'YOUR_PASSWORD_HERE',
+};
+
+// Clients
+const publicClient = generateClient();
```
9pace: this `publicClient` uses the default auth config which is authorized via userpool as soon as we sign in using `authenticateUser`

When you call generateClient() without specifying an auth mode, it uses the default from config, but it can also automatically attach the signed-in user's credentials when available.

So when a user signs in via signIn(), subsequent requests from that client will include the Cognito token in the request headers.

This means that calls like `publicClient.graphql `after signup are authenticated via userpool. 

Also, when you do:
`const authClient = generateClient({ authMode: 'apiKey' });`

This actually creates a non authenticated client, by using default apikey instead of userpool authentication. 

The naming is inverted:
- publicClient → actually makes authenticated (userPool) requests after sign-in
- authClient → actually makes public (apiKey) requests

amplify-migration-apps/discussions/gen2-test-script.ts
```diff
@@ -0,0 +1,579 @@
...
+  });
+  if (activities.length > 10) console.log(`   ... and ${activities.length - 10} more`);
+}
+
+// ============================================================
+// Topic Mutation Test Functions
+// ============================================================
+async function testCreateTopic(discussionId: string = 'tech'): Promise<string | null> {
+  console.log('\n🆕 Testing createTopic...');
+  const authClient = generateClient({ authMode: 'apiKey' });
```
9pace: It is unnecessary to re init the client in every function, we can do it outside of the functions at the module level once instead.

```
const client = generateClient();  // Uses userPool after sign-in 
const apiKeyClient = generateClient({ authMode: 'apiKey' });
```

amplify-migration-apps/discussions/gen2-test-script.ts
```diff
@@ -0,0 +1,579 @@
...
+  }
+}
+
+// ============================================================
+// Query Test Functions
+// ============================================================
+async function testListTopics(): Promise<string | null> {
+  console.log('\n📋 Testing listTopics...');
+  const result = await publicClient.graphql({ query: listTopics });
+  const topics = (result as any).data.listTopics.items;
```
9pace: why do we typecast every result as any? (in all functions)
What happens if the graphql query fails?

amplify-migration-apps/discussions/gen2-test-script.ts
9pace: Also unsure if the intent here is that we test graphql queries with userpool auth or public key?

amplify-migration-apps/discussions/gen2-test-script.ts
9pace: also, should we handle errors, if not any failure will crash our test script

amplify-migration-apps/discussions/gen2-test-script.ts
```diff
@@ -0,0 +1,579 @@
...
+    const time = new Date(a.timestamp).toLocaleString();
+    console.log(`   - ${a.activityType.replace('_', ' ').toUpperCase()} at ${time}`);
+  });
+  if (activities.length > 10) console.log(`   ... and ${activities.length - 10} more`);
+}
+
+// ============================================================
+// Topic Mutation Test Functions
+// ============================================================
+async function testCreateTopic(discussionId: string = 'tech'): Promise<string | null> {
```
9pace: i would remove the hardcoded `tech` default since we pass it in later.

amplify-migration-apps/discussions/gen2-test-script.ts
```diff
@@ -0,0 +1,579 @@
...
+// ============================================================
+// Query Test Functions
+// ============================================================
+async function testListTopics(): Promise<string | null> {
+  console.log('\n📋 Testing listTopics...');
+  const result = await publicClient.graphql({ query: listTopics });
+  const topics = (result as any).data.listTopics.items;
+  console.log(`✅ Found ${topics.length} topics:`);
+  topics.slice(0, 5).forEach((t: any) => {
+    const topicName = t.content.includes(':') ? t.content.split(':')[1] : t.content;
```
9pace: This would also throw here and break if topics returned null.

amplify-migration-apps/discussions/gen2-test-script.ts
sanjanaravikumar-az: This makes sense to me, i can switch the names to make it match better. As for the graphql queries, the Notes CRUD operations have owner-based authorization rules in the schema so it uses userPool, and the lambda function operations use api key because they are public and don't require user authentication.

amplify-migration-apps/discussions/gen2-test-script.ts
sanjanaravikumar-az: While this is true, I decided to keep it inside so that it is clear what auth each test is using without having to scroll to the top to find the module level declarations. We can also comment out some tests without worrying about the unused client. Also, if in the future we have to modify the apps, and the test scripts, its easier to add a new test in this format. The impact on leaving the client in the function is negligible.

amplify-migration-apps/discussions/gen2-test-script.ts
sanjanaravikumar-az: true, good point, i'll remove this

amplify-migration-apps/discussions/gen2-test-script.ts
sanjanaravikumar-az: if the topics returned null, it would get caught by the runTest() function, the test script wouldnt crash but the error would be displayed at the end

amplify-migration-apps/discussions/gen2-test-script.ts
```diff
@@ -0,0 +1,579 @@
...
+// Configure Amplify
+Amplify.configure(amplifyconfig);
+
+// Initialize test runner
+const { runTest, printSummary } = createTestRunner();
+
+// ============================================================
+// CONFIGURATION - Update with your test user credentials
+// ============================================================
+const TEST_USER = {
```
9pace, line 36: Maybe making this an argument?

amplify-migration-apps/discussions/gen2-test-script.ts
9pace: could move to 51?

amplify-migration-apps/discussions/gen2-test-script.ts
9pace:
```
generateClient()  is the private (userpool). Use for get and list
generateClient({ authMode: 'apiKey' }) is the public. Use for set, modify...
```

amplify-migration-apps/discussions/gen2-test-script.ts
9pace: let me know what happens if you just do `result.data`

sanjanaravikumar-az: it will be the same

amplify-migration-apps/discussions/gen2-test-script.ts
9pace: Don't need to support this for now. Worth considering only if necessary.

9pace (approved):
Nitpick: still don't like heavy use of `any`

Will review the other ones soon.

PR #14500, opened 2026-01-20
chore(gen2-migration): remove protective conditions that mask the dynamodb table mapping error

dgandhi62:
Currently, model name to dyanmoDB table name mappings are lost in certain situations. We are unaware of the real reason why.

This pr removes two protective `if` statements that could be potentially masking this error. This would allow the error to propagate and fail the migration, aiding detection.

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
@@ -203,6 +203,8 @@ export const generateDataSource = async (gen1Env: string, dataDefinition?: DataD
     const apiId = await getApiId(gen1Env);
     if (apiId) {
       tableMappings = createDataSourceMapping(dataDefinition.schema, apiId, gen1Env);
+    } else {
+      throw new Error(`Unable to find AppSync API for environment '${gen1Env}'. Ensure the API exists and is properly tagged.`);
     }
   }
 


packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
```diff
@@ -201,45 +201,41 @@ export const generateDataSource = async (gen1Env: string, dataDefinition?: DataD
   // Generate table mappings if not provided but schema is available
   if (!tableMappings && dataDefinition?.schema) {
     const apiId = await getApiId(gen1Env);
-    if (apiId) {
```
iliapolo, line 204: Ok I see that this executes only when `dataDefinition?.schema` is not null so we definitely expect an AppSync server. But in that case we should throw an error and not just remove the `if` statement - otherwise the `undefined` value will be passed to `createDataSourceMapping` and its unclear what will happen.

packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
```diff
@@ -201,45 +201,41 @@ export const generateDataSource = async (gen1Env: string, dataDefinition?: DataD
...
-      tableMappings = createDataSourceMapping(dataDefinition.schema, apiId, gen1Env);
-    }
+    tableMappings = createDataSourceMapping(dataDefinition.schema, apiId, gen1Env);
   }
 
-  if (tableMappings) {
```
iliapolo, line 209: This can be `undefined` if the customer has only REST API so think its a valid `if`

dgandhi62: Fair

packages/amplify-cli/src/commands/gen2-migration/generate/generators/data/index.ts
dgandhi62: Resolved

iliapolo (approved):

PR #14513, opened 2026-01-23
fix(gen2-migration): migrating backend only apps doesn't quite work

iliapolo:
There are some quirks currently with running migration on apps that don't contain a frontend:

- The `implications` printed in the `generate` step refer to non existing node related files (e.g `package-lock.json`).
- If the app was never deployed via the hosting service, an `amplify.yml` buildspec is not generated and the gen2 branch cannot be deployed via hosting.
- The `package.json` file we create has a hardcoded `my-gen2-app` name.
- Function dependencies don't migrate correctly.

#### Description of changes

- Fix `implications` in the `generate` command to consider whether a node based frontend exists or not. This required making the `implications` function `async`. 
- Revert https://github.com/aws-amplify/amplify-cli/pull/14426 since apparently the previous implementation was the correct one.
- In apps that contained a function with dependencies, remove those dependencies from the root `package.json` and place them in a dedicated `package.json` file per function. The existence of those dependencies in the root `package.json` masked the bug mentioned above.
- When an app doesn't have an existing `amplify.yml` file, it means that either 1) a frontend doesn't exist, or 2) the frontend is not deployed via the hosting service. For both cases we now generate an `amplify.yml` file that deploys the backend and an empty frontend webapp. 
- Replace the hardcoded `my-gen2-app` app name with the actual app and environment name.
- Add `Frontend Migration` section to explain how frontends are expected to work post migration.
- Restructure and rephrase the explanation about stateful vs stateless resources.
- Remove the `Supported Frameworks` section because there is now no apparent framework limitation. Customers are welcome to test all frameworks.
- Added the backend only app.

Manually deployed and migrated a backend only app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
@@ -329,15 +329,6 @@ export async function updateAmplifyYmlFile(logger: Logger, amplifyClient: Amplif
     }
   }
 
-  if (amplifyYml && !amplifyYml.frontend) {
-    // shouldn't really happen because it indicates a corrupt gen1 app.
-    // we validate this because a missing 'frontend' section will prevent deploying the gen2 app via hosting.
-    throw new AmplifyError('MigrationError', {
-      message: `Buildspec for app '${appId}' is missing a 'frontend' section`,
-      resolution: `Add a 'frontend' section to your 'amplify.yml' file and try again`,
-    });
-  }
-
   if (amplifyYml === undefined) {
     // this means the gen1 environment was never deployed through hosting, neither backend nor frontend.
     // since we currently instruct customers to deploy their Gen2 branch via hosting, we need to create
--- a/GEN2_MIGRATION_GUIDE.md
+++ b/GEN2_MIGRATION_GUIDE.md
@@ -389,18 +389,21 @@ Your schema is located in `./amplify/data/resource.ts`.
 
 ### 3. Deploy
 
-Deploying the generated Gen2 application is done via [fullstack-branch-deployments](https://docs.amplify.aws/flutter/deploy-and-host/fullstack-branching/branch-deployments/). First, push the code:
-
-> [!NOTE]
-> The migration tool generates an `amplify.yml` buildspec file that allows for 
-> branch deployments to deploy Gen2 backend applications even in the absence of a 
-> webapp published via amplify hosting.
+Deploying the generated Gen2 application is done via [fullstack-branch-deployments](https://docs.amplify.aws/flutter/deploy-and-host/fullstack-branching/branch-deployments/). 
+First, push the code:
 
 ```bash
 git add .
 git commit -m "feat: migrate to gen2"
 git push origin gen2-main
 ```
+
+> [!NOTE]
+> The migration tool generates an `amplify.yml` buildspec file that allows for 
+> branch deployments to deploy Gen2 backend applications even in the absence of a 
+> webapp published via amplify hosting. If you'd like to start using the hosting service to publish your Gen2 webapp, you'll 
+> need to manually add a `frontend` section to this file and provide the necessary configuration to build your webapp. 
+> See [Build specification reference](https://docs.aws.amazon.com/amplify/latest/userguide/yml-specification-syntax.html) for more details.
 
 Next, login to the AWS Amplify console and connect your new branch to the existing application:
 


amplify-migration-apps/project-boards/README.md
```diff
@@ -272,7 +272,7 @@ git checkout gen2-main
 
 ```console
 git add .
-git commit -m "fix: reuse gen1 storage bucket"
+git commit -m "chore: post refactor"
```
iliapolo, line 275: Just to be consistent. All other apps have this post refactor commit message.

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
```diff
@@ -296,32 +311,66 @@ const unsupportedCategories = async (
...
+              'npm ci --cache .npm --prefer-offline',
+              'npx ampx pipeline-deploy --branch $AWS_BRANCH --app-id $AWS_APP_ID',
+            ],
+          },
+        },
+      },
+      frontend: {
+        phases: {
+          build: {
+            commands: ['mkdir dist', 'touch dist/index.html'],
```
dgandhi62, line 354: `writeToAmplifyYmlFile` doesn't touch the frontend build command. Are these the commands we want in Gen2?

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
```diff
@@ -296,32 +311,66 @@ const unsupportedCategories = async (
...
+
+  if (await pathExists(amplifyYmlPath)) {
+    logger.info(`Repurposing local buildspec: ${amplifyYmlPath}`);
+    amplifyYml = yaml.parse(await fs.readFile(amplifyYmlPath, 'utf-8'));
+  } else {
+    const response = await amplifyClient.send(new GetAppCommand({ appId }));
+    if (response.app.buildSpec) {
+      logger.info(`Repurposing remote buildspec`);
+      amplifyYml = yaml.parse(response.app.buildSpec);
+      if (!amplifyYml.frontend) {
```
dgandhi62: I think we should have a similar error check for backend too

migration-guide-images/gen1-frontend-post-migration.png
dgandhi62, line 1: I'm not sure I fully follow these images / the purpose of these images

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
dgandhi62: Also -> any specific reason why this check is only for the remote buildspec and not in local?

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
iliapolo: These are the commands we want in case we couldn't find an existing yaml file. It allows for Gen2 hosting to deploy your backend without failing on a "non existing" frontend.

migration-guide-images/gen1-frontend-post-migration.png
iliapolo: They aim to clarify how frontend applications access your backend resources during the migration phases. Are they not clear enough?

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
iliapolo: > I think we should have a similar error check for backend too

Have you seen a case where backend is missing?

> Also -> any specific reason why this check is only for the remote buildspec and not in local?

Good point. Changed.

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
dgandhi62: This case is for when we don't have a buildspec on the console and no local yml file either right?

The cases I see for this are:
1. The entire app is corrupted
2. They used s3 & cloudfront from the cli locally / amplify hosting & publish. 

For case 1 we wouldn't want to migrate at all. For case 2, is it necessary they don't have a frontend? They may or may not acc to me, depending on the type of app.

migration-guide-images/gen1-frontend-post-migration.png
dgandhi62: If this image is intended to be customer facing, my thought would be that they wouldn't care as much for the detail. 

Just a personal opinion, but does specifying state matter here? The idea is simply to say that gen2 branch is using the same gen1 frontend through `amplifyconfiguration.json` and after we make the manual change, it will use `amplify_outputs` right? This feels detailed when the context for it in the migration guide just says "The following diagram  describes how existing frontend applications interact with your backend resources post migration". 

I feel it could be simpler & make clear what "post migration means" (if we are intending it to be "post generate" for this image)

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
dgandhi62: I have seen backend in the yml file go missing post `generate`. At that time, I didn't verify if the gen1 backend in the yml file was also missing too. But just adding a fail check if it is, shouldn't cause any unintended damage from what I can think of

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
iliapolo: As discussed offline, the empty frontend definition is there to facilitate backend only apps. Its true that in case a customer:

- Has a webapp in the same repo.
- Never published it through Gen1 hosting.
- Wants to start publishing it with Gen2 hosting.

Then this blank frontend definition will be wrong for them. But there's not much we can do here because I don't want to guess what the correct frontend configuration for their app would be. I added a note to mention this scenario and tell the customer they would need to manually construct their own frontend section in the YAML file in that case.

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
iliapolo: As discussed offline, since we don't have a good understanding on when this would happen, lets not add code to "handle" it. Its ok to just let it fail during the deploy time have the customer reach out to us and describe the issue (they would do that regardless since we don't have any recommendation for them in this case).

migration-guide-images/gen1-frontend-post-migration.png
iliapolo: As discussed offline, these are starter images, some people will find them helpful and some won't. Lets see and adjust as needed prior to GA.

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/command-handlers.ts
dgandhi62: Just a small modification (so that we can refer this thread later) - this is for the case where customer 

- Has a webapp in the same repo.
- Published it through Gen1 hosting, but not through ci-cd / github, or never published it through gen1 hosting.
- Wants to start publishing it with Gen2 hosting.

PR #14544, opened 2026-01-28
feat(gen2-migration): fitness tracker test scripts

sanjanaravikumar-az:
This script provides comprehensive frontend-backend testing for the Fitness Tracker app's Amplify Gen1 and Gen2 backend.

What it tests:

- GraphQL queries (listWorkoutPrograms, getWorkoutProgram, listExercises, getExercise, listMeals, getMeal)
- GraphQL mutations (CRUD operations for WorkoutPrograms, Exercises, and Meals)
- REST API nutrition logging via Lambda function

How it works:

- Authenticates via Cognito user pool
- Runs authenticated queries to verify data retrieval (WorkoutPrograms, Exercises)
- Runs public queries for Meals (using API key auth mode)
- Performs full CRUD cycle on WorkoutPrograms (create → update → delete)
- Performs full CRUD cycle on Exercises linked to WorkoutPrograms (create → update → delete)
- Performs full CRUD cycle on Meals (create → update → delete)
- Tests REST API nutrition logging endpoint
- Cleans up test data in reverse order (exercises → workout programs) and signs out

Users need to update TEST_USER credentials before running tests.

Note: Lambda DomainAllow is amazon.com, so either test creds need to be changed for this.

### DIFF
--- a/amplify-migration-apps/fitness-tracker/gen1-test-script.ts
+++ b/amplify-migration-apps/fitness-tracker/gen1-test-script.ts
@@ -19,8 +19,8 @@ import { Amplify } from 'aws-amplify';
 import { signIn, signOut, getCurrentUser } from 'aws-amplify/auth';
 import { post } from 'aws-amplify/api';
 import amplifyconfig from './src/amplifyconfiguration.json';
-import { TestRunner, provisionTestUser } from '../shared-test-utils/test-apps-test-utils';
-import testCredentials from '../shared-test-utils/test-credentials.json';
+import { TestRunner } from '../_test-common/test-apps-test-utils';
+import { provisionTestUser } from '../_test-common/signup';
 import { createTestFunctions, createTestOrchestrator } from './test-utils';
 
 // Configure Amplify
@@ -51,7 +51,6 @@ async function testNutritionLogAPI(): Promise<void> {
   console.log('   Message:', (response as any).message);
 }
 
-
 // ============================================================
 // Main Test Execution
 // ============================================================
@@ -64,7 +63,7 @@ async function runAllTests(): Promise<void> {
   console.log('  3. REST API Operations (Nutrition Logging)');
 
   // Provision user via SDK, then sign in here so tokens stay in this module's Amplify scope
-  const { signinValue, testUser } = await provisionTestUser(amplifyconfig, testCredentials);
+  const { signinValue, testUser } = await provisionTestUser(amplifyconfig);
 
   // Sign in from this module so the auth tokens are available to api/storage
   try {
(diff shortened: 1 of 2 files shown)


amplify-migration-apps/fitness-tracker/gen2-test-script.ts
```diff
@@ -0,0 +1,199 @@
...
+      content: `Test nutrition log via Gen2 REST API - Pizza and salad - ${Date.now()}`,
+    };
+    console.log('   Request body:', JSON.stringify(requestBody));
+
+    // Path relative to the stage (endpoint already includes /prod/)
+    const response = await makeSignedRequest('POST', 'nutrition/log', requestBody);
+
+    console.log('✅ Gen2 REST API Response:', response);
+    console.log('   Message:', response.message);
+  } catch (error: any) {
```
9pace: Nitpick: Inconsistent use of try catch, gen 1 has no try catch which works, this one has catch block. Would prefer just one style.

amplify-migration-apps/fitness-tracker/gen1-test-script.ts
```diff
@@ -0,0 +1,112 @@
...
+// Configure Amplify
+Amplify.configure(amplifyconfig);
+
+// Initialize test runner
+const { runTest, printSummary } = createTestRunner();
+
+// ============================================================
+// CONFIGURATION - Update with your test user credentials
+// ============================================================
+const TEST_USER = {
```
9pace: One thing I would like you to comment on: does it make sense / is it secure to store test user credentials in the test file itself? I think it is worth having a discussion over a better way to pass credentials. (Maybe worth bringing up to ian since he is the direct customer here)

amplify-migration-apps/fitness-tracker/test-utils.ts
```diff
@@ -0,0 +1,402 @@
...
+  } catch (error: any) {
+    if (error.name === 'UserAlreadyAuthenticatedException') {
+      const user = await getCurrentUser();
+      console.log(`✅ Already signed in as: ${user.username}`);
+      return true;
+    }
+    console.log('❌ Authentication failed:', error.message || error);
+    return false;
+  }
+}
```
iankhou: Did you encounter cases where the user is already signed in?

Anyway, I think this function should return the credentials. Maybe just make it a wrapper around `getCurrentUser`, such that it will sign in if necessary? Something like:

```ts
function getAuthedUser(): Promise<{ username: string, password: string }> {
 // try getCurrentUser
  // on success, return the creds
  // on failure, call signIn
}
```

amplify-migration-apps/fitness-tracker/gen1-test-script.ts
iankhou: As we discussed in standup, we can just pass dummy values here, like username: user0 and password: Password1!

We're not worried about keeping these credentials secured.

Issue #14526, opened 2026-01-26
(gen2-migration) `lock --rollback`


iliapolo:
### Describe the feature you'd like to request

Executing `gen2-migration lock` cannot be rolled back at the moment. We need for when:

- A customer decides to abandon migration and needs to regain access to the Gen1 environment.
- A customer needs an emergency update to their Gen1 environment mid migration.

### Describe the solution you'd like

Add a `--rollback` option to the `lock` command. 

### Implementation Notes

- Must revert **all** operations performed by the `lock` command.
- Must validate the Gen1 stack has not been refactored yet. If it has, the command should error out and instruct the customer to execute `gen2-migration refactor --rollback` first. 

### Describe alternatives you've considered

None

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

Issue #14570, opened 2026-01-31
(gen2-migration) `lock --rollback` command should validate the Gen1 environment integrity


iliapolo:
### Describe the feature you'd like to request

When running `gen2-migration lock --rollback`, we need to validate the Gen1 environment is in an expected state:

- Gen1 environment should be in a post refactor state. 

Otherwise, the customer may unknowingly start pushing updates to the Gen1 environment that will cause highly unexpected and undesired changes.

### Describe the solution you'd like

I'm not sure how to achieve this, my hope is that running drift detection will cover this. Needs testing.

### Describe alternatives you've considered

None

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

iliapolo:
Drift detection on the Gen1 environment should tell use if it was refactored or not.

PR #14571, opened 2026-01-31
feat(gen2-migration): `lock --rollback`

iliapolo:
#### Description of changes

Added support for rolling back the `lock` command. Since this is the first command to support rollback, this PR also introduces general support for the rollback flow.

**Example output:**

```console
 ❯ npx amplify gen2-migration lock --skip-validations --rollback

You are about to rollback 'lock' on environment 'd3g10pohpvht6z/main'.

Operations Summary

• Remove environment variable 'GEN2_MIGRATION_ENVIRONMENT_NAME'
• Set a policy on stack 'amplify-backendonly-main-c6f6c': {"Statement":[{"Effect":"Allow","Action":"Update:*","Principal":"*","Resource":"*"}]}

Implications

• You will be able to run 'amplify push' on environment 'main'
• You will be able to start migration of another environment

✔ Do you want to continue? (y/N) · yes

[2026-01-31T16:20:56.223Z] [lock] [backendonly/main] → Rolling back
[2026-01-31T16:20:56.492Z] [lock] [backendonly/main] • Removed GEN2_MIGRATION_ENVIRONMENT_NAME environment variable
[2026-01-31T16:20:56.713Z] [lock] [backendonly/main] • Successfully set policy on stack 'amplify-backendonly-main-c6f6c': {"Statement":[{"Effect":"Allow","Action":"Update:*","Principal":"*","Resource":"*"}]}
[2026-01-31T16:20:56.713Z] [lock] [backendonly/main] → Rollback complete
```

**Changes to `gen2-migration.md`**:

- Remove step specific details to focus on the key responsibilities of the `gen2-migration.ts` file. Step specific explanations should go into dedicated documentation files. (will be added in follow up PRs).
- Remove details about validations. They should be embedded in the step specific documentation files (will be added in follow up PRs).
- Change flowchart according to the changes above. Also make it horizontal for easier human parsing.
- Add details about new methods, interfaces, options, and flow pertaining to rollback capabilities.

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/14526

Manually migrated our [project-boards-backend-only](https://github.com/aws-amplify/amplify-cli/tree/gen2-migration/amplify-migration-apps/project-boards-backend-only) app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration.ts
@@ -152,7 +152,7 @@ export const run = async (context: $TSContext) => {
   const implementation: AmplifyMigrationStep = new step.class(logger, envName, appName, appId, stackName, region, context);
 
   if (validationsOnly) {
-    await validate(implementation, rollingBack, logger);
+    await validate(implementation, rollingBack, logger, context);
     return;
   }
 
@@ -176,7 +176,9 @@ export const run = async (context: $TSContext) => {
   printer.info(chalk.bold(chalk.underline('Implications')));
   printer.blankLine();
 
-  for (const implication of rollingBack ? await implementation.rollbackImplications() : await implementation.executeImplications()) {
+  const cachedStep = new CachedAmplifyMigrationStep(step);
+
+  for (const implication of rollingBack ? await cachedStep.rollback() : await cachedStep.execute()) {
     printer.info(`• ${implication}`);
   }
 
@@ -194,19 +196,19 @@ export const run = async (context: $TSContext) => {
   printer.blankLine();
 
   if (!skipValidations) {
-    await validate(implementation, rollingBack, logger);
+    await validate(implementation, rollingBack, logger, context);
     printer.blankLine();
   }
 
   if (rollingBack) {
-    await runRollback(implementation, logger);
+    await runRollback(cachedStep, logger);
     printer.blankLine();
     printer.success('Done');
     return;
   }
 
   try {
-    await runExecute(implementation, logger);
+    await runExecute(cachedStep, logger);
     printer.blankLine();
     printer.success('Done');
     return;
@@ -214,19 +216,27 @@ export const run = async (context: $TSContext) => {
     if (!disableAutoRollback) {
       printer.error(`Execution failed: ${error}`);
       printer.blankLine();
-      await runRollback(implementation, logger);
+      await runRollback(cachedStep, logger);
     }
 
     throw error;
   }
 };
 
-async function validate(step: AmplifyMigrationStep, rollback: boolean, logger: Logger) {
+async function validate(step: AmplifyMigrationStep, rollback: boolean, logger: Logger, context: $TSContext) {
   logger.envelope('Performing validations');
-  if (rollback) {
-    await step.rollbackValidate();
-  } else {
-    await step.executeValidate();
+  try {
+    if (rollback) {
+      await step.rollbackValidate();
+    } else {
+      await step.executeValidate();
+    }
+  } catch (e) {
+    const skipValidationsCommand = `amplify ${context.input.argv.join(' ').trim()} --skip-validations`;
+    throw new AmplifyError('MigrationError', {
+      message: `Validations failed: ${e.message}`,
+      resolution: `Resolve the validation errors or skip them by running '${skipValidationsCommand}'`,
+    });
   }
   logger.envelope('Validations complete');
 }
@@ -237,13 +247,13 @@ async function runOperations(operations: AmplifyMigrationOperation[]) {
   }
 }
 
-async function runRollback(step: AmplifyMigrationStep, logger: Logger) {
+async function runRollback(step: CachedAmplifyMigrationStep, logger: Logger) {
   logger.envelope('Rolling back');
   await runOperations(await step.rollback());
   logger.envelope('Rollback complete');
 }
 
-async function runExecute(step: AmplifyMigrationStep, logger: Logger) {
+async function runExecute(step: CachedAmplifyMigrationStep, logger: Logger) {
   logger.envelope('Executing');
   await runOperations(await step.execute());
   logger.envelope('Execution complete');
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/commands/gen2-migration.ts
```diff
@@ -125,58 +151,109 @@ export const run = async (context: $TSContext) => {
...
+  printer.blankLine();
+  printer.info(
+    chalk.yellow(`You are about to ${rollingBack ? 'rollback' : 'execute'} '${stepName}' on environment '${appId}/${envName}'.`),
+  );
+  printer.blankLine();
+
+  printer.info(chalk.bold(chalk.underline('Operations Summary')));
+  printer.blankLine();
+
+  for (const operation of rollingBack ? await implementation.rollback() : await implementation.execute()) {
```
9pace, line 168: First time calling execute. why not call execute once and store the operations

packages/amplify-cli/src/commands/gen2-migration.ts
```diff
@@ -125,58 +151,109 @@ export const run = async (context: $TSContext) => {
...
+  logger.envelope('Performing validations');
+  if (rollback) {
+    await step.rollbackValidate();
+  } else {
+    await step.executeValidate();
+  }
+  logger.envelope('Validations complete');
+}
+
+async function runOperations(operations: AmplifyMigrationOperation[]) {
```
9pace, line 244: Why run await runOperations(await step.execute()); calling execute again when we could store the operations needed in a variable the first time we call execute.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
```diff
@@ -1,76 +1,153 @@
...
+      execute: async () => {
+        const app = await this.amplifyClient().send(new GetAppCommand({ appId: this.appId }));
+        const environmentVariables = app.app.environmentVariables ?? {};
+        delete environmentVariables[GEN2_MIGRATION_ENVIRONMENT_NAME];
+        await this.amplifyClient().send(new UpdateAppCommand({ appId: this.appId, environmentVariables }));
+        this.logger.info(`Removed ${GEN2_MIGRATION_ENVIRONMENT_NAME} environment variable`);
+      },
+    });
+
+    const stackPolicy = JSON.stringify({
```
9pace, line 135: If the stack had a *custom* policy before the lock step for ex, a policy that only allowed updates to specific resources, the rollback would replace it with allow all policy.

Which would not restore it to its original state. Its not an expected configuration, but maybe it should be checked for in the lock step.

packages/amplify-cli/src/commands/gen2-migration.ts
```diff
@@ -125,58 +151,109 @@ export const run = async (context: $TSContext) => {
...
+    await validate(implementation, rollingBack, logger);
     printer.blankLine();
-    logger.envelope('Performing validations');
-    try {
-      await implementation.validate();
-    } catch (e) {
-      const skipValidationsCommand = `amplify ${context.input.argv.join(' ').trim()} --skip-validations`;
-      throw new AmplifyError('MigrationError', {
-        message: `Validations failed: ${e.message}`,
-        resolution: `Resolve the validation errors or skip them by running '${skipValidationsCommand}'`,
```
9pace, line 137: Are we no longer going to recommend running --skip-validations or do we just plan on mentioning this in the guide?

packages/amplify-cli/src/commands/gen2-migration/lock.ts
```diff
@@ -1,76 +1,153 @@
...
+
+    return operations;
+  }
 
-    const stackPolicy = {
+  public async rollback(): Promise<AmplifyMigrationOperation[]> {
+    const operations: AmplifyMigrationOperation[] = [];
+
+    // note that we don't disable deletion protection on the tables because we don't
+    // know what the original value was; to play it safe we leave it untouched.
```
9pace, line 109: This should be mentioned in rollbackImplications.

packages/amplify-cli/src/commands/gen2-migration.ts
```diff
@@ -125,58 +151,109 @@ export const run = async (context: $TSContext) => {
...
+}
+
 function shiftParams(context) {
   delete context.parameters.first;
   delete context.parameters.second;
   delete context.parameters.third;
   const { subCommands } = context.input;
-  /* eslint-disable */
   if (subCommands && subCommands.length > 1) {
     if (subCommands.length > 1) {
```
9pace: Nitpick: not part of the PR, but I noticed `if (subCommands.length > 1) {` is redundant due to the line above it.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
iliapolo: Yeah, this will be addressed by resolving these two issue:

- https://github.com/aws-amplify/amplify-cli/issues/14577
- https://github.com/aws-amplify/amplify-cli/issues/14570

packages/amplify-cli/src/commands/gen2-migration.ts
iliapolo: Fair. The issue here is that storing the operations in a variable may result in invoking a function we won't actually use. 

- If we are rolling back, no need to call `step.execute()` at all.
- If we are in a forward execution, `step.rollback()` is only needed if there was an error and the user didn't disable auto rollbacks. 

So basically this code calls some functions twice instead of possibly calling a function once that we won't need. 

To address both cases i've created an internal `CachedAmplifyMigrationStep` class that has a `getOrCreate` semantic for the operations.

packages/amplify-cli/src/commands/gen2-migration.ts
iliapolo: Addressed

packages/amplify-cli/src/commands/gen2-migration.ts
iliapolo: Good catch. Added back.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
iliapolo: A bit awkward to put a negative into implications..."Deletion protection is left untouched"? 

Instead i've added it to the operations, also weird, but I think more intuitive.

9pace: Sounds good, I think this is something that should be exposed to the user and this resolves that for me.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
9pace: Nice

9pace (approved):

PR #14573, opened 2026-02-01
chore(gen2-migration): mark features as "untested" in migration guide

iliapolo:
There are many features that we still haven't validated with proper testing, but may actually work. Right now, these features are marked with ❌ and stop people from trying it out. Given this is still a beta, we can afford to be less strict here. 

This PR introduces an "untested" marking to encourage folks to try it out and give us feedback.

----

### DIFF
--- a/GEN2_MIGRATION_GUIDE.md
+++ b/GEN2_MIGRATION_GUIDE.md
@@ -565,24 +565,24 @@ by the CLI setting that configures them.
 
 - 🟢 **Do you want to add User Pool Groups**
 
-- 🔴 **Do you want to add an admin queries API**
+- ⚠️ **Do you want to add an admin queries API**
 
 - ➤ **Multifactor authentication (MFA) user login options**
 
   - 🟢 `OFF`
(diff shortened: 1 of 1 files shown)


GEN2_MIGRATION_GUIDE.md
```diff
@@ -525,114 +525,112 @@ by the CLI setting that configures them.
...
-> - ❌ | Unsupported.
-> - ✅ | Fully automated.
-> - ⚠️ | Partially supported. Includes indication whether it lacks support for `generate` or `refactor`.
-
-- If a feature is not supported for `refactor` you will not be able to fully migrate the app. You can however still generate 
-and deploy it to test whether code generation works properly.
-- If a feature is not supported for `generate` you will be able to manually augment the generated code 
-to add the necessary configuration.
-- Unsupported features naturally don't support either command.
+> - 🔴 | Unsupported.
```
iliapolo, line 528: Decided to switch to standard stop light signs.

PR #14589, opened 2026-02-12
chore(gen2-migration): add shared test utils file for test apps

sanjanaravikumar-az:
Created a shared test-utils.ts file containing shared code among the amplify migration apps. It contains the main test runner and the authentication flow using Cognito APIs. The authentication flow is made to handle all different modes of sign in including email and phone number.

### DIFF
--- a/amplify-migration-apps/_test-common/README.md
+++ b/amplify-migration-apps/_test-common/README.md
@@ -1,15 +1,14 @@
-# _test-common
+# \_test-common
 
 Shared test utilities for the Amplify migration test apps. These helpers handle Cognito user provisioning, test execution, and common type definitions used across all apps in `amplify-migration-apps/`.
 
 ## Files
 
-| File | Description |
-|------|-------------|
-| `test-apps-test-utils.ts` | Shared type definitions (`AmplifyConfig`, `TestUser`, `TestCredentials`, etc.) and barrel re-exports for backwards compatibility. |
-| `signup.ts` | `provisionTestUser` — creates a test user via `AdminCreateUser` and sets a permanent password. Supports email, phone, and username signin patterns. Works even when self-signup is disabled on the user pool. |
-| `runner.ts` | `TestRunner` class — runs async test functions, collects failures, and prints a summary. |
-| `test-credentials.json` | Static test credentials (email, phone, username, password) consumed by test scripts. |
+| File                      | Description                                                                                                                                                                                                                                             |
+| ------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `test-apps-test-utils.ts` | Shared type definitions (`AmplifyConfig`, `TestUser`, `TestCredentials`, etc.) and barrel re-exports for backwards compatibility.                                                                                                                       |
+| `signup.ts`               | `provisionTestUser` — creates a test user via `AdminCreateUser` and sets a permanent password. Generates unique credentials dynamically. Supports email, phone, and username signin patterns. Works even when self-signup is disabled on the user pool. |
+| `runner.ts`               | `TestRunner` class — runs async test functions, collects failures, and prints a summary.                                                                                                                                                                |
 
 ## Usage
 
--- a/amplify-migration-apps/_test-common/signup.ts
+++ b/amplify-migration-apps/_test-common/signup.ts
@@ -4,7 +4,8 @@ import {
   AdminSetUserPasswordCommand,
   type AttributeType,
 } from '@aws-sdk/client-cognito-identity-provider';
-import type { AmplifyConfig, SigninIdentifier, SignupAttribute, TestCredentials, TestUser } from './test-apps-test-utils';
+import { randomBytes } from 'crypto';
+import type { AmplifyConfig, SigninIdentifier, SignupAttribute, TestUser } from './test-apps-test-utils';
 
 interface ResolvedAuthConfig {
   signinIdentifier: SigninIdentifier;
@@ -46,24 +47,46 @@ function resolveSignupAttributes(signupAttributes: string[]): SignupAttribute[] 
   return mapped.length > 0 ? mapped : ['email'];
 }
 
+interface GeneratedCredentials {
+  email?: string;
+  phoneNumber?: string;
+  username?: string;
+  password: string;
+}
+
+/**
+ * Generates only the credentials required by the resolved auth config.
+ * The union of signupAttributes and signinIdentifier determines which
+ * identity fields are needed. Password is always generated.
+ */
+function generateCredentials(resolved: ResolvedAuthConfig): GeneratedCredentials {
+  const needed = new Set<SignupAttribute | SigninIdentifier>([...resolved.signupAttributes, resolved.signinIdentifier]);
+  return {
+    email: needed.has('email') ? generateTestEmail() : undefined,
+    phoneNumber: needed.has('phone') ? generateTestPhoneNumber() : undefined,
+    username: needed.has('username') ? generateTestUsername() : undefined,
+    password: generateTestPassword(),
+  };
+}
+
 function buildAdminCreateUserInput(
   userPoolId: string,
   resolved: ResolvedAuthConfig,
-  credentials: { email: string; phoneNumber: string; username: string; password: string },
+  credentials: GeneratedCredentials,
 ): { UserPoolId: string; Username: string; TemporaryPassword: string; UserAttributes: AttributeType[]; MessageAction: 'SUPPRESS' } {
   const { signinIdentifier, signupAttributes } = resolved;
 
-  const identifierValueMap: Record<SigninIdentifier, string> = {
+  const identifierValueMap: Record<SigninIdentifier, string | undefined> = {
     email: credentials.email,
     phone: credentials.phoneNumber,
     username: credentials.username,
   };
-  const username = identifierValueMap[signinIdentifier];
+  const username = identifierValueMap[signinIdentifier] ?? '';
 
-  const attributeMap: Record<SignupAttribute, AttributeType> = {
-    email: { Name: 'email', Value: credentials.email },
-    phone: { Name: 'phone_number', Value: credentials.phoneNumber },
-    username: { Name: 'username', Value: credentials.username },
+  const attributeMap: Record<SignupAttribute, AttributeType | undefined> = {
+    email: credentials.email ? { Name: 'email', Value: credentials.email } : undefined,
+    phone: credentials.phoneNumber ? { Name: 'phone_number', Value: credentials.phoneNumber } : undefined,
+    username: credentials.username ? { Name: 'username', Value: credentials.username } : undefined,
   };
 
   // When phone or username is used as the Username, Cognito already
@@ -76,13 +99,14 @@ function buildAdminCreateUserInput(
       if (signinIdentifier === 'username' && attr === 'username') return false;
       return true;
     })
-    .map((attr) => attributeMap[attr]);
+    .map((attr) => attributeMap[attr])
+    .filter((a): a is AttributeType => a !== undefined);
 
   // Mark email/phone as verified so the user can sign in immediately
-  if (signupAttributes.includes('email')) {
+  if (credentials.email) {
     userAttributes.push({ Name: 'email_verified', Value: 'true' });
   }
-  if (signupAttributes.includes('phone')) {
+  if (credentials.phoneNumber) {
     userAttributes.push({ Name: 'phone_number_verified', Value: 'true' });
   }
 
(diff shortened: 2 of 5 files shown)


amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+  aws_user_pools_id?: string;
+  aws_user_pools_web_client_id?: string;
+  aws_cognito_region?: string;
+  signinIdentifier?: SigninIdentifier;
+  signupAttributes?: SignupAttribute[];
+}
+
+export interface TestFailure {
+  name: string;
+  error: string;
```
iankhou: Can we include the stacktrack from the error too? To give us more insight into what went wrong.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  return {
+    ClientId: clientId,
+    Username: username,
+    Password: credentials.password,
+    UserAttributes: userAttributes,
+  };
+}
+
+export function createTestRunner(): {
```
iankhou: Why is this a function and not a class? Seems odd to me that `printSummary` is a util function that's returned by a function call, instead of a method being called on an instance of a class.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
iankhou: Can we include the error name and the stacktrace? This will help us when debugging failures.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+const TEST_EMAIL = 'user1@test.example.com';
+const TEST_PHONE_NUMBER = '+15551234567';
+const TEST_USERNAME = 'testuser1';
+
+function resolveAuthConfig(config: AmplifyConfig): ResolvedAuthConfig {
+  return {
+    signinIdentifier: config.signinIdentifier ?? 'email',
+    signupAttributes: config.signupAttributes ?? ['email'],
+  };
+}
```
iankhou: Not sure why this is done, could you add a comment? It's also only used once, so might not be worth abstracting it out.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
iankhou: Ah, I see we're using `getErrorMessage` to fill this value in most cases. I think this key is misleading. Error is a specific thing, and a message is a component of an error. Reconsider this name.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  const attributeMap: Record<SignupAttribute, AttributeType> = {
+    email: { Name: 'email', Value: credentials.email },
+    phone: { Name: 'phone_number', Value: credentials.phoneNumber },
+    username: { Name: 'username', Value: credentials.username },
+  };
+
+  // Cognito requires email in UserAttributes even when it's also the Username.
+  // Only phone/username are excluded since Cognito treats them as implicit
+  // when they appear in the Username field.
```
iankhou: I don't understand this comment, could you provide clarification on what it means by "implicit when they appear in the Username field"?

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+
+export type SignupAttribute = 'email' | 'phone' | 'username';
+
+export interface AmplifyConfig {
+  aws_user_pools_id?: string;
+  aws_user_pools_web_client_id?: string;
+  aws_cognito_region?: string;
+  signinIdentifier?: SigninIdentifier;
+  signupAttributes?: SignupAttribute[];
+}
```
iankhou: Why all optional keys? We use this in both `createTestUser` and `resolveAuthConfig`, how come we don't have separate interfaces for these two functions, since the keys that each of these functions use do not intersect at all?

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  const resolved = resolveAuthConfig(config);
+  const { signupAttributes } = resolved;
+
+  const signUpInput = buildSignUpInput(clientId ?? '', resolved, {
+    email: TEST_EMAIL,
+    phoneNumber: TEST_PHONE_NUMBER,
+    username: TEST_USERNAME,
+    password: TEST_PASSWORD,
+  });
```
iankhou: I don't feel great about this not being configurable, although I also think the test creds should be shared. Can we keep the test creds in a separate JSON file, parameterize the credential inputs, and pass them in when we call `createTestUser`?

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: yes, that makes sense, I'll add an optional Stack field.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: Yes I left it in because this pattern was pre-existing, but I think the class approach might be more readable

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: i get your point, i've changed the name to message instead of error

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: thinking it over, i think it might be better to inline it directly into ```createTestUser```. Keeping the ```ResolvedAuthConfig``` type might be better since it guarantees non-optional fields for ```buildSignUpInput```

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: Yes, what I meant was phone/username are already associated via the ```Username``` field so they're excluded from ```UserAttributes```, but email is required in ```UserAttributes``` for verification even when it's also the ```Username```. This is Cognito behavior, don't know the exact reason why. But I'll update the comment to explain better.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: The fields are all optional because the three Cognito fields match the Amplify config JSON shape where they're optional. The auth pattern fields (```signinIdentifier```, ```signupAttributes```) are optional so existing callers don't have to change anything, they just get email-only by default. Splitting it up would mean callers have to break apart the same config object into multiple pieces, and since ```createTestUser``` is the only thing reading it, I don't think we'd gain much from that.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: i think thats a good idea, moved the test creds into a ```test-credentials.json``` and made ```createTestUser``` take a ```credentials``` param instead of using hardcoded constants. Callers just import the JSON and pass it in

sanjanaravikumar-az:
> Could you accompany this change with a change to a test-script? That will help give context on how this code is actually used.

Added the project boards test script which uses this logic.

amplify-migration-apps/_test-common/test-credentials.json
iliapolo, line 1: Why do we need this file? The signup function should just generate this on the fly and return the necessary information to the test.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,220 @@
...
+  }
+}
+
+/**
+ * Provisions a test user via Cognito signup and admin-confirms them.
+ * Does NOT sign in — the caller should handle signIn in its own module scope
+ * so the Amplify auth singleton has the tokens available for API/Storage calls.
+ * Returns the username to use for signIn.
+ */
+export async function provisionTestUser(
```
iliapolo: Lets put this function in a dedicated file called `signup.ts`. Its better for us (as well as AI) to have multiple small files, rather than one large one.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  return {
+    ClientId: clientId,
+    Username: username,
+    Password: credentials.password,
+    UserAttributes: userAttributes,
+  };
+}
+
+export class TestRunner {
```
iliapolo: Split all the functions and classes related to test runner to a separate file called `runner.ts`

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
iliapolo, line 1: Rename the directory to `_test-common`. We use `_` to differentiate this directory from normal apps. Also add a README.md with some basic explanation to what this directory contains.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
```diff
@@ -0,0 +1,220 @@
...
+  const signUpInput = buildSignUpInput(clientId ?? '', resolved, credentials);
+  const signinValue = signUpInput.Username;
+
+  console.log(`\n🔑 Creating test user: ${signUpInput.Username}`);
+
+  const cognitoClient = new CognitoIdentityProviderClient({ region });
+
+  // Step 1: SignUp
+  try {
+    await cognitoClient.send(new SignUpCommand(signUpInput));
```
iliapolo: Lets use `AdminCreateUserCommand` instead so that it also works with user pools that don't have self-sign up enabled.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: made the change

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: done!

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: done

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
sanjanaravikumar-az: updated the script

amplify-migration-apps/_test-common/test-credentials.json
sanjanaravikumar-az: It was initially configured that way but I received a comment on it:

"I don't feel great about this not being configurable, although I also think the test creds should be shared. Can we keep the test creds in a separate JSON file, parameterize the credential inputs, and pass them in when we call ```createTestUser```"

So we put the creds in a separate file

amplify-migration-apps/_test-common/README.md
```diff
@@ -0,0 +1,22 @@
+# _test-common
+
+Shared test utilities for the Amplify migration test apps. These helpers handle Cognito user provisioning, test execution, and common type definitions used across all apps in `amplify-migration-apps/`.
+
+## Files
+
+| File | Description |
+|------|-------------|
+| `test-apps-test-utils.ts` | Shared type definitions (`AmplifyConfig`, `TestUser`, `TestCredentials`, etc.) and barrel re-exports for backwards compatibility. |
+| `signup.ts` | `provisionTestUser` — creates a test user via `AdminCreateUser` and sets a permanent password. Supports email, phone, and username signin patterns. Works even when self-signup is disabled on the user pool. |
```
iankhou: yippee, no need to manually enable self-signup in CI!

amplify-migration-apps/_test-common/signup.ts
```diff
@@ -0,0 +1,164 @@
...
+          return first.message;
+        }
+      }
+    }
+    return JSON.stringify(error, null, 2);
+  }
+  return String(error);
+}
+
+function resolveSigninIdentifier(config: AmplifyConfig): SigninIdentifier {
```
iankhou: This function doesn't need the whole config, only `aws_cognito_username_attributes`.

amplify-migration-apps/_test-common/signup.ts
```diff
@@ -0,0 +1,164 @@
...
+  const usernameAttrs = config.aws_cognito_username_attributes ?? [];
+  if (usernameAttrs.includes('PHONE_NUMBER')) return 'phone';
+  if (usernameAttrs.includes('EMAIL')) return 'email';
+  return 'username';
+}
+
+function resolveSignupAttributes(config: AmplifyConfig): SignupAttribute[] {
```
iankhou: This function doesn't need the whole config, only `aws_cognito_signup_attributes`

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
iankhou: This function is defined both here and in `amplify-migration-apps/_test-common/signup.ts`, did you intend to remove this function here?

amplify-migration-apps/_test-common/runner.ts
```diff
@@ -0,0 +1,61 @@
...
+      const result = await testFn();
+      return result;
+    } catch (error: unknown) {
+      const stack = error instanceof Error ? error.stack : undefined;
+      this.failures.push({ name, message: getErrorMessage(error), stack });
+      return null;
+    }
+  }
+
+  printSummary(): void {
```
iankhou, line 41: This function is defined twice in this PR (also in `amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts`).

amplify-migration-apps/_test-common/runner.ts
```diff
@@ -0,0 +1,61 @@
...
+    }
+    return JSON.stringify(error, null, 2);
+  }
+  return String(error);
+}
+
+export class TestRunner {
+  readonly failures: TestFailure[] = [];
+
+  async runTest<T>(name: string, testFn: () => Promise<T>): Promise<T | null> {
```
iankhou, line 30: Same issue as my comment below on `printSummary()`, it is defined twice.

amplify-migration-apps/shared-test-utils/test-apps-test-utils.ts
iankhou: Did you mean to delete this class definition here?

amplify-migration-apps/_test-common/test-apps-test-utils.ts
```diff
@@ -0,0 +1,29 @@
+export type SigninIdentifier = 'email' | 'phone' | 'username';
+
+export type SignupAttribute = 'email' | 'phone' | 'username';
+
+export interface AmplifyConfig {
```
iankhou, line 5: This is being used in `resolveSigninIdentifier`, `resolveSignupAttributes`, `Amplify.configure`, and `provisionTestUser`. I feel like it should only be used for `Amplify.configure`, and `provisionTestUser`. For `resolveSigninIdentifier` and `resolveSignupAttributes` you can just pass the one value that they each need.

iankhou (changes requested):

amplify-migration-apps/_test-common/test-apps-test-utils.ts
sanjanaravikumar-az: made the change

amplify-migration-apps/_test-common/README.md
```diff
@@ -0,0 +1,22 @@
...
+| `runner.ts` | `TestRunner` class — runs async test functions, collects failures, and prints a summary. |
+| `test-credentials.json` | Static test credentials (email, phone, username, password) consumed by test scripts. |
```
iliapolo: Why do we need this to be static though? Can't the `provisionTestUser` function return this object dynamically to the caller? generating a different password each time?

PR #14596, opened 2026-02-16
feat(gen2-migration): add geo category code generation for gen2-migration 

sai-ray:
This PR adds geo category (`Map`, `PlaceIndex`, `GeofenceCollection`) codegen support to the gen2-migration pipeline. When a Gen1 app has geo resources, the `generate` step now produces per-resource CDK L1 constructs via `cdk-from-cfn`, wraps each in its own independent CloudFormation stack, and wires them into `backend.ts` through a top-level `defineGeo(backend)` aggregator. 

### Description of changes

1. #### **Geo Definition Fetcher** (`app_geo_definition_fetcher.ts`)

- `AppGeoDefinitionFetcher` Reads the `geo` key from the deployed `amplify-meta.json` in S3 and returns a record of `GeoResourceDefinition` entries keyed by resource name. Follows the same pattern as `AppAnalyticsDefinitionFetcher` but uses explicit `if`/`throw` instead of `assert`.

2. #### **Per-Resource Generator** (`geo-resource-renderer.ts`)

- `renderGeoResource` generates a `resource.ts` for each geo resource that creates its own independent stack via `backend.createStack()`, adds a dependency on the auth stack, and instantiates the construct with dynamic props (`branchName`, `authRoleName`, `unauthRoleName`, `userPoolId`, group roles from `backend.auth.resources`) and static props (`mapName`, `mapStyle`, `indexName`, etc. from deployed parameters). Handles all three service types with a switch on `serviceName`.

3. #### **Top-Level Aggregator** (`geo-renderer.ts`)

- `renderGeo` generates the top-level `amplify/geo/resource.ts` that imports all per-resource `define*` functions, calls each one assigning the return value to a variable using the resource name, then calls `backend.addOutput()` with geo configuration using construct output properties (`.name`, `.region`, `.style`) for maps, search indices, and geofence collections.

4. #### **Backend Synthesizer** (`synthesizer.ts`)

- Added `geo` to `BackendRenderParameters` with an `importFrom` path. When `geo` is present, the `render()` method adds `import { defineGeo }` and `const geo = defineGeo(backend)` after `defineBackend()`, following the analytics pattern.

5. #### **Migration Pipeline** (`migration-pipeline.ts`)

- Added `geo` to `Gen2RenderingOptions`. When geo resources are present, `createGen2Renderer()` creates the `amplify/geo/` directory, iterates over each resource creating per-resource directories and renderers (calling `generateGeoL1Code()` then `renderGeoResource()`), collects all render params, then creates a renderer for the top-level `renderGeo()` aggregator.

6. #### **Command Handlers** (`command-handlers.ts`)

- Removed `geo` from the `unsupportedCategories` map. Added `AppGeoDefinitionFetcher` instantiation in `prepare()` and wired it through `CodegenCommandParameters` to `generateGen2Code()`, which fetches geo definitions and passes them to `Gen2RenderingOptions`.

7. #### `cdk-from-cfn.ts`

- Type Interfaces : Added `GeoResourceDefinition` to model geo entries from `amplify-meta.json` (service type, S3 template URL, logical ID). Added `GeoCodegenResult` as a discriminated union of `MapCodegenResult`, `PlaceIndexCodegenResult`, and `GeofenceCollectionCodegenResult`, each carrying typed service-specific fields (`mapName`/`mapStyle`, `indexName`/`dataProvider`/`dataSourceIntendedUse`, `collectionName`) plus common auth parameter metadata (`userPoolIdParamName`, `groupRoles`).

- `postTransmute`  Fn::FindInMap` Token Crash Fix: The `cdk-from-cfn` library translates CFN `Fn::FindInMap` into plain JS dictionary lookups that crash at CDK synth time because `this.region` is a CDK Token, not a concrete string. `postTransmute` replaces `Record<string, Record<string, string>>` declarations with `cdk.CfnMapping` constructors and bracket lookups with `.findInMap()` calls, producing proper `Fn::FindInMap` intrinsics in synthesized CloudFormation.

- `preTransmute` `Fn::Join` GroupRole Fix: Gen1 geo templates construct group role names via `Fn::Join` of `UserPoolId` and a `GroupRole` suffix. `cdk-from-cfn` translates this into `.join('-')` which produces a plain string, breaking CloudFormation's cross-stack dependency detection. The fix replaces these `Fn::Join` patterns with a direct `Ref` to the corresponding `GroupRole` parameter, preserving CDK token chains for correct deployment ordering.

- Geo L1 Code Generation: `generateGeoL1Code` Fetches the CFN template from S3, fetches deployed stack parameters, applies `preTransmute` (`env` rename, condition resolution, `GroupRole` fix), calls `cdk_from_cfn.transmute()`, applies `postTransmute` (`FindInMap` fix), writes the construct file, extracts the class name, and categorizes deployed parameters into the typed `GeoCodegenResult`.

### Generated Output Structure

```
amplify/
  geo/
    resource.ts                        # defineGeo() aggregator + backend.addOutput()
    MyMap/
      MyMap-construct.ts               # L1 CDK code (cdk-from-cfn + postTransmute)
      resource.ts                      # defineMyMap() (independent stack)
    MySearch/
      MySearch-construct.ts
      resource.ts                      # defineMySearch() (independent stack)
    MyGeofence/
      MyGeofence-construct.ts
      resource.ts                      # defineMyGeofence() (independent stack)
  backend.ts                           # import { defineGeo } + defineGeo(backend)
```

#### `cdk-from-cfn` reference

The `cdk-from-cfn` library converts CloudFormation templates into CDK code. It exposes transmute(template, language, className, classType?) function. It takes a CFN template as a JSON string, a target language ('typescript' in our case), a class name (we pass the nested stack's logical ID), and an class type ('construct'). Returns a string of generated TypeScript containing a CDK construct class with L1 resources matching the original template. All CFN parameters become construct props, all resources become L1 constructs.

Our `CdkFromCfn` wrapper class in `en2-migration/generate/unsupported/cdk-from-cfn.ts` adds pre- and post-processing around `transmute()` because the library's output needs fixes for our use cases:

- `preTransmute` — modifies the CFN template before passing it to transmute(): renames env → branchName, replaces Fn::Join patterns for group roles with direct Refs (to preserve CDK tokens), and resolves CFN conditions using deployed stack parameters (since `cdk-from-cfn` generates broken TypeScript for conditions).

- `postTransmute` — fixes the generated TypeScript after `transmute()` returns: replaces plain dictionary lookups for Fn::FindInMap with cdk.CfnMapping + findInMap() calls, because the library generates mapping[this.region]['key'] which fails at synth time since this.region is a CDK token, not a concrete string.

Geo uses all three preTransmute fixes and the postTransmute fix, unlike kinesis which used only the preTransmute ones.

Dependencies -
- [x] To fully support geo codegen migration, the `cdk-from-cfn` version needs to be bumped to include `Custom::LambdaCallout` resource type support. Until that version is published, the `transmute()` call will fail on geo CFN templates.

This PR covers the `npx amplify gen2-migration generate` codegen step for geo category migration. Refactor for geo (migrating existing Gen1 resources into Gen2 stacks) will be a separate follow-up PR.

#### Checklist

- [ ] PR description included
- [ ] `yarn test` passes


packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
dgandhi62, line 1: This needs to be within `generator` or one of the existing folders. Lets not add more folders

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
```diff
@@ -37,6 +37,59 @@ export interface AnalyticsCodegenResult {
   streamName: string;
 }
 
+/**
+ * Definition for Geo resource from Gen1 amplify-meta.json
+ */
+export interface GeoResourceDefinition {
```
dgandhi62, line 43: This interface seems to be missing things from the amplify-meta file provided in the pr description. Is that intentional?

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
```diff
@@ -37,6 +37,59 @@ export interface AnalyticsCodegenResult {
...
+  /** The file name of the generated construct without extension */
+  constructFileName: string;
+  /** The resource name used for construct ID */
+  resourceName: string;
+  /** The full parameter name for the user pool ID prop on the construct */
+  userPoolIdParamName: string;
+  /** Group role parameters with extracted group names */
+  groupRoles: Array<{ paramName: string; groupName: string }>;
+  /** Whether this is the default resource */
+  isDefault: string;
```
dgandhi62, line 70: Why is this string and not boolean? amplify-meta has it as boolean as per the example provided

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
```diff
@@ -167,6 +220,95 @@ export class CdkFromCfn {
     };
   }
 
+  public async generateGeoL1Code(definition: GeoResourceDefinition): Promise<GeoCodegenResult> {
```
dgandhi62, line 223: This only uses 3 properties from what i see. 

1. name
2. s3TemplateURL
3. logicalId

What is the entire interface defined above used for? I don't fully understand

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
dgandhi62, line 1: We shouldn't club kinesis and geo in one file

dgandhi62 (changes requested):
Clarifying questions. Also still a bit confused on the how resource naming is being done.

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
sai-ray: Yes, the narrow interface is intentional. We only model the fields the pipeline actually reads from amplify-meta.json. The other fields are read from the deployed CloudFormation stack parameters inside `generateGeoL1Code()`. The CFN stack parameters are what's actually live in AWS, so reading from the stack is safer for migration. That said, for `service` and `providerMetadata` we do read from amplify-meta because we need those before we can even reach the stack.

Each field on the interface has a specific job:
- `service` to determine which geo type we're dealing with (Map, PlaceIndex, or GeofenceCollection) and extract the right params from the stack
- `s3TemplateURL` to download the CFN template so cdk-from-cfn can convert it into a TypeScript L1 construct
- `logicalId` to look up the nested stack's deployed parameters and pass to cdk-from-cfn as the stack identifier
- `name` for output file paths, construct filenames, and construct IDs (added context for naming below)

**For the naming question:**

When a user runs `amplify add geo` in Gen1, the CLI prompts them to provide a name (e.g. "Provide a name for the Map:"). Say they type `myMap`. That name becomes the key in amplify-meta.json :

```json
{
  "geo": {
    "myMap": {
      "service": "Map",
      "providerMetadata": { ... }
    }
  }
}
```

The pipeline copies it onto the definition (`geoObj.name = geoName`) before passing it to `generateGeoL1Code()`. From there, `resourceName` drives all the naming:

```typescript
// generateGeoL1Code() in cdk-from-cfn.ts

const resourceName = definition.name ?? 'geo';
// -> "myMap"

const constructFileName = `${resourceName}-construct`;
// -> "myMap-construct"

const filePath = path.join(this.dir, 'amplify', 'geo', resourceName, `${constructFileName}.ts`);
// -> "<outputDir>/amplify/geo/myMap/myMap-construct.ts"
```

Then in migration-pipeline.ts, the per-resource directory is also based on the name:

```typescript
// migration-pipeline.ts

for (const geoName of Object.keys(geo)) {
  const geoObj = geo[geoName];
  geoObj.name = geoName;  // copy key onto definition

  const resourceDir = path.join(geoDir, geoName);
  // -> "<outputDir>/amplify/geo/myMap/"

  renderers.push(new EnsureDirectory(resourceDir));
  // ... renders myMap-construct.ts and resource.ts into this dir
}
```

So the user-provided name controls the file/directory structure.

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
sai-ray: I agree. As discussed, the refactor exercise should handle this.

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
sai-ray: Nice observation. isDefault is a string (not boolean) because it comes from CloudFormation stack parameters, which are always strings. Downstream it's compared as === 'true' and emitted as a string literal into the generated construct props.

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
sai-ray: I hope the above reply for interface question makes it clearer.

packages/amplify-cli/src/commands/gen2-migration/generate/unsupported/cdk-from-cfn.ts
sai-ray: As discussed, the refactor exercise should handle this.

PR #14616, opened 2026-02-24
feat: add snapshots for fitnesstracker

dgandhi62:
NA


iliapolo:
Stale, we already the snapshot.

Issue #14603, opened 2026-02-19
(gen2-migration) `generate` command should fetch necessary gen1 attributes for REST API


iliapolo:
### Describe the feature you'd like to request

Currently, we instruction our customers to manually extract the following Gen1 REST API resource attributes after codegen is complete:

- `restApiId`
- `rootResourceId`

These are required in order to build the following IAM policy:

```ts
const gen1RestApi = RestApi.fromRestApiAttributes(restApiStack, "Gen1RestApi", {
    restApiId: '<gen1-rest-api-id>',
    rootResourceId: '<gen1-root-resource-id>',
})
const gen1RestApiPolicy = new Policy(restApiStack, "Gen1RestApiPolicy", {
    statements: [
        new PolicyStatement({
            actions: ["execute-api:Invoke"],
            resources: [`${gen1RestApi.arnForExecuteApi("*", "/*")}`]
        })
    ]
});
backend.auth.resources.authenticatedUserIamRole.attachInlinePolicy(gen1RestApiPolicy);
``` 

Which allows the auth role to keep accessing the Gen1 REST API.

> See https://github.com/aws-amplify/amplify-cli/blob/gen2-migration/GEN2_MIGRATION_GUIDE.md#post-generate--rest-api

Lets automate that!

### Describe the solution you'd like

We can grab the value for those properties during codegen and hardcode them into the Gen2 app.

### Describe alternatives you've considered

None

### Additional context

Similar issue: https://github.com/aws-amplify/amplify-cli/issues/14518

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

PR #14619, opened 2026-02-24
feat(gen2-migration): add dynamic fetching of rest api id and resource id for apps with rest apis

dgandhi62:
Solves https://github.com/aws-amplify/amplify-cli/issues/14603

- rest api id is fetched from amplify-meta.json
- resource id is fetched from api gateway 

## Input
```json
"output": {
        "ApiName": "nutritionapi",
        "RootUrl": "https://55iigxjh1f.execute-api.us-east-1.amazonaws.com/dev",
        "ApiId": "55iigxjh1f"
      },
```
Screenshot 2026-02-24 at 6 10 41 PM

## Output
```typescript
const gen1nutritionapiApi = RestApi.fromRestApiAttributes(nutritionapiStack, "Gen1nutritionapiApi", {
    restApiId: "55iigxjh1f",
    rootResourceId: "4faao8kuha"
});
```

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
@@ -405,10 +405,16 @@ export class DataDefinitionFetcher {
    */
   private getGen1ApiResourceId = async (restApiId: string): Promise<string | undefined> => {
     const client = new APIGatewayClient({});
-    const response = await client.send(new GetResourcesCommand({ restApiId }));
+    let position: string | undefined;
 
-    const rootResource = response.items?.find((resource) => resource.path === '/');
-    return rootResource?.id;
+    do {
+      const response = await client.send(new GetResourcesCommand({ restApiId, position }));
+      const rootResource = response.items?.find((resource) => resource.path === '/');
+      if (rootResource) return rootResource.id;
+      position = response.position;
+    } while (position);
+
+    return undefined;
   };
 
   /**


reports/report.xml
dgandhi62, line 1: Unsure what this file is / how this came. Looking into it (maybe its unnecessary?)

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
```diff
@@ -387,6 +400,16 @@ export class DataDefinitionFetcher {
       throw new Error(`Failed to fetch logging config from AWS: ${error.message}`);
     }
   };
+  /**
+   * Fetches root resource ID from AWS API Gateway
+   */
+  private getGen1ApiResourceId = async (restApiId: string): Promise<string | undefined> => {
+    const client = new APIGatewayClient({});
+    const response = await client.send(new GetResourcesCommand({ restApiId }));
```
9pace: nit: Maybe catch error if this fails like what `getAdditionalAuthProvidersFromConsole` does

reports/report.xml
9pace: Its auto generated from jest. if you run  test with npx jest from repository root this gets generated. Not needed to commit

packages/amplify-cli/src/commands/gen2-migration/generate/backend/synthesizer.ts
```diff
@@ -1633,11 +1633,8 @@ export class BackendSynthesizer {
                     factory.createStringLiteral(`Gen1${restApi.apiName}Api`),
                     factory.createObjectLiteralExpression(
                       [
-                        factory.createPropertyAssignment('restApiId', factory.createStringLiteral(`<gen1-${restApi.apiName}-api-id>`)),
-                        factory.createPropertyAssignment(
-                          'rootResourceId',
-                          factory.createStringLiteral(`<gen1-${restApi.apiName}-root-resource-id>`),
-                        ),
+                        factory.createPropertyAssignment('restApiId', factory.createStringLiteral(restApi.gen1RestApiId)),
+                        factory.createPropertyAssignment('rootResourceId', factory.createStringLiteral(restApi.gen1ApiResourceId)),
```
9pace, line 1637: is it okay to access restApi.gen1ApiResourceId without null check?

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
9pace: requires pagination (unless you can find reliable documentation that says / is always returned in the first page, i couldnt)

dgandhi62: We need to remove try-catches. in this scenario, a failure here would mean the api call failed, and we would want to fail the command then anyway

packages/amplify-cli/src/commands/gen2-migration/generate/backend/synthesizer.ts
dgandhi62: Both are required properties on a rest api - https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.RestApiAttributes.html

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
dgandhi62: It is a required property - https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_apigateway.RestApiAttributes.html

dgandhi62: Done

Issue #14559, opened 2026-01-28
(gen2-migration) `generate` command should handle custom resolvers


iliapolo:
### Describe the feature you'd like to request

Gen1 offers a mechanism for customers to write their own [custom AppSync resolvers](https://docs.amplify.aws/gen1/javascript/tools/cli-legacy/overwrite-customize-resolvers/). We need to migrate those resolvers as well.

### Describe the solution you'd like

In Gen2, configuring customer resolvers is done via the CDK by accessing the underlying AppSync API construct from `backend.data`. For example:

```ts
backend.data.addResolver("custom", {...});
```

### Describe alternatives you've considered

None

### Additional context

https://github.com/aws-amplify/amplify-category-api/tree/main/packages/amplify-graphql-api-construct

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

PR #14624, opened 2026-02-26
feat (gen2-migration): add migration support for overriding default resolver

dgandhi62:
Linked Issue (partially addressed with this pr) - https://github.com/aws-amplify/amplify-cli/issues/14559

This pr adds migration support for overriding default pipeline resolvers created by Gen1. Gen1 & Gen2 use pipeline resolvers. Therefore we need to access the functions inside these resolvers and modify their responseMappingTemplates. Furthermore, we need to set accessing them from s3 as undefined and take them locally instead. 

## Input
amplify/backend/api/productcatalog/resolvers/Query.listProducts.res.vtl
```vtl
## Add 10% discount to all products
#foreach($item in $ctx.result.items)
  #if($item.price)
    #set($discount = $item.price * 0.10)
    #set($discountedPrice = $item.price - $discount)
    $util.qr($item.put("discountedPrice", $discountedPrice))
    $util.qr($item.put("savings", $discount))
  #end
#end

$util.toJson($ctx.result)
```

## Output

amplify/backend.ts
```typescript
import { readFileSync, readdirSync } from "fs";
import { join, dirname } from "path";
import { fileURLToPath } from "url";

const __dirname = dirname(fileURLToPath(import.meta.url));
const resolversDir = join(__dirname, "data/resolvers");
const resolverFiles = readdirSync(resolversDir).filter(f => f.endsWith(".res.vtl"));
for (const file of resolverFiles) {
    const [typeName, fieldName] = file.replace(".res.vtl", "").split(".");
    const functionId = `${typeName}${fieldName.charAt(0).toUpperCase() + fieldName.slice(1)}DataResolverFn`;
    const pipelineFunction = backend.data.resources.cfnResources.cfnFunctionConfigurations[functionId];
    if (pipelineFunction) {
        const template = readFileSync(join(resolversDir, file), "utf8");
        pipelineFunction.responseMappingTemplateS3Location = undefined;
        pipelineFunction.responseMappingTemplate = template;
    }
}
```

amplify/data/resolvers/Query.listProducts.res.vtl
```vtl
## Add 10% discount to all products
#foreach($item in $ctx.result.items)
  #if($item.price)
    #set($discount = $item.price * 0.10)
    #set($discountedPrice = $item.price - $discount)
    $util.qr($item.put("discountedPrice", $discountedPrice))
    $util.qr($item.put("savings", $discount))
  #end
#end

$util.toJson($ctx.result)
```


packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
```diff
@@ -111,10 +115,10 @@ export interface CorsConfiguration {
   maxAge?: number;
 }
 
-import { BackendEnvironmentResolver } from './backend_environment_selector';
-import { BackendDownloader } from './backend_downloader';
-import { fileOrDirectoryExists } from './directory_exists';
-import { AppSyncClient, GetGraphqlApiCommand } from '@aws-sdk/client-appsync';
+// Add locally in the fetcher
```
sai-ray, line 118: Do we need this comment? If I am not wrong this just tells that the imports were moved up?

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
```diff
@@ -111,10 +115,10 @@ export interface CorsConfiguration {
...
+export interface ResolverConfig {
```
sai-ray, line 119: Nit: ResolverConfig wraps a single boolean. Could we just add `hasResolvers?: boolean` directly to DataDefinition instead? Saves the extra interface, the wrapping/unwrapping, and the import. Simpler to read too.

packages/amplify-cli/src/commands/gen2-migration/generate/core/migration-pipeline.ts
```diff
@@ -116,6 +118,29 @@ export interface Gen2RenderingOptions {
...
+const copyResolverFiles = (outputDir: string): Renderer => ({
+  render: async () => {
+    const rootDir = pathManager.findProjectRoot();
+    const projectName = getProjectName();
+    const resolversPath = path.join(rootDir, 'amplify', 'backend', 'api', projectName, 'resolvers');
+    const targetPath = path.join(outputDir, 'amplify', 'data', 'resolvers');
+
+    if (require('fs').existsSync(resolversPath)) {
+      const files = require('fs').readdirSync(resolversPath);
+      const vtlFiles = files.filter((file: string) => file.endsWith('.vtl'));
```
sai-ray, line 133: I noticed copyResolverFiles copies all `.vtl` files but `backend.ts` only processes `.res.vtl` does that mean `.req.vtl` files get copied but silently ignored at deploy time? Do we handle `.req.vtl` as well or is it in plan for a different PR?

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
```diff
@@ -4,8 +4,12 @@ import glob from 'glob';
 import assert from 'node:assert';
 
 import { DataDefinition } from '../core/migration-pipeline';
-import { AdditionalAuthProvider } from '../generators/data';
+import { AdditionalAuthProvider, getProjectName } from '../generators/data';
```
sai-ray, line 7: `data_definition_fetcher.ts` and `generators/data/index.ts` import from each other. `getProjectName` makes this a runtime circular dependency. Ideally, I believe we should move but it should be okay to defer this to the refactor exercise if you prefer.

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
```diff
@@ -138,6 +142,22 @@ export class DataDefinitionFetcher {
    */
   constructor(private backendEnvironmentResolver: BackendEnvironmentResolver, private ccbFetcher: BackendDownloader) {}
 
+  /**
+   * Checks if GraphQL API has resolvers directory with VTL files
+   */
+  private hasResolvers = (): boolean => {
+    const rootDir = pathManager.findProjectRoot();
+    const projectName = getProjectName();
```
sai-ray, line 150: Correct me if I am wrong,`getProjectName()` can return undefined and if that happens, path.join would produce .../api/undefined/resolvers and resolver migration would be silently skipped. Is a guard needed here, or is it safe to say that the project name always exists at this point?"

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
```diff
@@ -138,6 +142,22 @@ export class DataDefinitionFetcher {
...
+
+    const resolversPath = path.join(rootDir, 'amplify', 'backend', 'api', projectName, 'resolvers');
+
+    if (!require('fs').existsSync(resolversPath)) return false;
+
+    const files = require('fs').readdirSync(resolversPath);
```
sai-ray, line 156: The rest of the file uses fileWriter for writing files (makes it easy to mock in tests), but this uses require('fs').copyFileSync directly. Any reason not to use fileWriter here too for consistency?

packages/amplify-cli/src/commands/gen2-migration/generate/core/migration-pipeline.ts
dgandhi62: This is a missing functionality, I'll add it

packages/amplify-cli/src/commands/gen2-migration/generate/codegen-head/data_definition_fetcher.ts
dgandhi62: I think we can defer structural issues to refactor for now (as long as the code is passing the tests)

PR #14604, opened 2026-02-19
feat(gen2-migration): implement holding stack for Gen2 resource retention during refactor

9pace:
#### Description of changes
During forward migration, Gen2 stateful resources are now moved to a temporary holding stack instead of being deleted. This preserves test data that customers may have created while testing the Gen2 deployment.

#### Issue #, if available
#14531

#### Description of how you validated changes
Ran manual flow of forward refactor and rollback on discussions app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are [changed or added](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)
- [x] Relevant documentation is changed or added (and PR referenced)
- [x] New AWS SDK calls or CloudFormation actions have been added to relevant test and service IAM policies
- [x] [Pull request labels](https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#labels) are added

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #14638, opened 2026-03-02
fix(gen2-migration): refactor throws unexpected error

iliapolo:
Followup to https://github.com/aws-amplify/amplify-cli/pull/14604 which introduced a couple of minor bugs.

#### Description of changes

See PR comments.

Manul

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/refactor/generators/template-generator.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/refactor/generators/template-generator.ts
@@ -425,7 +425,6 @@ class TemplateGenerator {
     try {
       // Resolve gen 2 template
       const { newTemplate, parameters: gen2StackParameters } = await categoryTemplateGenerator.generateGen2PreProcessTemplate();
-      assert(gen2StackParameters);
       this.logger.info(`Updating Gen 2 ${this.getStackCategoryName(category)} stack...`);
 
       const gen2StackUpdateStatus = await tryUpdateStack(this.cfnClient, destinationCategoryStackId, gen2StackParameters, newTemplate);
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/commands/gen2-migration/refactor/generators/template-generator.ts
```diff
@@ -425,7 +425,6 @@ class TemplateGenerator {
     try {
       // Resolve gen 2 template
       const { newTemplate, parameters: gen2StackParameters } = await categoryTemplateGenerator.generateGen2PreProcessTemplate();
-      assert(gen2StackParameters);
```
iliapolo, line 428: This was never asserted on before and can actually be `undefined` - a case that is gracefully handled in subsequent code.

packages/amplify-cli/src/commands/gen2-migration/refactor/holding-stack.ts
```diff
@@ -36,7 +36,7 @@ export async function findHoldingStack(cfnClient: CloudFormationClient, stackNam
     }
     return null;
   } catch (error: unknown) {
-    if (error instanceof StackNotFoundException) {
```
iliapolo, line 39: This isn't thrown on `DescribeStacksCommand` apparently.

packages/amplify-cli/src/commands/gen2-migration/refactor/generators/template-generator.ts
9pace: how is it handled? Is it fine to not pass params to UpdateStackCommand? Will it exit in failure and then tryUpdateStack throws?

iliapolo: I haven't dug through but the fact that removing this line made things work means something is handling it. Good enough for me at the moment.

Issue #14577, opened 2026-02-06
(gen2-migration) `lock` command overrides existing root stack policy


iliapolo:
### How did you install the Amplify CLI?

npm install --no-save @aws-amplify/cli-internal-gen2-migration-experimental-alpha

### Amplify CLI Version

0.6.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

The `lock` command currently overrides the policy the root stack:

https://github.com/aws-amplify/amplify-cli/blob/4cf7c8deb857455066b83f35bcbac9c1412fc583/packages/amplify-cli/src/commands/gen2-migration/lock.ts#L50-L67

Any existing stack policy statements are lost, which is not good since we customer may still need them. Its gets even worse because during `lock --rollback`, we set the same policy with an `Allow` instead of `Deny`, so we might end up with a policy thats more permissive than the customer originally had.

### Expected behavior

Instead of overriding, we should **add** our statement to it by first reading the existing policy with `GetStackPolicyCommand`.

### Reproduction steps

1. Deploy a gen1 app.
2. Manually add some stack policy to the root stack.
3. Run `npx amplify gen2-migration lock`

Observe the policy added in (2) being removed.

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Additional information

This also means `--rollback` needs to do the reverse, subtract instead of overriding.

### Before submitting, please confirm:

- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [x] I have removed any sensitive information from my code snippets and submission.

PR #14648, opened 2026-03-06
feat(gen2-migration): preserve existing stack policies in lock step

sai-ray:
#### Description of changes

The lock command was overwriting the entire CloudFormation stack policy with a deny-all, losing any existing customer statements. On rollback, it set an allow-all, potentially leaving the stack more permissive than it was originally.

In `lock.ts`:

- `execute()` now reads the existing stack policy and appends the lock statement instead of overwriting. 
- `rollback()` finds and removes only the lock statement, preserving customer statements. Falls back to allow-all only when users didn't have any stack policy set before migration.

#### Issue #, if available

#14577 

- Added unit tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/commands/gen2-migration/lock.ts
```diff
@@ -93,29 +111,21 @@ export class AmplifyMigrationLockStep extends AmplifyMigrationStep {
...
-        },
-      ],
-    });
-
     operations.push({
       describe: async () => {
-        return [`Set a policy on stack '${this.rootStackName}': ${stackPolicy}`];
+        return [`Add lock statement to stack policy on '${this.rootStackName}': ${JSON.stringify(LOCK_STATEMENT)}`];
       },
       execute: async () => {
```
9pace, line 93: execute is not idempotent. lets say user ctrl + c right after this line executes, and then runs lock again, the result is 

"Statement": [
  { "Effect": "Deny", "Action": "Update:*", "Principal": "*", "Resource": "*" },
  { "Effect": "Deny", "Action": "Update:*", "Principal": "*", "Resource": "*" }

This technically doesn't change anything, but if you then UNLOCK, we would only remove one copy. And that would leave the n-1 extra policy statements.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
```diff
@@ -163,29 +173,32 @@ export class AmplifyMigrationLockStep extends AmplifyMigrationStep {
...
-        },
-      ],
-    });
-
     operations.push({
       describe: async () => {
-        return [`Set a policy on stack '${this.rootStackName}': ${stackPolicy}`];
+        return [`Remove lock statement from stack policy on '${this.rootStackName}': ${JSON.stringify(LOCK_STATEMENT)}`];
       },
       execute: async () => {
```
9pace, line 152: Same issue here where it would only remove the first copy of the lock statement, if there is a case that multiple statements got added (which is currently possible). This issue should be solved by fixing the first idempotency issue.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
```diff
@@ -163,29 +173,32 @@ export class AmplifyMigrationLockStep extends AmplifyMigrationStep {
...
+            statement.Effect === LOCK_STATEMENT.Effect &&
+            statement.Action === LOCK_STATEMENT.Action &&
+            statement.Principal === LOCK_STATEMENT.Principal &&
+            statement.Resource === LOCK_STATEMENT.Resource,
+        );
+        if (index !== -1) {
+          existingPolicy.Statement.splice(index, 1);
+        }
+        const restoredPolicy = existingPolicy.Statement.length > 0 ? JSON.stringify(existingPolicy) : JSON.stringify(ALLOW_ALL_POLICY);
         await this.cfnClient().send(
```
9pace, line 161: this would write back the same policy if lock wasnt found? causes an unecessary api call.

if the lock wasn't found, skip the write entirely.

packages/amplify-cli/src/__tests__/commands/gen2-migration/lock.test.ts
```diff
@@ -0,0 +1,223 @@
...
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('execute stack policy merge', () => {
+    it('should append lock statement to empty stack policy', async () => {
+      mockCfnSend.mockResolvedValueOnce({ StackPolicyBody: undefined }).mockResolvedValueOnce({});
+
+      const operations = await lockStep.execute();
+      await operations[3].execute();
```
9pace: I don't really like operations[3], i would prefer a different way to identify operations if possible or at least 
a comment that tells me what operations[3] is. (also anywhere else where you just index operations.)
It is also susceptible to a change in operations order etc.

sai-ray: Its a good call and I agree. Replaced all operations[N] with a findOperation helper that looks up operations by description substring. This is much more readable, and more importantly order-independent. However, It's coupled to the description text, but I believe that's the right tradeoff as a description change would break the test loudly compare to an index shift that would just test a wrong operation.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
sai-ray: Fixed. Execute now checks if the lock statement already exists before appending. If it's already there, we skip. This prevents duplicates from Ctrl+C re-runs, which also resolves the rollback only-removes-one-copy issue since there can only ever be one copy.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
sai-ray: Resolved by the idempotency fix above.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
sai-ray: Fixed. Rollback now returns early when the lock statement isn't found. No `SetStackPolicy` call.

9pace (approved):
Lgtm, I may add that we maybe should be explicit about the behavior for unlock. (that it removes that specific policy)

PR #14659, opened 2026-03-10
feat: add exponential backoff retry to SSMClientWrapper for function secrets

sanjanaravikumar-az:
This PR adds exponential backoff retry logic and increases the page size for SSM `GetParametersByPath` calls in `SSMClientWrapper`, the class responsible for fetching Lambda function secrets during Amplify CLI deployments.

Customers with a high number of SSM parameters are experiencing intermittent "Rate exceeded" errors during CI/CD backend builds. The error originates from `SSMClientWrapper.getSecretNamesByPath`, which is called during environment initialization and pre-push to fetch function secrets from SSM Parameter Store.

SSM Parameter Store has a shared throughput limit of 40 TPS across `GetParameter`, `GetParameters`, and `GetParametersByPath` operations. The `getSecretNamesByPath` method paginates through parameters in a tight loop, and with a small page size of 10, a customer with hundreds of parameters generates many rapid API calls that easily exhaust this budget — especially when other deployment operations are hitting SSM concurrently. With zero retry logic, a single throttled request immediately fails the entire deployment.

#### A retry mechanism exists:

The Amplify CLI monorepo has two independent SSM code paths in separate packages:

1. `amplify-provider-awscloudformation` (`src/utils/ssm-utils/`) handles CloudFormation environment parameters (uploading/downloading CFN parameter values to SSM, deleting environment parameters). This code uses `executeSdkPromisesWithExponentialBackOff` from `exp-backoff-executor.ts`, which retries on `ThrottlingException` with up to 5 retries and jittered exponential backoff.

2. `amplify-category-function` (`src/provider-utils/awscloudformation/secrets/`) handles Lambda function secrets (SecureString parameters). Its `SSMClientWrapper` makes raw `ssmClient.send()` calls with no retry logic.

These two packages have no dependency relationship — `amplify-category-function` cannot import from `amplify-provider-awscloudformation` without introducing a cross-cutting dependency. The backoff executor was never ported to or replicated in the function category package, leaving function secret operations unprotected against throttling.

`SSMClientWrapper` in `amplify-category-function` is the only code path for function secret SSM operations. It's called by `FunctionSecretsStateManager` during:

- **Pre-push**: `prePushHandler` → `ensureNewLocalSecretsSyncedToCloud` → `getCloudFunctionSecretNames` → `getSecretNamesByPath`
- **Environment init/clone**: `cloneSecretsOnEnvInitHandler` → `getEnvCloneDeltas` → `getCloudFunctionSecretNames` → `getSecretNamesByPath`
- **Environment removal**: `deleteAllEnvironmentSecrets` → `getSecretNamesByPath`

#### Code changes:

1. **Adds `executeWithExponentialBackOff`** — a local retry utility that wraps an async operation with jittered exponential backoff on `ThrottlingException`/`Throttling` errors. Configured with 8 max retries and a 30-second max backoff, providing a total backoff window of ~75 seconds to handle sustained throttling in high-parameter-count accounts.

2. **Wraps the `GetParametersByPath` SDK call** inside `getSecretNamesByPath` with this retry utility, so each paginated request is individually retried on throttle.

3. **Increases `MaxResults` from 10 to 50** , reducing the number of API calls and directly lowering throttling pressure.

This PR is meant to solve [this](https://t.corp.amazon.com/V2114824403/) issue.

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
+++ b/packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
@@ -14,26 +14,26 @@ import {
  */
 const executeWithExponentialBackOff = async <T>(operation: () => Promise<T>): Promise<T> => {
   const MAX_RETRIES = 8;
-  const MAX_BACK_OFF_IN_MS = 30 * 1000;
-  const MIN_BACK_OFF_IN_MS = 1000;
+  const MAX_BACK_OFF_IN_MS = 10 * 1000; // 10 seconds
+  const MIN_BACK_OFF_IN_MS = 1000; // 1 second
   let backOffSleepTimeInMs = 500;
-  let consecutiveRetries = 0;
+  let lastError: Error | undefined;
 
-  // eslint-disable-next-line no-constant-condition
-  while (true) {
+  for (let attempt = 0; attempt <= MAX_RETRIES; attempt++) {
     try {
       return await operation();
     } catch (e) {
-      if ((e?.name === 'ThrottlingException' || e?.name === 'Throttling') && consecutiveRetries < MAX_RETRIES) {
-        ++consecutiveRetries;
+      lastError = e;
+      if ((e?.name === 'ThrottlingException' || e?.name === 'Throttling') && attempt < MAX_RETRIES) {
         await new Promise((resolve) => setTimeout(resolve, backOffSleepTimeInMs));
-        backOffSleepTimeInMs = 2 ** consecutiveRetries * backOffSleepTimeInMs;
+        backOffSleepTimeInMs = 2 ** (attempt + 1) * backOffSleepTimeInMs;
         backOffSleepTimeInMs = Math.max(Math.min(Math.random() * backOffSleepTimeInMs, MAX_BACK_OFF_IN_MS), MIN_BACK_OFF_IN_MS);
         continue;
       }
       throw e;
     }
   }
+  throw lastError;
 };
 
 /**


packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
```diff
@@ -9,6 +9,33 @@ import {
   SSMClient,
 } from '@aws-sdk/client-ssm';
 
+/**
+ * Executes an async operation with exponential backoff retry on throttling exceptions.
+ */
+const executeWithExponentialBackOff = async <T>(operation: () => Promise<T>): Promise<T> => {
+  const MAX_RETRIES = 8;
+  const MAX_BACK_OFF_IN_MS = 30 * 1000;
```
ShadowCat567: Why 30 seconds for the Max exponential back off time, that is a pretty long time to wait if that ends up being less than `Math.random() * backOffSleepTimeInMs`. Any reason why you did not go for 10 seconds (10 * 1000) as the Max backoff (like the other exponential backoff implementation: https://github.com/aws-amplify/amplify-cli/blob/05dec5900a9052c1d48ad26e77eccb7d6db3a67b/packages/amplify-provider-awscloudformation/src/utils/ssm-utils/exp-backoff-executor.ts#L3)?

packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
```diff
@@ -9,6 +9,33 @@ import {
...
+  const MIN_BACK_OFF_IN_MS = 1000;
+  let backOffSleepTimeInMs = 500;
+  let consecutiveRetries = 0;
+
+  // eslint-disable-next-line no-constant-condition
+  while (true) {
```
ShadowCat567: I agree with the linter, I'm not a huge fan of this `while (true)`. Could this be `while (consecutiveRetries < MAX_RETRIES)` instead?

packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
sanjanaravikumar-az: That is true, in this case the customer had a really busy SSM account, but I think it's fair to keep it to 10 seconds for now and see if that works alright.

packages/amplify-category-function/src/provider-utils/awscloudformation/secrets/ssmClientWrapper.ts
sanjanaravikumar-az: Ah yes, so the loop needs to run at least once (for the initial attempt before any retries), and on success it returns immediately. If we just swap to ```while (consecutiveRetries < MAX_RETRIES)```, the initial attempt ```(when consecutiveRetries is 0)``` works fine, but after the 8th retry (```consecutiveRetries``` becomes 8, which equals ```MAX_RETRIES```), the loop exits silently without throwing. We'd need to handle that final throw after the loop. So I'll update the code to handle this.

Issue #14524, opened 2026-01-26
(gen2-migration) `decommission` command attempts to delete DynamoDB model tables


iliapolo:
### How did you install the Amplify CLI?

npm install --no-save @aws-amplify/cli-internal-gen2-migration-experimental-alpha

### Amplify CLI Version

0.6.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

Running `decommission` post migration results in the following validation error:

```console
[2026-01-26T01:13:49.468Z] [decommission] [mediavault/main] • Stateful resources scheduled for deletion

┌──────────┬──────────────────────┬──────────────────────────────────────┐
│ Category │ Resource Type        │ Physical ID                          │
├──────────┼──────────────────────┼──────────────────────────────────────┤
│ Api      │ AWS::DynamoDB::Table │ Note-l7ho3fum4nfxbc6rwg6icdnrnq-main │
└──────────┴──────────────────────┴──────────────────────────────────────┘

🛑 Validations failed: Decommission will delete stateful resources.
```

This is good since it correctly identifies that DynamoDB tables are incorrectly scheduled for deletion. Running this command this `--skip-validation` won't actually delete the tables because they have deletion protection enabled; but the command itself will fail because CFN won't be able to delete the stack.

### Expected behavior

By the time a customer runs the `decommission` command, these tables should not be scheduled for deletion and the command should complete successfully.

### Reproduction steps

- Migrate any app that contains GraphQL.
- Run `decommission`

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Additional information

What happens is that because model tables don't undergo refactoring, they stay managed in the Gen1 stack (in addition to the new Gen2 stack).

The solution is would be for the _lock_ command to change the CFN [DeletionPolicy](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-deletionpolicy.html) of those tables to `RETAIN`. 

### Before submitting, please confirm:

- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [x] I have removed any sensitive information from my code snippets and submission.

PR #14662, opened 2026-03-11
fix(gen2-migration): retain GraphQL model tables during `decommission`

sai-ray:
Fixes: #14524

#### Description of changes

Running `amplify gen2-migration decommission` after migrating an Amplify app with api DynamoDB tables, fails with `Validations failed: Decommission will delete stateful resources`. The refactor step moves auth, storage, and analytics resources to Gen2 but not GraphQL model tables. Gen2 imports those via `modelNameToTableNameMapping`. So when decommission tries to delete the Gen1 stack, those tables are still there.

#### Changes

Lock step (`lock.ts`):
- Added `findApiCategoryStacks()`: identifies API nested stacks using `extractCategory` on logical resource IDs from `DescribeStackResources`.
- Added `setDeletionPolicyRetainOnDynamoTables()` for each API stack: lists its model nested stacks, fetches each template via `GetTemplate`, sets `DeletionPolicy: Retain` on `AWS::DynamoDB::Table` resources.
- Added changeset validation before applying the update: creates a changeset, validates that every change is a Modify on `AWS::DynamoDB::Table` or `AWS::IAM::Policy` (known side effect), aborts if unexpected changes are found. The changeset is deleted after validation, then the update is applied via `UpdateStackCommand` + `waitUntilStackUpdateComplete`.
- Replaced tryUpdateStack (removed in #14698) with inline UpdateStackCommand + SDK waiter, matching the Cfn.update() pattern.

Validation (`validations.ts`):
- In `getStatefulResources`, when a `AWS::DynamoDB::Table` is found in a nested stack, fetches the stack template and checks `DeletionPolicy`. If `Retain`, skips flagging it. Tables without `Retain` are still flagged.

By submitting this pull request, I confirm that my contribution is
made under the terms of the Apache 2.0 license.

- Ran the updated `lock` and `decomission` commands on a sample Gen1 app.
- Added unit tests.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/commands/gen2-migration/lock.ts
```diff
@@ -250,6 +273,68 @@ export class AmplifyMigrationLockStep extends AmplifyMigrationStep {
...
+
+          // Pass existing parameters through unchanged
+          const describeResponse = await this.cfnClient().send(new DescribeStacksCommand({ StackName: modelStackId }));
+          const parameters = (describeResponse.Stacks?.[0]?.Parameters ?? []).map((p) => ({
+            ParameterKey: p.ParameterKey,
+            UsePreviousValue: true,
+          }));
+
+          this.logger.info(`Updating DeletionPolicy for table '${logicalId}'...`);
+          await tryUpdateStack(this.cfnClient(), modelStackId, parameters, template);
```
9pace: `tryUpdateStack` takes a params object. It should look like: 
```
await tryUpdateStack({
  cfnClient: this.cfnClient(),
  stackName: modelStackId,
  parameters,
  templateBody: template,
});
```
(I think this was a recent change to the code so merge from gen2-migration to see what i mean.)

9pace:
For `lock --rollback`, what do we do? Leave them with the retain policy? 

Also, this would modify our deployed cfn template. The proposed rollback validation is running template drift detection which would conflict here. Just flagging for discussion.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
```diff
@@ -250,6 +273,68 @@ export class AmplifyMigrationLockStep extends AmplifyMigrationStep {
...
   }
 
+  private async findApiCategoryStacks(): Promise<string[]> {
+    const response = await this.cfnClient().send(new DescribeStackResourcesCommand({ StackName: this.rootStackName }));
+    const stackResources = response.StackResources ?? [];
+    return stackResources
+      .filter(
+        (resource) =>
+          resource.ResourceType === 'AWS::CloudFormation::Stack' &&
+          resource.LogicalResourceId?.startsWith('api') &&
```
9pace: Maybe better to use `extractCategory` and check if it is 'api'.

packages/amplify-cli/src/commands/gen2-migration/lock.ts
sai-ray: Updated

packages/amplify-cli/src/commands/gen2-migration/lock.ts
sai-ray: Updated. I was intentionally using startsWith('api') to avoid the includes false positive risk. But yeah its better to centralize switched to extractCategory here. I think we should prioritze that some time soon.

9pace: agreed

9pace (changes requested):
Before tryUpdateStack, we should add a changeset validation step similar to what we do in the refactor step.

Right now, `setDeletionPolicyRetainOnDynamoTables` modifies the template and calls tryUpdateStack directly without verifying that the only changes CloudFormation will apply are DeletionPolicy updates. 
for the lock step, we should implement:

 1. Before calling tryUpdateStack, create a changeset on the stack
 2. Describe the changeset and inspect Changes
 3. Validate that every change is a Modify on a AWS::DynamoDB::Table resource, scoped to a deletion policy update that we want
 4. If any unexpected changes are present, abort with a clear error

This is the same pattern as in _validations.ts. See: https://github.com/aws-amplify/amplify-cli/blob/a0015a7aef7466e87475b74973ff7b42879d2144/packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/category-refactorer.ts#L147-L154

sai-ray:
> Before tryUpdateStack, we should add a changeset validation step similar to what we do in the refactor step.
> 
> Right now, `setDeletionPolicyRetainOnDynamoTables` modifies the template and calls tryUpdateStack directly without verifying that the only changes CloudFormation will apply are DeletionPolicy updates. for the lock step, we should implement:
> 
>  1. Before calling tryUpdateStack, create a changeset on the stack  2. Describe the changeset and inspect Changes  3. Validate that every change is a Modify on a AWS::DynamoDB::Table resource, scoped to a deletion policy update that we want  4. If any unexpected changes are present, abort with a clear error
> 
> This is the same pattern as in _validations.ts. See:
> 
> https://github.com/aws-amplify/amplify-cli/blob/a0015a7aef7466e87475b74973ff7b42879d2144/packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/category-refactorer.ts#L147-L154

@9pace 
Added changeset validation before the stack update in `setDeletionPolicyRetainOnDynamoTables`.

The flow is now:

1. Create a changeset with the modified template
2. Wait for creation to complete
3. Describe the changeset and validate the changes
4. Delete the changeset (inspection only, not executed)
5. Apply the update via `UpdateStackCommand` + `waitUntilStackUpdateComplete`

The validation allows only Modify actions on `AWS::DynamoDB::Table` (our intended DeletionPolicy change) and `AWS::IAM::Policy` (known side effect, CloudFormation recalculates the IAM policy because it references the table). Any other action type (Add, Remove) or resource type (AppSync resolvers, IAM roles, etc.) aborts with a descriptive error including a resolution field.

Also replaced the tryUpdateStack import (removed in #14698) with inline `UpdateStackCommand `+ SDK waiter, matching the `Cfn.update()` pattern.

Added 4 tests:

1. Happy path: DynamoDB table + IAM policy Modify, proceeds
2. Add action, aborts
3. Remove action, aborts
4. Modify on unexpected resource type (AppSync resolver), aborts

Validated on a deployed app, lock completes successfully with the changeset validation passing.

9pace:
Still interested in https://github.com/aws-amplify/amplify-cli/pull/14662#issuecomment-4126662503
I will handle the drift scenario, but what about rollback?

9pace (approved):

Issue #14551, opened 2026-01-28
(gen2-migration) refactor the `generate` command code


iliapolo:
### Describe the feature you'd like to request

We need to improve the code quality of the `generate` command. Here are some issues I see with it right now:

- The [directory structure](https://github.com/aws-amplify/amplify-cli/tree/gen2-migration/packages/amplify-cli/src/commands/gen2-migration/generate) is unclear. There is no coherent strategy to organize the code, making it hard to navigate.
- Code that issues AWS calls is scattered throughout many different files that play different roles. We need to clearly separate the part that queries AWS into a single place thats easy to mock.
- The collection of Gen1 information is scattered and unclear. Sometimes we use `cli-inputs.json`, sometimes we use the CFN template, sometimes we query AWS resources directly. We need to take a look at those cases and determine the right approach for each. Also create a clean separation for components that rely on external sources vs components that perform local computation. This is kind of already modeled with our fetchers, but not quite.
- Generation if individual `resource.ts` files doesn't have access to the generation of the `backend.ts` file even though it is often needed. For example, the generation of `storage/resource.ts` requires adding an escape hatch to `backend.ts`. Currently this is done separately and creates a duplication of input extraction that can lead to subtle bugs and makes it hard to navigate the code.

### Describe the solution you'd like

Just resolve the issues mentioned above.

### Describe alternatives you've considered

None

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

PR #14664, opened 2026-03-11
chore(gen2-migration): refactor the generate command

iliapolo:
#### Description of changes

Rewrites the `generate` command's code generation pipeline into a new `generate-new/` module that follows the project's `CODING_GUIDELINES.md`. The old code in `generate/` is left in place (still used by the old test path) but is no longer the active code path for `generate.ts`.

**Why:** The old pipeline (`command-handlers.ts`, `migration-pipeline.ts`, `BackendSynthesizer`) accumulated significant guideline violations documented in the adr — god modules, unjustified layer boundaries, scattered API calls, excessive mutability, and optional-everywhere interfaces. Adding new categories or fixing bugs required changes across 4–5 files.

#### Issue #, if available

https://github.com/aws-amplify/amplify-cli/issues/14551

- All 9 generate snapshot tests pass
- Added standard unit tests.
- 1 live execution on the project-boards app.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -43709,7 +43709,14 @@ __metadata:
   languageName: node
   linkType: hard
 
-"xstate@npm:^4.14.0, xstate@npm:^4.33.6":
+"xstate@npm:^4.14.0":
+  version: 4.26.1
+  resolution: "xstate@npm:4.26.1"
+  checksum: f221af3400a9c19de4eb6c0b0080364b63a3a352b1af0965b54855a6a41d13a3e104878267ea5f19be8ecbb318559b0930237de68c20b065d58112d7ec23ea78
+  languageName: node
+  linkType: hard
+
+"xstate@npm:^4.33.6":
   version: 4.38.3
   resolution: "xstate@npm:4.38.3"
   checksum: 8a2063743517390107275113bca0e757dba99102e7d57d40cf656b5cc03a6f2c5e10fbf3752294d9d29fbe1d8757bb9a54f54c934a22f205a237956dd10dcd0f


packages/amplify-cli/src/commands/gen2-migration/generate/backend/__snapshots__/synthesizer.test.ts.snap
dgandhi62, line 1: Changes to the `generate` folder should not exist

packages/amplify-cli/src/commands/gen2-migration/generate-new/_infra/aws-fetcher.ts
```diff
@@ -0,0 +1,190 @@
...
+export class AwsFetcher {
+  private readonly clients: AwsClients;
+
+  public constructor(clients: AwsClients) {
+    this.clients = clients;
+  }
+
+  // ── Auth (Cognito) ──────────────────────────────────────────────
+
+  public async fetchUserPool(userPoolId: string): Promise<UserPoolType> {
```
dgandhi62, line 45: I argue all these functions should not be in `input`. If these are just `fetcher` functions, then they should be categorized accordingly (or just merged with `generator`). It violates one of the coding guidelines which mentions files and folders should be named appropriately.

Can ignore this in the interest of time

dgandhi62: Extension of thought: I don't fully understand the `input` and `output` naming fully. I think names here are not fully appropriate

amplify-migration-apps/backend-only/_snapshot.post.generate/amplify/data/resource.ts
dgandhi62, line 1: I'm confused on why there exist snapshot updates with this pr. Wasn't the idea to keep snapshots unchanged and just add code?

dgandhi62: Do the input and output snapshot folders also need to be deleted?

dgandhi62 (changes requested):
There are a bunch of small issues/improvements I see which I am not addressing in the review in the interest of time. Have focused on larger problems. I can do a more finegrained review if needed. 

I don't fully understand the snapshot changes, since I was under the impression that the snapshots would need to be fixed and verified first for testing before we refactor.

I also see some snapshot output issues. For example, `fitnesstracker` has an output schema in `amplify-migration-apps/fitness-tracker/_snapshot.post.generate` which I am unable to verify. I see this pattern in the other apps too. Kiro claims that this is a compiled graphql schema and is valid, but I don't see the point since we are unable to verify it visually first. 

My opinion is that a manual double check of all the finalized output snapshots is needed before merge. I am happy to  spend some time and do this if needed once we have the snapshots finalized.

amplify-migration-apps/backend-only/_snapshot.post.generate/amplify/data/resource.ts
iliapolo: Great catch - these snapshot changes should not exist.

packages/amplify-cli/src/commands/gen2-migration/generate-new/_infra/aws-fetcher.ts
iliapolo: > I don't fully understand the input and output naming fully. I think names here are not fully appropriate

Fair enough. I'll flatten it.

iliapolo: > If these are just fetcher functions, then they should be categorized accordingly (or just merged with generator)

Not necessarily. The important point here is that the fetcher doesn't add a conversion layer, it merely serves as the single place for AWS interaction - but it returns SDK types as is. Easier to mock, easier to track.

iliapolo: I decided to stick with the directory level but renamed.

`_input` -> `_infra` to represent just utility classes that help the code itself (infra for the code).
`_output` -> `amplify` to mimc the structure of the generated gen2 app.

amplify-migration-apps/mood-board/_snapshot.post.generate/amplify/backend.ts
```diff
@@ -84,3 +65,22 @@ backend.moodboardKinesisReader.addEnvironment(
...
+      'kinesis:ListStreamConsumers',
+      'kinesis:DescribeStream',
+      'kinesis:DescribeStreamSummary',
+      'kinesis:DescribeStreamConsumer',
+      'kinesis:GetRecords',
+      'kinesis:GetShardIterator',
+      'kinesis:SubscribeToShard',
+      'kinesis:DescribeLimits',
+      'kinesis:ListTagsForStream',
+      'kinesis:SubscribeToShard',
```
dgandhi62: `'kinesis:SubscribeToShard'` is mentioned twice here.

iliapolo: Interesting. Good catch.

iliapolo: fixed

PR #14685, opened 2026-03-17
chore: update feature coverage section of the GEN2_MIGRATION_GUIDE.md

dgandhi62:
This pr updates the feature coverage section to match the latest updates by the engineering team

### DIFF
--- a/GEN2_MIGRATION_GUIDE.md
+++ b/GEN2_MIGRATION_GUIDE.md
@@ -687,7 +687,7 @@ by the CLI setting that configures them.
   - 🟢 `Do you want to add a sort key to your table`
   - 🟢 `Do you want to add global secondary indexes to your table`
   - 🟢 `Do you want to add a sort key to your global secondary index`
-  - ⚠️ `Do you want to add a Lambda Trigger for your Table`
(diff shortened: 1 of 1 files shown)


iliapolo (changes requested):
We need to make sure all these changes are included in our alpha release, otherwise the guide will be misleading.

GEN2_MIGRATION_GUIDE.md
```diff
@@ -687,7 +687,7 @@ by the CLI setting that configures them.
   - 🟢 `Do you want to add a sort key to your table`
   - 🟢 `Do you want to add global secondary indexes to your table`
   - 🟢 `Do you want to add a sort key to your global secondary index`
-  - 🔴 `Do you want to add a Lambda Trigger for your Table`
+  - ⚠️ `Do you want to add a Lambda Trigger for your Table`
```
iliapolo: I don't think "untested" is right here. If it was released in the alpha then i'm pretty sure tested it - if it wasn't - it shouldn't be in the guide anyway.

dgandhi62: Have removed it

PR #14686, opened 2026-03-17
feat: update feature coverage of GEN2_MIGRATION_GUIDE.md

dgandhi62:
This pr updates the feature coverage section to match the latest updates by the engineering team.

Source of truth - https://docs.amplify.aws/gen1/javascript/tools/cli/hosting/

### DIFF
--- a/GEN2_MIGRATION_GUIDE.md
+++ b/GEN2_MIGRATION_GUIDE.md
@@ -533,6 +533,11 @@ by the CLI setting that configures them.
 > to manually augment the generated code to add the necessary configuration.
 > - ⚠️ | Untested. You're welcome to try it out and let us know!
 
+## Types of frontend deployment
+  - ⚠️ Continuous deployment / GitHub 
+  - ⚠️ Manual deployment
+  - ⚠️ s3 and Cloudfront
+
 ## Auth
 
 ### `amplify add auth`
--- a/GEN2_MIGRATION_GUIDE.md
+++ b/GEN2_MIGRATION_GUIDE.md
@@ -533,11 +533,6 @@ by the CLI setting that configures them.
 > to manually augment the generated code to add the necessary configuration.
 > - ⚠️ | Untested. You're welcome to try it out and let us know!
 
-## Types of frontend deployment
-  - ⚠️ Continuous deployment / GitHub 
-  - ⚠️ Manual deployment
-  - ⚠️ s3 and Cloudfront
-
 ## Auth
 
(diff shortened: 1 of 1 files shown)


GEN2_MIGRATION_GUIDE.md
```diff
@@ -533,6 +533,11 @@ by the CLI setting that configures them.
 > to manually augment the generated code to add the necessary configuration.
 > - ⚠️ | Untested. You're welcome to try it out and let us know!
 
+## Types of frontend deployment
```
iliapolo: We have a `hosting` section at the bottom: https://github.com/aws-amplify/amplify-cli/blob/gen2-migration/GEN2_MIGRATION_GUIDE.md#amplify-add-hosting-

Maybe update that based on the CLI command / inputs? just like all the other categories do it.

iliapolo (changes requested):

GEN2_MIGRATION_GUIDE.md
```diff
@@ -792,6 +792,17 @@ by the CLI setting that configures them.
 
 ### `amplify add hosting` 🔴
 
+- ➤ **Select the plugin module to execute**
```
iliapolo, line 795: What are we trying to convey with this messaging? I'm concerned this communicates that there are hosting equivalents for each of these selections in Gen2 - when in fact Gen2 doesn't offer anything except full CI/CD deployments...

I think this needs some more thought. Lets discuss offline.

dgandhi62: I don't see the user being confused here since `amplify add hosting` is a command exclusive to Gen1, as are the prompts mentioned after it. By saying it is `untested`, or `unsupported`, or `supported`, the claim is not that Gen2 supports this. The claim is about our handling of it. 

Example in point - dynamodb. By claiming we are `supporting` dynamodb does not claim that Gen2 supports it. 

I am open to editing the feature coverage preface by making this explicit that the following section highlights what we handle during migration and does not make any claims regarding Gen2 native support. Just that this applies to every command and option mentioned in this section.

iliapolo: > Example in point - dynamodb. By claiming we are supporting dynamodb does not claim that Gen2 supports it.

But it does support, via CDK. Thats kind of the point - marking it as "supported" means that we provide a way for you to get the same functionality in Gen2. With hosting its not like that. 

> I am open to editing the feature coverage preface by making this explicit.

What did you have in mind?

Issue #14689, opened 2026-03-17
(gen2-migration): `refactor` command fails after rollback and redeploy


iliapolo:
### How did you install the Amplify CLI?

Local

### Amplify CLI Version

Local

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

When running the following sequence:

1. `refactor --to <gen2-stack-name>`
2. `refactor --to <gen2-stack-name> --rollback`
3. Redeploy Gen1
4. Redeploy Gen2
5. `refactor --to <gen2-stack-name>`

We get the following error:

```console
→ Planning failed

🛑 [AuthCognitoForwardRefactorer] unable to find source stack
```

### Expected behavior

Not to fail

### Reproduction steps

See above

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Additional information

The problem is that after these sequence of operations, our auth refactorer fails to classify the auth stack correctly:

https://github.com/aws-amplify/amplify-cli/blob/e989847f6c4bac3d8f60195221470f21f448fad6/packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-utils.ts#L35-L50

This is because for some reason, the `Description` of the auth stack changed.

**Before the first refactor it was**

`{"createdOn":"Mac","createdBy":"Amplify","createdWith":"14.2.5","stackType":"auth-Cognito","metadata":{"whyContinueWithGen1":"Prefer not to answer"}}`

**After the sequence it is**

`Amplify Cognito Stack for AWS Amplify CLI`

Not sure whats going on but it looks like the Gen1 CLI has some special handling with respect to stack description:

https://github.com/aws-amplify/amplify-cli/blob/13f1a5f4e89fc9314c81e0aa03b2cd570c0250e2/packages/amplify-provider-awscloudformation/src/template-description-utils.ts#L26-L35

My suspicion:

1. During `amplify push` the template description changes from what's written in the template to this custom description.
2. First rollback works fine.
3. Redeployment of Gen1 via hosting reverts the description to the one in the template (the wrong one).
4. Second rollback fails.

Im not sure why the hosting deployment would change the description. We need to figure out why this is happening and then change the classification logic to inspect the resources inside the template instead of relying on the description.

### Before submitting, please confirm:

- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [x] I have removed any sensitive information from my code snippets and submission.

Issue #14657, opened 2026-03-10
(gen2-migration) `refactor` command fails when user pool groups are configured


iliapolo:
### How did you install the Amplify CLI?

local

### Amplify CLI Version

local

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

```console
[2026-03-10T03:30:41.816Z] [refactor] [mediavault/main] · Stack outputs count: 10
[2026-03-10T03:30:41.816Z] [refactor] [mediavault/main] · Describing Gen2 stack resources...
[2026-03-10T03:30:41.944Z] [refactor] [mediavault/main] · Gen2 Stack Resources count: 51
[2026-03-10T03:30:41.944Z] [refactor] [mediavault/main] · Resolving Gen2 dependencies...
[2026-03-10T03:30:41.945Z] [refactor] [mediavault/main] · Resolving Gen2 output references...
[2026-03-10T03:30:41.947Z] [refactor] [mediavault/main] · Finished resolving Gen2 output references...
[2026-03-10T03:30:41.947Z] [refactor] [mediavault/main] • Updating Gen 2 auth stack...
[2026-03-10T03:31:12.721Z] [refactor] [mediavault/main] • Updated Gen 2 auth stack successfully
[2026-03-10T03:31:12.721Z] [refactor] [mediavault/main] • Moving Gen2 auth resources to holding stack...
[2026-03-10T03:31:12.722Z] [refactor] [mediavault/main] · moveGen2ResourcesToHoldingStack: Moving Gen2 resources to holding stack
[2026-03-10T03:31:12.789Z] [refactor] [mediavault/main] · Deleting resources from template...
[2026-03-10T03:31:12.789Z] [refactor] [mediavault/main] · Deleting resource: amplifyAuthUserPool4BA7F805
[2026-03-10T03:31:12.789Z] [refactor] [mediavault/main] · Deleting resource: amplifyAuthUserPoolUserPoolDomain1F688B5B
[2026-03-10T03:31:12.789Z] [refactor] [mediavault/main] · Deleting resource: amplifyAuthUserPoolNativeAppClient79534448
[2026-03-10T03:31:12.789Z] [refactor] [mediavault/main] · Deleting resource: amplifyAuthUserPoolAppClient2626C6F8
[...]

Issue #14530, opened 2026-01-27
(gen2-migration) `refactor` command should validate stacks integrity before execution


iliapolo:
### Describe the feature you'd like to request

Prior to attempting a refactor, the command should validate that both the Gen1 and the Gen2 stacks are in an expected state. 

**For example:**

What happens if auth is configured in the Gen1 app but the stack doesn't contain a UserPool? I'm pretty sure the refactor will just be skipped for auth and continue to the other categories. 

This isn't the correct behavior because it will be acting upon a very unexpected state, possibly causing some very unexpected consequences. 

We won't be able to come up with all possible corrupt state scenarios, but lets at least think of what the most likely or dangerous scenarios for refactor are validate for those. 

### Describe the solution you'd like

When encountering an unexpected state, the `refactor` command should fail on a validation error. Customers will have the option to skip validations if they deem the state to actually be expected and safe.

### Describe alternatives you've considered

None

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

Issue #14529, opened 2026-01-27
(gen2-migration) `refactor` command should invoke the stateful resources validation before updating stacks


iliapolo:
### Describe the feature you'd like to request

During `gen2-migration refactor --to <gen2-stack>` we update both Gen1 and Gen2 stacks.

Due to a bug or some unforeseen scenario, these updates could accidentally delete stateful resources; we ned to prevent that.

### Describe the solution you'd like

In `cfn-stack-updater.ts` we call the `UpdateStackCommand`:

https://github.com/aws-amplify/amplify-cli/blob/5eb652c115fa807df17668cf7b345122c0280af3/packages/amplify-cli/src/commands/gen2-migration/refactor/cfn-stack-updater.ts#L27-L35

Before we do that we should first create a CFN change set and then reuse our [stateful resources validation](https://github.com/aws-amplify/amplify-cli/blob/gen2-migration/packages/amplify-cli/src/commands/gen2-migration/_validations.ts#L92) to make sure this change set doesn't delete any stateful resource. 

Then, since we are already creating a changeset, we could **execute** that same change set instead of calling `UpdateStack` directly. Having changeset is better for traceability and recovery since it preserves all version of a template.

### Describe alternatives you've considered

None

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

PR #14698, opened 2026-03-24
fix(gen2-migration): `refactor` breaks for user pool groups

iliapolo:
### Description of changes

Fixes the stale template bug in the refactor workflow when two Gen1 stacks map to the same Gen2 stack (auth Cognito + UserPoolGroups both targeting the Gen2 auth stack). The second refactorer would operate on a Gen2 template that the first had already mutated, causing incorrect resource mappings and CloudFormation errors.

#### Deferred template resolution

`RefactorBlueprint` now carries only mappings and stack IDs — no templates. `Cfn.refactor()` fetches and manipulates templates internally at execution time. `beforeMove` and `afterMove` independently discover resources from the current stack template rather than relying on pre-computed blueprint data.

This is the core fix. Previously, all templates were resolved at plan time and baked into the blueprint. When two refactorers targeted the same Gen2 stack, the second one's blueprint contained a stale template snapshot from before the first refactorer modified the stack. By deferring template reads to execution time, each operation sees the stack as it actually is.

#### Shared Cfn instance with plan-time update deduplication

A single `Cfn` instance is shared across all refactorers. `updateSource`/`updateTarget` call `cfn.claimUpdate()` at plan time so the second refactorer targeting the same stack skips the update. Without this, the second refactorer would push its plan-time resolved template — which still contains resources that the first refactorer already moved away — effectively reverting the first refactorer's work. We keep resolution at plan time (rather than deferring to execution) because the changeset validation that runs during the plan phase needs the resolved template to detect unexpected changes.

#### Rollback now resolves and updates both stacks

Rollback previously skipped `resolveSource`/`resolveTarget` and `updateSource`/`updateTarget`. This caused `Fn::GetAtt` dangling reference errors when rolling back after a Gen2 redeploy — the Gen2 stack template had fresh intrinsic references that needed resolving before resources could be removed.

#### Unconditional DependsOn stripping

`resolveDependencies` now strips all `DependsOn` from templates instead of selectively filtering by resource ID. Since `updateSource`/`updateTarget` only runs once per stack (the first refactorer wins), every refactorer's resolution must produce the same template for a given stack. The old `resolveDependencies` took a refactorer-specific resource ID list, meaning two refactorers targeting the same stack would produce different resolved templates. Stripping all `DependsOn` unconditionally makes resolution deterministic regardless of which refactorer runs first. This is safe because `DependsOn` only controls deployment ordering, which is irrelevant during refactor since all resources already exist.

#### Changeset-based update validations

`updateSource` and `updateTarget` now create a CloudFormation changeset before execution and validate that the resolved template introduces no unexpected changes. Previously these validations were stubs that always returned `valid: true`. The changeset report is shown to the user during the plan phase so they can review what will change before confirming.

#### SDK ResourceMapping type

Replaced the custom `MoveMapping` interface with the SDK's `ResourceMapping` type. This removes a redundant interface and aligns the codebase with the CloudFormation API types directly.

#### Test reorganization

Moved test files into subdirectories mirroring the source structure (`resolvers/`, `auth/`, `workflow/`). Merged scattered test files that tested the same source file. Made the CloudFormation mock stateful with a template map that tracks UpdateStack and CreateStackRefactor mutations.

### Issue #, if available

Resolves:

- https://github.com/aws-amplify/amplify-cli/issues/14657
- https://github.com/aws-amplify/amplify-cli/issues/14529
- https://github.com/aws-amplify/amplify-cli/issues/14528
- https://github.com/aws-amplify/amplify-cli/issues/14530
- https://github.com/aws-amplify/amplify-cli/issues/14689

### Description of how you validated changes

- Live tested forward refactor and rollback with auth (Cognito + UserPoolGroups) and storage categories
- All 14 refactor test suites pass (92 tests)
- Snapshot tests updated and passing for all 5 migration apps
- Added refactor snapshot test for fitness-tracker (covers auth with UserPoolGroups)

### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
```diff
@@ -1,32 +1,38 @@
...
-export const AUTH_RESOURCE_TYPES = [
+export const GEN1_NATIVE_APP_CLIENT = 'UserPoolClient';
+export const GEN1_WEB_CLIENT = 'UserPoolClientWeb';
+
+export const GEN2_NATIVE_APP_CLIENT = 'UserPoolNativeAppClient';
+export const GEN2_WEB_CLIENT = 'UserPoolAppClient';
+
+export const USER_POOL_CLIENT_TYPE = 'AWS::Cognito::UserPoolClient';
+
+export const RESOURCE_TYPES = [
```
iliapolo, line 28: Put all resource types in a `const` and reuse in the rollback refactorer

packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/rollback-category-refactorer.ts
```diff
@@ -89,117 +82,97 @@ export abstract class RollbackCategoryRefactorer extends CategoryRefactorer {
...
+        resource: this.resource,
         validate: () => undefined,
-        describe: async () => [`Update ${holdingStackName} to include placeholder resource`],
+        describe: async () => {
+          return [`Update holding stack '${extractStackNameFromId(holdingStackName)}' with placeholder resource`];
+        },
         execute: async () => {
-          await tryUpdateStack({
-            cfnClient: this.clients.cloudFormation,
+          const holdingTemplate = await this.cfn.fetchTemplate(holdingStackName);
```
iliapolo: Needs to happen in forward as well since Gen2 storage stacks contains a single table (excluding CDK metadata which we don't want to rely on).

packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/forward-category-refactorer.ts
```diff
@@ -85,14 +121,12 @@ export abstract class ForwardCategoryRefactorer extends CategoryRefactorer {
   protected async resolveTarget(stackId: string): Promise<ResolvedStack> {
     const facade = this.gen2Branch;
     const originalTemplate = await facade.fetchTemplate(stackId);
-    const description = await facade.fetchStackDescription(stackId);
+    const description = await facade.fetchStack(stackId);
```
iliapolo:
```suggestion
    const stack = await facade.fetchStack(stackId);
```

packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/rollback-category-refactorer.ts
```diff
@@ -89,117 +82,97 @@ export abstract class RollbackCategoryRefactorer extends CategoryRefactorer {
...
-
-    const holdingTemplateResponse = await this.clients.cloudFormation.send(
-      new GetTemplateCommand({ StackName: holdingStackName, TemplateStage: 'Original' }),
-    );
-    if (!holdingTemplateResponse.TemplateBody) {
-      throw new AmplifyError('InvalidStackError', {
-        message: `Holding stack '${holdingStackName}' returned an empty template`,
-      });
+  protected async afterMove(gen2StackId: string): Promise<AmplifyMigrationOperation[]> {
+    const gen2StackName = extractStackNameFromId(gen2StackId);
```
iliapolo, line 105: Need to check for `REVIEW_IN_PROGRESS` of holding stack and delete it if needed.

packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/rollback-category-refactorer.ts
```diff
@@ -89,117 +82,97 @@ export abstract class RollbackCategoryRefactorer extends CategoryRefactorer {
...
-    if (resourcesToRestore.length === 0) {
-      return [this.buildDeleteHoldingStackOp(holdingStackName)];
+    this.debug(`Fetching template of holding stack: ${holdingStackName}`);
+    const holdingStackTemplate = await this.gen2Branch.fetchTemplate(holdingStackName);
+    const resources = this.filterResourcesByType(holdingStackTemplate);
+    this.debug(`Found ${resources.size} resources to move from stack: ${holdingStackName}`);
+
+    const resourceMappings: ResourceMapping[] = [];
+    for (const logicalId of resources.keys()) {
+      this.debug(`Registering ${logicalId} to move from ${holdingStackName} to ${gen2StackName}`);
```
iliapolo, line 142: Throw an error if the logical id already exists in the Gen2 stack - we should not be able to get here if the Gen2 -> Gen1 move failed.

packages/amplify-cli/src/commands/gen2-migration/_operation.ts
```diff
@@ -1,3 +1,23 @@
+import { DiscoveredResource } from './generate/_infra/gen1-app';
+
+export const NO_OP_MESSAGE = 'No-op\n';
+
+/**
+ * Creates a no-op operation for a resource with nothing to migrate.
+ */
+export function buildNoopOperation(resource: DiscoveredResource): AmplifyMigrationOperation {
```
iliapolo: Unused - delete.

packages/amplify-cli/src/commands/gen2-migration/_plan.ts
```diff
@@ -60,23 +60,41 @@ export class Plan {
...
-      descriptions.push(...(await op.describe()));
+      const lines = await op.describe();
+      if (lines.length === 0) continue;
+      const label = op.resource ? `Resource: ${op.resource.category}/${op.resource.resourceName} (${op.resource.service})` : 'Project';
+      if (!grouped.has(label)) grouped.set(label, []);
+      grouped.get(label)!.push(...lines);
     }
 
-    if (descriptions.length > 0) {
+    let hasRealImplications = false;
```
iliapolo: not needed - we don't create NO_OP operations

packages/amplify-cli/src/commands/gen2-migration/_validations.ts
```diff
@@ -97,7 +97,6 @@ export class AmplifyGen2MigrationValidations {
     const statefulRemoves: Array<{ category: string; resourceType: string; physicalId: string }> = [];
     for (const change of changeSet.Changes) {
       if (change.Type === 'Resource' && change.ResourceChange?.Action === 'Remove' && change.ResourceChange?.ResourceType) {
-        // Skip deployment bucket only when explicitly requested (e.g., during decommission)
```
iliapolo, line 100: bring this back

9pace (approved):

PR #14736, opened 2026-04-01
fix(gen2-migration): update category initializer to take specific auth modes in api init

sanjanaravikumar-az:
Refactors GraphQL API auth mode initialization in `category-initializer.ts` to build `authTypesConfig` in the order specified by `migration-config.json`'s authModes array. Previously, auth modes were added in a hardcoded order with a separate `addApiWithBlankSchema` fallback for the no-auth case. Now the first entry in `authModes` becomes the default auth mode (since `addApi` uses the first key as default), with a fallback to API key if no modes are configured. Removes the unused `addApiWithBlankSchema` import.

### DIFF
--- a/packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
+++ b/packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
@@ -201,17 +201,22 @@ export class CategoryInitializer {
         IAM: 'IAM',
         API_KEY: 'API key',
         COGNITO_USER_POOLS: 'Amazon Cognito User Pool',
+        AMAZON_COGNITO_USER_POOLS: 'Amazon Cognito User Pool',
       };
 
       const authTypesConfig: Record<string, Record<string, unknown>> = {};
       for (const mode of apiConfig.authModes ?? []) {
         const mapped = authModeMap[mode];
-        if (mapped) authTypesConfig[mapped] = {};
+        if (!mapped) {
+          throw new Error(
+            `Unsupported auth mode '${mode}' in migration-config.json. Supported modes: ${Object.keys(authModeMap).join(', ')}`,
+          );
+        }
+        authTypesConfig[mapped] = {};
       }
 
-      // Fallback: ensure at least API key is present
       if (Object.keys(authTypesConfig).length === 0) {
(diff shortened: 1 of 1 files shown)


packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
```diff
@@ -196,23 +195,28 @@ export class CategoryInitializer {
...
+      const authModeMap: Record<string, string> = {
+        IAM: 'IAM',
+        API_KEY: 'API key',
+        COGNITO_USER_POOLS: 'Amazon Cognito User Pool',
+      };
+
+      const authTypesConfig: Record<string, Record<string, unknown>> = {};
+      for (const mode of apiConfig.authModes ?? []) {
+        const mapped = authModeMap[mode];
+        if (mapped) authTypesConfig[mapped] = {};
```
iliapolo: If `mapped` is undefined it means the migration config is invalid - right? we should throw an error.

packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
```diff
@@ -196,23 +195,28 @@ export class CategoryInitializer {
...
+      }
+
+      // Fallback: ensure at least API key is present
+      if (Object.keys(authTypesConfig).length === 0) {
```
iliapolo, line 218: Why have this fallback? do we have an app that doesn't specify `authModes`? Prefer to throw an error and be more strict rather than allowing unexpected scenarios.

iliapolo (changes requested):

packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
sanjanaravikumar-az: true, i'll update

packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
sanjanaravikumar-az: yes, every migration-config has an auth mode specified, we should throw an error

iliapolo (approved):

Issue #14739, opened 2026-04-02
(gen2-migration) `generate` fails when Rest API names have hyphens in them


dgandhi62:
### Amplify CLI Version

NA

### What operating system are you using?

NA

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

NA

### Describe the bug

`generate` fails when Rest API names have hyphens in them. This does not seem like something we should be failing over. 

`api/cli-inputs`
```typescript
{
  "version": 1,
  "paths": {
    "/auth-test": {
      "name": "/auth-test",
      "lambdaFunction": "app2f4d22a3b",
      "permissions": {
        "setting": "private",
        "auth": [
          "read"
        ]
      }
    }
  }
}
```

`cfn template`
```typescript
"Resource": [
                {
                  "Fn::Join": [
                    "",
                    [
                      "arn:aws:execute-api:",
                      {
                        "Ref": "AWS::Region"
                      },
                      ":",
                      {
                        "Ref": "AWS::AccountId"
                      },
                      ":",
                      {
                        "Ref": "api3c3877d7"
                      },
                      "/",
                      {
                        "Fn::If": [
                          "ShouldNotCreateEnvResources",
                          "Prod",
                          {
                            "Ref": "env"
                          }
                        ]
                      },
                      "/GET/auth-test/*"
                    ]
                  ]
                },
```

Failure message 
```typescript

🛑 Failed: SyntaxError: ',' expected. (46:11)
  44 | });
  45 | backend.auth.resources.authenticatedUserIamRole.attachInlinePolicy(gen1api3c3877d7Policy);
> 46 | const auth-test = api3c3877d7Api.root.addResource("auth-test", {
     |           ^
  47 |     defaultMethodOptions: {
  48 |         authorizationType: AuthorizationType.IAM
  49 |     },

🛑 Not Implemented
```

### Expected behavior

Should not fail over a name assumption

### Reproduction steps

Add rest api with hypened name

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Before submitting, please confirm:

- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [x] I have removed any sensitive information from my code snippets and submission.

PR #14740, opened 2026-04-02
fix(gen2-migration): sanitize generated variable names in rest api

dgandhi62:
Solves https://github.com/aws-amplify/amplify-cli/issues/14739

#### Description of changes

The REST API renderer in the gen2 migration tool uses path names and API names directly as JavaScript variable names in the generated `backend.ts`. When these names contain non-alphanumeric characters (e.g. `auth-test`), the generated code produces invalid identifiers like `const auth-test = ...`, which fails to compile.

This strips non-alphanumeric characters from path segments and API names before using them as variable names, matching the sanitization pattern already used by the policy renderers in the same file.

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/amplify/rest-api/rest-api.renderer.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/amplify/rest-api/rest-api.renderer.ts
@@ -67,10 +67,11 @@ export class RestApiRenderer {
    */
   public renderApi(restApi: RestApiDefinition): ts.Statement[] {
     const statements: ts.Statement[] = [];
-    const stackVarName = `${restApi.apiName}Stack`;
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/commands/gen2-migration/generate/amplify/rest-api/rest-api.renderer.ts
iliapolo, line 1: Missing unit test.

iliapolo (changes requested):

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/rest-api/rest-api.renderer.ts
dgandhi62: Added

iliapolo (approved):

PR #14743, opened 2026-04-02
feat(gen2-migration): add apiAccess and secrets support for Lambda functions

sanjanaravikumar-az:
Adds support for Lambda functions that need API access (GraphQL endpoint env vars, appsync:GraphQL IAM permission) and secrets (SSM parameter access) in the migration e2e system. This enables the product-catalog app's `lowstockproducts` function to call the AppSync API and read SSM secrets at runtime.

### Problem

The `lowstockproducts` Lambda in product-catalog makes IAM-signed GraphQL queries and reads SSM secrets, but the e2e system had no way to wire up the required permissions, env vars, or SSM access. The function was deployed without API endpoint env vars, without `appsync:GraphQL` IAM permission, without SSM access, and without AppSync admin role authorization — causing all `checkLowStock` tests to fail.

### Approach

The core challenge is that `addFunction` with `additionalPermissions` sets `dependsOn` to auth (not API), so the deployed Lambda gets `AUTH_*` env vars but not `API_*` env vars. The gen2-migration generate tool, however, needs API `dependsOn` to emit `grantQuery`/`grantMutation`. This creates a timing conflict: push needs auth `dependsOn`, but the committed state needs API `dependsOn`.

The solution splits the work across three phases:

1. **During category init** (before push): `addFunction` with `additionalPermissions` wires auth `dependsOn` + `patchFunctionAppsyncPermission` adds the `appsync:GraphQL` IAM policy to the CFN template
2. **After push** (before gen1 test): `aws lambda update-function-configuration` merges API env vars into the deployed Lambda by reading the GraphQL endpoint from `amplify-meta.json`
3. **After gen1 test** (before git commit): `patchRegularFunctionApiAccess` replaces auth `dependsOn` with API `dependsOn` in backend-config/amplify-meta/function-parameters/CFN template so gen2-migration generate sees the right dependency graph

### Changes

- **category-initializer.ts**: Two-pass function init (without/with API access), secrets support via `secretsConfig`, `patchFunctionAppsyncPermission` for IAM policy, `patchRegularFunctionApiAccess` for post-push config patching, `discoverS3TriggerFunctionName` helper
- **types/index.ts**: Added `secrets` and `apiAccess` fields to `LambdaFunction` interface
- **cli.ts**: Post-push Lambda env var update via AWS CLI, `resolveCustomRolesJson` for AppSync admin role authorization, `patchRegularFunctionApiAccess` call after push, `npm install` before `ampx sandbox`
- **gen2-migration-executor.ts**: Added `--skip-validations` to `lock` and `generate` (local drift from post-push patching)
- **migration-config.json** (product-catalog): Added `secrets` and `apiAccess` to `lowstockproducts`
- **configure.sh** (product-catalog): Dynamic API name resolution instead of hardcoded `productcatalog`
- **lowstockproducts.js** (product-catalog): Dynamic GraphQL endpoint env var discovery instead of hardcoded name


packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
```diff
@@ -98,6 +119,11 @@ export class CategoryInitializer {
       await this.initializeApiCategory(appPath, categories.api, categories.function, result);
     }
 
+    // Initialize regular functions that need API access (after API exists)
+    if (categories.function && categories.api) {
```
iliapolo: This will result in calling `initializeRegularFunctions` twice. Instead, `initializeRegularFunctions` should handle configuring access by looking at the function definition (does it have `apiAccess`).

packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
```diff
@@ -608,6 +666,172 @@ export class CategoryInitializer {
...
+    }
+    return entries.find((name) => name.startsWith(triggerNamePrefix)) ?? null;
+  }
+
+  /**
+   * Patch a function's CFN template to add appsync:GraphQL permission.
+   * Required for functions that make IAM-signed callbacks to AppSync.
+   * Note: API env vars are wired by patchRegularFunctionApiAccess called from cli.ts.
+   */
+  private patchFunctionAppsyncPermission(appPath: string, functionName: string): void {
```
iliapolo: The point of the category initializer is to use the CLI, not create the files ourselves because it would mean duplicating what amplify is already doing and introducing more bugs. Why can't this go through the normal CLI invocation mechanism?

packages/amplify-gen2-migration-e2e-system/src/core/category-initializer.ts
```diff
@@ -608,6 +666,172 @@ export class CategoryInitializer {
...
+
+    fs.writeFileSync(cfnPath, JSON.stringify(cfn, null, 2) + '\n', 'utf-8');
+    this.logger.debug(`Patched ${functionName} CFN template with appsync:GraphQL permission`);
+  }
+
+  /**
+   * Patch backend-config.json and CFN template for a function with API access.
+   * Replaces auth dependsOn with API dependsOn and adds API env vars.
+   */
+  public patchRegularFunctionApiAccess(appPath: string, functionName: string, apiName: string, operations: string[]): void {
```
iliapolo: same

packages/amplify-gen2-migration-e2e-system/src/cli.ts
```diff
@@ -544,16 +577,99 @@ async function initializeAppFromCLI(params: InitializeAppFromCLIParams): Promise
...
     // Push the initialized app to AWS
     logger.info(`Pushing ${deploymentName} to AWS...`);
     await amplifyPush(targetAppPath);
     logger.info(`Successfully pushed ${deploymentName} to AWS`);
 
+    // Update deployed Lambda env vars for functions with API access.
+    // addFunction with additionalPermissions sets dependsOn to auth, so the deployed
+    // Lambda gets AUTH_* env vars but not API_* env vars. We read the GraphQL endpoint
+    // from amplify-meta.json and update the Lambda configuration directly.
+    if (config.categories?.function && config.categories?.api) {
```
iliapolo: Why is this needed and cannot be configured via the CLI?

packages/amplify-gen2-migration-e2e-system/src/core/gen2-migration-executor.ts
```diff
@@ -80,8 +80,9 @@ export class Gen2MigrationExecutor {
    * Enables deletion protection on DynamoDB tables, sets a deny-all stack policy,
    * and adds GEN2_MIGRATION_ENVIRONMENT_NAME env var to the Amplify app.
    */
-  public async lock(appPath: string): Promise<void> {
-    await this.executeStep('lock', appPath);
+  public async lock(appPath: string, skipValidations = false): Promise<void> {
```
iliapolo: Where are you actually passing `skipValidations = true`?

iliapolo (changes requested):

PR #14744, opened 2026-04-02
feat(gen2-migration): allow generate to skip non-JS Lambda functions

### LINKED ISSUE #14723: (gen2-migration) `generate` should allow skipping over non JS functions

_No response_

_No response_

Being able to run `npx amplify gen2-migration generate` even if I have Non JS functions in my Gen1 app. These functions can later be hand coded by the user.

Instead of throwing an error, surface the unsupported functions as part of the assessment validation - this would allow `--skip-validations` to skip over this as well.

None

### Additional context

_No response_

iliapolo:
#### Description of changes

Resolves #14723 — `gen2-migration generate` now allows skipping
non-JS Lambda functions instead of throwing an error.

Non-JS runtimes (Python, Java, .NET, etc.) are detected during
assessment by reading the function's local CloudFormation template.
The `FunctionAssessor` marks these resources as unsupported for
`generate`, which causes the assessment validation to fail with a
suggestion to use `--skip-validations`. When that flag is set, the
generate orchestrator skips unsupported resources entirely — no
generator is instantiated for them.

Key design decisions:

- Runtime detection uses the local CFN template, not AWS API calls,
  so it works in the assessor without async.
- Non-JS runtime is a resource-level unsupported status, not a
  feature-level one — the entire function can't be migrated.
- `Assessment.of(resource, step)` is a new general-purpose method
  for per-resource support lookup, used by the generate orchestrator
  to skip any unsupported resource before the switch statement.

#### Issue #, if available

Closes #14723

- 6 new unit tests in `function.assessor.test.ts` covering Python,
  Java, dotnet runtimes, Node.js as supported, no feature entry for
  non-JS, and missing template as an error.
- 1 new unit test in `assess.test.ts` verifying end-to-end
  assessment failure for non-JS functions.
- Updated existing test mocks to throw on missing files (matching
  real `gen1App.json()` behavior with `throwIfNotExist: true`).
- Full gen2-migration test suite passes (431 tests, 55 suites).

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/assess/function/function.assessor.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/assess/function/function.assessor.ts
@@ -4,7 +4,7 @@ import { Gen1App, DiscoveredResource, KNOWN_FEATURES } from '../../generate/_inf
 
 /**
  * Assesses migration readiness for a single Lambda function resource.
- * Detects custom-policies.json usage.
+ * Detects non-JS runtimes and custom-policies.json usage.
  */
 export class FunctionAssessor implements Assessor {
   public constructor(private readonly gen1App: Gen1App, private readonly resource: DiscoveredResource) {}
@@ -13,9 +13,13 @@ export class FunctionAssessor implements Assessor {
    * Records resource-level and feature-level support for this function.
    */
   public record(assessment: Assessment): void {
+    const templatePath = `function/${this.resource.resourceName}/${this.resource.resourceName}-cloudformation-template.json`;
+    const template = this.gen1App.json(templatePath);
+    const runtime = template.Resources.LambdaFunction.Properties.Runtime;
+
     assessment.recordResource({
       resource: this.resource,
-      generate: supported(),
+      generate: this.isNonJsRuntime(runtime) ? unsupported(`uses non-JS runtime '${runtime}'`) : supported(),
       refactor: supported(),
     });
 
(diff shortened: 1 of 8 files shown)


dgandhi62:
Does this flag skip unsupported resources which are non-lambda related too?

packages/amplify-cli/src/commands/gen2-migration/assess/function/function.assessor.ts
```diff
@@ -13,9 +13,13 @@ export class FunctionAssessor implements Assessor {
...
    */
   public record(assessment: Assessment): void {
+    const templatePath = `function/${this.resource.resourceName}/${this.resource.resourceName}-cloudformation-template.json`;
+    const template = this.gen1App.json(templatePath);
+    const runtime: string | undefined = template?.Resources?.LambdaFunction?.Properties?.Runtime;
+
     assessment.recordResource({
       resource: this.resource,
-      generate: supported(),
+      generate: this.isNonJsRuntime(runtime) ? unsupported(`uses non-JS runtime '${runtime}'`) : supported(),
```
dgandhi62, line 22: What happens if `runtime` is `undefined`?

dgandhi62 (commented):
Clarifying Questions

packages/amplify-cli/src/commands/gen2-migration/assess/function/function.assessor.ts
iliapolo: You're right. It can never be `undefined` - removed that case. It will fail somewhere else if its undefined.

dgandhi62 (approved):

Issue #14751, opened 2026-04-05
(gen2-migration) `refactor` with `--rollback` is broken


iliapolo:
### How did you install the Amplify CLI?

Local

### Amplify CLI Version

Local

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

Running `npx gen2-migration refactor --to <stack-name> --rollback` results in the following error:

```console
Resource '${logicalId}' already exists in Gen2 stack '${gen2StackName}' — the Gen2 → Gen1 move should have removed it
```

### Expected behavior

Not to fail

### Reproduction steps

1. Refactor any app.
2. Run with `--rollback`

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Additional information

This is a bug in the planning phase, where it prematurely inspects the target stack template:

https://github.com/aws-amplify/amplify-cli/blob/64edeeb975da01359039b633bf8f3f6ce602baf6/packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/rollback-category-refactorer.ts#L135-L147

During planning, the Gen2 -> Gen1 hasn't happened yet, so of course the resource will be there. This validation should happen only during execution. 

The bug was introduced in this [PR](https://github.com/aws-amplify/amplify-cli/pull/14698). The offending validation was added only after the manual live testing phase - thinking its harmless. It wasn't caught in the snapshot tests because they don't exercise rollback scenarios. 

### Before submitting, please confirm:

- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [x] I have removed any sensitive information from my code snippets and submission.

Issue #14750, opened 2026-04-05
(gen2-migration) `refactor` command deletes user pool triggers


iliapolo:
### How did you install the Amplify CLI?

Local

### Amplify CLI Version

Local

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

Running `refactor` will cause the Gen1 app to loose its auth triggers, breaking functionality.

### Expected behavior

Auth triggers are preserved

### Reproduction steps

1. Run the entire migration on the fitness-tracker app.
2. Try signing up with an invalid email.
3. Observe preSignup trigger is not triggered, and the user is allowed to singup.

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Additional information

**Initial investigation:**

- In Gen1 the auth triggers is defined as a [custom resource in a separate stack](https://github.com/aws-amplify/amplify-cli/blob/gen2-migration/amplify-migration-apps/fitness-tracker/_snapshot.pre.generate/amplify/backend/auth/fitnesstracker33f5545533f55455/build/auth-trigger-cloudformation-template.json). 
- In Gen2 it is defined using the standard [`LambdaConfig`](https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-cognito-userpool.html#cfn-cognito-userpool-lambdaconfig) property on the user pool resource.
- The `refactor` command performs an `UpdateStackCommand` on the Gen1 auth stack. It seems like even though the update results in an empty change-set, it still triggers a user pool update command. Since the `LambdaConfig` property is not present in the updated template (since its defined in a different stack), the triggers are nullified.

**More context:**

- The Gen2 deployment should restore the trigger because it will redeploy the user pool with Gen2 props.
- A possible fix is to use `ExecuteChangeSet` instead of `UpdateStackCommand`.

### Before submitting, please confirm:

- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [x] I have removed any sensitive information from my code snippets and submission.

iliapolo:
This was the problem:

https://github.com/aws-amplify/amplify-cli/blob/64edeeb975da01359039b633bf8f3f6ce602baf6/packages/amplify-cli/src/commands/gen2-migration/refactor/cfn.ts#L100

Wiping out tags triggers an `UPDATE` event on each resource in the template. Then, a combination of two things causes the triggers to be deleted:

[...]

Issue #14560, opened 2026-01-28
(gen2-migration) `lock` command sometimes doesn't enable deletion protection on DynamoDB model tables


iliapolo:
### How did you install the Amplify CLI?

npm install --no-save @aws-amplify/cli-internal-gen2-migration-experimental-alpha

### Amplify CLI Version

0.6.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

When running `gen2-migration lock`, the command normally prints out that it had enabled deletion protection for the DynamoDB model tables:

```console
[2026-01-28T18:52:30.811Z] [lock] [backendonly/main] → Executing
[2026-01-28T18:52:30.983Z] [lock] [backendonly/main] • Environment 'main' has been marked for migration
[2026-01-28T18:52:31.444Z] [lock] [backendonly/main] • Enabled deletion protection on table 'Project-lroca7kqyredxom2szijcpaw2m-main'
[2026-01-28T18:52:31.483Z] [lock] [backendonly/main] • Enabled deletion protection on table 'Todo-lroca7kqyredxom2szijcpaw2m-main'
[2026-01-28T18:52:31.520Z] [lock] [backendonly/main] • Enabled deletion protection on table 'randomItem-lroca7kqyredxom2szijcpaw2m-main'
[2026-01-28T18:52:31.840Z] [lock] [backendonly/main] • Root stack 'amplify-backendonly-main-9dc7d' has been locked
[2026-01-28T18:52:31.840Z] [lock] [backendonly/main] → Execution complete
```

Sometimes, these printouts don't exist and indeed the tables don't have deletion protection enabled. It looks like there is some scenario where the command is unable to discover the tables it needs.

### Expected behavior

Command should always enable deletion protection on the tables. If no tables are found but are expected, the command should throw an error and terminate the migration process.

### Reproduction steps

Unclear yet. 

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Before submitting, please confirm:

- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [x] I have removed any sensitive information from my code snippets and submission.

sarayev:
[...]

PR #14753, opened 2026-04-05
feat(gen2-migration): full e2e execution of all apps

iliapolo:
## Description of changes

Rewrites the gen2-migration e2e system and restructures all migration apps. The goal is to make the e2e workflow simpler to understand, debug, and extend.

---

#### 1. Replace category initializer with snapshot restore

**Before:**

Each migration app had a `migration-config.json` — a custom DSL that described the app's categories in a JSON schema the e2e system invented:

```json
{
  "categories": {
    "auth": { "signInMethods": ["phone"], "socialProviders": [] },
    "api": { "type": "GraphQL", "authModes": ["API_KEY"], "schema": "schema.graphql" },
    "storage": { "buckets": [...], "tables": [{ "name": "activity", "partitionKey": "id", "gsi": [...] }] },
    "function": { "functions": [{ "name": "fetchuseractivity", "runtime": "nodejs", "template": "hello-world" }] }
  }
}
```

The `CategoryInitializer` (644 lines) parsed this DSL and translated each entry into the corresponding `@aws-amplify/amplify-e2e-core` helper call (`addAuthWithEmail`, `addApi`, `addS3Storage`, `addDynamoDBWithGSIWithSettings`, `addFunction`, `addLambdaTriggerWithModels`, etc.). This required maintaining a mapping layer between the DSL and the e2e-core prompt-driven helpers — every new category feature or CLI prompt change required updating both the DSL schema and the translator. The DSL also couldn't express everything the CLI supports (auth triggers, custom roles, complex storage configurations), so apps that needed those features required workarounds.

**After:**

The e2e system restores the pre-generate snapshot (`_snapshot.pre.generate/amplify/`) directly into the `amplify/` directory after `amplify init`. This gives us the exact Gen1 project state — all categories, all resources, all configuration — without running any `amplify add` commands. The `App.configure()` method handles merging the init-generated metadata (CloudFormation provider info, environment config) with the snapshot's backend definitions.

This eliminates `CategoryInitializer`, `ConfigurationLoader`, the `types/index.ts` DSL definitions (227 lines), and all `migration-config.json` files across every app.

#### 2. Consolidate into a single `App` class

**Before:**

The e2e system was spread across 15+ files: `cli.ts` (657 lines of orchestration), `AmplifyInitializer`, `CategoryInitializer`, `ConfigurationLoader`, `AppSelector`, `CDKAtmosphereIntegration`, `Gen2MigrationExecutor`, `EnvironmentDetector`, `DirectoryManager`, `FileManager`, `AWSProfileManager`, `Logger`, `types/index.ts`, plus unit tests and e2e tests for each. Total: ~4,400 lines.

**After:**

Three files: `App` (owns the full lifecycle), `Git` (commit/checkout), `Logger`. Total: ~650 lines. The migration workflow in `cli.ts` reads as a flat sequence of method calls:

```typescript
await app.init();
await app.configure();
await app.push();
await app.testGen1();
await app.assess();
await app.lock();
await app.generate();
await app.deployGen2Sandbox();
await app.refactor(gen2StackName);
// ...
```

#### Atmosphere support removed

The `CDKAtmosphereIntegration` class and all atmosphere-related code paths are removed. We haven't determined whether atmosphere will actually be used in the migration workflow, and carrying its maintenance overhead — wiring, configuration surface, test coverage — isn't justified while that decision is still open. The code is in git history and can be restored straightforwardly if we decide to adopt it.

#### E2E test ordering

`testGen1` and `testGen2` now run immediately after `refactor`, before `gitCheckoutGen2`/`postRefactor`. This validates both environments right after the refactor completes.

---

#### 3. Replace custom test scripts with Jest suites

**Before:**

Each app had three files for testing: `test-utils.ts` (a factory that created test functions and an orchestrator), `gen1-test-script.ts`, and `gen2-test-script.ts`. The gen1/gen2 scripts were nearly identical — they differed only in which config file they imported. The test-utils files were 300-700 lines each, with a custom `TestRunner` class, manual console logging, and `process.exit(1)` on failure. A shared `_test-common/` directory provided `signup.ts` and `runner.ts` that all apps imported via relative paths.

**After:**

Each app has a `tests/` directory with standard Jest test files (`api.test.ts`, `storage.test.ts`, etc.) and its own `signup.ts` tailored to its auth config. The same test files run against both Gen1 and Gen2 — the `APP_CONFIG_PATH` env var controls which config is loaded. npm scripts `test:gen1` and `test:gen2` set this variable. The shared `_test-common/` directory is removed.

#### 4. Expose migration scripts as npm scripts

**Before:**

The e2e system dynamically imported each app's `post-generate.ts` and `post-refactor.ts` from the source app directory using `import()`. The scripts accepted an options object (`{ appPath, envName }`) and the e2e system had to know the function signature.

**After:**

Each app's `package.json` defines npm scripts (`post-generate`, `post-refactor`, `post-push`) that the e2e system runs via `npm run <script>`. Apps that don't need a step set it to `"true"`. The e2e system doesn't need to know anything about the scripts' internals.

#### 5. Restructure backend assets

Backend source files (GraphQL schemas, Lambda function code, configure scripts) moved from the app root into a `backend/` subdirectory. This separates backend assets from test files, migration scripts, and app source code.

---

#### Fix `lock` failing to find GraphQL API (#14560)

`fetchGraphQLApiId` matched APIs by name using the pattern `${appName}-${envName}`, which isn't guaranteed to match the actual AppSync API name. When it didn't match, the API was silently missed and DynamoDB model tables never got deletion protection enabled. The new `findGraphQLApiId` reads the API ID directly from `amplify-meta.json` (the source of truth) instead of scanning `ListGraphqlApis` with a name heuristic.

#### Fix `refactor` deleting user pool triggers (#14750)

The `update()` method passed `Tags: []` to `UpdateStackCommand`, which wiped out existing stack tags and triggered an UPDATE event on every resource in the template. For Cognito user pools, this was destructive: `LambdaConfig` (auth triggers) is defined in a separate stack and not present in the auth stack template, and Cognito's `UpdateUserPool` API doesn't support PATCH semantics — it replaces the entire configuration. The combination caused auth triggers to be silently deleted. The fix removes the `Tags: []` parameter. Additionally, stack updates now use `ExecuteChangeSet` instead of `UpdateStackCommand`, so a truly empty change set results in a no-op rather than a spurious update.

#### Fix `refactor --rollback` failing with false duplicate-resource error (#14751)

`refactor --rollback` failed with "Resource already exists in Gen2 stack" because `RollbackCategoryRefactorer.afterMove()` checked whether the target stack already contained the resource being moved. This check ran during planning, before any moves had executed — so the resource was naturally still present in the Gen2 stack. The fix removes the premature check from the rollback planner and moves the duplicate-resource guard into `Cfn.refactor()`, which runs at execution time when the stack state is current. The bug was introduced in PR #14698 and wasn't caught because snapshot tests don't exercise rollback scenarios.

#### Cursor restore on SIGINT

`SpinningLogger` now registers a SIGINT handler that stops the spinner before exiting. Previously, Ctrl-C during a spinner-active migration step left the terminal cursor hidden.

#### Output formatting

Minor improvements to `assess`, `lock`, and `refactor` CLI output (spacing, colors).

#### Issue #, if available

#14560, #14750, #14751

## Description of how you validated changes

Ran the full e2e migration workflow against all migration apps.

## Checklist

- [x] PR description included
- [x] `yarn test` passes

---


packages/amplify-gen2-migration-e2e-system/src/cli.ts
```diff
@@ -1,657 +1,127 @@
...
- * Copies _test-common and installs deps before executing.
- */
-async function runGen2TestScript(targetAppPath: string, migrationTargetPath: string, sourceAppsBasePath: string): Promise<void> {
-  const testScriptName = 'gen2-test-script.ts';
-
-  // Check if the gen2 test script exists
-  if (!fs.existsSync(path.join(targetAppPath, testScriptName))) {
-    logger.debug(`No ${testScriptName} found in ${targetAppPath}, skipping`);
+async function migrate(app: App): Promise<void> {
+  app.logger.info(`Staring migration`);
```
dgandhi62:
```suggestion
  app.logger.info(`Starting migration`);
```

packages/amplify-cli/src/commands/gen2-migration/lock.ts
```diff
@@ -223,7 +217,7 @@ export class AmplifyMigrationLockStep extends AmplifyMigrationStep {
 
   private async dynamoTableNames(): Promise<string[]> {
     if (!this._dynamoTableNames) {
-      const graphQLApiId = await this.fetchGraphQLApiId();
+      const graphQLApiId = await this.findGraphQLApiId();
       this._dynamoTableNames = await this.fetchGraphQLModelTables(graphQLApiId);
```
dgandhi62: `findGraphQLApiId()` can be undefined. Which means that `graphQLApiId` could be undefined. Is passing an undefined value in `fetchGraphQLModelTables` intended?

packages/amplify-cli/src/commands/gen2-migration/refactor/cfn.ts
```diff
@@ -230,25 +236,54 @@ export class Cfn {
...
+      await waitUntilChangeSetCreateComplete(
+        { client: this.client, maxWaitTime: 120 },
+        { StackName: stackName, ChangeSetName: changeSetName },
       );
-    } finally {
-      await this.client.send(new DeleteChangeSetCommand({ StackName: stackName, ChangeSetName: changeSetName }));
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    } catch (e: any) {
+      if (e.message?.includes(`The submitted information didn't contain changes`)) {
+        await this.client.send(new DeleteChangeSetCommand({ StackName: stackName, ChangeSetName: changeSetName }));
```
dgandhi62, line 246: The previous code seemed to always delete the changeset under the `finally` block. We are now only deleting it under a condition. This seems to be a behavioral shift (although I don't know how significant of one)

packages/amplify-gen2-migration-e2e-system/src/core/logger.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const entry: LogEntry = { timestamp: new Date(), level, message, error };
+    const formatted = this.formatMessage(entry);
+
+    console.log(formatted);
+
+    try {
+      fs.appendFileSync(this.logFilePath, formatted + '\n');
+    } catch {
+      // Avoid infinite recursion
```
dgandhi62: Empty catch block

dgandhi62 (changes requested):

packages/amplify-cli/src/commands/gen2-migration/lock.ts
iliapolo: The old code also did that but its definitely not intended. great catch.

packages/amplify-cli/src/commands/gen2-migration/refactor/cfn.ts
iliapolo: Yes. Previously the change-set was deleted because it was only used to create the report - now its actually being executed. We keep it there so the customer has access to what changed their stack (another benefit of using change sets).

packages/amplify-gen2-migration-e2e-system/src/core/logger.ts
iliapolo: removed

dgandhi62 (approved):

PR #14764, opened 2026-04-08
fix(gen2-migration): handle lock rollback

9pace:
#### Description of changes

Template drift detection creates changesets with `IncludeNestedStacks: true`, but currently discards any changeset that comes back as FAILED. After gen2-migration refactor moves resources between stacks, nested changesets fail EarlyValidation (`ResourceExistenceCheck`) or Template format error (duplicate exports from `Fn::Join`). CFN still populates the Changes array before these validation steps fail so the data is there, we were just throwing it away.

This PR reads changes from those FAILED changesets instead of discarding them. It also handles the race condition where the root changeset fails immediately but some nested changesets are still `CREATE_IN_PROGRESS` we poll those to terminal status before reading their changes. Stacks that genuinely can't be analyzed are tracked as `incompleteStacks` so callers know the result is partial.

On the lock side, `rollback()` now validates drift before reverting DeletionPolicy changes. The lock step intentionally adds `DeletionPolicy: Retain` to DynamoDB tables, so those changes (and the cascading IAM Policy re-evaluations CFN generates) are expected. The filter distinguishes these from real drift that would indicate the environment was modified outside the migration flow.

See ADR-005 for the full design rationale and 8 empirical findings.

#### Issue #, if available

#14570

- 30 unit tests (lock rollback drift validation, DeletionPolicy filter, cascading IAM, nested tree walk, edge cases)
- 7 E2E scenarios against a deployed Amplify app (FAILED changeset reading, filter correctness, negative tests with injected drift)
- TypeScript compiles clean

#### Checklist

- [x] PR description included
- [x] `yarn test` passes


packages/amplify-cli/src/commands/drift-detection/detect-template-drift.ts
```diff
@@ -238,30 +244,51 @@ async function analyzeChangeSet(
...
+      if (!changeSet.Changes || changeSet.Changes.length === 0) {
+        print.warn(`Recoverable failure but no Changes populated: ${changeSet.StatusReason}`);
+        return {
+          changes: [],
+          skipped: true,
+          skipReason: `Changeset failed with no usable changes: ${changeSet.StatusReason}`,
+        };
+      }
+      print.warn(`Nested changeset FAILED (recoverable): ${changeSet.StatusReason}`);
+    } else if (isRoot) {
```
sai-ray: Can the root changeset fail for a root-level reason rather than a nested changeset failure? If so, the `else if (isRoot)` branch would fall through, find zero changes, and return `{ changes: [], skipped: false }`, which would report "no drift" when we actually couldn't check. Should we verify the `StatusReason` mentions a nested changeset before falling through, and treat other failures as `skipped: true`?

packages/amplify-cli/src/commands/drift-detection/detect-template-drift.ts
```diff
@@ -291,13 +318,34 @@ async function analyzeChangeSet(
...
+        // but other nested changesets may still be CREATE_IN_PROGRESS.
+        let nestedChangeSet = await cfn.send(
           new DescribeChangeSetCommand({
             StackName: stackName,
             ChangeSetName: changeSetName,
           }),
         );
+        const terminalStatuses = ['CREATE_COMPLETE', 'FAILED'];
+        const NESTED_POLL_MAX_ATTEMPTS = 30;
+        const NESTED_POLL_INTERVAL_MS = 2000;
```
sai-ray: Curious, how long did this polling take in your testing? Just wondering if the sequential approach adds noticeable wait time with multiple nested stacks.

9pace: Maybe a couple of seconds max (from what I can remember, i didn't really measure it). But I don't think that it's necessarily sequential, includeNestedStacks will try to create multiple changesets at once (I assume its like BFS) and poll for completion just exists because I started to see that sometimes if i instantly checked for the nested stack changeset completion, they would be in progress.

packages/amplify-cli/src/commands/drift-detection/detect-template-drift.ts
9pace: Yes i assume this is possible and should be accounted for.

sai-ray (approved):

PR #14776, opened 2026-04-13
chore: add Gen 1 maintenance mode warning

Simone319:
## What

Adds a non-suppressible deprecation warning to the Amplify Gen 1 CLI that is printed on **every invocation**.

## Why

AWS Amplify Gen 1 enters **maintenance mode on May 1, 2026** and reaches **end of life on May 1, 2027**. During maintenance mode only critical bug fixes and security patches will be provided. Users must be proactively notified to migrate to Gen 2.

## Details

- Warning is printed to **stderr** (not stdout) so it does not interfere with JSON output, piped commands, or scripted automation that parses stdout.
- Output uses **ANSI yellow** (`\x1b[33m`) for visual emphasis, with a reset (`\x1b[0m`) at the end.
- Runs **unconditionally** at the top of `run()` (before `deleteOldVersion()`). It cannot be suppressed or silenced.
- Includes a direct link to the Gen 2 migration guide.

### Warning text

```
⚠️  WARNING: AWS Amplify Gen 1 CLI is in maintenance mode and will reach end of life on May 1, 2027.
During maintenance mode, only critical bug fixes and security patches will be provided.
Migrate to Amplify Gen 2: https://docs.amplify.aws/react/start/migrate-to-gen2/
```

## Testing

```bash
node packages/amplify-cli/bin/amplify version
# Warning appears on stderr before normal output
```

## Checklist

- [x] Change is scoped to a single file (`packages/amplify-cli/src/index.ts`)
- [x] No new dependencies
- [x] Warning uses stderr to avoid breaking automation
- [x] Migration link included

### DIFF
--- a/packages/amplify-cli/src/index.ts
+++ b/packages/amplify-cli/src/index.ts
@@ -63,6 +63,13 @@ const disableCDKDeprecationWarning = () => {
  * Command line entry point
  */
 export const run = async (startTime: number): Promise<void> => {
+  printer.blankLine();
+  printer.warn(
+    'AWS Amplify Gen 1 CLI is in maintenance mode and will reach end of life on May 1, 2027.\n' +
+      'During maintenance mode, only critical bug fixes and security patches will be provided.\n' +
+      'Migrate to Amplify Gen 2: https://docs.amplify.aws/react/start/migrate-to-gen2/\n',
+  );
+
   deleteOldVersion();
 
   //TODO: This is a temporary suppression for CDK deprecation warnings, which should be removed after the migration is complete
--- a/packages/amplify-cli-npm/index.ts
+++ b/packages/amplify-cli-npm/index.ts
@@ -16,4 +16,4 @@ export const install = async (): Promise<void> => {
   return binary.install();
 };
 
-// force version bump to 14.3.0
+// force version bump to 14.4.0


packages/amplify-cli/src/index.ts
```diff
@@ -63,6 +63,13 @@ const disableCDKDeprecationWarning = () => {
  * Command line entry point
  */
 export const run = async (startTime: number): Promise<void> => {
+  // Gen 1 CLI deprecation warning — prints to stderr to avoid interfering with JSON/stdout parsing
+  process.stderr.write(
+    '\n\x1b[33m⚠️  WARNING: AWS Amplify Gen 1 CLI is in maintenance mode and will reach end of life on May 1, 2027.\n' +
```
adrianjoshua-strutt: Could we use "chalk.yellow"? It seems like it is already part of the repo: https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-cli/package.json#L91

Simone319: Good call, will use the AmplifyPrinter which uses chalk.yellow.

svidgen (dismissed):

adrianjoshua-strutt (dismissed):

adrianjoshua-strutt (approved):

iliapolo (approved):

PR #14789, opened 2026-04-18
chore(gen2-migration): shared data tests

iliapolo:
### Description of changes

This PR adds shared-data integration tests for all gen2-migration test apps and extends the code generator with two new IAM policy features.

#### Shared-data and shared-auth tests

Each migration app now has two test files that validate stateful resource sharing between gen1 and gen2 configurations:

- `shared-auth.test.ts` — verifies that a user created in one config can sign in via the other (bidirectional). Each test signs in and out independently.
- `shared-data.test.ts` — verifies that data written via one config is readable via the other (bidirectional). Signs in once in `beforeAll` and swaps `configureAmplify` between tests without sign-out, reducing test runtime significantly.

New npm scripts: `test:shared-data`, `test:shared-auth`, `test:shared` (runs both). The `test:gen1`/`test:gen2` scripts exclude shared tests via `--testPathIgnorePatterns='shared'`.

The `configureAmplify` helper in each app's `signup.ts` was refactored to match the discussions app pattern: optional param with `APP_CONFIG_PATH` fallback, no module-level side effects.

#### Gen1 group path policies for REST APIs

`RestApiRenderer` now generates a second `attachInlinePolicy` call for each group-restricted API path, granting the same group role access to the equivalent gen1 API resource. This ensures Cognito group roles retain access to both gen2 and gen1 REST APIs during the migration transition period.

#### Gen1 IAM policies for geo resources

`GeoRenderer` now generates a `gen1AuthPolicy` in each geo `resource.ts` file. The policy grants access to the gen1 Location Service resource (map, place index, or geofence collection) using actions extracted from the Gen1 CloudFormation template. The policy attaches to the same roles the gen2 construct receives.

#### Post-refactor tag

`BackendGenerator` appends commented-out `Tags.of(backend.stack).add('gen2-migration/post-refactor', 'true')` lines to every generated `backend.ts`. Post-refactor scripts uncomment them, forcing a CloudFormation redeployment after migration. New `post-refactor.ts` scripts were added for apps that previously had no-op `"post-refactor": "true"` entries.

#### E2E system

`app.ts` updated to use `test:shared` (runs both auth and data tests) and to run from the gen1 branch pulling in the gen2 config file, rather than the reverse.

### Issue #, if available

N/A

### Description of how you validated changes

- All generate and refactor snapshot tests pass (`yarn test` in `packages/amplify-cli/`).
- Shared-data and shared-auth tests validated against live deployed stacks for discussions, fitness-tracker, mood-board, product-catalog, project-boards, and store-locator apps.
- Snapshots updated to reflect new codegen output.

### Checklist

- [x] PR description included
- [x] `yarn test` passes

---


amplify-migration-apps/store-locator/tests/shared-auth.test.ts
```diff
@@ -0,0 +1,43 @@
+/**
+ * Validates that the Cognito user pool is shared between gen1 and gen2 configurations.
+ * Tests both directions: gen1→gen2 and gen2→gen1.
+ *
+ * Note: Geo resources (search, geocode) are stateless and geofences are not supported
```
sai-ray:
```suggestion
 * Note: Geo resources (search, map) are stateless and geofences are not supported
```

sai-ray (approved):
lgtm

PR #14791, opened 2026-04-20
feat(gen2-migration): codebuild e2e workflow

iankhou:
### CodeBuild Infrastructure

Added the complete CodeBuild pipeline for running Gen2 migration E2E tests:

- `gen2_migration_pr_workflow.yml` — batch workflow that builds the CLI, publishes to Verdaccio, builds the linux pkg binary, then runs all migration apps in parallel
- `run_gen2_migration_e2e.yml` — per-app buildspec that loads the cached binary and runs a single migration test
- `upload_pkg_binaries_linux_only.yml` — uploads only the linux binary (the migration workflow doesn't need macOS/Windows)
- `cloud-gen2-migration.sh` — CI entry points for triggering the workflow
- `_runGen2MigrationE2E` in `shared-scripts.sh` — configures AWS credentials via a named profile, unsets credential env vars to avoid SDK conflicts, bootstraps CDK, configures git identity, and runs the migration test
- `_uploadPkgBinariesLinuxOnly` in `shared-scripts.sh` — caches only the linux binary to S3

### Resource Teardown

Added a `--teardown` CLI option that cleans up all deployed resources after test execution:

- Deletes Gen1 root CFN stacks matching the deployment name prefix via the CloudFormation SDK
- Deletes the Amplify console app via the Amplify SDK
- Deletes the Gen2 sandbox via `ampx sandbox delete --yes`
- Finds and deletes holding stacks (CFN stacks with `-holding` suffix created during refactor)

This uses direct SDK calls instead of `amplify delete` because the migration mutates the workspace and breaks `amplify delete`. The teardown runs in a `finally` block so it executes whether the test passes or fails. All migration app `package.json` files pass `${TEARDOWN:+--teardown}`, and the buildspec sets `TEARDOWN=1`.

### `TEST_ACCOUNT_ROLE`
If `TEST_ACCOUNT_ROLE` is set to an IAM Role ARN, the Gen 2 Migration CLI uses it to assume a role and get/refresh credentials for running the migration.

- May not be used with `--profile` option
- Must have administrator rights

### Git Init Fix

Changed `git init` to `git init --initial-branch=main` in `git.ts` so the default branch is always `main`. This fixes checkout failures in CodeBuild where the default branch name is `master`.

### Binary Packaging Fix (https://github.com/yao-pkg/pkg/issues/195, https://github.com/yao-pkg/pkg/issues/189)

A `yarn.lock` drift bumped `get-intrinsic` from `1.3.0` to `1.3.1`, which added `async-function`, `async-generator-function`, and `generator-function` as new dependencies. These packages ship with a `"module-sync"` export condition that points to `.mjs` entry points. Inside `pkg`'s snapshot filesystem, the ESM module resolution triggered by `module-sync` fails with `ERR_MODULE_NOT_FOUND`.

The fix patches the affected packages during `generatePkgCli` by stripping the `"module-sync"` line from their `package.json` via `sed`. Node then falls through to the `"default"` condition which points to a CJS file that `pkg` handles correctly.

CodeBuild runs in an account dedicated to E2E workflow runs. Uses existing CodeBuild job AmplifyCLI-E2E-Testing.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Tests are added


iankhou:
Codebuild: https://tiny.amazon.com/nhf82d4a/IsenLink

Not reported to this PR due to rebase.

scripts/cloud-gen2-migration.sh
iliapolo, line 1: Call it `scripts/cloud-e2e-gen2-migration.sh` and add `e2e` to the yaml files as well. Just for clarity.

.circleci/local_publish_helpers_codebuild.sh
```diff
@@ -64,6 +64,18 @@ function generatePkgCli {
   cp ../yarn.lock ./
   yarn workspaces focus --production
 
+  # Workaround for yao-pkg/pkg#195: pkg's snapshot filesystem can't resolve
+  # ESM imports from .mjs files referenced by the "module-sync" export condition.
+  # We strip just the "module-sync" line from each affected package.json so pkg
+  # falls back to the "default" (CJS) condition instead.
+  for esm_pkg in async-function async-generator-function generator-function; do
```
iliapolo: Lets use `resolutions` instead and add `README.md` here explaining why its needed and providing this patch as a fast workaround in case its necessary.

codebuild_specs/run_gen2_migration_e2e.yml
```diff
@@ -0,0 +1,25 @@
...
+      - echo "AMPLIFY_PATH=$AMPLIFY_PATH"
+      - echo "MIGRATION_APP=$MIGRATION_APP"
+      - source ./shared-scripts.sh && _runGen2MigrationE2E
+
+  post_build:
+    commands:
+      - echo "Migration E2E test completed for $MIGRATION_APP"
+
+# 45 minute timeout to accommodate ~30 min migration + buffer
+# This is set at the CodeBuild project level, but documented here for reference
```
iliapolo: Can remove to avoid being out of sync

packages/amplify-gen2-migration-e2e-system/src/cli.ts
```diff
@@ -24,7 +25,14 @@ async function main(): Promise<void> {
     })
     .option('profile', {
       type: 'string',
-      description: 'AWS profile to use',
+      description: 'AWS profile to use (mutually exclusive with --roleArn)',
+      string: true,
+    })
+    .option('roleArn', {
```
iliapolo: Remove this option. We won't use it during local dev and for CI you're already relying on the default behavior of picking up `process.env.TEST_ACCOUNT_ROLE`.

packages/amplify-gen2-migration-e2e-system/src/core/credentials.ts
```diff
@@ -0,0 +1,191 @@
...
+      process.env.AWS_ACCESS_KEY_ID = creds.AccessKeyId;
+      process.env.AWS_SECRET_ACCESS_KEY = creds.SecretAccessKey;
+      process.env.AWS_SESSION_TOKEN = creds.SessionToken;
+      delete process.env.AWS_PROFILE;
+
+      // Satellite: named profile file, required by `amplify init --profile`.
+      this.writeCredentialsFile(creds.AccessKeyId, creds.SecretAccessKey, creds.SessionToken);
+
+      this.logger.info('Credentials refreshed');
+    } catch (e) {
```
iliapolo: Should only protect the `assumeRoleAsOriginalIdentity` which may be affected by sporadic network issues. Other errors are not acceptable at this point and should fail the process.

packages/amplify-gen2-migration-e2e-system/src/core/app.ts
```diff
@@ -439,10 +462,243 @@ export class App {
...
+  // ============================================================
+  // Teardown
+  // ============================================================
+
+  /**
+   * Delete all deployed resources (Gen1 backend + Gen2 sandbox + holding stacks).
+   * Runs in a best-effort manner — logs errors but does not throw.
+   */
+  public async teardown(): Promise<void> {
+    this.logger.info('Starting teardown...');
```
iliapolo: call assume role here to get fresh credentials and use them to configure the SDK clients.

packages/amplify-gen2-migration-e2e-system/src/core/credentials.ts
```diff
@@ -0,0 +1,191 @@
...
+      if (!creds?.AccessKeyId || !creds?.SecretAccessKey || !creds?.SessionToken) {
+        throw new Error('STS AssumeRole returned incomplete credentials');
+      }
+
+      // Primary: env vars. Picked up by SDK default chain and inherited by
+      // subprocesses. Clear AWS_PROFILE so a stale profile can't shadow them.
+      process.env.AWS_ACCESS_KEY_ID = creds.AccessKeyId;
+      process.env.AWS_SECRET_ACCESS_KEY = creds.SecretAccessKey;
+      process.env.AWS_SESSION_TOKEN = creds.SessionToken;
+      delete process.env.AWS_PROFILE;
```
iliapolo: Should not be needed. Avoid mutating `process.env` - pass the credentials object you get from `assumeRoleAsOriginalIdentity` to whoever needs it.

packages/amplify-gen2-migration-e2e-system/src/core/credentials.ts
```diff
@@ -0,0 +1,191 @@
...
+      throw new Error('assumeRoleAsOriginalIdentity called outside role mode');
+    }
+    const saved = {
+      AWS_ACCESS_KEY_ID: process.env.AWS_ACCESS_KEY_ID,
+      AWS_SECRET_ACCESS_KEY: process.env.AWS_SECRET_ACCESS_KEY,
+      AWS_SESSION_TOKEN: process.env.AWS_SESSION_TOKEN,
+    };
+    delete process.env.AWS_ACCESS_KEY_ID;
+    delete process.env.AWS_SECRET_ACCESS_KEY;
+    delete process.env.AWS_SESSION_TOKEN;
```
iliapolo: Should not be needed

packages/amplify-gen2-migration-e2e-system/src/core/credentials.ts
```diff
@@ -0,0 +1,191 @@
...
+          RoleArn: this.source.roleArn,
+          RoleSessionName: `gen2-migration-e2e-${Date.now()}`,
+          DurationSeconds: SESSION_DURATION_SECONDS,
+        }),
+      );
+    } finally {
+      if (saved.AWS_ACCESS_KEY_ID !== undefined) process.env.AWS_ACCESS_KEY_ID = saved.AWS_ACCESS_KEY_ID;
+      if (saved.AWS_SECRET_ACCESS_KEY !== undefined) process.env.AWS_SECRET_ACCESS_KEY = saved.AWS_SECRET_ACCESS_KEY;
+      if (saved.AWS_SESSION_TOKEN !== undefined) process.env.AWS_SESSION_TOKEN = saved.AWS_SESSION_TOKEN;
+    }
```
iliapolo: Should not be needed.

packages/amplify-gen2-migration-e2e-system/src/core/app.ts
```diff
@@ -439,10 +462,243 @@ export class App {
...
 
+  // ============================================================
+  // Teardown
+  // ============================================================
+
+  /**
+   * Delete all deployed resources (Gen1 backend + Gen2 sandbox + holding stacks).
+   * Runs in a best-effort manner — logs errors but does not throw.
+   */
+  public async teardown(): Promise<void> {
```
iliapolo: Lets create a `Teardown` class that accepts the `App` class and has a `clean` method.

packages/amplify-gen2-migration-e2e-system/src/core/app.ts
```diff
@@ -439,10 +462,243 @@ export class App {
...
+            await this.emptyStackBuckets(cfnClient, stack.StackName);
+            await this.deleteStackWithRetainOnFailure(cfnClient, stack.StackName);
+          }
+        }
+      }
+    } catch (e) {
+      this.logger.info(`Holding stack cleanup failed: ${(e as Error).message} (continuing teardown)`);
+    }
+
+    this.logger.info('Teardown complete');
```
iliapolo: Should error out if some cleanup error occurred (cummulative) so we know to cleanup the resources.

packages/amplify-gen2-migration-e2e-system/src/core/app.ts
```diff
@@ -439,10 +462,243 @@ export class App {
...
+        this.logger.info(`Emptying bucket: ${bucket}`);
+        let continuationToken: string | undefined;
+        do {
+          const page = await s3.send(new ListObjectsV2Command({ Bucket: bucket, ContinuationToken: continuationToken }));
+          if (page.Contents && page.Contents.length > 0) {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            const objects = page.Contents.filter((o) => o.Key).map((o) => ({ Key: o.Key! }));
+            await s3.send(new DeleteObjectsCommand({ Bucket: bucket, Delete: { Objects: objects, Quiet: true } }));
+          }
+          continuationToken = page.NextContinuationToken;
```
iliapolo: Use paginators from the SDK.

packages/amplify-gen2-migration-e2e-system/src/core/app.ts
```diff
@@ -439,10 +462,243 @@ export class App {
...
+    }
+  }
+
+  /**
+   * Recursively delete any nested stacks of `stackName` that are in
+   * `DELETE_FAILED`. Each recursive call can itself retain problem leaf
+   * resources, so after this returns the parent's retry only needs to retain
+   * nested-stack logical IDs that remained stuck.
+   */
+  private async cleanupNestedFailedStacks(cfnClient: CloudFormationClient, stackName: string): Promise<void> {
```
iliapolo: Shouldn't need this. Just run a "discover-and-delete-with-retain" loop 5 times. So the logic is:

1. Empty all buckets.
2. List and delete all stacks.
3. Wait for DELETE success/failure
4. Run the discover-and-delete-with-retain

packages/amplify-gen2-migration-e2e-system/README.md
```diff
@@ -8,20 +8,32 @@ Automation system for end-to-end testing of the Amplify Gen1-to-Gen2 migration w
 # Set AMPLIFY_PATH to your development Amplify CLI (optional - falls back to monorepo build, then global install)
 export AMPLIFY_PATH=$(pwd)/.bin/amplify-dev
 
-# Migrate an app using a named AWS profile
+# Local dev: migrate an app using a named AWS profile
 npx tsx src/cli.ts --app project-boards --profile default
 
+# CI: migrate using an IAM role that gets re-assumed before each long step
+npx tsx src/cli.ts --app project-boards --roleArn arn:aws:iam::123456789012:role/TestRole
```
iliapolo: Remove

iliapolo (changes requested):

packages/amplify-gen2-migration-e2e-system/src/core/app.ts
```diff
@@ -61,13 +62,17 @@ export class App {
...
+  private readonly credentials: CredentialManager;
 
   public readonly logger: Logger;
   public readonly targetAppPath: string;
 
-  private readonly git: Git;
+  readonly git: Git;
+
+  /** Latest assumed-role credentials, or undefined in profile mode. */
+  private credentialEnv: AwsCredentials | undefined;
```
iliapolo: Shouldn't be needed at all. Use the profile everywhere so we have a single credential source.

iliapolo (changes requested):

iliapolo (approved):

PR #14794, opened 2026-04-21
chore(gen2-migration): turn on strict mode, and associated fixes

iankhou:
#### Description of changes

Turn strict mode back on in `packages/amplify-cli`.

https://tiny.amazon.com/7q8zqcl9/IsenLink

Created a new branch that contains Gen 2 Migration E2E CodeBuild running logic, and rebased from this commit to bring in the changes. Then, ran `yarn cloud-gen2-migration` to validate that the migration tool still works for all apps we are testing it on.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes
- [X] Tests are changed(https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md#tests)

### DIFF
--- a/packages/amplify-cli/src/commands/drift-detection/services/cloudformation-service.ts
+++ b/packages/amplify-cli/src/commands/drift-detection/services/cloudformation-service.ts
@@ -86,7 +86,7 @@ export class CloudFormationService {
 
       try {
         // Download the latest backend state from S3
-        currentCloudBackendZip = await downloadZip(s3, tempDir, S3BackendZipFileName, '');
+        currentCloudBackendZip = await downloadZip(s3, tempDir, S3BackendZipFileName, context.amplify.getEnvInfo().envName);
       } catch (err: any) {
         if (err?.name === 'NoSuchBucket') {
           // Environment not deployed yet, nothing to sync
--- a/packages/amplify-cli/src/commands/drift-detection/services/cloudformation-service.ts
+++ b/packages/amplify-cli/src/commands/drift-detection/services/cloudformation-service.ts
@@ -86,7 +86,7 @@ export class CloudFormationService {
 
       try {
         // Download the latest backend state from S3
-        currentCloudBackendZip = await downloadZip(s3, tempDir, S3BackendZipFileName, undefined);
+        currentCloudBackendZip = await downloadZip(s3, tempDir, S3BackendZipFileName, '');
       } catch (err: any) {
         if (err?.name === 'NoSuchBucket') {
           // Environment not deployed yet, nothing to sync
--- a/packages/amplify-cli/src/commands/gen2-migration/_infra/aws-clients.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/_infra/aws-clients.ts
@@ -3,7 +3,7 @@ import { AppSyncClient } from '@aws-sdk/client-appsync';
 import { CloudFormationClient } from '@aws-sdk/client-cloudformation';
 import { CognitoIdentityProviderClient } from '@aws-sdk/client-cognito-identity-provider';
 import { CognitoIdentityClient } from '@aws-sdk/client-cognito-identity';
-import { S3Client } from '@aws-sdk/client-s3';
+import { S3Client, S3ClientConfig } from '@aws-sdk/client-s3';
 import { LambdaClient } from '@aws-sdk/client-lambda';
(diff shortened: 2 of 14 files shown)


packages/amplify-cli/src/commands/drift-detection/services/cloudformation-service.ts
```diff
@@ -86,7 +86,7 @@ export class CloudFormationService {
 
       try {
         // Download the latest backend state from S3
-        currentCloudBackendZip = await downloadZip(s3, tempDir, S3BackendZipFileName, undefined);
```
iliapolo, line 89: This seems strange...apparently we were supposed to pass the environment name? pretty sure we can just do that.

packages/amplify-cli/src/commands/gen2-migration/_infra/aws-clients.ts
```diff
@@ -52,7 +52,7 @@ export class AwsClients {
     this.cloudFormation = new CloudFormationClient(config);
     this.cognitoIdentityProvider = new CognitoIdentityProviderClient(config);
     this.cognitoIdentity = new CognitoIdentityClient(config);
-    this.s3 = new S3Client(config);
+    this.s3 = new S3Client(config as S3ClientConfig);
```
iliapolo, line 55: How come just this client needs this and not the others?

packages/amplify-cli/src/commands/gen2-migration/_infra/validations.ts
```diff
@@ -29,7 +29,7 @@ export class AmplifyGen2MigrationValidations {
     const result = await new AmplifyDriftDetector(this.context, this.logger).detect();
     if (result.code !== 0) {
       throw new AmplifyError('MigrationError', {
-        message: result.report.trim() ?? 'Drift detected',
+        message: result.report!.trim() ?? 'Drift detected',
```
iliapolo: Why not:


```suggestion
        message: result.report?.trim() ?? 'Drift detected',
```

We are already accepting that a default value should be used.

iliapolo (changes requested):

packages/amplify-cli/src/commands/gen2-migration/_infra/aws-clients.ts
iankhou: S3Client asks for `S3ClientConfig`, which extends S3-only mixins (`S3InputConfig`, `EventStreamSerdeInputConfig`, `FlexibleChecksumsInputConfig`, plus a stricter requestHandler shape) that aren't part of AmplifyClientConfig, so TS rejects the shared config without a cast. At runtime every v3 client accepts the same base shape we're passing (credentials, customUserAgent, requestHandler).

https://github.com/aws/aws-sdk-js-v3/blob/main/clients/client-s3/src/S3Client.ts#L844

packages/amplify-cli/src/commands/gen2-migration/_infra/validations.ts
iankhou: Missed this; thanks, will change.

packages/amplify-cli/src/commands/drift-detection/services/cloudformation-service.ts
iankhou: It was probably written with `undefined` before because there is existing code that does it:

https://github.com/aws-amplify/amplify-cli/blob/e366e6f68ca9be0a83f44236fc8e54bcc632805b/packages/amplify-provider-awscloudformation/src/initialize-env.ts#L31

However, it expects a string (an environment name). Passing it is only effective for [logging](https://github.com/aws-amplify/amplify-cli/blob/e366e6f68ca9be0a83f44236fc8e54bcc632805b/packages/amplify-provider-awscloudformation/src/zip-util.ts#L22) purposes. Under the hood, it's used in [`attachBucketToParams`](https://github.com/aws-amplify/amplify-cli/blob/e366e6f68ca9be0a83f44236fc8e54bcc632805b/packages/amplify-provider-awscloudformation/src/aws-utils/aws-s3.ts#L117), but when `envName` is falsy, that function gets the envName using `getEnvInfo`, anyway.

I think either way it will not make much of a difference, but passing the envName won't do any harm or cost us anything, since we have it available at the call site through `context`.

iliapolo (approved):

PR #14826, opened 2026-04-27
chore: restrict Gen 1 new app creation to existing customers

Simone319:
## Summary

Restricts `amplify init` so only existing Gen 1 customers can create new Gen 1 apps. This is part of the Gen 1 maintenance mode.

### Definition of "existing customer"
An AWS account is considered an existing Gen 1 customer if it has at least one Amplify app with backend environments in the same region.

### Behavior
| Scenario | Result |
|----------|--------|
| `--quickstart` | Unconditionally blocked |
| New customer (no apps with backend envs) | Blocked with deprecation message |
| Existing customer (has apps with backend envs) | Allowed |
| `--appId` provided (re-init) | Restriction check skipped |

### Files changed
- `packages/amplify-cli/src/commands/init.ts` — blocks `--quickstart` unconditionally
- `packages/amplify-provider-awscloudformation/src/gen1-new-customer-restriction.ts` — restriction logic (ListApps + ListBackendEnvironments)
- `packages/amplify-provider-awscloudformation/src/initializer.ts` — calls restriction before CreateApp
- `packages/amplify-cli/src/__tests__/commands/init.test.ts` — unit tests for quickstart block
- `packages/amplify-provider-awscloudformation/src/__tests__/gen1-new-customer-restriction.test.ts` — unit tests for restriction logic
- `packages/amplify-provider-awscloudformation/src/__tests__/initializer.test.ts` — unit tests for initializer integration
- `packages/amplify-e2e-tests/src/__tests__/init-gen1-restriction.test.ts` — E2E test

### Testing
- 13 unit tests pass (8 restriction + 2 init command + 3 initializer)
- Manually tested: us-west-2 (8 apps, 6 with backend envs) = ALLOWED; eu-central-1 (3 apps, 0 backend envs) = BLOCKED


packages/amplify-provider-awscloudformation/src/initializer.ts
```diff
@@ -62,6 +64,15 @@ export const run = async (context: $TSContext): Promise<void> => {
 
     await configurePermissionsBoundaryForInit(context);
 
+    // Block new Gen 1 app creation for non-existing customers
+    const hasAppId = context.exeInfo?.inputParams?.amplify?.appId;
+    if (!hasAppId) {
+      const amplifyClient = await getConfiguredAmplifyClient(context, awsConfigInfo);
```
adrianjoshua-strutt, line 70: Is there any potential realistic scenario where "getConfiguredAmplifyClient()" returns null and we silently skip the validation?

packages/amplify-provider-awscloudformation/src/gen1-new-customer-restriction.ts
```diff
@@ -0,0 +1,66 @@
...
+        nextToken,
+        maxResults: 25,
+      }),
+    );
+
+    const apps = listAppsResponse.apps ?? [];
+
+    for (const app of apps) {
+      const envResponse = await amplifyClient.send(
+        new ListBackendEnvironmentsCommand({
```
adrianjoshua-strutt: Could this potentially be rate limited?

packages/amplify-provider-awscloudformation/src/gen1-new-customer-restriction.ts
```diff
@@ -0,0 +1,66 @@
...
+  } while (nextToken);
+
+  return false;
+};
+
+/**
+ * Enforce the Gen 1 new-customer restriction.
+ *
+ * Calls {@link isExistingGen1Customer} and, if the account is NOT an existing
+ * customer, prints a yellow warning to stderr and throws an `AmplifyError` to
```
adrianjoshua-strutt, line 57: Maybe I am missing it, but where is the yellow warning being printed?

packages/amplify-e2e-tests/src/__tests__/init-gen1-restriction.test.ts
```diff
@@ -0,0 +1,22 @@
...
+  });
+
+  afterEach(() => {
+    deleteProjectDir(projRoot);
+  });
+
+  it('should block amplify init --quickstart for new Gen 1 customers', async () => {
+    await expect(
+      spawn(getCLIPath(), ['init', '--quickstart'], { cwd: projRoot, stripColors: true }).runAsync(),
+    ).rejects.toThrowError('Process exited with non zero exit code 1');
```
adrianjoshua-strutt, line 22: If I am not mistaken, any init failure would pass. Should we be more precise here and check for a specific message?

packages/amplify-provider-awscloudformation/src/gen1-new-customer-restriction.ts
Simone319: AmplifyError thrown on line 67 will be caught by the CLI [error handler](https://github.com/aws-amplify/amplify-cli/blob/dev/packages/amplify-cli/src/amplify-exception-handler.ts) and printed.

packages/amplify-provider-awscloudformation/src/initializer.ts
Simone319: `amplify init` command makes several calls using `amplifyClient`, in addition to what we are using here, to check for existing app, create app, etc. If the amplifyClient is null, the app creation will fail anyways.

packages/amplify-provider-awscloudformation/src/gen1-new-customer-restriction.ts
Simone319: Amplify provides 20,000 tokens/second per account, per region (shared across all Amplify APIs) and the quota is adjustable. See: https://docs.aws.amazon.com/amplify/latest/userguide/quotas-chapter.html. Even it's unlikely that an account will run into the limit, will add an try catch block to fail safe.

packages/amplify-e2e-tests/src/__tests__/init-gen1-restriction.test.ts
Simone319: Updated the test with verifying the messaging.

svidgen (dismissed):

adrianjoshua-strutt (dismissed):

adrianjoshua-strutt (approved):

adrianjoshua-strutt (approved):

PR #14829, opened 2026-04-28
chore: pass cloud-e2e CodeBuild jobs

iankhou:
#### Description of changes

Various changes needed to make code from the `gen2-migration` branch pass `yarn cloud-e2e` checks.

#### Checklist

- [X] PR description included
- [X] `yarn test` passes

### DIFF
--- a/.circleci/local_publish_helpers_codebuild.sh
+++ b/.circleci/local_publish_helpers_codebuild.sh
@@ -28,7 +28,8 @@ function uploadPkgCliCodeBuild {
     set -e
 
     cd out/
-    export version=$(./amplify-pkg-linux-x64 --version)
+    # --version can emit warnings before the version line; keep only the last line
+    export version=$(./amplify-pkg-linux-x64 --version | tail -n 1)
 
     # validate that version is uploaded in right build
     if [[ "$version" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
--- a/.circleci/local_publish_helpers_codebuild.sh
+++ b/.circleci/local_publish_helpers_codebuild.sh
@@ -140,7 +140,7 @@ function verifyPkgCli {
     }
 
     # TODO: After V3 migrations are done, decrease 1095 back to 930 and 875 back to 750
-    verifySinglePkg "amplify-pkg-linux-x64" "amplify-pkg-linux-x64.tgz" $((1095 * 1024 * 1024))
+    verifySinglePkg "amplify-pkg-linux-x64" "amplify-pkg-linux-x64.tgz" $((1130 * 1024 * 1024))
     verifySinglePkg "amplify-pkg-macos-x64" "amplify-pkg-macos-x64.tgz" $((1105 * 1024 * 1024))
     verifySinglePkg "amplify-pkg-win-x64.exe" "amplify-pkg-win-x64.tgz" $((1095 * 1024 * 1024))
     verifySinglePkg "amplify-pkg-linux-arm64" "amplify-pkg-linux-arm64.tgz" $((875 * 1024 * 1024))
--- a/packages/amplify-cli-npm/index.ts
+++ b/packages/amplify-cli-npm/index.ts
@@ -16,4 +16,4 @@ export const install = async (): Promise<void> => {
   return binary.install();
 };
 
-// force version bump to 14.3.0
+// force version bump to 14.4.0
--- a/packages/amplify-cli/src/commands/diagnose.ts
+++ b/packages/amplify-cli/src/commands/diagnose.ts
@@ -1,4 +1,11 @@
-import { stateManager, pathManager, spinner, DiagnoseReportUploadError, projectNotInitializedError } from '@aws-amplify/amplify-cli-core';
+import {
+  stateManager,
+  pathManager,
+  spinner,
+  DiagnoseReportUploadError,
+  projectNotInitializedError,
+  AmplifyException,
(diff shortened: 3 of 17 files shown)


.circleci/local_publish_helpers_codebuild.sh
```diff
@@ -28,7 +28,8 @@ function uploadPkgCliCodeBuild {
     set -e
 
     cd out/
-    export version=$(./amplify-pkg-linux-x64 --version)
+    # --version can emit warnings before the version line; keep only the last line
+    export version=$(./amplify-pkg-linux-x64 --version | tail -n 1)
```
iankhou, line 33: Fixes this silly thing:

```log
Unknown options: AWS,Amplify,Gen,1,CLI,is,in,maintenance,mode,and,will,reach,end,of,life,on,May,1,,2027.,During,maintenance,mode,,only,critical,bug,fixes,and,security,patches,will,be,provided.,Migrate,to,Amplify,Gen,2:,https://docs.amplify.aws/react/start/migrate-to-gen2/,14.4.0-dev.8ca9ebdc46194759a107471776bf8690a56452fa.0/amplify-pkg-win-x64.tgz
```

.circleci/local_publish_helpers_codebuild.sh
```diff
@@ -140,7 +141,7 @@ function verifyPkgCli {
     }
 
     # TODO: After V3 migrations are done, decrease 1095 back to 930 and 875 back to 750
-    verifySinglePkg "amplify-pkg-linux-x64" "amplify-pkg-linux-x64.tgz" $((1095 * 1024 * 1024))
+    verifySinglePkg "amplify-pkg-linux-x64" "amplify-pkg-linux-x64.tgz" $((1130 * 1024 * 1024))
```
iankhou, line 144: Linux binary grew past this hardcoded limit with the addition of our gen2-migration code and dependencies.

packages/amplify-cli-npm/index.ts
```diff
@@ -16,4 +16,4 @@ export const install = async (): Promise<void> => {
   return binary.install();
 };
 
-// force version bump to 14.3.0
+// force version bump to 14.4.0
```
iankhou, line 19: Need version bump to keep internal and npm packages in sync.

packages/amplify-cli/src/commands/diagnose.ts
```diff
@@ -241,3 +249,23 @@ const getAppId = (): string => {
...
+/**
+ * Convert an error into a JSON-safe shape for inclusion in the diagnose report.
+ *
+ * Caught AWS SDK errors attach the raw `IncomingMessage`/`ClientRequest` via `$response`,
+ * which contains a circular `req <-> res` reference. A naive `JSON.stringify(error)` throws
+ * `TypeError: Converting circular structure to JSON` and aborts report creation. Routing
+ * through `SerializableError` guarantees a cycle-free payload and redacts ARNs and
+ * home-directory paths as a side benefit.
+ */
+const serializeErrorForReport = (error: Error): object => {
```
iankhou, line 262: Sometimes during failures, the error will have a circular reference, which causes the error reporting mechanism to fail.

packages/amplify-e2e-core/src/init/initProjectHelper.ts
```diff
@@ -107,6 +140,91 @@ export function initJSProjectWithProfile(cwd: string, settings?: Partial<typeof
   return chain.wait(/Try "amplify add api" to create a backend API and then "amplify (push|publish)" to deploy everything/).runAsync();
 }
 
+export function initJSProjectWithProfileGen2Migration(cwd: string, settings?: Partial<typeof defaultSettings>): Promise<void> {
```
iankhou, line 143: We author a special version of this function that is "more correct" than the original, which doesn't actually choose the profile that's passed in (always chooses the first available profile). This new version is not backwards-compatible because it throws when there are no profiles configured. However, our existing E2Es depend on that behavior.

iankhou:
cloud-e2e: https://tiny.amazon.com/15rnkk0bw/IsenLink

packages/amplify-cli/src/commands/diagnose.ts
```diff
@@ -163,7 +171,7 @@ const createZip = async (context: Context, error: Error | undefined): Promise<st
   }
 
   if (error) {
-    zipper.append(JSON.stringify(error, null, 4), {
+    zipper.append(JSON.stringify(serializeErrorForReport(error), null, 4), {
```
iliapolo, line 174: Is this necessary or a fly by fix?

iankhou: It's a bug. When certain code fails, the error report also fails. This prevents the error report from failing so we can properly diagnose the problem from the artifact.

It is possible that we need it, and possible that we don't - for this change in particular. But we can't know without running this code (and encountering a failure that uses it).

iankhou: - w.r.t.: Gen 2 Migration code
  - Suppose: It breaks something
    - Suppose: We get a helpful error message
      - NO we don't need this code
    - Suppose: We do not get a helpful error message
      - Suppose: We retry and the code works - NO we don't need this code
      - Suppose: We retry and the code doesn't work - **YES we need this code**
  - Suppose: it doesn't break anything - NO we don't need this code

iliapolo (dismissed):

iliapolo (approved):

9pace (approved):

PR #14843, opened 2026-04-30
feat(cli): add `retain` command for gen2-migration

### LINKED ISSUE #14790: (gen2-migration) implement `retain` command and add it to the E2E migration flow

_No response_

_No response_

Add a new `retain` subcommand to `amplify gen2-migration` and extend the E2E flow to cover it. Two concerns, one delivery:

#### 1. Implement the `retain` subcommand

`amplify gen2-migration retain` applies `DeletionPolicy: Retain` and `UpdateReplacePolicy: Retain` to every resource in every Gen1 CloudFormation stack (root + all nested, recursively). The purpose is to protect the Gen2 environment from unintended impact caused by the Gen1 stacks after migration.

#### 2. Add `retain` to the E2E flow

The E2E system currently runs deploy → assess → lock → generate → sandbox → refactor → test and stops after refactor. There's no automated coverage for `retain`.

#### 1. For behaviour of `retain`:

- Walks the Gen1 stack hierarchy via paginated `ListStackResources` calls (leaf-first).
- For each stack, fetches the template, sets `DeletionPolicy: Retain` / `UpdateReplacePolicy: Retain` on every resource entry, creates a changeset, and executes it.
- Strict per-stack whitelist validation: the changeset must only contain `DeletionPolicy` / `UpdateReplacePolicy` → `Retain` edits. Anything else fails validation with the rendered changeset as the report.
- Each operation's `describe()` surfaces the CloudFormation changeset console URL so the user can inspect exactly what will change before approving.
- Rollback throws descriptive one-way operation message.

#### 2. For wiring it in e2e migration flow:

Extend `App.migrate()` to run `retain` after the final `testSharedData()` call, followed by another `testGen2()` to confirm the Gen2 app still behaves correctly once the Gen1 stacks are marked for retention.

In `App.migrate()`, after the final `testSharedData()` call, add:

```ts
await this.retain();
await this.testGen2();
```

Where `retain()` follows the same pattern as `refactor()`:
```ts
public async retain(): Promise<void> {
  await this.runMigrationStep('retain');
}
```

This validates that after retention, the Gen2 app is fully self-contained and is still functional.

None

### Additional context

_No response_

sai-ray:
Fixes #14790 

#### Description of changes

Adds a new `amplify gen2-migration retain` step and removes dead code left behind by the upstream decommission removal.

#### `retain` subcommand

After migration, Gen1 CloudFormation stacks are defunct but still contain custom resources that may reference resources now owned by Gen2. Changes to those stacks can cause unintended impact on the Gen2 environment.

`retain` walks every Gen1 stack below the root top-down and applies `DeletionPolicy: Retain` and `UpdateReplacePolicy: Retain` to every resource except `AWS::CloudFormation::Stack` references. Once applied, the user can manually delete the Gen1 root stack and every underlying AWS resource survives as an orphan.

##### Skip root

Updating the root post-refactor risks cascading `TemplateURL` reconciliation through the entire tree. The user manually deletes the root after `retain` completes; CloudFormation cascades through the already-retained children and every resource below root gets `DELETE_SKIPPED`.

##### Skip `AWS::CloudFormation::Stack` references

Adding retain to a nested-stack reference would be a no-op for child protection — the child's own retain state is what matters when the cascade delete hits it. Leaving the reference entry untouched keeps the parent changeset strictly additive on non-stack attributes.

##### Per-operation changeset flow

Each stack's `fetchTemplate → createChangeSet → executeChangeSet` round-trip happens inside the operation's `execute()` rather than at plan time. Creating all changesets up front would produce `OBSOLETE` changesets on every parent execute.

##### Strict allowlist

Every changeset is validated against a two-shape whitelist: direct `DeletionPolicy` / `UpdateReplacePolicy` edits targeting `Retain`, or CloudFormation's own Automatic/Dynamic re-evaluations on `AWS::CloudFormation::Stack` references emitted on every parent update. Anything else (property edits, `Add`, `Remove`, `Replacement: True`) throws `MigrationError`.

##### One-way operation

`--rollback` is rejected at the dispatcher and `step.rollback()` throws `NotImplementedFault`. To undo, the user edits the CloudFormation templates directly.

#### Dead code removal

The upstream decommission removal left `validateStatefulResources` and its private helper `getStatefulResources` in `_common/validations.ts` with no remaining callers, plus 619 lines of tests, a `bottleneck` dependency used only by those methods, and three unimplemented `logger.warn('Not implemented')` stubs (`validateDeploymentVersion`, `validateIsolatedEnvironment`, `validateIngressTraffic`). All removed.

#### Issue #, if available

#14790 

- Added 13 unit tests covering the walker, the per-stack operation lifecycle, and the changeset whitelist.
- End-to-end validated against both `mood-board` and `project-boards` in `amplify-migration-apps/`. After running the full migration workflow and invoking `retain`, the user-triggered manual root-stack delete completed successfully — all nested stack resources transitioned to `DELETE_SKIPPED` and every underlying AWS resource (DynamoDB, Lambda, S3, Cognito, AppSync, Kinesis) survived as an orphan.
- Post-delete `test:gen1` verified the retained resources are still reachable and the app continues to function.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes (for the changed packages)
- [x] Tests are added
- [x] Relevant documentation is changed or added


packages/amplify-cli/src/commands/gen2-migration/retain.ts
```diff
@@ -0,0 +1,343 @@
...
+  /**
+   * Name of the AppSync `@model` type owning this stack.
+   * Present for per-model nested stacks like Board, Todo, MoodItem.
+   */
+  readonly modelName?: string;
+
+  /**
+   * Logical id of the AppSync infrastructure sub-stack.
+   */
+  readonly subStackLabel?: string;
```
iliapolo: Not needed. This information is already contained within the stack name, albeit less clear, but still.

packages/amplify-cli/src/commands/gen2-migration/retain.ts
```diff
@@ -0,0 +1,343 @@
...
+            this.logger.push(ctx.subStackLabel);
+            pushed++;
+          }
+        }
+        try {
+          const template = await cfn.fetchTemplate(stackId);
+
+          const targetEntries = Object.values(template.Resources).filter((r) => r.Type !== 'AWS::CloudFormation::Stack');
+          if (targetEntries.length === 0) {
+            this.logger.info(`${stackName} — no non-nested-stack resources to retain`);
```
iliapolo:
```suggestion
            this.logger.info(`${stackName} — no resources to retain`);
```

packages/amplify-cli/src/commands/gen2-migration/retain.ts
```diff
@@ -0,0 +1,343 @@
...
+            this.logger.info(`${stackName} — no retain changes needed`);
+            return;
+          }
+
+          for (const resource of targetEntries) {
+            resource.DeletionPolicy = 'Retain';
+            resource.UpdateReplacePolicy = 'Retain';
+          }
+
+          const describeResponse = await this.gen1App.clients.cloudFormation.send(new DescribeStacksCommand({ StackName: stackId }));
```
iliapolo: use `cfn.describeStack`

iliapolo (changes requested):

packages/amplify-cli/src/commands/gen2-migration/retain.ts
sai-ray: Updated and removed related helpers.

iliapolo (approved):

PR #14850, opened 2026-05-01
fix(cli): dynamo table mappings are sometimes omitted in `gen2-migration generate` 

### LINKED ISSUE #14487: (gen2-migration) `generate` command sometimes omits dynamo table model mappings

npm install --no-save @aws-amplify/cli-internal-gen2-migration-experimental-alpha

_No response_

### Amplify CLI Version

0.4.0

Mac

No

### Describe the bug

When running `npx amplify gen2-migration generate` the data resource sometimes looks like this:

```ts
  authorizationModes: {
    defaultAuthorizationMode: "iam",
    apiKeyAuthorizationMode: { expiresInDays: 7 }
  },
  schema
});
```

Its missing the `migratedAmplifyGen1DynamoDbTableMappings` property which tells amplify which tables should be reused for which model.

### Expected behavior

Data resource should always include the mapping if an app has a GraphQL schema:

```ts
  migratedAmplifyGen1DynamoDbTableMappings: [{
    //The "branchname" variable needs to be the same as your deployment branch if you want to reuse your Gen1 app tables
    branchName: "main",
    modelNameToTableNameMapping: { User: "User-vzljmnp34bajrbgjzq7gowln6q-main", Product: "Product-vzljmnp34bajrbgjzq7gowln6q-main", Comment: "Comment-vzljmnp34bajrbgjzq7gowln6q-main" }
  }],
  authorizationModes: {
    defaultAuthorizationMode: "iam",
    apiKeyAuthorizationMode: { expiresInDays: 7 }
  },
  schema
});
```

### Reproduction steps

This behavior is sporadic and probably doesn't relate to a specific app. For me though it happened while following the migration guide of the [fitness-tracker app](https://github.com/aws-amplify/amplify-cli/tree/gen2-migration/amplify-migration-apps/fitness-tracker)

### Project Identifier

_No response_

### Log output

<details>

```


```

</details>

### Additional information

_No response_

dgandhi62:
Fixes #14487.

## Issue Summary

The `generate` command sometimes omits `migratedAmplifyGen1DynamoDbTableMappings` from the data resource. This is caused by the model-detection regex `type\s+(\w+)\s+@model` failing when `@model` is not the first directive on a type (e.g., `type Todo @auth(...) @model`).

## Reasoning

1. The issue describes the bug as "sporadic," but investigation revealed it's actually deterministic based on directive ordering in the GraphQL schema.
2. Read `data.generator.ts` — found `createTableMappings()` which uses `type\s+(\w+)\s+@model` to find model types. This regex requires `@model` to immediately follow the type name, so `type Todo @auth(...) @model` doesn't match.
3. The compiled `build/schema.graphql` (produced by the Amplify transformer) is more reliable — it expands each `@model` type into a `Model<Name>Connection` type. Matching `type Model(\w+)Connection` against the build schema is directive-order-independent.
4. The fix reads `build/schema.graphql` when available and falls back to the raw schema regex when it's not.

## Solution

- **`data.generator.ts`** — `createTableMappings()` now first checks for `build/schema.graphql` in the API resource directory. If present, it extracts model names by matching `type Model(\w+)Connection` patterns. Falls back to the original `@model` regex when the build schema is absent.
- **`data.generator.test.ts`** — Added test `'renders table mappings when @model is not the first directive'` with a schema where `@auth` and `@key` precede `@model`, verifying both types are correctly mapped.

## Example

**Input (Gen 1 — schema.graphql):**
```graphql
type Todo @auth(rules: [{ allow: public }]) @model {
  id: ID!
}
type Post @key(name: "byUser") @model {
  id: ID!
}
```

**Output — before fix (resource.ts):**
```ts
  // migratedAmplifyGen1DynamoDbTableMappings is MISSING
  schema,
});
```

**Output — after fix (resource.ts):**
```ts
  migratedAmplifyGen1DynamoDbTableMappings: [
    {
      branchName: 'main',
      modelNameToTableNameMapping: {
        Todo: 'Todo-abc-main',
        Post: 'Post-abc-main',
      },
    },
  ],
  schema,
});
```

### DIFF
--- a/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/data/data.generator.test.ts
+++ b/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/data/data.generator.test.ts
@@ -26,16 +26,13 @@ const dataResource: DiscoveredResource = {
   key: 'api:AppSync',
 };
 
-/** Generates a minimal build schema containing ModelXConnection types for each model name. */
-function buildSchemaFor(...modelNames: string[]): string {
-  return modelNames.map((m) => `type Model${m}Connection { items: [${m}]! nextToken: String }`).join('\n');
-}
-
 /**
- * Mocks gen1App.file() to return the build schema for the build path
- * and the raw schema for all other paths.
+ * Mocks gen1App.file() so it returns the raw schema for the source
+ * path and a synthetic build schema (with ModelXConnection types)
+ * for the build path.
  */
-function mockFiles(gen1App: Gen1App, rawSchema: string, buildSchema: string): void {
+function mockSchema(gen1App: Gen1App, rawSchema: string, modelNames: string[]): void {
+  const buildSchema = modelNames.map((m) => `type Model${m}Connection { items: [${m}]! nextToken: String }`).join('\n');
   jest.spyOn(gen1App, 'file').mockImplementation((p: string) => {
     if (p.includes('build/schema.graphql')) return buildSchema;
     return rawSchema;
@@ -63,7 +60,7 @@ describe('DataGenerator', () => {
         },
       },
     });
-    mockFiles(gen1App, 'type Todo @model { id: ID! }', buildSchemaFor('Todo'));
+    mockSchema(gen1App, 'type Todo @model { id: ID! }', ['Todo']);
 
     const generator = new DataGenerator(gen1App, backendGenerator, outputDir, dataResource);
     await expect(generator.plan()).rejects.toThrow('GraphQLAPIIdOutput');
@@ -79,7 +76,7 @@ describe('DataGenerator', () => {
         },
       },
     });
-    mockFiles(gen1App, 'type Todo @model { id: ID! }', buildSchemaFor('Todo'));
+    mockSchema(gen1App, 'type Todo @model { id: ID! }', ['Todo']);
     jest.spyOn(gen1App.aws, 'fetchGraphqlApi').mockResolvedValue(undefined);
 
     const generator = new DataGenerator(gen1App, backendGenerator, outputDir, dataResource);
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/data/data.generator.test.ts
```diff
@@ -659,4 +659,59 @@ describe('DataGenerator', () => {
...
+    const buildSchema = [
+      'type Todo { id: ID! createdAt: AWSDateTime! updatedAt: AWSDateTime! }',
+      'type Post { id: ID! createdAt: AWSDateTime! updatedAt: AWSDateTime! }',
+      'type ModelTodoConnection { items: [Todo]! nextToken: String }',
+      'type ModelPostConnection { items: [Post]! nextToken: String }',
+    ].join('\n');
+    jest.spyOn(gen1App, 'fileExists').mockImplementation((p: string) => p.includes('build/schema.graphql'));
+    jest.spyOn(gen1App, 'file').mockImplementation((p: string) => {
+      if (p.includes('build/schema.graphql')) return buildSchema;
+      return rawSchema;
```
iliapolo: this means that any file requested from the app except for `build/schema.graphql` will return the raw schema - is that correct? this should throw instead, we don't expect any other file to be requested in this test.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/data/data.generator.ts
```diff
@@ -80,7 +80,30 @@ export class DataGenerator implements Planner {
...
+   * transformer expands each @model type into a `Model<Name>Connection`
+   * type. This is more reliable than regex-matching `@model` in the
+   * raw schema, which is sensitive to directive ordering.
+   *
+   * Falls back to the raw schema regex when build/schema.graphql is
+   * absent.
+   */
   private createTableMappings(schema: string, apiId: string): Record<string, string> {
+    const buildSchemaPath = path.join('api', this.resource.resourceName, 'build', 'schema.graphql');
+    if (this.gen1App.fileExists(buildSchemaPath)) {
```
iliapolo: does this represent a valid state? why would this file not exist but the raw schema file would?

iliapolo (changes requested):

iliapolo (approved):

Issue #14523, opened 2026-01-26
(gen2-migration) `refactor` commands breaks social auth providers deployment


iliapolo:
### How did you install the Amplify CLI?

npm install --no-save @aws-amplify/cli-internal-gen2-migration-experimental-alpha

### Amplify CLI Version

0.6.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

When an a Gen1 application configures social auth providers, deploying the Gen2 app post refactor fails.

First, it encounters this error:

```console
CREATE_FAILED | AWS::Cognito::UserPoolDomain | auth/amplifyAuth/UserPool/UserPoolDomain (amplifyAuthUserPoolUserPoolDomain1F688B5B) Resource handler returned message: "Invalid request provided: AWS::Cognito::UserPoolDomain" (RequestToken: 907cfef3-530e-84af-bf46-76cf38436b0c, HandlerErrorCode: InvalidRequest)
```

I _think_ this happens because the in Gen1 user pool domains are managed outside of CloudFormation. During refactor, the Gen2 domain is deleted and then recreated during the subsequent deployment. This recreation fails.

This specific error can be avoided by uncommenting the following line from the generated code:

```diff
- backend.auth.resources.userPool.node.tryRemoveChild("UserPoolDomain");
```

Redeploying again will now have this error:

```console
2026-01-26T01:00:49.674Z [INFO]: [CloudFormationDeploymentError] The CloudFormation deployment has failed.
                                 ∟ Caused by: [ToolkitError] The stack named amplify-d264qfo090j6p0-gen2main-branch-b9403b9a3c failed to deploy: UPDATE_ROLLBACK_COMPLETE: 

Resource handler returned message: "Resource of type 'AWS::Cognito::UserPoolIdentityProvider' with identifier 'Facebook' already exists." (RequestToken: e1117462-b53b-8fef-0981-07ca8a7f57fb, HandlerErrorCode: AlreadyExists), 
Resource handler returned message: "Resource of type 'AWS::Cognito::UserPoolIdentityProvider' with identifier 'Google' already exists." (RequestToken: d0482450-6345-eb97-557a-73a79aa97cac, HandlerErrorCode: AlreadyExists), 
[...]

Issue #14521, opened 2026-01-25
(gen2-migration) `generate` command creates non compiling social auth providers configuration


iliapolo:
### How did you install the Amplify CLI?

npm install --no-save @aws-amplify/cli-internal-gen2-migration-experimental-alpha

### Amplify CLI Version

0.6.0

### What operating system are you using?

Mac

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

No

### Describe the bug

When a Gen1 app is configured with a social auth provider, we generate:

```ts
const userPoolClient = userPool.addClient("NativeAppClient", {
  ...,
  flows: ["code"],
  ...,
});
```

The `flow` property causes a compilation error:

```console
Object literal may only specify known properties, and 'flows' does not exist in type 'UserPoolClientOptions'
```

### Expected behavior

Not sure. We definitely need to remove the property but I don't know if we need to replace it with something. 

### Reproduction steps

Follow the [media-vault](https://github.com/aws-amplify/amplify-cli/tree/gen2-migration/amplify-migration-apps/media-vault) app migration guide.

### Log output

<details>

```
# Put your logs below this line


```

</details>

### Before submitting, please confirm:

- [x] I have done my best to include a minimal, self-contained set of instructions for consistently reproducing the issue.
- [x] I have removed any sensitive information from my code snippets and submission.

PR #14853, opened 2026-05-01
fix(cli): support social auth in gen2-migration

9pace:
## Summary

`amplify gen2-migration refactor` didn't work for apps with social login (Google, Facebook). This PR fixes that with an **orphan + import** approach that preserves the Gen1 UserPool domain and IDPs through the refactor with zero downtime.

## The problem 

Gen1 represents social auth via Lambda-backed custom resources (CFN doesn't see them as native Cognito types); Gen2 represents them as native `AWS::Cognito::UserPoolDomain` + `UserPoolIdentityProvider` resources that `Fn::GetAtt` into a stack-local `AmplifySecretFetcherResource`. Moving IDPs through the holding stack breaks those refs and tries (and fails) to recreate resources that already exist.

## The fix: orphan + import

**Don't move social auth resources through the holding stack.** Instead:

| Phase | What happens |
|-------|---|
| `beforeMove()` | Move the 4 core Cognito resources (UserPool, 2× UserPoolClient, IdentityPool, IdentityPoolRoleAttachment) to holding. Separately **orphan** UserPoolDomain + IDPs from Gen2. 
| `move()` | Move Gen1 core resources into Gen2 via StackRefactor. Then **import** the Gen1 domain + IDPs into Gen2 under the Gen2 logical IDs via `CreateChangeSet(IMPORT)`. |

Rollback mirrors the forward flow in reverse:

| Phase | What happens |
|-------|---|
| `move()` | **Orphan** the imported UserPoolDomain + IDPs from Gen2. Then move the 4 core Cognito resources back to Gen1 via StackRefactor. |
| `afterMove()` | Restore Gen2's original UserPoolDomain + IDPs from the holding stack via `CreateChangeSet(IMPORT)` under the same Gen2 logical IDs they had pre-refactor. |

**Why CFN import works here:** Gen1's IDPs/domain live on the pool as physical resources but CFN tracks them as `Custom::LambdaCallout` invocations, so they're eligible for adoption. Import only checks `ResourceIdentifier`; template property values are metadata, so dummy values are accepted.

## Safety: `DeletionPolicy: Retain` at three points

Orphaning removes resources from the template. Without Retain, CFN would physically delete them. Retain is applied at:

1. **Lock step**: on Gen1's `HostedUICustomResourceInputs` + `HostedUIProvidersCustomResourceInputs` (matched by logical ID, not type, to avoid unrelated `Custom::LambdaCallout` resources). This also prevents the Gen1 Lambda delete handlers from firing when the Gen1 stack is decommissioned.
2. **Generate step**: CDK escape hatches set Retain on Gen2's `UserPoolDomain` + `UserPoolIdentityProvider` so every deploy preserves it.
3. **Import step**: Retain set inline on all imported resource definitions.

Orphan also has an execute-time guard that throws if any targeted resource is missing Retain.

## Adjacent fixes
Small fixes surfaced during the work:

- **Output resolver** (`cfn-output-resolver.ts`): Ref fallback to `PhysicalResourceId`; skip `Custom::` resources in `GetAtt`; avoid ARN double-nesting.
- **NoEcho parameters** (`cfn-parameter-resolver.ts`): `DescribeStacks` masks them as `"****"`, which crashed downstream Lambda custom resources on `UpdateStack`. Now sends `UsePreviousValue: true` for NoEcho params across all 4 resolve paths.
- **Generate auth** (`auth.renderer.ts`):
  - `authorized_scopes` → `authorize_scopes` (matches Cognito API).
  - Stopped filtering IDP scopes against a hardcoded allowlist.
  - Preserved custom attribute mappings (e.g., `username: 'sub'`).
  - Emits `addPropertyDeletionOverride('SupportedLoginProviders')` on IdentityPool (Gen1 doesn't set it; leaving it causes drift).
  - Emits `cfnUserPoolDomain.domain = '<gen1-domain>'` as post refactor step to preserve the Gen1 domain prefix after import.

And the very nice: 
- **Dead code removal**: deleted `oauth-values-retriever.ts` and its `resolveOAuthParameters` hook so *real OAuth secrets are not needed during refactor.*

## Issues

- https://github.com/aws-amplify/amplify-cli/issues/14523
- https://github.com/aws-amplify/amplify-cli/issues/14521

## Test plan

- [x] `npx tsc --noEmit` passes
- [x] Unit tests pass
- [x] **Live e2e** against `media-vault` (Google + Facebook OAuth) full flow: Gen1 push → Gen2 sandbox → refactor → `test:gen1` + `test:gen2`
- [x] Full manual testing on media-vault, with rollback flow validated

## Notes

- ADR-005 (`packages/amplify-cli/adr/005-refactor-social-auth-idp-domain-handling.md`) has the full architecture walkthrough and the rejected alternatives.


packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-rollback.ts
```diff
@@ -67,7 +61,10 @@ export class AuthCognitoRollbackRefactorer extends RollbackCategoryRefactorer {
       const gen2StackName = extractStackNameFromId(gen2StackId);
       baseOps.push({
         resource: this.resource,
-        validate: () => undefined,
+        validate: () => ({
+          description: `Deletion Protection: ${gen2StackName}`,
```
9pace: "deletion protection on social providers:"

packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/category-refactorer.ts
```diff
@@ -81,11 +81,11 @@ export abstract class CategoryRefactorer implements Planner {
     const source = await this.resolveSource(sourceStackId);
     const target = await this.resolveTarget(destStackId);
 
-    const sourceResources = this.filterResourcesByType(source.resolvedTemplate);
-    const targetResources = this.filterResourcesByType(target.resolvedTemplate);
+    const sourceResources = filterResourcesByTypes(source.resolvedTemplate, this.resourceTypes());
```
9pace: This can be changed back to     const sourceResources = this.filterResourcesByType(source.resolvedTemplate);


only ever used with this.resourceTypes

packages/amplify-cli/src/commands/gen2-migration/_common/aws-fetcher.ts
```diff
@@ -88,6 +102,26 @@ export class AwsFetcher {
     return this.clients.cognitoIdentity.send(new DescribeIdentityPoolCommand({ IdentityPoolId: identityPoolId }));
   }
 
+  /**
+   * Returns domain + IDP summaries for a UserPool, or undefined if the pool
+   * has no domain or no IDPs (non-social-auth app).
+   */
+  public async fetchSocialAuthConfig(userPoolId: string): Promise<SocialAuthConfig | undefined> {
```
9pace: this can be moved to stack-facade

packages/amplify-cli/src/commands/gen2-migration/_common/aws-fetcher.ts
```diff
@@ -26,6 +27,19 @@ import { GetAppCommand } from '@aws-sdk/client-amplify';
...
+/** Non-null name + type pair for a Cognito identity provider. */
+interface IdpConfig {
+  readonly providerName: string;
+  readonly providerType: string;
+}
+
+/** A UserPool's domain and identity-provider summaries. */
+export interface SocialAuthConfig {
+  readonly userPoolId: string;
+  readonly domain: string;
```
9pace: domainName

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
```diff
@@ -286,6 +287,96 @@ export class Cfn {
...
   }
 
+  /**
+   * Orphans resources from a stack by removing them from the template and
+   * an UpdateStack. Target resources must have `DeletionPolicy: Retain`
+   */
+  public async orphan(params: {
+    readonly stackName: string;
+    readonly logicalIds: string[];
+    readonly resource?: DiscoveredResource;
```
9pace: should not be optional for orphan

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
```diff
@@ -286,6 +287,96 @@ export class Cfn {
...
+
+  /**
+   * Imports existing physical resources into a stack via CreateChangeSet(IMPORT).
+   * The template must include resource definitions matching the physical state.
+   */
+  public async importResources(params: {
+    readonly stackName: string;
+    readonly templateAdditions: Record<string, CFNResource>;
+    readonly resourcesToImport: ResourceToImport[];
+    readonly resource?: DiscoveredResource;
```
9pace: resource should not be optional for import

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
```diff
@@ -286,6 +287,96 @@ export class Cfn {
...
+      templateBody: template,
+      parameters: stack.Parameters ?? [],
+      resource,
+    });
+  }
+
+  /**
+   * Imports existing physical resources into a stack via CreateChangeSet(IMPORT).
+   * The template must include resource definitions matching the physical state.
+   */
```
9pace, line 346: comment that explains templateadditions, resourcestoimport jsdoc

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
```diff
@@ -286,6 +287,96 @@ export class Cfn {
...
+  public async orphan(params: {
+    readonly stackName: string;
+    readonly logicalIds: string[];
+    readonly resource?: DiscoveredResource;
+  }): Promise<void> {
+    const { stackName, logicalIds, resource } = params;
+    const displayName = extractStackNameFromId(stackName);
+    const template = await this.fetchTemplate(stackName);
+
```
9pace, line 311: Requires orphan snapshot (similar to import)

9pace: can be done by passing a snapshot prefix arg to update, which will then write the snapshot. Currently it would overwrite the prev update snapshots

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
```diff
@@ -286,6 +287,96 @@ export class Cfn {
...
+   */
+  public async importResources(params: {
+    readonly stackName: string;
+    readonly templateAdditions: Record<string, CFNResource>;
+    readonly resourcesToImport: ResourceToImport[];
+    readonly resource?: DiscoveredResource;
+  }): Promise<void> {
+    const { stackName, templateAdditions, resourcesToImport, resource } = params;
+    const displayName = extractStackNameFromId(stackName);
+    const changeSetName = `import-resources-${Date.now()}`;
```
9pace: add gen2-migration prefix

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
```diff
@@ -19,56 +19,223 @@ export const USER_POOL_TYPE = 'AWS::Cognito::UserPool';
...
+    .filter(([, r]) => r.Type === USER_POOL_DOMAIN_TYPE || r.Type === USER_POOL_IDENTITY_PROVIDER_TYPE)
+    .map(([id]) => id);
+  return ids.length > 0 ? ids : undefined;
+}
+
+/**
+ * Extracts the Gen2-original logical IDs to import physical resources
+ * under: UserPoolDomain logical ID and `providerName → logicalId` map
+ */
+export function extractImportLogicalIds(
```
9pace: rename to extractSocialAuthLogicalIds() 

and use the same function in orphan

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
```diff
@@ -19,56 +19,223 @@ export const USER_POOL_TYPE = 'AWS::Cognito::UserPool';
...
+  }
+
+  /**
+   * Moves Gen1 resources into Gen2
+   *
+   * For social auth:
+   * Imports Gen1's UserPoolDomain / UserPoolIdentityProvider under the Gen2-original logical IDs.
+   * Runs after super.move() so the UserPool is already in Gen2 when the import references it.
+   */
+  protected override async move(blueprint: RefactorBlueprint): Promise<AmplifyMigrationOperation[]> {
```
9pace, line 208:
```
  protected override async move(blueprint: RefactorBlueprint): Promise<AmplifyMigrationOperation[]> {
    const baseOps = await super.move(blueprint);

    const gen2StackId = blueprint.targetStackId;
    const template = await this.cfn.fetchTemplate(gen2StackId);

    const userPoolId = this.gen1App.resourceMetaOutput(this.resource, 'UserPoolId');
    const socialAuthConfig = await this.gen1App.aws.fetchSocialAuthConfig(userPoolId);
    if (socialAuthConfig){
      const importTargets = extractImportLogicalIds(template);
      if (!importTargets){
        throw new AmplifyError('MigrationError', {
          message: `Unable to determine logical IDs for social auth import`,
        });
      }

      const { domainLogicalId, idpLogicalIds } = importTargets;


      const { resourcesToImport, templateAdditions } = buildImportSpec(socialAuthConfig, domainLogicalId, idpLogicalIds);
      const gen2StackName = extractStackNameFromId(gen2StackId);

      baseOps.push({
        resource: this.resource,
        validate: () => undefined,
        describe: async () => [renderImportTable(resourcesToImport, gen2StackName)],
        execute: () =>
          this.cfn.importResources({
            stackName: gen2StackId,
            templateAdditions,
            resourcesToImport,
            resource: this.resource,
          }),
      });
    }


    return baseOps;
  }

```

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
```diff
@@ -19,56 +19,223 @@ export const USER_POOL_TYPE = 'AWS::Cognito::UserPool';
...
+      domainLogicalId = logicalId;
+    } else if (resource.Type === USER_POOL_IDENTITY_PROVIDER_TYPE) {
+      const providerName = resource.Properties.ProviderName as string;
+      if (providerName) {
+        idpLogicalIds.set(providerName, logicalId);
+      }
+    }
+  }
+
+  if (!domainLogicalId || idpLogicalIds.size === 0) return undefined;
```
9pace: remove this line. caller can determine to throw
if (!importTargets.domain)
if (!importTargets.logicalids)

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
```diff
@@ -19,56 +19,223 @@ export const USER_POOL_TYPE = 'AWS::Cognito::UserPool';
...
+    },
+  });
+
+  for (const provider of config.providers) {
+    const logicalId = idpLogicalIds.get(provider.providerName);
+    if (!logicalId) {
+      throw new AmplifyError('MigrationError', {
+        message:
+          `Identity provider '${provider.providerName}' exists on the UserPool but has no matching ` +
+          `UserPoolIdentityProvider resource in the Gen2 template. Add it to amplify/auth/resource.ts ` +
```
9pace:
```suggestion
```

remove resolution message

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-rollback.ts
```diff
@@ -32,6 +42,99 @@ export class AuthCognitoRollbackRefactorer extends RollbackCategoryRefactorer {
...
+
+    const { domainLogicalId, idpLogicalIds } = importTargets;
+    const gen2StackName = extractStackNameFromId(gen2StackId);
+
+    // The Gen2-original UserPool sits in the holding stack at plan time
+    // (super.afterMove() restores it to Gen2 only at execute time), so we
+    // read the pool ID from there.
+    const holdingStackName = this.getHoldingStackName(gen2StackName);
+    const holdingStack = await this.cfn.findStack(holdingStackName);
+    if (!holdingStack) return baseOps;
```
9pace, line 96: move up so this is the first early return. This is the gate to see if we have already rolled back

9pace: also apply same refactor as forward

9pace (commented):
enable e2e + enable unit test snapshot in refactor.test.ts

packages/amplify-cli/src/commands/gen2-migration/_common/aws-fetcher.ts
9pace: Keeping `domain`. `SocialAuthConfig` field names mirror the SDK: `userPoolId ↔ UserPoolId`, `providerName ↔ ProviderName`, `providerType ↔ ProviderType`, `domain ↔ Domain`. The CFN property on `AWS::Cognito::UserPoolDomain` is also `Domain` — consumers pass `Domain: config.domain` directly.

packages/amplify-cli/src/commands/gen2-migration/_common/aws-fetcher.ts
9pace: moved `fetchSocialAuthConfig` (+ `SocialAuthConfig` / `IdpConfig` interfaces) to `StackFacade`. StackFacade already has the Cognito client via its `AwsClients` bag, and the new `fetchSocialAuthConfig(stackId)` signature fits the facade's stack-scoped pattern (like `fetchUserPoolId`). Landed in 9454099a6a.

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-rollback.ts
9pace: done — 2955c56. now `Deletion Protection (social auth): ${gen2StackName}` in both forward and rollback. framework default stays `Deletion Protection: ${stackName}`.

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
9pace: done — a64311e

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
9pace: done — f96488e. same reasoning as orphan.

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
9pace: done — ad4a4ae.

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
9pace: done — template+params share the update path via a `snapshotPrefix` arg (orphan passes `'orphan'`). added sidecar `orphan.<stack>.logicalIds.json` in b907be7 for the unique operator-facing datum (what IDs got removed), symmetric with import's `resources.json`.

9pace: done — 2244665. `update()` gained `snapshotPrefix?: string` (default `'update'`). orphan passes `'orphan'`. also parameterized writeImportSnapshot symmetrically so both helpers share the same interface.

packages/amplify-cli/src/commands/gen2-migration/_common/cfn.ts
9pace: done — 954357c. now `gen2-migration-import-${Date.now()}`, matches the sibling at line 231.

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
9pace: done — 638fff6. single `extractSocialAuthLogicalIds(template)` returning `{ domainLogicalId: string | undefined, idpLogicalIds: Map<string,string> }` (always populated-or-empty). orphan sites flatten what's present; import sites validate per-field.

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
9pace: done in 0b560fa. took this shape. throw became narrower (`!domainLogicalId`) because  unified extractor never returns undefined the "can't determine logical IDs" case collapsed into per-field checks.

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
9pace: done — 638fff6 (same commit as C11). extractor is now pure; caller destructures and gates on `!domainLogicalId` / `idpLogicalIds.size`.

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-forward.ts
9pace: done — 12619b5. trimmed to `Identity provider '${providerName}' exists on the UserPool but has no matching UserPoolIdentityProvider resource in the Gen2 template.` — no resolution text.

packages/amplify-cli/src/commands/gen2-migration/refactor/auth/auth-cognito-rollback.ts
9pace: done — 9015650. `!holdingStack` is now the first early-return in `afterMove`. matches the base class pattern in `rollback-category-refactorer.ts`. also caught a retry bug — mid-rollback where orphan ran but import didn't, the old order would silently skip re-import because orphan deletes the template entries.

9pace: done — 3f2ce3a. wrapped-if shape on `socialAuthConfig`, matches forward `move()`. rollback keeps its extra `!holdingStack` gate at the top (forward has no equivalent).

packages/amplify-cli/src/commands/gen2-migration/refactor/workflow/category-refactorer.ts
9pace: update: implemented — 122367e. now `this.filterResourcesByType(template)` at all 3 call sites.

amplify-migration-apps/media-vault/_snapshot.post.generate/amplify/auth/resource.ts
```diff
@@ -18,17 +19,26 @@ export const auth = defineAuth({
...
         clientSecret: secret('FACEBOOK_CLIENT_SECRET'),
+        scopes: ['email', 'public_profile'],
         attributeMapping: {
           email: 'email',
+          custom: {
+            username: 'id',
+          },
         },
       },
+      // Add the Gen2 Amplify Hosting URL (e.g. https://<branch>.<gen2-appId>.amplifyapp.com/) after the gen2-main branch is deployed.
```
iliapolo: Put the comment twice, before each prop.

amplify-migration-apps/media-vault/_snapshot.post.generate/amplify/auth/resource.ts
```diff
@@ -103,6 +115,7 @@ export function applyEscapeHatches(backend: Backend) {
       UserPoolClientIdentityProvider.COGNITO,
     ],
     oAuth: {
+      // Add the Gen2 Amplify Hosting URL (e.g. https://<branch>.<gen2-appId>.amplifyapp.com/) after the gen2-main branch is deployed.
```
iliapolo: Same

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
```diff
@@ -97,13 +97,13 @@ export interface FunctionAccess {
  */
 export interface AuthRenderOptions {
   readonly userPool: UserPoolType;
-  readonly webClient: UserPoolClientType;
-  readonly nativeClient: UserPoolClientType;
   readonly identityPool?: IdentityPool;
   readonly identityProviders?: readonly IdentityProviderType[];
   readonly identityGroups?: readonly GroupType[];
+  readonly webClient?: UserPoolClientType;
```
iliapolo: Revert to https://github.com/aws-amplify/amplify-cli/pull/14857/changes#diff-56c69f3a44a042251f3231ce8f24713f5cb2a28a2d51c2af3c2c6faae563d8d7

iliapolo (changes requested):

iliapolo (dismissed):

iliapolo (approved):

PR #14856, opened 2026-05-02
chore(e2e): gen2-migration packaged binary support and placeholder app provisioning

iliapolo:
#### Description of changes

Improvements to the gen2-migration E2E infrastructure: support for
running against a packaged CLI binary, automatic Gen1 placeholder app
provisioning for fresh accounts, and better error diagnostics.

#### Gen1 placeholder app

The Gen1 new-customer restriction blocks `amplify init` in accounts
that have no existing Amplify app with a backend environment. E2E test
accounts can end up in this state after cleanup scripts run.

A shared `ensureGen1PlaceholderApp` utility is added to `amplify-e2e-core`
that creates an app named `gen1-placeholder-do-not-delete` with a `main`
backend environment if it doesn't already exist. The app name is exported
as `GEN1_PLACEHOLDER_APP_NAME` and used by both cleanup scripts (replacing
hardcoded strings) so the placeholder is never deleted.

The gen2-migration E2E runner calls this utility before `amplify init`.
A standalone script (`ensure-gen1-placeholder-apps`) is also added to
`amplify-e2e-tests` that provisions the placeholder across all org
accounts and regions.

#### Improved nexpect error output

When a spawned process exits with a non-zero code, the error now includes
the last 10 lines of terminal output. Previously only the exit code was
shown, making it hard to diagnose the underlying cause of `amplify init`
failures.

#### Support for running E2Es against a packaged CLI binary

Previously the gen2-migration runner hardcoded `getCLIPath(true)`, which
always resolved to the local source tree `amplify-cli/bin/amplify` script.
This meant E2Es would always run against the development codebase, even
when they weren't supposed to.

Now `getCLIPath` is called with `testingWithLatestCodebase` set based on
whether `AMPLIFY_PATH` is defined. When `AMPLIFY_PATH` points to a
packaged binary, the runner uses that binary instead. This also required
adding an `amplify pull` step before refactor — the generate step wipes
the amplify local directory on the gen1 branch, and when running against
a packaged binary this causes credentials resolution errors during
refactor. Pulling restores the local amplify directory so refactor can
resolve credentials correctly. (In source code mode this doesn't happen
because the credential resolution path differs.)

Related changes:

- New `pull()` method wraps `amplifyPullNonInteractive` with the correct
  appId and profile
- `amplifyPullNonInteractive` now accepts an optional `profile` parameter
  instead of always using the test profile constant

#### Other changes

- Removed leftover debug `console.log` calls from `initProjectHelper`
  and `cli.ts`
- Consistent log formatting in the gen2-migration app runner with timing
- Added `posix_spawnp` troubleshooting entry to `LOCAL_E2E_TESTING.md`

#### Issue #, if available

N/A

- Verified TypeScript compiles cleanly with no diagnostics
- All pre-commit hooks pass (prettier, eslint, commitlint)
- Manually tested gen2-migration E2E runner with the placeholder app
  creation flow

#### Checklist

- [x] PR description included
- [x] `yarn test` passes
- [x] Relevant documentation is changed or added (LOCAL_E2E_TESTING.md)

-----


packages/amplify-e2e-core/src/utils/nexpect.ts
```diff
@@ -439,7 +439,23 @@ function chain(context: Context): ExecutionContext {
...
+          : 'No output';
+        const output = [
+          '<><><><><><><><><><><><><><><>',
+          'Process Output (last 10 lines)',
+          '<><><><><><><><><><><><><><><>',
+          '',
+          exitLastScreen,
+          '',
+        ].join('\n');
+        return onError(new Error(`Process exited with non zero exit code ${code}\n\n${output}`), false);
```
iankhou, line 458: Nice, do you have an example of the output?

Would we expect to see, for example, the last prompt for an interactive CLI, or something else?

iankhou (approved):

PR #14857, opened 2026-05-03
fix(cli): native user pool client dropped during gen2-migration

iliapolo:
#### Description of changes

When the auth generator creates a `NativeAppClient` via `userPool.addClient()`, the new client was not being registered with the Cognito Identity Pool. This meant the identity pool didn't know about the native app client, which could cause authentication failures for native app users after migration.

This PR fixes that by always registering the native app client as a Cognito identity provider on the identity pool. The generated `applyEscapeHatches` function now includes a block that pushes the native client's ID and provider name into `cfnIdentityPool.cognitoIdentityProviders`.

#### Identity pool provider registration

The generated `applyEscapeHatches` now always includes:

```ts
const cognitoProviders = backend.auth.resources.cfnResources.cfnIdentityPool.cognitoIdentityProviders;
if (cognitoProviders && Array.isArray(cognitoProviders)) {
  cognitoProviders.push({
    clientId: nativeUserPoolClient.userPoolClientId,
    providerName: `cognito-idp.${backend.auth.stack.region}.amazonaws.com/${userPool.userPoolId}`,
  });
}
```

This ensures the identity pool recognizes the native app client for federated identity resolution.

#### Auth render options tightened

`webClient` and `nativeClient` are now required fields on `AuthRenderOptions` (previously optional). The old `userPoolClient` property was renamed to `nativeClient` for clarity. `fetchUserPoolClient` now throws `AmplifyError` instead of returning `undefined`, matching the fail-fast pattern used by `fetchUserPool`. The generator no longer conditionally fetches clients — both web and native clients are always fetched.

#### Removed `hasIdentityProviders` threading

The `hasIdentityProviders` boolean was passed through `render()` → `renderCdkImports()` → `renderApplyEscapeHatches()` → `buildEscapeHatchStatements()` → `buildAdditionalImports()`. This has been replaced by reading `nativeClient.SupportedIdentityProviders` directly where needed, reducing parameter threading.

#### E2E logging improvements

Log output from `App.runAmplify` and `App.runMigrationStep` now uses `path.basename(this.amplifyPath)` for shorter, more readable log lines. The format was standardized to `(→)` for start and `(✔)` for completion.

#### Snapshot updates

All migration app snapshots updated to include the new `cognitoIdentityProviders.push(...)` block. The store-locator app snapshots were re-recorded against a fresh test environment (new resource IDs throughout).

Unit tests updated — all `fetchUserPoolClient` mocks changed from returning `undefined` to returning `{}`, matching the new non-optional return type. Inline snapshots updated to include the new identity pool registration block.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

----


packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
```diff
@@ -991,17 +988,13 @@ export class AuthRenderer {
...
       statements.push(TS.assignProp('cfnUserPoolClient', 'allowedOAuthFlows', options.webClient.AllowedOAuthFlows));
     }
 
-    if (options.userPoolClient) {
-      statements.push(...this.buildUserPoolClientStatements(options.userPoolClient, hasIdentityProviders));
-    }
-
-    if (hasIdentityProviders) {
-      statements.push(...this.buildProviderSetupStatements());
+    if (options.nativeClient) {
```
9pace: This param is required. I guess this is just here because it existed for options.userPoolClient (which was actually optional)

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.generator.ts
```diff
@@ -46,16 +46,16 @@ export class AuthGenerator implements Planner {
     const userPool = await this.gen1App.aws.fetchUserPool(userPoolId);
 
     const appClientIdWeb = this.gen1App.resourceMetaOutput(this.resource, 'AppClientIDWeb');
-    const appClientId = this.gen1App.resourceMetaOutput(this.resource, 'AppClientID');
+    const appClientIdNative = this.gen1App.resourceMetaOutput(this.resource, 'AppClientID');
     const identityPoolId = this.gen1App.resourceMetaOutput(this.resource, 'IdentityPoolId');
```
9pace, line 50: Ig this would break if userpool only in gen 1?

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
9pace:
```suggestion
```

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.generator.ts
iliapolo: Yeap. Known issue. We don't support Gen1 apps with only a user pool.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
iliapolo: Fixed. Good catch.

9pace (approved):

PR #14858, opened 2026-05-03
feat(cli): concrete error reporting in `gen2-migration` commands

iliapolo:
#### Description of changes

Replace the generic `MigrationError` error type throughout the gen2-migration module with specific, descriptive error types. Additionally, record the gen2-migration subcommand name in usage data telemetry so that metrics distinguish between `assess`, `lock`, `generate` and `refactor` instead of reporting all of them as `core`.

#### Issue #, if available

N/A

- `yarn build` passes for both `amplify-cli` and `amplify-cli-core`
- All existing tests updated to pass the new `SpinningLogger` parameter and assert against the new specific error names

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

----


packages/amplify-cli-core/src/errors/amplify-exception.ts
```diff
@@ -216,6 +241,9 @@ export type AmplifyFaultType =
   | 'PullBackendFault'
   | 'PushResourcesFault'
   | 'ResourceAddFault'
+  | 'RegexMatchingFault'
+  | 'ResourceAddFault'
```
dgandhi62: Duplicate (2 lines above)

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
```diff
@@ -296,7 +296,7 @@ export class Gen1App {
 
     const response = await s3Client.send(new GetObjectCommand({ Key: zipKey, Bucket: bucket }));
     if (!response.Body) {
-      throw new AmplifyError('MigrationError', { message: 'S3 GetObject response body is empty' });
+      throw new AmplifyError('BucketNotFoundError', { message: 'S3 GetObject response body is empty' });
```
dgandhi62: Reaching this point means that the bucket is found right? The text indicates that the response body on the s3 is empty

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
```diff
@@ -115,15 +115,15 @@ export class Gen1App {
...
+      throw new AmplifyError('TpiEnvironmentNotFoundError', {
         message: `Environment ${envName} does not exist in ${tpiRelPath}`,
         resolution: `Checkout to the branch corresponding to environment ${envName} and rerun the command`,
       });
     }
 
     const cfnProvider = envInfo.awscloudformation;
     if (!cfnProvider?.StackName || !cfnProvider?.DeploymentBucketName) {
-      throw new AmplifyError('MigrationError', {
+      throw new AmplifyError('TpiEnvironmentNotFoundError', {
```
dgandhi62: The environment is found here, from what I understand

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
```diff
@@ -214,11 +214,11 @@ export class Gen1App {
   public singleResourceName(category: string, service: string): string {
     const categoryBlock = this.categoryMeta(category);
     if (!categoryBlock) {
-      throw new AmplifyError('MigrationError', { message: `Category '${category}' not found in amplify-meta.json` });
+      throw new AmplifyError('CategoryMetaNotFoundError', { message: `Category '${category}' not found in amplify-meta.json` });
```
dgandhi62, line 217: A lot of throws have resolution fields while a lot don't. Just a nit, not an issue

dgandhi62 (changes requested):
Questions

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
iliapolo: Yeah, we'll accept that...

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
iliapolo: Yes good catch. I'll change.

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
iliapolo: yeah, i'll change.

dgandhi62 (dismissed):

dgandhi62 (approved):

PR #14875, opened 2026-05-10
feat(cli): `gen2-migration refactor` supports cdk custom resources

### LINKED ISSUE #14545: (gen2-migration) `refactor` command should handle custom resources

_No response_

_No response_

Stateful resources in Gen1 custom resource stacks need to be moved to the Gen2 stack in the same exact way that normal categories are moved. 

The solution should:

### 1. Inspect the `backend-config.json` file to extract the names of the custom resources.

```json
"custom": {
  "processingQueue": {
    "providerPlugin": "awscloudformation",
    "service": "customCDK"
  },
  "publisher": {
    "providerPlugin": "awscloudformation",
    "service": "customCDK"
  }
},
```

The keys of the "custom" property correspond to the Gen1 nested stack name of each custom resource. The example above will create two nested stacks:

- `customprocessingQueue`
- `customPublisher`

These will also be the names of the nested stacks in Gen2 (we will code generate them this way).

### 2. Discover the stateful resources in each nested stack.

Since custom resource stacks can contain any CloudFormation resource, this is tricky because it requires we have an authoritative and exhaustive list of all possible stateful resources in CFN template. 

We can't afford to miss any resource here so we need to prompt the customer and ask them for this information.

**Proposal:**

Prompt the user with a list of all resources in the stack. Something like:

```console
Your "customprocessingQueue" stack contains the following resources. Please mark the stateful ones you want to refactor:

<> `AWS::SQS::Queue (https://sqs.us-east-1.amazonaws.com/185706627232/sqs-queue-projectboards-main)`
<> `AWS::IAM::Policy (my-policy)`
```

The customer would then select the SQS queue and we will store its logical ID in memory.

> Note: To make this process a bit easier, we can use our [stateful-resources list](https://github.com/aws-amplify/amplify-cli/blob/gen2-migration/packages/amplify-cli/src/commands/gen2-migration/stateful-resources.ts) to do initial discovery; but we still need the customer prompt for confirmation.

### 3. Discover the expected Gen2 logical IDs.

Now we need to figure out which Gen2 resources we need to map to. We should use the following heuristic:

- If there is only a single resource of the expected type in the Gen2 stack - assume that is the correct resource.
- If there are multiple resources (for example 2 SQS queus) - prompt the customer again and ask for the correct Gen2 logical ID for each resource.

### 4. Proceed with standard refactor flow

At this point we should have all the information we need:

- The nested stack name.
- The Gen1 logical IDs of all stateful resources in each stack.
- The Gen2 logical IDs of those same resources.

We should be able to just reuse the existing refactor logic.

None

### Additional context

_No response_

iliapolo:
#### Description of changes

This PR adds end-to-end support for migrating custom CDK resources during the gen2-migration refactor step. Previously, `custom:customCDK` resources were skipped entirely — now they are handled with both forward and rollback paths.

For known categories (auth, storage, analytics), rollback mapping is deterministic because Gen1 logical IDs are static and known. Custom CDK resources don't have this property — users define arbitrary resources with arbitrary logical IDs, and a stack can contain multiple resources of the same type (e.g. two SQS queues). During forward, when multiple candidates match, the user is prompted to disambiguate. During rollback, we don't want to ask again — so the forward step now records its resource mappings in the holding stack's template metadata, and rollback reads them back to reverse the operation.

Since this mechanism is generic, it replaces the per-category `targetLogicalId` overrides that previously existed in every rollback class — those were just hardcoded versions of what the metadata now provides automatically.

#### Custom CDK Refactoring

New `CustomCDKForwardRefactorer` and `CustomCDKRollbackRefactorer` move stateful resources between Gen1 custom CDK stacks and Gen2. When multiple Gen2 candidates match a Gen1 resource type, the user is prompted to disambiguate using physical resource IDs. The refactorers use `gen1App.statefulResourceTypes` to determine which resource types to move — since custom stacks can contain any combination of stateful resources.

#### Dynamic Stateful Resource Types

The static `STATEFUL_RESOURCES` set is replaced by a `statefulResourceTypes` instance field on `Gen1App`, seeded from defaults and extensible via a new `--additional-stateful-resource-types <path>` CLI flag that accepts a JSON file. All consumers (lock, refactorers, generate) now read from this field instead of importing a static set.

#### CloudControl-Based GetAtt Resolution

`cfn-output-resolver.ts` now resolves `Fn::GetAtt` references at runtime using CloudControl `GetResource` instead of building ARNs from a hardcoded service-type map. This removes the brittle `buildArn()` helper. The tree walker is now async to support this.

#### Generated Resource Wrapper with Retention Policies

`custom.generator.ts` emits a `resource.ts` wrapper that applies `DeletionPolicy: Retain` and `UpdateReplacePolicy: Retain` to all stateful CfnResources in the construct. Stack naming changed to `custom${resourceName}` to match the nested stack convention used by refactor.

#### Simplified Rollback Mapping

The `targetLogicalId` abstract method is removed from `RollbackCategoryRefactorer`. Previously each rollback subclass had to hardcode the Gen2→Gen1 logical ID mapping (e.g. "UserPool maps to UserPool", "KinesisStream maps to KinesisStream"). Now the base class reads the forward mappings from the holding stack metadata (described above) and reverses them. This eliminates per-category rollback mapping code in analytics, auth, storage-s3, and storage-dynamo rollback classes.

#### Issue #, if available

Resolves https://github.com/aws-amplify/amplify-cli/issues/14545

- `npx jest --testPathPattern=gen2-migration` — 48 suites, 411 tests pass
- Snapshot comparison across all migration apps (finance-tracker exercises the custom CDK path)
- New `customfinance.ts` backend with SQS queues and Lambda exercises the custom CDK refactoring end-to-end

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

----


packages/amplify-cli/src/commands/gen2-migration/refactor/custom/custom-cdk-forward.ts
```diff
@@ -0,0 +1,64 @@
...
+      const physicalId = gen2Resources.find((r) => r.LogicalResourceId === logicalId)?.PhysicalResourceId;
+      if (!physicalId) {
+        throw new AmplifyError('MigrationError', {
+          message: `Unable to find physical ID of resource ${logicalId} in stack ${extractStackNameFromId(gen2StackId)}`,
+        });
+      }
+      return physicalId;
+    }
+
+    const choices = Array.from(targetResources.keys()).map((r) => `${r}${CHOICE_SEPERATOR}(${gen2PhysicalResourceId(r)})`);
```
iliapolo: Should build selection from `candidates`, not from the `targetResources`.

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
```diff
@@ -129,8 +136,12 @@ export class Gen1App {
       });
     }
 
+    const additionalStatefulResourceTypes = additionalStatefulResourceTypesPath
+      ? JSON.parse(await fs.readFile(additionalStatefulResourceTypesPath, { encoding: 'utf-8' }))
```
iliapolo, line 140: Validate the result of JSON.parse is an array + add some unit tests.

iliapolo: Added

packages/amplify-cli/src/commands/gen2-migration/refactor/custom/custom-cdk-forward.ts
iliapolo: Fixed

9pace (dismissed):

9pace (approved):

PR #14892, opened 2026-05-18
fix(cli-internal): use consistent model name casing for DynamoDB table references

### LINKED ISSUE #14561: (gen2-migration) `generate` command creates incorrect env variable declaration for a function that accesses a non pascal cased model name table

npm install --no-save @aws-amplify/cli-internal-gen2-migration-experimental-alpha

_No response_

### Amplify CLI Version

0.6.0

Mac

No

### Describe the bug

When a model is named in a non pascal case fashion, i.e:

```graphql
type randomItem @model {...}
```

And a function is configured to have access to this model table, the generated (snippet) code is incorrect:

```ts
backend.quotegenerator.addEnvironment("API_BACKENDONLY_RANDOMITEMTABLE_NAME", backend.data.resources.tables["Randomitem"].tableName);
backend.data.resources.tables["randomItem"].grant(backend.quotegenerator.resources.lambda, "dynamodb:Put*", "dynamodb:Create*", "dynamodb:BatchWriteItem", "dynamodb:PartiQLInsert", "dynamodb:Get*", "dynamodb:BatchGetItem", "dynamodb:List*", "dynamodb:Describe*", "dynamodb:Scan", "dynamodb:Query", "dynamodb:PartiQLSelect", "dynamodb:Update*", "dynamodb:RestoreTable*", "dynamodb:PartiQLUpdate", "dynamodb:Delete*", "dynamodb:PartiQLDelete");
```

Notice the value of the env variable is `backend.data.resources.tables["Randomitem"].tableName`, but our model is named `randomitem`. Also notice that the next line does actually use the **correct casing**:

```ts
backend.data.resources.tables["randomItem"].grant(...)
```

Deploying this code results in:

```console
2026-01-28T22:48:01.995Z [INFO]: [BackendBuildError] Unable to deploy due to CDK Assembly Error
                                 ∟ Caused by: [AssemblyError] Assembly builder failed
                                 ∟ Caused by: [TypeError] Cannot read properties of undefined (reading 'tableArn')
```

### Expected behavior

```ts
backend.quotegenerator.addEnvironment("API_BACKENDONLY_RANDOMITEMTABLE_NAME", backend.data.resources.tables["randomitem"].tableName);
backend.data.resources.tables["randomItem"].grant(backend.quotegenerator.resources.lambda, "dynamodb:Put*", "dynamodb:Create*", "dynamodb:BatchWriteItem", "dynamodb:PartiQLInsert", "dynamodb:Get*", "dynamodb:BatchGetItem", "dynamodb:List*", "dynamodb:Describe*", "dynamodb:Scan", "dynamodb:Query", "dynamodb:PartiQLSelect", "dynamodb:Update*", "dynamodb:RestoreTable*", "dynamodb:PartiQLUpdate", "dynamodb:Delete*", "dynamodb:PartiQLDelete");
```

### Reproduction steps

1. Add a non pascal named model name to your GraphQL schema:

```graphql
type randomItem @model @auth(rules: [
  { allow: public, operations: [read] },
  { allow: owner, operations: [create, read, update, delete] }
]) {
  id: ID!
  name: String!
  description: String
  images: [String]
  projectID: ID
}
```

2. Configure a function to have access to this model:
3. Run `gen2-migration generate`
4. Deploy the generated code.

### Project Identifier

_No response_

### Log output

<details>

```


```

</details>

### Additional information

_No response_

adrianjoshua-strutt:
## Description

When a GraphQL model uses non-PascalCase naming (e.g., `randomItem`), the generated code previously used naive capitalization to reconstruct the model name from the uppercase environment variable (`API_MYAPI_RANDOMITEMTABLE_ARN` → `Randomitem`). This caused deployment failures because `backend.data.resources.tables` is keyed by the original model name from the schema.

## Root Cause

The `extractTableName` function applied `charAt(0).toUpperCase() + slice(1).toLowerCase()` to the captured segment from the env var name. Since Gen1 uppercases model names in env vars, this transformation cannot reconstruct names like `randomItem` or `MealPlan` — it produces `Randomitem` and `Mealplan` instead.

## Fix

The fix reads model names from the GraphQL schema (using the same regex pattern already used by `DataGenerator`) and performs a case-insensitive lookup to recover the original casing. Both the environment variable escape hatches and the table grant statements now consistently use the correct model name.

The lookup is cached per `FunctionGenerator` instance to avoid redundant schema reads.

## Testing

- Added a test exercising `randomItem` and `MealPlan` models to verify correct casing in generated output
- All 242 existing tests in `gen2-migration/generate` continue to pass

Fixes #14561


soberm (dismissed):

sarayev (dismissed):

sarayev (dismissed):

sharonyajain (dismissed):

soberm (dismissed):

soberm (dismissed):

sharonyajain (dismissed):

soberm (dismissed):

sharonyajain (dismissed):

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
```diff
@@ -447,3 +461,95 @@ function resolveAuthAccess(cognitoActions: string[]): { permissions: AuthPermiss
...
+  const buildSchemaPath = path.join('api', apiName, 'build', 'schema.graphql');
+  try {
+    const buildSchema = gen1App.file(buildSchemaPath);
+    const connectionRegex = /type\s+Model(\w+)Connection\b/g;
+    const names: string[] = [];
+    let match: RegExpExecArray | null;
+    while ((match = connectionRegex.exec(buildSchema)) !== null) {
+      names.push(match[1]);
+    }
+    if (names.length > 0) return names;
```
sarayev: Nice approach reading names from the schema instead of guessing. One gap though: we bail out with `if (names.length > 0) return names;` as soon as the `ModelXConnection` regex finds anything. But the transformer only emits a `Connection` type for models that have a list query, so a model declared `@model(queries: null)` won't show up here. Since one match is enough to short-circuit, that model quietly falls through to the naive-casing path, which is the exact bug we're fixing. Could we merge the build-schema names with the raw-schema fallback instead of returning early, so nothing gets missed?

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
```diff
@@ -513,11 +517,25 @@ function nonNull(expr: ts.Expression): ts.Expression {
...
+ * When model names are provided, performs a case-insensitive lookup to
+ * recover the original casing (e.g., `RANDOMITEM` → `randomItem`).
+ * Falls back to capitalizing the first letter when no match is found.
+ *
+ * @example
+ * extractTableName('API_MYAPI_RANDOMITEMTABLE_ARN', ['randomItem', 'Meal']) // → 'randomItem'
+ * extractTableName('API_MYAPI_MEALTABLE_ARN', ['randomItem', 'Meal'])       // → 'Meal'
+ */
+export function extractTableName(envVar: string, modelNames: readonly string[] = []): string | undefined {
   const match = envVar.match(/API_.*_(.+?)TABLE_/);
```
sarayev: Two small things on the extraction here. The regex `/API_.*_(.+?)TABLE_/` is greedy on the front, so for a model name with an underscore (like `API_MYAPI_MY_MODELTABLE_ARN`) the `.*` eats through `MY_` and we capture `MODEL` instead of `MY_MODEL`. Anchoring to the last segment before `TABLE_` would be safer, something like `/_([A-Za-z0-9]+)TABLE_(?:ARN|NAME)$/`. Also, `find()` returns the first case-insensitive match, so if two models collide when uppercased (say `randomItem` and `randomitem`) we'd resolve to whichever comes first in schema order, which isn't necessarily the one the env var meant. Unlikely in a real schema, but might be worth a quick comment noting the assumption?

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
```diff
@@ -447,3 +461,95 @@ function resolveAuthAccess(cognitoActions: string[]): { permissions: AuthPermiss
...
+    for (let i = 0; i < afterName.length; i++) {
+      const ch = afterName[i];
+      if (ch === '(') {
+        depth++;
+      } else if (ch === ')') {
+        depth--;
+      } else if (depth === 0 && ch === '{') {
+        break;
+      }
+      if (depth === 0 && afterName.slice(i).startsWith('@model')) {
```
sarayev: `startsWith('@model')` will also match things like `@models` or `@modelFoo`. A word boundary would avoid the false positive, e.g. `/^@model(?![A-Za-z0-9])/.test(afterName.slice(i))`. Minor, but cheap to harden.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
```diff
@@ -1,5 +1,7 @@
 import path from 'node:path';
 import fs from 'node:fs/promises';
+import { existsSync } from 'node:fs';
+import { globSync } from 'glob';
```
sarayev, line 4: Is `glob` a declared runtime dependency of this package, or is it coming in transitively? If it's only hoisted, this can pass locally and then break the build. Worth a quick check on the package.json.

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/function/function.generator.test.ts
```diff
@@ -1264,4 +1264,312 @@ describe('FunctionGenerator', () => {
...
+    });
+    jest.spyOn(gen1App.aws, 'fetchFunctionSchedule').mockResolvedValue(undefined);
+
+    const generator = createFunctionGenerator({ gen1App, backendGenerator, packageJsonGenerator, outputDir });
+    const ops = await generator.plan();
+    await ops[0].execute();
+
+    const resourceTs = writtenFile('resource.ts');
+    // Without model names, falls back to naive capitalization
+    expect(resourceTs).toContain('Some');
```
sarayev: This `toContain('Some')` check is a bit loose, `Some` is a short substring that could show up by accident and doesn't really prove we hit the naive-fallback path. Asserting the exact generated reference (like the full `tables['Some']` expression) would make this test mean what it says.

packages/amplify-cli/adr/006-social-auth.md
```diff
@@ -1,346 +0,0 @@
-# ADR-005: Social Auth IDP & Domain Handling During Refactor
```
sarayev, line 1: This 346-line ADR deletion looks unrelated to the casing fix and isn't mentioned in the description, looks like it rode in on a base-branch merge. Just flagging so it doesn't get attributed here, could we rebase it out to keep the diff focused, or is it intentional?

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
sharonyajain: Thank you for pointing this out. A @model(queries: null) model has no ModelXConnection in the build schema, so the early return would skip it. Reworked to merge both sources into a Set instead of short-circuiting: build schema first, then union the raw-schema names, so nothing is missed and duplicates collapse. Added a regression test (recovers a @model(queries: null) model missing from the build schema by merging the raw schema).

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
sharonyajain: Both valid. Since we already have the model names, I switched the primary path to match each known name directly against the env var (_<MODEL>TABLE_), which sidesteps the greedy-prefix problem entirely and correctly handles model names containing underscores. 
The fallback regex is now anchored to the last segment:` /_([A-Za-z0-9]+)TABLE_(?:ARN|NAME)$/.`
Added unit tests in function.renderer.test.ts covering the underscore case, suffix-overlap (Item/LineItem), and the fallback.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
sharonyajain: Checked. glob is a declared runtime dependency of amplify-cli ("glob": "^11.0.1" in packages/amplify-cli/package.json), not just hoisted, so the build is safe.

packages/amplify-cli/adr/006-social-auth.md
sharonyajain: The 346-line delete of packages/amplify-cli/adr/006-social-auth.md comes from commit e7dd867 ("chore(cli): remove duplicate social-auth ADR (#14951)") — it's genuinely in this branch's diff and not yet on dev. 

If #14951 merges to dev on its own, the deletion drops out of this diff automatically. If not, we can rebase/drop it to keep this PR focused on the casing fix. It looks like intentional cleanup that rode in via a base merge rather than anything belonging to this PR.

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/function/function.generator.test.ts
sharonyajain: Tightened to assert the exact reference: expect(resourceTs).toContain("backend.data.resources.tables['Some']")

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
sharonyajain: Hardened to /^@model(?![A-Za-z0-9])/.test(...) so @models/@modelFoo no longer match.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
```diff
@@ -447,3 +461,107 @@ function resolveAuthAccess(cognitoActions: string[]): { permissions: AuthPermiss
...
+    const afterName = schema.slice(match.index + match[0].length);
+    let depth = 0;
+    let foundModel = false;
+    for (let i = 0; i < afterName.length; i++) {
+      const ch = afterName[i];
+      if (ch === '(') {
+        depth++;
+      } else if (ch === ')') {
+        depth--;
+      } else if (depth === 0 && ch === '{') {
```
sarayev, line 570: Small edge case here: the inner scan breaks on `{` at depth 0, which correctly stops at the type body. But if a `type X` had no `@model` and no `{` body before the next `type` or EOF, the scan would keep walking `afterName` into the following type looking for `@model`. Every real GraphQL type has a `{...}` body so the brace break fires first, so this is theoretical, but would it be worth bounding the search to the current type (slice up to the next `\btype\s` or first `{`) to make it airtight?

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
```diff
@@ -447,3 +461,107 @@ function resolveAuthAccess(cognitoActions: string[]): { permissions: AuthPermiss
...
+  return [...names];
+}
+
+/**
+ * Collects all user-authored GraphQL schema content from either the single
+ * `schema.graphql` file or the `schema/` directory (multi-file pattern).
+ */
+function collectUserSchema(gen1App: Gen1App, apiName: string): string {
+  const schemaFilePath = path.join('api', apiName, 'schema.graphql');
+  if (gen1App.fileExists(schemaFilePath)) {
```
sarayev, line 532: `collectUserSchema` reads the single `schema.graphql` OR the `schema/` directory, but never both. If an app happened to have both, directory models would be missed. Is single-file-XOR-directory a guaranteed invariant for Gen1 apps here? If so a one-line comment noting it would settle it; if not, unioning both (like `readSchemaModelNames` now unions its two sources) would be safer.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
```diff
@@ -447,3 +461,107 @@ function resolveAuthAccess(cognitoActions: string[]): { permissions: AuthPermiss
...
+  // Source 1: build/schema.graphql (transformer standardised output).
+  try {
+    const buildSchema = gen1App.file(path.join('api', apiName, 'build', 'schema.graphql'));
+    const connectionRegex = /type\s+Model(\w+)Connection\b/g;
+    let match: RegExpExecArray | null;
+    while ((match = connectionRegex.exec(buildSchema)) !== null) {
+      names.add(match[1]);
+    }
+  } catch {
+    // build schema not available — rely on the raw schema below
```
sarayev, line 501: Nit: these two `catch {}` blocks swallow errors silently (intentional, since a missing schema is normal). A `logger.debug(...)` in each would make it much easier to diagnose a miscased table down the line, the class already has a `SpinningLogger`. Totally optional.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
sharonyajain: Good catch — bounded it. The scan now slices `afterName` up to the next top-level `type` keyword before walking, so it can never cross a type boundary even in the bodyless-type case you describe. The brace break still fires first for every well-formed type, so this is defense-in-depth for an unreachable case (a bodyless `type` is invalid GraphQL and wouldn't compile in Gen1). Added a guard test that a non-model `type Query` preceding real models is never attributed as a table. Pushed in 06258ef.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
sharonyajain: Single-file-XOR-directory is a guaranteed Gen1 invariant: `amplify add api` scaffolds a single `schema.graphql`, and the multi-file `schema/` directory is the alternate layout — the Gen1 transformer's own schema loader reads one or the other (directory preferred), never a concatenation of both. So mirroring that XOR here is correct; unioning would actually diverge from Gen1 transformer behaviour. Added a comment documenting the invariant (your suggested resolution). Pushed in 06258ef.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
sharonyajain: Done — added `printer.debug(...)` in both catch blocks (with the api name and the error), so a miscased table is diagnosable with `--debug`. Used `printer` from `@aws-amplify/amplify-prompts` directly since these are free functions with no `this.logger`; it's the same debug channel `SpinningLogger` delegates to and respects the same `isDebug` global. Pushed in 06258ef.

sarayev (approved):

PR #14894, opened 2026-05-18
fix(cli-internal): add lambda authorizer import and use dynamic region in data resource

### LINKED ISSUE #14813: (gen2-migration): `generate` has bug in lambda authorizer for additional auth mode

_No response_

_No response_

### Amplify CLI Version

na

na

na

### Description

When migrating an Amplify Gen1 app that uses a Lambda authorizer as an additional auth mode on a GraphQL API, the generated Gen2 `amplify/data/resource.ts` references the Lambda function in `lambdaAuthorizationMode` without importing it. This results in a build failure.

### Gen1 State

The Gen1 app had a GraphQL API (`AppSync`) configured with:
- **Default auth**: API Key (7-day expiry)
- **Additional auth**: AWS Lambda authorizer (`graphQlLambdaAuthorizerd351d098`) with a 300s TTL

This was defined across multiple Gen1 config files:

**`amplify/backend/api/lambdaauthconfig/cli-inputs.json`**:
```json
{
  "additionalAuthTypes": [
    {
      "mode": "AWS_LAMBDA",
      "lambdaFunction": "graphQlLambdaAuthorizerd351d098",
      "ttlSeconds": "300"
    }
  ]
}
```

**`amplify/backend/backend-config.json`** linked the API to the function and both resources deployed correctly.

The GraphQL schema used both auth modes:
```graphql
type Todo @model @auth(rules: [
  { allow: public, provider: apiKey },
  { allow: custom }
]) {
  id: ID!
  name: String!
  description: String
}
```

### Issue in Generated Gen2 Code

The generated `amplify/data/resource.ts` references `graphQlLambdaAuthorizerd351d098` in the `lambdaAuthorizationMode` config but does not include an import statement for it:

```typescript
import { defineData } from '@aws-amplify/backend';

const schema = `...`;

  authorizationModes: {
    defaultAuthorizationMode: 'apiKey',
    apiKeyAuthorizationMode: { expiresInDays: 7, description: 'sdfgadfv' },
    lambdaAuthorizationMode: {
      function: graphQlLambdaAuthorizerd351d098, // ❌ undefined — no import
      timeToLiveInSeconds: 300,
    },
  },
  schema,
});
```

The function is correctly defined in `amplify/function/graphQlLambdaAuthorizerd351d098/resource.ts` and correctly imported in `amplify/backend.ts`, but the import is missing in `data/resource.ts`.

### Expected Behavior

The migration should generate `data/resource.ts` with the necessary import:

```typescript
import { defineData } from '@aws-amplify/backend';
import { graphQlLambdaAuthorizerd351d098 } from '../function/graphQlLambdaAuthorizerd351d098/resource';
```

### Impact

- The Gen2 app will fail to build/deploy
- The Lambda authorizer auth mode is non-functional
- The `@auth(rules: [{ allow: custom }])` rule on the Todo model cannot be enforced

### Steps to Reproduce

1. Create a Gen1 Amplify app with a GraphQL API
2. Add API Key as default auth and Lambda as an additional auth type
3. Run the Gen1 to Gen2 migration
4. Observe that `amplify/data/resource.ts` references the Lambda function without importing it

---

### Description

When migrating an Amplify Gen1 app to Gen2, the generated function `resource.ts` files hardcode the `REGION` environment variable to a specific region string (`us-east-1`) instead of resolving it dynamically at deploy time. This causes incorrect behavior if the app is deployed to a different region.

### Gen1 State

In Gen1, both Lambda functions had their `REGION` environment variable set dynamically via CloudFormation intrinsic functions:

**`graphQlLambdaAuthorizerd351d098-cloudformation-template.json`**:
```json
"Environment": {
  "Variables": {
    "ENV": { "Ref": "env" },
    "REGION": { "Ref": "AWS::Region" }
  }
}
```

`AWS::Region` is a CloudFormation pseudo-parameter that resolves to whatever region the stack is deployed in. This ensured the function always had the correct region regardless of deployment target.

### Issue in Generated Gen2 Code

Both generated function `resource.ts` files hardcode the region:

**`amplify/function/graphQlLambdaAuthorizerd351d098/resource.ts`**:
```typescript
export const graphQlLambdaAuthorizerd351d098 = defineFunction({
  environment: { ENV: `${branchName}`, REGION: 'us-east-1' }, // ❌ hardcoded
});
```

**`amplify/function/recurring/resource.ts`**:
```typescript
export const recurring = defineFunction({
  environment: { ENV: `${branchName}`, REGION: 'us-east-1' }, // ❌ hardcoded
});
```

### Expected Behavior

The migration should either:
1. Use the CDK `Stack.of(this).region` to resolve the region dynamically, or
2. Omit the `REGION` variable entirely since `process.env.AWS_REGION` is automatically available in all Lambda execution environments

### Impact

- The Lambda authorizer uses `process.env.AWS_REGION` to construct ARNs for `deniedFields`. If `REGION` is hardcoded incorrectly, the `AWS_REGION` env var still works, but the custom `REGION` env var would be misleading if any code references it.
- Deploying to any region other than `us-east-1` would result in an incorrect `REGION` environment variable.

### Steps to Reproduce

1. Create a Gen1 Amplify app with Lambda functions in any region
2. Run the Gen1 to Gen2 migration
3. Observe that `REGION` is hardcoded to `us-east-1` in the generated `resource.ts` files

### Expected behavior

na

### Reproduction steps

na

### Project Identifier

_No response_

### Log output

<details>

```


```

</details>

### Additional information

_No response_

adrianjoshua-strutt:
## Description

Fixes #14813.

Two bugs in the gen2-migration code generator that caused build failures when migrating Gen1 apps with Lambda authorizer auth modes:

1. **Missing import for Lambda authorizer function** — The generated `data/resource.ts` referenced the Lambda function identifier (e.g., `myAuthFn`) without emitting an import statement, producing a TypeScript compilation error.

2. **Hardcoded REGION environment variable** — The `REGION` env var was placed in `defineFunction()`'s static `environment` object with the Gen1 deployment region (e.g., `us-east-1`). This is incorrect for Gen2 where the deployment region may differ from the original Gen1 region.

#### Data resource generation

`DataRenderer` now accepts an optional `lambdaAuthFunction` reference describing the function name and relative import path. When present, it adds the import to the generated `data/resource.ts`.

`DataGenerator` extracts the Lambda function name from `authorizationModes` (searching both `defaultAuthentication` and `additionalAuthenticationProviders`) and passes it to the renderer.

#### Dynamic REGION resolution

`FunctionGenerator.resolve()` now intercepts the `REGION` key after `classifyEnvVars()` runs. Instead of leaving it as a retained (hardcoded) env var, it moves REGION to an escape hatch that emits `backend.{funcName}.stack.region` — the CDK token that resolves to the actual deployment region at synth time.

## Testing

- Unit tests: `data.renderer.test.ts` updated with Lambda auth import assertion
- Integration snapshots: All 9 fixture apps regenerated and verified (12 snapshot tests pass)
- Function generator tests: 19 tests pass
- Full test suite: 55 tests across 5 suites pass

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/amplify/data/data.generator.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/amplify/data/data.generator.ts
@@ -9,7 +9,7 @@ import { AmplifyMigrationOperation } from '../../../_common/operation';
 import { BackendGenerator } from '../backend.generator';
 import { Gen1App, DiscoveredResource } from '../../../_common/gen1-app';
 import { TS } from '../../ts';
-import { DataRenderer } from './data.renderer';
+import { DataRenderer, LambdaAuthFunctionRef } from './data.renderer';
 import { SpinningLogger } from '../../../_common/spinning-logger';
 
 // ── Resolver Utility Types ─────────────────────────────────────────────
@@ -170,6 +170,7 @@ export class DataGenerator implements Planner {
       graphqlApi.additionalAuthenticationProviders !== undefined && graphqlApi.additionalAuthenticationProviders.length > 0;
     const hasAuth = this.gen1App.categoryMeta('auth') !== undefined;
     const authorizationModes = supplementOidcConfig(this.gen1App.resourceMetaOutput(this.resource, 'authConfig'), graphqlApi);
+    const lambdaAuthFunction = extractLambdaAuthFunction(authorizationModes);
     const hasIamAuth = this.detectIamAuth(authorizationModes, graphqlApi);
     const vtlFiles = this.findResolverVtlFiles(this.resource.resourceName);
     const hasResolvers = vtlFiles.length > 0;
@@ -191,6 +192,7 @@ export class DataGenerator implements Planner {
             hasAuth,
             apiId,
             classifiedResolvers,
+            lambdaAuthFunction,
           });
 
           const content = TS.printNodes(nodes);
@@ -313,4 +315,44 @@ function supplementOidcConfig(authConfig: any, graphqlApi: GraphqlApi): any {
   }
 
   return result;
+}
+
+/**
+ * Extracts the Lambda authorizer function reference from the authConfig.
+ * Searches both defaultAuthentication and additionalAuthenticationProviders.
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- untyped authConfig from amplify-meta.json
+function extractLambdaAuthFunction(authorizationModes: any): LambdaAuthFunctionRef | undefined {
+  if (!authorizationModes) return undefined;
+
+  const lambdaFunctionName =
+    findLambdaFunctionName(authorizationModes.defaultAuthentication) ??
+    findLambdaFunctionNameInProviders(authorizationModes.additionalAuthenticationProviders);
+
+  if (!lambdaFunctionName) return undefined;
+
+  return {
+    name: lambdaFunctionName,
+    importPath: `../function/${lambdaFunctionName}/resource`,
+  };
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- untyped authConfig provider
+function findLambdaFunctionName(provider: any): string | undefined {
+  if (provider?.authenticationType === 'AWS_LAMBDA' && provider.lambdaAuthorizerConfig?.lambdaFunction) {
+    return provider.lambdaAuthorizerConfig.lambdaFunction;
+  }
+  return undefined;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- untyped authConfig providers array
+function findLambdaFunctionNameInProviders(providers: any[] | undefined): string | undefined {
+  if (!providers) return undefined;
+  // AppSync supports only one Lambda authorizer per API, but we iterate defensively
+  // in case the config structure has multiple providers listed
+  for (const provider of providers) {
+    const name = findLambdaFunctionName(provider);
+    if (name) return name;
+  }
+  return undefined;
 }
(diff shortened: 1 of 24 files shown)


packages/amplify-cli/src/commands/gen2-migration/generate/amplify/data/data.generator.ts
```diff
@@ -275,3 +277,41 @@ export class DataGenerator implements Planner {
...
+  if (provider?.authenticationType === 'AWS_LAMBDA' && provider.lambdaAuthorizerConfig?.lambdaFunction) {
+    return provider.lambdaAuthorizerConfig.lambdaFunction;
+  }
+  return undefined;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any -- untyped authConfig providers array
+function findLambdaFunctionNameInProviders(providers: any[] | undefined): string | undefined {
+  if (!providers) return undefined;
+  for (const provider of providers) {
```
sarayev, line 353: question: Is it possible to have multiple Lambda providers?

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
```diff
@@ -78,6 +78,19 @@ export class FunctionGenerator implements Planner {
...
     const { literalEnvVars, dynamicEnvVars } = classifyEnvVars(config.Environment?.Variables ?? {});
 
+    // REGION should resolve dynamically to the deployment region rather than
+    // being hardcoded. Move it from retained env vars to an escape hatch that
+    // uses the CDK stack region token.
+    const finalLiteralEnvVars = { ...literalEnvVars };
+    const finalDynamicEnvVars = [...dynamicEnvVars];
+    if ('REGION' in finalLiteralEnvVars) {
+      delete finalLiteralEnvVars.REGION;
+      finalDynamicEnvVars.push({
```
sarayev, line 88: Can we have tests for this?

adrianjoshua-strutt: I added region tests

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/data/data.generator.ts
adrianjoshua-strutt: AppSync supports only one Lambda authorizer per API, but we iterate defensively in case the config structure has multiple providers listed

PR #14898, opened 2026-05-26
fix(e2e): disable self-registration in Cognito User Pools created by e2e tests

adrianjoshua-strutt:
## Description

All e2e test auth helper functions (`addAuthWithDefault`, `addAuthWithDefaultSocial`, `addAuthWithGroupTrigger`, etc.) now write an `override.ts` file that sets `AdminCreateUserConfig.AllowAdminCreateUserOnly = true` on the User Pool.

## Motivation

This addresses a security finding (Mirador: `palisade.udd.cognito.userpool.selfregistration_enabled`) where Cognito User Pools created by e2e/canary tests allow public self-sign-up. Since all test user creation already uses `AdminCreateUserCommand` (never the SignUp API), disabling self-registration does not break any existing test functionality.

## Changes

- Added `addAuthOverrideToDisableSelfSignup(cwd)` helper function to `packages/amplify-e2e-core/src/categories/auth.ts`
- Called this helper at the end of every `addAuth*` function that creates a new auth resource
- The override uses the existing Amplify CLI override mechanism (typed `override.ts` written to the auth resource directory)
- `buildOverrideDir` automatically picks up the override during `amplify push`

## Impact

- Tests that set up their own custom overrides (e.g., `auth_6.test.ts`) will simply overwrite this file with their own content — no conflict
- The `addAuthOverrideToDisableSelfSignup` function is exported so tests adding auth through non-standard mechanisms can also call it explicitly
- No changes to CLI product behavior — only the e2e test helpers are affected

## Testing

- TypeScript compilation verified (no new errors in `auth.ts`)
- All changes are in the e2e test helper layer (`amplify-e2e-core`), so existing e2e test suites validate correctness

### DIFF
--- a/packages/amplify-e2e-core/src/categories/auth.ts
+++ b/packages/amplify-e2e-core/src/categories/auth.ts
@@ -1,4 +1,6 @@
 import _ from 'lodash';
+import * as fs from 'fs-extra';
+import * as path from 'path';
 import { getCLIPath, getSocialProviders, KEY_DOWN_ARROW, KEY_UP_ARROW, nspawn as spawn, setTransformerVersionFlag } from '..';
 
 export type AddAuthUserPoolOnlyNoOAuthSettings = {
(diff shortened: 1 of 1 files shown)


sarayev (approved):
Have we validated that none of these tests need self-registration be enabled (any tests that test signUp functionality)?

adrianjoshua-strutt:
@sarayev Running the e2e tests should reveal this. But yes, some tests will need self registration.

sarayev (approved):

PR #14900, opened 2026-06-02
fix(cli): handle unsupported S3 acceleration in migration

### LINKED ISSUE #14876: Gen1 to Gen2 migration fails with MethodNotAllowed error for S3 Transfer Acceleration in unsupported regions (eu-north-1)

npm

v20.10.0

### Amplify CLI Version

14.4.0

Windows 11

Some Lambda resources had previously been modified manually in AWS Console:
- Lambda runtimes were updated to nodejs22.x
- Lambda execution roles had inline permissions added for Secrets Manager
- WebsocketMessageForwarder also had execute-api:ManageConnections permission

Before retrying migration, I updated the local Amplify Gen 1 function CloudFormation templates/custom-policies.json files and ran amplify push so the runtime/policy drift was mostly reconciled.

The current generate failure appears unrelated to those changes and occurs during S3 storage planning.

### Describe the bug

I am migrating an Amplify Gen 1 app to Amplify Gen 2 using the official gen2-migration commands.

`amplify gen2-migration lock --skip-validations` succeeds, but `amplify gen2-migration generate` fails during the planning phase with:

"The specified method is not allowed against this resource."

Running with `--debug` shows the error comes from the S3 storage generator:

- AwsFetcher.fetchBucketAccelerate
- S3Generator.plan
- gen2-migration/generate/amplify/storage/s3.generator.js

The generator appears to call `GetBucketAccelerateConfiguration` on an S3 bucket in eu-north-1. That API returns `MethodNotAllowed` for this bucket/region.

The same error can be reproduced directly with AWS CLI:

`aws s3api get-bucket-accelerate-configuration --bucket tiemerkintalaite-storage-for-vehicle111901-dev --region eu-north-1`

This returns:

`An error occurred (MethodNotAllowed) when calling the GetBucketAccelerateConfiguration operation: The specified method is not allowed against this resource.`

Trying to set acceleration to suspended also fails:

`aws s3api put-bucket-accelerate-configuration --bucket tiemerkintalaite-storage-for-vehicle111901-dev --accelerate-configuration Status=Suspended --region eu-north-1`

This also returns `MethodNotAllowed`.

Expected behavior: the migration generator should handle this S3 API error gracefully, assume acceleration is unavailable/disabled, and continue generating the Gen 2 backend.

### Expected behavior

`amplify gen2-migration generate` should complete successfully.

If `GetBucketAccelerateConfiguration` is not supported for a bucket/region, the migration tool should treat Transfer Acceleration as unavailable/disabled and continue generating the Gen 2 storage resource.

### Reproduction steps

1. Have an Amplify Gen 1 app in eu-north-1 with an Amplify-managed S3 storage resource.
2. Run `amplify gen2-migration assess`.
3. Run `amplify gen2-migration lock --skip-validations`.
4. Run `amplify gen2-migration generate`.
5. The command fails during planning with:
   `The specified method is not allowed against this resource.`
6. Run `amplify gen2-migration generate --debug`.
7. Debug output shows failure in:
   - AwsFetcher.fetchBucketAccelerate
   - S3Generator.plan

### Project Identifier

Region: eu-north-1
Amplify Gen 1 environment: dev
Storage resource name: tiemerkintastorageforvehicle
Physical S3 bucket: tiemerkintalaite-storage-for-vehicle111901-dev
Migration command: amplify gen2-migration generate
Session Identifier: 5a0e67be-d621-4c5c-91c7-5925e581b348

### Log output

<details>

```
UnknownFault: The specified method is not allowed against this resource.
MethodNotAllowed: The specified method is not allowed against this resource.
    at throwDefaultError (...@smithy\smithy-client...)
    at de_CommandError (...@aws-sdk\client-s3...)
    at AwsFetcher.fetchBucketAccelerate (...@aws-amplify\cli-internal\lib\commands\gen2-migration\_common\aws-fetcher.js...)
    at S3Generator.plan (...@aws-amplify\cli-internal\lib\commands\gen2-migration\generate\amplify\storage\s3.generator.js...)
    at AmplifyMigrationGenerateStep.forward (...gen2-migration\generate.js...)

```

</details>

### Additional information

_No response_

hariharan077:
#### Description of changes

Fixes #14876.

`amplify gen2-migration generate` currently fails when S3 returns `MethodNotAllowed` for `GetBucketAccelerateConfiguration`, which can happen for buckets where transfer acceleration is unsupported. This updates `AwsFetcher.fetchBucketAccelerate()` to treat only `MethodNotAllowed` as an unavailable acceleration state and return `undefined`, while still propagating other S3 errors.

The regression test uses a real `S3Client` against a mocked HTTP 405 REST-XML response. This exercises the AWS SDK deserializer and proves that `<Code>MethodNotAllowed</Code>` becomes an S3 service error whose `name` is `MethodNotAllowed`.

- Reproduced the failure through the real AWS SDK path before the patch: the HTTP 405 REST-XML response deserialized to `S3ServiceException` with `name === 'MethodNotAllowed'`, and `fetchBucketAccelerate()` rejected.
- Verified the same reproduction after the patch resolves to `undefined`.
- Focused AwsFetcher regression: 3/3 tests passed.
- S3 migration generator suite: 13/13 tests and 11/11 snapshots passed.
- Full `@aws-amplify/cli-internal` suite: 131/131 suites, 802/802 tests, and 138/138 snapshots passed.
- `@aws-amplify/cli-internal` build passed.
- ESLint, Prettier, and diff checks passed.
- Repository pre-commit and pre-push hooks passed without bypasses.

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/amplify-cli/src/__tests__/commands/gen2-migration/_common/aws-fetcher.test.ts
+++ b/packages/amplify-cli/src/__tests__/commands/gen2-migration/_common/aws-fetcher.test.ts
@@ -1,0 +1,51 @@
+import { GetBucketAccelerateConfigurationCommand, S3Client } from '@aws-sdk/client-s3';
+import nock from 'nock';
+import { AwsFetcher } from '../../../../commands/gen2-migration/_common/aws-fetcher';
+import { AwsClients } from '../../../../commands/gen2-migration/_common/aws-clients';
+
+describe('AwsFetcher', () => {
+  it('returns the bucket accelerate status when available', async () => {
+    const send = jest.fn();
+    const fetcher = new AwsFetcher({ s3: { send } } as unknown as AwsClients);
+    send.mockResolvedValueOnce({ Status: 'Enabled' });
+
+    await expect(fetcher.fetchBucketAccelerate('my-bucket')).resolves.toBe('Enabled');
+    expect(send).toHaveBeenCalledWith(expect.any(GetBucketAccelerateConfigurationCommand));
+  });
+
+  it('returns undefined for an SDK-deserialized MethodNotAllowed response', async () => {
+    const endpoint = 'https://s3.example.com';
+    const scope = nock(endpoint)
+      .get(/.*/)
+      .reply(
+        405,
+        '<?xml version="1.0" encoding="UTF-8"?><Error><Code>MethodNotAllowed</Code><Message>The specified method is not allowed against this resource.</Message></Error>',
+        { 'Content-Type': 'application/xml' },
+      );
+    const s3 = new S3Client({
+      region: 'us-east-1',
+      endpoint,
+      forcePathStyle: true,
+      credentials: { accessKeyId: 'test', secretAccessKey: 'test' },
+    });
+    const fetcher = new AwsFetcher({ s3 } as unknown as AwsClients);
+
+    try {
+      await expect(fetcher.fetchBucketAccelerate('my-bucket')).resolves.toBeUndefined();
+      scope.done();
+    } finally {
+      s3.destroy();
+      nock.cleanAll();
+    }
+  });
+
+  it('propagates non-MethodNotAllowed errors', async () => {
+    const send = jest.fn();
+    const fetcher = new AwsFetcher({ s3: { send } } as unknown as AwsClients);
+    const error = new Error('Access denied');
+    error.name = 'AccessDenied';
+    send.mockRejectedValueOnce(error);
+
+    await expect(fetcher.fetchBucketAccelerate('my-bucket')).rejects.toBe(error);
+  });
+});
--- a/packages/amplify-cli/src/__tests__/commands/gen2-migration/_common/aws-fetcher.test.ts
+++ b/packages/amplify-cli/src/__tests__/commands/gen2-migration/_common/aws-fetcher.test.ts
@@ -1,0 +1,37 @@
+import { GetBucketAccelerateConfigurationCommand } from '@aws-sdk/client-s3';
+import { AwsFetcher } from '../../../../commands/gen2-migration/_common/aws-fetcher';
+import { AwsClients } from '../../../../commands/gen2-migration/_common/aws-clients';
+
+describe('AwsFetcher', () => {
+  describe('fetchBucketAccelerate', () => {
+    const send = jest.fn();
+    const fetcher = new AwsFetcher({ s3: { send } } as unknown as AwsClients);
+
+    beforeEach(() => {
+      send.mockReset();
+    });
+
+    it('returns the bucket accelerate status when available', async () => {
+      send.mockResolvedValueOnce({ Status: 'Enabled' });
+
+      await expect(fetcher.fetchBucketAccelerate('my-bucket')).resolves.toBe('Enabled');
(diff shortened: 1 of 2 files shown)


packages/amplify-cli/src/__tests__/commands/gen2-migration/_common/aws-fetcher.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      send.mockResolvedValueOnce({ Status: 'Enabled' });
+
+      await expect(fetcher.fetchBucketAccelerate('my-bucket')).resolves.toBe('Enabled');
+      expect(send).toHaveBeenCalledWith(expect.any(GetBucketAccelerateConfigurationCommand));
+    });
+
+    it('returns undefined when bucket acceleration is not supported for the bucket', async () => {
+      const error = new Error('The specified method is not allowed against this resource.');
+      error.name = 'MethodNotAllowed';
+      send.mockRejectedValueOnce(error);
```
sharonyajain: **Regression test fabricates the S3 error name, so it never proves production errors actually expose name === 'MethodNotAllowed'**

The 'not supported' test hand-constructs a plain Error and sets error.name = 'MethodNotAllowed' to drive the branch. It validates the branch logic but not that a real @aws-sdk/client-s3 GetBucketAccelerateConfiguration failure is surfaced with that exact discriminator. The PR body's 'reproduced locally with the new regression test' likewise exercises the synthetic error, not the live SDK shape.

```typescript
const error = new Error('The specified method is not allowed against this resource.');
error.name = 'MethodNotAllowed';
send.mockRejectedValueOnce(error);
```

**Why it matters:** synthetic error name (cause) -> test and local repro pass on a value the test itself picked (mechanism) -> if the SDK ever surfaces this condition via a wrapping ServiceException name or only $metadata.httpStatusCode 405, the e.name === 'MethodNotAllowed' branch never matches in production (mechanism) -> the user in an unsupported region still hits the original crash the PR set out to fix (consequence). Probability is low because AWS SDK v3 sets a service error's name to its error code and the sibling fetchFunctionSchedule relies on the identical convention, which is why this is should-fix, not blocking.

**Suggestion:**  We can anchor the discriminator to reality: assert against an error shaped like the real S3ServiceException (e.g. construct it from the SDK's exception class or add a comment citing the SDK error-code == name convention), or add one integration-style check that the actual GetBucketAccelerateConfiguration rejection carries name === 'MethodNotAllowed'.

sharonyajain (dismissed):

packages/amplify-cli/src/__tests__/commands/gen2-migration/_common/aws-fetcher.test.ts
hariharan077: Fixed in `f50c7d4fe5`. The regression now uses a real `S3Client` against a mocked HTTP 405 REST-XML response, so the AWS SDK itself deserializes `<Code>MethodNotAllowed</Code>` before `AwsFetcher` handles it. The focused AwsFetcher test, S3 generator suite, full cli-internal suite, build, lint, formatting, and repository hooks all pass.

sharonyajain (approved):

PR #14952, opened 2026-07-01
fix(amplify-category-auth): clean up OIDC provider on auth stack teardown

ahmedhamouda78:
#### Description of changes

The auth category's `OpenId` custom resource Lambda (`openIdLambda.js`) only handled `Create`/`Update` events. On stack deletion CloudFormation sends a `Delete` event, which fell through to a no-op — so the account-global IAM OIDC provider (`accounts.google.com`) was never removed. Orphaned providers accumulate in the account on every teardown of an auth resource configured with Google/OpenID federation.

This PR:
- Adds a `Delete` handler to `openIdLambda.js`. Because the provider is account-global and keyed by URL (the create path reuses it and appends client IDs), the handler removes only the client IDs this resource registered and deletes the provider **only once no client IDs remain** — preserving providers shared by other Amplify apps in the same account. All IAM calls tolerate an already-removed provider/client ID, so the handler is idempotent.
- Grants the OpenId Lambda role `iam:RemoveClientIDFromOpenIDConnectProvider` and `iam:DeleteOpenIDConnectProvider` (scoped to the `accounts.google.com` provider ARN) in `auth-cognito-stack-builder.ts`.
- Adds unit tests for create, sole-owner delete, shared-provider delete, and idempotent delete.

#### Issue #, if available

N/A

- New Jest unit test `openIdLambda.test.js` passes (6/6), driving the real handler through: create, update-appends, sole-owner delete, shared-provider delete, concurrent-deletion race, and idempotent delete.
- Confirmed the IAM policy change does not alter existing stack-builder snapshots.

#### Checklist

- [x] PR description included

### DIFF
--- a/packages/amplify-category-auth/resources/auth-custom-resource/__tests__/openIdLambda.test.js
+++ b/packages/amplify-category-auth/resources/auth-custom-resource/__tests__/openIdLambda.test.js
@@ -2,80 +2,88 @@
 
 // In-memory IAM OIDC provider state driven by the mocked client.
 // Must be prefixed with `mock` so jest allows referencing it inside the jest.mock factory.
-const mockState = { providers: {} };
+let mockState = { providers: {} };
 
-jest.mock('cfn-response', () => ({
-  SUCCESS: 'SUCCESS',
-  FAILED: 'FAILED',
-  send: jest.fn((event, context, status, data) => {
-    context.__result = { status, data };
+jest.mock(
+  'cfn-response',
+  () => ({
+    SUCCESS: 'SUCCESS',
+    FAILED: 'FAILED',
+    send: jest.fn((event, context, status, data) => {
+      context.__result = { status, data };
+    }),
   }),
-}), { virtual: true });
+  { virtual: true },
+);
 
-jest.mock('@aws-sdk/client-iam', () => {
-  class NoSuchEntityException extends Error {
-    constructor(message) {
-      super(message);
-      this.name = 'NoSuchEntityException';
-    }
-  }
-  const mkCmd = (name) =>
-    class {
-      constructor(input) {
-        this.input = input;
-        this.__name = name;
+jest.mock(
+  '@aws-sdk/client-iam',
+  () => {
+    class NoSuchEntityException extends Error {
+      constructor(message) {
+        super(message);
+        this.name = 'NoSuchEntityException';
       }
+    }
+    const mkCmd = (name) =>
+      class {
+        constructor(input) {
+          this.input = input;
+          this.__name = name;
+        }
+      };
+    const send = jest.fn(async (cmd) => {
+      switch (cmd.__name) {
+        case 'List':
+          return { OpenIDConnectProviderList: Object.keys(mockState.providers).map((Arn) => ({ Arn })) };
+        case 'Create': {
+          const host = cmd.input.Url.replace('https://', '');
+          const Arn = `arn:aws:iam::123456789012:oidc-provider/${host}`;
+          mockState.providers[Arn] = { ClientIDList: [...cmd.input.ClientIDList] };
+          return { OpenIDConnectProviderArn: Arn };
+        }
+        case 'Get': {
+          const p = mockState.providers[cmd.input.OpenIDConnectProviderArn];
+          if (!p) throw new NoSuchEntityException('no provider');
+          return { ClientIDList: [...p.ClientIDList] };
+        }
+        case 'Add': {
+          const p = mockState.providers[cmd.input.OpenIDConnectProviderArn];
+          if (!p) throw new NoSuchEntityException('no provider');
+          if (!p.ClientIDList.includes(cmd.input.ClientID)) p.ClientIDList.push(cmd.input.ClientID);
+          return {};
+        }
+        case 'Remove': {
+          const p = mockState.providers[cmd.input.OpenIDConnectProviderArn];
+          if (!p) throw new NoSuchEntityException('no provider');
+          const i = p.ClientIDList.indexOf(cmd.input.ClientID);
+          if (i === -1) throw new NoSuchEntityException('no client id');
+          p.ClientIDList.splice(i, 1);
+          return {};
+        }
+        case 'Delete': {
+          if (!mockState.providers[cmd.input.OpenIDConnectProviderArn]) throw new NoSuchEntityException('no provider');
+          delete mockState.providers[cmd.input.OpenIDConnectProviderArn];
+          return {};
+        }
+        default:
+          throw new Error(`unexpected command ${cmd.__name}`);
+      }
+    });
+    return {
+      NoSuchEntityException,
+      IAMClient: jest.fn(() => ({ send })),
+      ListOpenIDConnectProvidersCommand: mkCmd('List'),
+      CreateOpenIDConnectProviderCommand: mkCmd('Create'),
+      GetOpenIDConnectProviderCommand: mkCmd('Get'),
+      AddClientIDToOpenIDConnectProviderCommand: mkCmd('Add'),
+      RemoveClientIDFromOpenIDConnectProviderCommand: mkCmd('Remove'),
+      DeleteOpenIDConnectProviderCommand: mkCmd('Delete'),
+      __send: send,
     };
-  const send = jest.fn(async (cmd) => {
-    switch (cmd.__name) {
-      case 'List':
-        return { OpenIDConnectProviderList: Object.keys(mockState.providers).map((Arn) => ({ Arn })) };
-      case 'Create': {
-        const host = cmd.input.Url.replace('https://', '');
-        const Arn = `arn:aws:iam::123456789012:oidc-provider/${host}`;
-        mockState.providers[Arn] = { ClientIDList: [...cmd.input.ClientIDList] };
-        return { OpenIDConnectProviderArn: Arn };
-      }
-      case 'Get': {
-        const p = mockState.providers[cmd.input.OpenIDConnectProviderArn];
-        if (!p) throw new NoSuchEntityException('no provider');
-        return { ClientIDList: [...p.ClientIDList] };
-      }
-      case 'Add': {
-        const p = mockState.providers[cmd.input.OpenIDConnectProviderArn];
-        if (!p) throw new NoSuchEntityException('no provider');
-        if (!p.ClientIDList.includes(cmd.input.ClientID)) p.ClientIDList.push(cmd.input.ClientID);
-        return {};
-      }
-      case 'Remove': {
-        const p = mockState.providers[cmd.input.OpenIDConnectProviderArn];
-        if (!p) throw new NoSuchEntityException('no provider');
-        const i = p.ClientIDList.indexOf(cmd.input.ClientID);
-        if (i === -1) throw new NoSuchEntityException('no client id');
-        p.ClientIDList.splice(i, 1);
-        return {};
-      }
-      case 'Delete': {
-        if (!mockState.providers[cmd.input.OpenIDConnectProviderArn]) throw new NoSuchEntityException('no provider');
-        delete mockState.providers[cmd.input.OpenIDConnectProviderArn];
-        return {};
-      }
-      default:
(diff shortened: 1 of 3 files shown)


packages/amplify-category-auth/resources/auth-custom-resource/openIdLambda.js
```diff
@@ -65,5 +75,46 @@ async function handleEvent(event) {
...
+      const existingValue = await findProviderArn(url);
+      if (!existingValue) {
+        return {};
+      }
+      const getOpenIDConnectProviderResponse = await iam.send(
+        new GetOpenIDConnectProviderCommand({ OpenIDConnectProviderArn: existingValue }),
+      );
+      const currentAudiences = getOpenIDConnectProviderResponse.ClientIDList || [];
+      const remainingAudiences = currentAudiences.filter((clientID) => !clientIdList.includes(clientID));
+      if (remainingAudiences.length === 0) {
```
soberm, line 95: [major] The shared-provider protection has a concurrent-deletion race that can leave a provider with zero client IDs permanently in the account.

When two stacks sharing this provider (e.g., one with `clientIdList = ['a']`, another with `['b']`) both receive their Delete events close together:

1. Both call `GetOpenIDConnectProvider` and observe `ClientIDList = ['a', 'b']`
2. Stack A computes `remainingAudiences = ['b']` (non-zero) and calls `RemoveClientID` for `a`
3. Stack B computes `remainingAudiences = ['a']` (non-zero) and calls `RemoveClientID` for `b`
4. Provider ends up with `ClientIDList = []` but neither stack called `DeleteOpenIDConnectProvider`

A re-read after the Remove loop would cover most cases:

```js
const refreshed = await iam.send(new GetOpenIDConnectProviderCommand({ OpenIDConnectProviderArn: existingValue }));
if ((refreshed.ClientIDList || []).length === 0) {
  await iam.send(new DeleteOpenIDConnectProviderCommand({ OpenIDConnectProviderArn: existingValue }));
}
```

The existing `NoSuchEntityException` catch already handles the case where a concurrent deleter wins the final race and deletes first.

packages/amplify-category-auth/resources/auth-custom-resource/openIdLambda.js
```diff
@@ -22,22 +24,30 @@ async function tryHandleEvent(event, context) {
...
 
+// Returns the ARN of the OIDC provider matching the given url, or undefined if none exists.
+async function findProviderArn(url) {
+  const listOpenIDConnectProvidersResponse = await iam.send(new ListOpenIDConnectProvidersCommand({}));
+  if (
+    listOpenIDConnectProvidersResponse.OpenIDConnectProviderList &&
+    listOpenIDConnectProvidersResponse.OpenIDConnectProviderList.length > 0
+  ) {
+    const vals = listOpenIDConnectProvidersResponse.OpenIDConnectProviderList.map((x) => x.Arn);
+    return vals.find((i) => i.split('/')[1] === url.split('//')[1]);
```
soberm: [minor] The URL matching here is fragile. `i.split('/')[1]` works for the current ARN format (`arn:...:oidc-provider/accounts.google.com`), and `url.split('//')[1]` works for plain `https://` URLs, but both break silently if the URL ever has a path component (e.g., `https://idp.example.com/realms/main`). Now that this logic lives in a named shared abstraction, worth hardening:

```js
const providerHost = new URL(url).host;
return vals.find((arn) => arn.endsWith(`/oidc-provider/${providerHost}`));
```

packages/amplify-category-auth/src/provider-utils/awscloudformation/auth-stack-builder/auth-cognito-stack-builder.ts
```diff
@@ -1065,7 +1065,13 @@ export class AmplifyAuthCognitoStack extends cdk.Stack implements AmplifyAuthCog
...
           {
             Effect: 'Allow',
-            Action: ['iam:CreateOpenIDConnectProvider', 'iam:GetOpenIDConnectProvider', 'iam:AddClientIDToOpenIDConnectProvider'],
+            Action: [
+              'iam:CreateOpenIDConnectProvider',
+              'iam:GetOpenIDConnectProvider',
+              'iam:AddClientIDToOpenIDConnectProvider',
+              'iam:RemoveClientIDFromOpenIDConnectProvider',
+              'iam:DeleteOpenIDConnectProvider',
+            ],
```
soberm, line 1074: [minor] `iam:ListOpenIDConnectProviders` is not in this statement. Both the Create/Update and new Delete paths call `ListOpenIDConnectProvidersCommand` via `findProviderArn`, and that action requires `Resource: *` (it is account-level, not scoped to a specific provider ARN). If it is granted via a separate statement or a managed policy on the role, a comment noting that would prevent future confusion. If it is not granted anywhere, the Delete handler will fail at runtime with an authorization error.

packages/amplify-category-auth/resources/auth-custom-resource/__tests__/openIdLambda.test.js
```diff
@@ -0,0 +1,133 @@
...
+// The handler is intentionally fire-and-forget (returns void, calls cfn-response.send when done).
+// Flush pending microtasks until cfn-response.send populates the result.
+async function invoke(event) {
+  const context = {};
+  handler(event, context);
+  for (let i = 0; i < 100 && !context.__result; i++) {
+    // eslint-disable-next-line no-await-in-loop
+    await new Promise((r) => setImmediate(r));
+  }
+  return context.__result;
```
soberm, line 105: [minor] If the handler never calls `response.send` (e.g., due to an unexpected mock failure or uncaught throw), `invoke` silently returns `undefined` after 100 iterations and the test fails with a cryptic "Cannot read properties of undefined (reading 'status')" rather than pointing at the actual problem. Adding a guard before the return makes failures much easier to trace:

```js
if (!context.__result) throw new Error('handler never called response.send');
return context.__result;
```

packages/amplify-category-auth/resources/auth-custom-resource/__tests__/openIdLambda.test.js
```diff
@@ -0,0 +1,133 @@
...
+    RemoveClientIDFromOpenIDConnectProviderCommand: mkCmd('Remove'),
+    DeleteOpenIDConnectProviderCommand: mkCmd('Delete'),
+    __send: send,
+  };
+}, { virtual: true });
+
+const iamMock = require('@aws-sdk/client-iam');
+const { handler } = require('../openIdLambda');
+
+const URL = 'https://accounts.google.com';
```
soberm: [nit] `URL` shadows the global `URL` constructor from Node.js. Won't cause a problem here since none of the tests use the built-in, but it is a confusing collision. `GOOGLE_OIDC_URL` or `TEST_URL` would be cleaner.

packages/amplify-category-auth/resources/auth-custom-resource/__tests__/openIdLambda.test.js
```diff
@@ -0,0 +1,133 @@
...
+  const context = {};
+  handler(event, context);
+  for (let i = 0; i < 100 && !context.__result; i++) {
+    // eslint-disable-next-line no-await-in-loop
+    await new Promise((r) => setImmediate(r));
+  }
+  return context.__result;
+}
+
+describe('openIdLambda custom resource', () => {
```
soberm, line 108: [nit] The Create/Update branch handles both request types but only `Create` is exercised by a test. A case that calls `Create` first, then `Update` with a different `clientIdList`, and checks that the new client IDs are appended to the existing provider would close this gap.

ahmedhamouda78:
Thanks for the thorough review @soberm — all six points addressed in b16780b:

- **[major] Concurrent-deletion race:** After removing our client IDs I now re-read the provider and delete it if the list is empty, so two stacks racing on a shared provider can't leave it orphaned with zero client IDs. The existing `NoSuchEntityException` catch handles the case where the other deleter wins the final delete. Added a `deletes the provider when a concurrent stack empties it during our Delete` test that simulates the interleaving.
- **[minor] Fragile URL matching:** `findProviderArn` now derives the host via `new URL(url).host` and matches the provider ARN suffix. One correction to the suggested snippet: real OIDC provider ARNs use `:oidc-provider/<host>` (colon), e.g. `arn:aws:iam::<account>:oidc-provider/accounts.google.com`, so I matched on `:oidc-provider/${providerHost}` rather than `/oidc-provider/...` — the latter never matches.
- **[minor] ListOpenIDConnectProviders:** It's granted in the adjacent statement (account-level, `Resource: *`) since it can't be scoped to a single provider ARN. Added a comment there noting it's required by `findProviderArn`.
- **[minor] invoke() guard:** Now throws `handler never called response.send` instead of returning `undefined`.
- **[nit] URL shadowing:** Renamed the test constant to `GOOGLE_OIDC_URL`.
- **[nit] Update coverage:** Added a test that creates, then updates with an additional client ID and asserts it's appended.

All six unit tests pass locally under CI-like conditions (runtime-only deps mocked virtually).

soberm (dismissed):

sarayev (dismissed):
## Review — Approve with comments ✅

Solid fix for a real latent leak: the OpenId custom-resource Lambda never handled `Delete`, so the `accounts.google.com` OIDC provider (+ IAM resources) leaked on stack teardown. The `Delete` handler (removes only this stack's client IDs, deletes the provider only when empty), scoped IAM perms, idempotency (`NoSuchEntityException` tolerance), shared-provider safety, and the `new URL(url).host` refactor are all good. 6-scenario test file is a nice addition.

### Should-fix (cheap, none break the happy path)
1. **Inaccurate comment — `auth-cognito-stack-builder.ts:~1073`.** Says `ListOpenIDConnectProviders` must use `Resource '*'`, but the code scopes to `arn:aws:iam::<acct>:oidc-provider/*` (via `selector: '*'`). Misleading — reword to describe the wildcard provider ARN.
2. **Asymmetric race — `openIdLambda.js:~93` (sole-owner branch).** Shared-provider branch re-reads before deleting; sole-owner deletes from the stale initial read. A concurrent `Create` could be destroyed. Mirror the re-read (Get → confirm empty → delete).
3. **Missing `url` null-guard.** `clientIdList` got a guard this PR; `url` didn't → `new URL(undefined)` throws `TypeError`. Add `event.ResourceProperties.url ?? ''` (or validate early).

### Nits
4. PR body says 4 tests; file has 6 — update the count.
5. `openIdLambda.test.js:5` mutates module-level `mockState` (guideline: reset a `let` in `beforeEach`).
6. Shared-provider test injects state directly; prefer a 2nd `Create` to set up shared state (also exercises add-client-ID path).
7. Pre-existing gap (not this PR): no snapshot coverage of the OpenId IAM policy.

Please run the **full split e2e + all PR checks** and get them green before merge.

ahmedhamouda78:
Thanks @sarayev — addressed in d153b19:

**Should-fix**
1. **Inaccurate IAM comment:** Reworded — `ListOpenIDConnectProviders` is granted against the wildcard provider ARN (`oidc-provider/*`) in this account, not `Resource '*'`.
2. **Asymmetric race (sole-owner branch):** The sole-owner branch now re-reads immediately before deleting (Get → confirm nothing but our client IDs remain → delete). If a concurrent `Create` added client IDs (or re-created the provider) after our first read, we fall back to removing only our client IDs instead of deleting. Both branches are now symmetric.
3. **`url` null-guard:** `findProviderArn` returns early when `url` is falsy, so `new URL(undefined)` can no longer throw.

**Nits**
4. Updated the PR description (6 tests, not 4).
5. `mockState` is now reset via reassignment (`let` + `beforeEach`).
6. The shared-provider test now sets up shared state with a real second `Create` (which finds the existing provider and adds its client ID), exercising the add-client-ID path instead of injecting state.
7. Snapshot coverage of the OpenId IAM policy — agreed it's a pre-existing gap; leaving it out of this PR's scope.

All 6 unit tests pass, and prettier/lint is now clean (the earlier `_lint` failure was a prettier formatting miss on the test file, fixed here).

sarayev (approved):
Thanks for addressing the feedback — approving, this looks good to me.

soberm (approved):

PR #14953, opened 2026-07-02
fix(cli): scope migrated private/protected S3 access to the caller identity

ahmedhamouda78:
#### Description of changes

The `gen2-migration generate` storage generator rendered Gen1 authenticated
access on the per-user paths `private/{entity_id}/*` and
`protected/{entity_id}/*` using `allow.authenticated`. With that rule the
`{entity_id}` token is not bound to the requesting user, so the generated
`defineStorage` access did not match the per-user scoping of the original
Gen1 configuration.

This change maps Gen1 authenticated access on `private/` and `protected/` to
`allow.entity('identity')`, so `{entity_id}` resolves to the caller's own
Cognito identity — matching the Gen1 behavior. `protected/` additionally
keeps `allow.authenticated` read access (Gen1 protected semantics), and
`public/*` is unchanged. Group and function access mappings are unchanged.

Files:
- `packages/amplify-cli/.../generate/amplify/storage/s3.renderer.ts` — access-rule mapping + a small `entity('identity')` render helper.
- `packages/amplify-cli/.../__tests__/.../storage/s3.generator.test.ts` — new coverage + updated inline snapshots.

- `yarn jest s3.generator` — 12 tests pass, snapshots updated.
- Added a test asserting the per-user paths render `allow.entity('identity')`
  and never grant write/delete via `allow.authenticated`.
- Deployed a sandbox with a representative migrated config and confirmed the
  authenticated-role IAM policy scopes `private`/`protected` object access to
  the caller's Cognito identity.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
+++ b/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
@@ -189,8 +189,11 @@ describe('S3Generator', () => {
         name: \`myBucket-main-\${branchName}\`,
         access: (allow) => ({
           'public/*': [allow.authenticated.to(['read', 'write'])],
-          'protected/{entity_id}/*': [allow.authenticated.to(['read', 'write'])],
-          'private/{entity_id}/*': [allow.authenticated.to(['read', 'write'])],
+          'protected/{entity_id}/*': [
+            allow.entity('identity').to(['read', 'write']),
+            allow.authenticated.to(['read']),
+          ],
+          'private/{entity_id}/*': [allow.entity('identity').to(['read', 'write'])],
         }),
       });
 
@@ -216,6 +219,43 @@ describe('S3Generator', () => {
       }
       "
     `);
+  });
+
+  // private/ and protected/ are per-user paths: authenticated access must be scoped to the
+  // caller's own identity via allow.entity('identity'), matching the Gen1 configuration.
+  it('scopes authenticated private/protected access to the caller identity', async () => {
+    const gen1App = await createGen1App({
+      providers: { awscloudformation: { StackName: 'amplify-test-main-123456', Region: 'us-east-1' } },
+      storage: { myBucket: { service: 'S3', output: { BucketName: 'myBucket-main-abc123' } } },
+    });
+    jest.spyOn(gen1App, 'cliInputs').mockReturnValue({
+      authAccess: ['READ', 'CREATE_AND_UPDATE', 'DELETE'],
+      guestAccess: [],
+    });
+    jest.spyOn(gen1App.aws, 'fetchBucketAccelerate').mockResolvedValue(undefined);
+    jest.spyOn(gen1App.aws, 'fetchBucketVersioning').mockResolvedValue(undefined);
+    jest.spyOn(gen1App.aws, 'fetchBucketEncryption').mockResolvedValue(undefined);
+
+    const generator = new S3Generator(
+      gen1App,
+      backendGenerator,
+      outputDir,
+      { category: 'storage', resourceName: 'myBucket', service: 'S3', key: 'storage:S3' },
+      logger,
+    );
+    const ops = await generator.plan();
+    await ops[0].execute();
+    const content = writtenFile('resource.ts');
+    const normalized = content.replace(/\s+/g, ' ');
+
+    // Per-user paths use entity('identity') so {entity_id} maps to the caller's identity.
+    expect(normalized).toMatch(
+      /'private\/\{entity_id\}\/\*':\s*\[\s*allow\.entity\('identity'\)\.to\(\['read', 'write', 'delete'\]\)\s*,?\s*\]/,
+    );
+    expect(normalized).toMatch(/'protected\/\{entity_id\}\/\*':\s*\[\s*allow\.entity\('identity'\)\.to\(\['read', 'write', 'delete'\]\)/);
+
+    // Enforce per-user scoping: private/protected must not grant write/delete via allow.authenticated.
+    expect(content).not.toMatch(/'(private|protected)\/\{entity_id\}\/\*': \[[^\]]*allow\.authenticated\.to\(\[[^\]]*'(write|delete)'/s);
   });
 
   it('renders guest access patterns', async () => {
@@ -324,10 +364,11 @@ describe('S3Generator', () => {
             allow.authenticated.to(['read', 'write', 'delete']),
           ],
           'protected/{entity_id}/*': [
-            allow.authenticated.to(['read', 'write', 'delete']),
+            allow.entity('identity').to(['read', 'write', 'delete']),
+            allow.authenticated.to(['read']),
           ],
           'private/{entity_id}/*': [
-            allow.authenticated.to(['read', 'write', 'delete']),
+            allow.entity('identity').to(['read', 'write', 'delete']),
           ],
         }),
       });
@@ -402,11 +443,12 @@ describe('S3Generator', () => {
             allow.groups(['admin']).to(['read', 'write', 'delete']),
           ],
           'protected/{entity_id}/*': [
+            allow.entity('identity').to(['read']),
             allow.authenticated.to(['read']),
             allow.groups(['admin']).to(['read', 'write', 'delete']),
           ],
           'private/{entity_id}/*': [
-            allow.authenticated.to(['read']),
+            allow.entity('identity').to(['read']),
             allow.groups(['admin']).to(['read', 'write', 'delete']),
           ],
         }),
--- a/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
+++ b/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
@@ -248,10 +248,51 @@ describe('S3Generator', () => {
     const content = writtenFile('resource.ts');
     const normalized = content.replace(/\s+/g, ' ');
 
-    expect(normalized).toMatch(
-      /'private\/\{entity_id\}\/\*':\s*\[\s*allow\.entity\('identity'\)\.to\(\['read', 'write', 'delete'\]\)\s*,?\s*\]/,
-    );
-    expect(normalized).toMatch(/'protected\/\{entity_id\}\/\*':\s*\[\s*allow\.entity\('identity'\)\.to\(\['read', 'write', 'delete'\]\)/);
+    // Positive assertion: full generated file locks the per-user mapping (owner scoped to
+    // entity('identity'); protected/ retains authenticated read).
+    expect(content).toMatchInlineSnapshot(`
+      "import { defineStorage } from '@aws-amplify/backend';
+      import { CfnResource } from 'aws-cdk-lib';
+      import type { Backend } from '../backend';
+
+      const branchName = process.env.AWS_BRANCH ?? 'sandbox';
+
+      export const storage = defineStorage({
+        name: \`myBucket-main-\${branchName}\`,
+        access: (allow) => ({
+          'public/*': [allow.authenticated.to(['read', 'write', 'delete'])],
+          'protected/{entity_id}/*': [
+            allow.entity('identity').to(['read', 'write', 'delete']),
+            allow.authenticated.to(['read']),
+          ],
+          'private/{entity_id}/*': [
+            allow.entity('identity').to(['read', 'write', 'delete']),
+          ],
+        }),
+      });
+
+      export function postRefactor(backend: Backend) {
+        const s3Bucket = backend.storage.resources.cfnResources.cfnBucket;
+        s3Bucket.bucketName = 'myBucket-main-abc123';
+      }
+
+      export function applyEscapeHatches(backend: Backend) {
+        const s3Bucket = backend.storage.resources.cfnResources.cfnBucket;
+        for (const cfnResource of backend.storage.stack.node
+          .findAll()
+          .filter(
+            (c) =>
+              CfnResource.isCfnResource(c) &&
+              ['AWS::S3::Bucket', 'Custom::S3AutoDeleteObjects'].includes(
+                c.cfnResourceType
+              )
+          )) {
+          (cfnResource as CfnResource).addOverride('UpdateReplacePolicy', 'Retain');
+          (cfnResource as CfnResource).addOverride('DeletionPolicy', 'Retain');
+        }
+      }
+      "
+    `);
 
     // private/ and protected/ must not grant write/delete via allow.authenticated (public/*
     // legitimately can, so the check is scoped to those paths). Rules render in the order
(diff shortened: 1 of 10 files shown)


soberm (commented):
## Code Review

### Summary

The `gen2-migration generate` storage renderer was emitting `allow.authenticated` for all three path prefixes (`public/*`, `protected/{entity_id}/*`, `private/{entity_id}/*`), so the `{entity_id}` placeholder was never bound to the requesting user — silently granting any authenticated user full write/delete access to every other user's private files. This PR correctly maps the two per-user prefixes to `allow.entity('identity')` and preserves Gen1 `protected/` semantics by retaining `allow.authenticated.to(['read'])` alongside the entity rule.

### Overall Assessment

**✅ Approve with minor notes.** The fix is semantically correct, the core logic is clean, and test coverage is headed in the right direction. Two issues deserve attention before merge: a **test regex that gives false security guarantees**, and a **redundant IAM rule** in the `protected/` read-only case.

---

### 🟡 SUGGESTION — Test false negative: negative assertion won't catch the regression it guards against

**File:** `s3.generator.test.ts` (new test, lines ~250–253)

```ts
expect(content).not.toMatch(
  /'(private|protected)\/\{entity_id\}\/\*': \[[^\]]*allow\.authenticated\.to\(\[[^\]]*'(write|delete)'/s
);
```

**The problem:** `[^\]]*` halts at the first `]` it encounters. In the correct emitted output, `allow.entity('identity').to(['read', 'write', 'delete'])` contains a `]` **before** the `allow.authenticated` call, so the regex can never scan past it to find `allow.authenticated.to(...)`. The `not.toMatch` therefore passes unconditionally whenever an entity rule precedes the authenticated call — exactly the layout the fixed code produces. A future regression that adds `allow.authenticated.to(['write'])` after an entity rule would go **undetected**.

Additionally, this assertion uses the raw `content` string while the two positive assertions immediately above it use `normalized`. The inconsistency is confusing.

**Suggested replacement** (using the already-normalized string):
```ts
// Any rule granting write/delete through allow.authenticated is a per-user isolation violation.
expect(normalized).not.toMatch(/allow\.authenticated\.to\(\[[^\]]*'(write|delete)/);
```

---

### 🟡 SUGGESTION — Redundant `allow.entity('identity').to(['read'])` on `protected/` when auth is read-only

**File:** `s3.renderer.ts` lines ~287–292; also visible in the groups-test snapshot (diff line ~443)

When `authAccess` is `['read']` only, the protected path generates:
```ts
'protected/{entity_id}/*': [
  allow.entity('identity').to(['read']),  // owner reads via entity scope
  allow.authenticated.to(['read']),        // any authenticated user reads
]
```
The entity `['read']` rule is **strictly subsumed** by `allow.authenticated.to(['read'])` (the owner is also an authenticated user), producing an unnecessary IAM policy statement. The entity rule only adds value when it grants something the broader authenticated rule does not (i.e., `write` or `delete`). Worth either trimming the entity rule when auth is read-only, or adding a comment explaining why the redundancy is intentional.

---

### 🟡 SUGGESTION — No CHANGELOG / migration guide entry

The PR checklist has "Relevant documentation is changed or added" **unchecked**. This is a security-relevant behavioral change: users who already ran `gen2-migration generate` have broken `resource.ts` files and need to re-run the generator. At minimum, a `CHANGELOG.md` entry should describe the old (broken) output, the corrected output, and the remediation step for anyone who already migrated.

---

### 🟡 SUGGESTION — Add a regression guard that `public/*` stays on `allow.authenticated`

The new dedicated test verifies private/protected paths but never asserts the `public/*` path was left unchanged. A single line like:
```ts
expect(normalized).toMatch(/'public\/\*':\s*\[\s*allow\.authenticated\.to\(\[/);
```
documents the intended asymmetry and guards against accidentally switching `public/` to an entity rule in the future.

---

### 🔵 NITPICK — Redundant comment in the new test

```ts
// Per-user paths use entity('identity') so {entity_id} maps to the caller's identity.
expect(normalized).toMatch(...);
```
This comment restates what the `it(...)` description and the assertion already convey. Can be removed.

---

### 🔵 NITPICK — `createEntityPattern` does not use `this`; consider `private static`

**File:** `s3.renderer.ts` line ~357. The method has no `this` references and depends only on its parameters and the module-level `factory` import. If `createAllowPattern` and `createResourcePattern` share this trait it would be worth making all three `private static` for clarity.

---

### ✅ Praise

- **Incidental AST-node-sharing fix.** The old code stored one `pattern` node and pushed it into all three path-access arrays. TypeScript AST nodes must not be shared across parents in transformation contexts; the new code correctly allocates a fresh node for each use.
- **Gen1 semantics faithfully reproduced.** `protected/` → entity write/delete + authenticated global read; `private/` → entity only; `public/` → unchanged. Matches Gen1's documented access-level contract precisely.
- **`createEntityPattern` is clean** and follows the established factory-helper pattern; the JSDoc accurately describes the generated output.
- **Commit message is exemplary** — explains the old behavior, root cause, and fix in one paragraph, exactly what `git log` readers need when investigating an IAM regression.
- **All affected snapshots updated** — no stale passing tests left behind.

---

soberm (commented):
## Code Review

### Summary

The `gen2-migration generate` storage renderer was emitting `allow.authenticated` for all three path prefixes (`public/*`, `protected/{entity_id}/*`, `private/{entity_id}/*`), so the `{entity_id}` placeholder was never bound to the requesting user — silently granting any authenticated user full write/delete access to every other user's private files. This PR correctly maps the two per-user prefixes to `allow.entity('identity')` and preserves Gen1 `protected/` semantics by retaining `allow.authenticated.to(['read'])` alongside the entity rule.

### Overall Assessment

**Approve with minor notes.** The fix is semantically correct, the core logic is clean, and test coverage is headed in the right direction. Two issues deserve attention before merge: a **test regex that gives false security guarantees**, and a **redundant IAM rule** in the `protected/` read-only case.

---

### [major] Test false negative: negative assertion will not catch the regression it guards against

**File:** `s3.generator.test.ts` (new test, negative assertion)

```ts
expect(content).not.toMatch(
  /'(private|protected)\/\{entity_id\}\/\*': \[[^\]]*allow\.authenticated\.to\(\[[^\]]*'(write|delete)'/s
);
```

`[^\]]*` halts at the first `]` it encounters. In the correct emitted output, `allow.entity('identity').to(['read', 'write', 'delete'])` contains a `]` **before** the `allow.authenticated` call, so the regex can never scan past it to find `allow.authenticated.to(...)`. The `not.toMatch` therefore passes unconditionally whenever an entity rule precedes the authenticated call — exactly the layout the fixed code produces. A future regression that reintroduces `allow.authenticated.to(['write'])` after an entity rule would go **undetected**.

Additionally, this assertion uses the raw `content` string while the two positive assertions immediately above it use `normalized`, which is inconsistent.

Suggested replacement using the already-normalized string:
```ts
// Any allow.authenticated granting write/delete is a per-user isolation violation.
expect(normalized).not.toMatch(/allow\.authenticated\.to\(\[[^\]]*'(write|delete)/);
```

---

### [minor] Redundant `allow.entity('identity').to(['read'])` on `protected/` when auth is read-only

**File:** `s3.renderer.ts` lines ~287-292; also visible in the groups-test snapshot

When `authAccess` is `['read']` only, the protected path generates:
```ts
'protected/{entity_id}/*': [
  allow.entity('identity').to(['read']),  // owner reads via entity scope
  allow.authenticated.to(['read']),        // any authenticated user reads
]
```
The entity `['read']` rule is strictly subsumed by `allow.authenticated.to(['read'])` (the owner is also an authenticated user), producing a superfluous IAM policy statement. The entity rule only adds value when it grants something the broader authenticated rule does not (i.e. `write` or `delete`). Consider either omitting the entity rule when auth is read-only, or adding a comment explaining why the redundancy is intentional.

---

### [minor] No CHANGELOG / migration guide entry

The PR checklist has "Relevant documentation is changed or added" unchecked. This is a security-relevant behavioral change: any user who already ran `gen2-migration generate` has a broken `resource.ts` and needs to re-run the generator. A `CHANGELOG.md` entry describing the old (broken) output, the corrected output, and the remediation step would help affected users discover the fix.

---

### [nit] Add a regression guard that `public/*` stays on `allow.authenticated`

The new dedicated test verifies private/protected paths but never asserts the `public/*` path was left unchanged. A single line like:
```ts
expect(normalized).toMatch(/'public\/\*':\s*\[\s*allow\.authenticated\.to\(\[/);
```
documents the intended asymmetry and prevents accidentally switching `public/` to an entity rule in future.

---

### [nit] Redundant comment in the new test

```ts
// Per-user paths use entity('identity') so {entity_id} maps to the caller's identity.
expect(normalized).toMatch(...);
```
This comment restates what the `it(...)` description and the assertion already convey. Can be removed.

---

### [nit] `createEntityPattern` does not use `this`; consider `private static`

**File:** `s3.renderer.ts` new method. The method has no `this` references and depends only on its parameters and the module-level `factory` import. If `createAllowPattern` and `createResourcePattern` share this trait, making all three `private static` would be a cleaner signal.

---

### Praise

- **Incidental AST-node-sharing fix.** The old code stored one `pattern` node and pushed it into all three path-access arrays. TypeScript AST nodes must not be shared across parents; the new code correctly allocates a fresh node for each use.
- **Gen1 semantics faithfully reproduced.** `protected/` gets entity write/delete + authenticated global read; `private/` gets entity only; `public/` is unchanged. Matches Gen1's documented access-level contract precisely.
- **`createEntityPattern` is clean** and follows the established factory-helper pattern; the JSDoc accurately describes the generated output.
- **Commit message is exemplary** — explains the old behavior, root cause, and fix in one paragraph, which is exactly what `git log` readers need when investigating an IAM regression.
- **All affected snapshots updated** — no stale passing tests left behind.

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
```diff
@@ -218,6 +221,43 @@ describe('S3Generator', () => {
...
+    const normalized = content.replace(/\s+/g, ' ');
+
+    // Per-user paths use entity('identity') so {entity_id} maps to the caller's identity.
+    expect(normalized).toMatch(
+      /'private\/\{entity_id\}\/\*':\s*\[\s*allow\.entity\('identity'\)\.to\(\['read', 'write', 'delete'\]\)\s*,?\s*\]/,
+    );
+    expect(normalized).toMatch(/'protected\/\{entity_id\}\/\*':\s*\[\s*allow\.entity\('identity'\)\.to\(\['read', 'write', 'delete'\]\)/);
+
+    // Enforce per-user scoping: private/protected must not grant write/delete via allow.authenticated.
+    expect(content).not.toMatch(/'(private|protected)\/\{entity_id\}\/\*': \[[^\]]*allow\.authenticated\.to\(\[[^\]]*'(write|delete)'/s);
```
soberm: [major] `[^\]]*` stops at the first `]` it sees, and the entity pattern's permissions array (`.to(['read', 'write', 'delete'])`) contains one well before `allow.authenticated` appears. In practice this regex never scans far enough to find the authenticated call -- a regression that re-introduces `allow.authenticated.to(['write'])` after an entity rule would pass right through undetected.

Also using `content` (raw) here while the two positive assertions above use `normalized` is inconsistent. A version that actually catches the problem:

```ts
expect(normalized).not.toMatch(/allow\.authenticated\.to\(\[[^\]]*'(write|delete)/);
```

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
```diff
@@ -218,6 +221,43 @@ describe('S3Generator', () => {
...
+      outputDir,
+      { category: 'storage', resourceName: 'myBucket', service: 'S3', key: 'storage:S3' },
+      logger,
+    );
+    const ops = await generator.plan();
+    await ops[0].execute();
+    const content = writtenFile('resource.ts');
+    const normalized = content.replace(/\s+/g, ' ');
+
+    // Per-user paths use entity('identity') so {entity_id} maps to the caller's identity.
```
soberm: [nit] This just restates what the test name already says -- can drop it.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/storage/s3.renderer.ts
```diff
@@ -280,10 +280,15 @@ export class S3Renderer {
...
-      publicPathAccess.push(pattern);
-      protectedPathAccess.push(pattern);
-      privatePathAccess.push(pattern);
+      // public/* is a shared path: authenticated users get the Gen1 auth permission set.
+      publicPathAccess.push(this.createAllowPattern(allowIdentifier, 'authenticated', accessPatterns.auth));
+      // private/ and protected/ are per-user paths. Use allow.entity('identity') so the
+      // {entity_id} token resolves to the caller's own Cognito identity, matching the
+      // per-user scoping of the Gen1 configuration.
+      privatePathAccess.push(this.createEntityPattern(allowIdentifier, accessPatterns.auth));
+      protectedPathAccess.push(this.createEntityPattern(allowIdentifier, accessPatterns.auth));
```
soberm: [minor] When `authAccess` is read-only (`['read']` only), this generates both `allow.entity('identity').to(['read'])` and the `allow.authenticated.to(['read'])` added below for the same protected path. The entity rule is fully subsumed (the owner is also an authenticated user), so you get an IAM policy statement that does nothing extra. Worth either skipping the entity rule when the permission set contains only `read`, or leaving a comment here explaining why the overlap is intentional.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/storage/s3.renderer.ts
```diff
@@ -352,6 +357,25 @@ export class S3Renderer {
     );
   }
 
+  /**
+   * Renders `allow.entity('identity').to([...])`, which scopes an {entity_id} path to the
+   * caller's own Cognito identity in the generated IAM policy.
+   */
+  private createEntityPattern(allowIdentifier: ts.Identifier, permissions: readonly Permission[]): CallExpression {
```
soberm: [nit] Nothing touches instance state here -- this could be `private static createEntityPattern(...)`. Same goes for `createAllowPattern` and `createResourcePattern` if those are also purely functional.

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
ahmedhamouda78: Good catch — the `[^\]]*` stopped at the entity array's `]` and never reached the authenticated call, so the guard was dead. Fixed in `662bea5`. I scoped it to the protected/private segments rather than a repo-wide `allow.authenticated` check, since `public/*` legitimately grants authenticated write/delete and would false-positive. Verified it now flags a regression and passes on correct output.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/storage/s3.renderer.ts
ahmedhamouda78: Done in `662bea5` — all three are now `private static`.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/storage/s3.renderer.ts
ahmedhamouda78: Added a comment explaining the overlap is intentional (`662bea5`) — the authenticated read subsumes the owner read only in the read-only case, and keeping the entity rule uniform avoids special-casing. Happy to switch to skip-logic if you'd prefer.

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
ahmedhamouda78: Removed in `662bea5`.

soberm (dismissed):

sarayev (dismissed):
## Review — Approve with comments ✅

Correct security fix: `gen2-migration generate` was emitting `allow.authenticated.to([...])` for `private/{entity_id}/*` and `protected/{entity_id}/*`, granting every authenticated user the same access and defeating per-user `{entity_id}` scoping. Mapping them to `allow.entity('identity').to([...])` (keeping `allow.authenticated.to(['read'])` on `protected/` for Gen1 cross-user read) is right. Separate AST nodes per path + `private static` helpers are clean; all 8 fixture snapshots + 3 inline snapshots updated consistently.

### Should-fix (test quality, not correctness)
1. **`s3.generator.test.ts` new test — use a full snapshot for positive assertions.** The `toMatch()` regex + `normalized.slice(indexOf(...))` positives conflict with the repo guideline (`toMatchInlineSnapshot()` on complete output) and the slice-to-EOF is fragile (captures postRefactor/applyEscapeHatches; breaks if path ordering changes). Convert positives to one `toMatchInlineSnapshot()` on `writtenFile('resource.ts')`; keep the `not.toMatch` negative guards as-is.
2. **Missing edge case: write-only auth on `protected/`** (`authAccess: ['CREATE_AND_UPDATE']`). Code unconditionally appends `allow.authenticated.to(['read'])` to `protected/{entity_id}/*` — correct Gen1 semantics but untested; add a test locking owner=`entity('identity').to(['write'])` + public=`authenticated.to(['read'])`.

### Nits
3. `s3.renderer.ts` createEntityPattern: multi-line JSDoc on a `private static` helper → use a `//` comment per the style guide.
4. `buildAccessProperty`: 3 inline comment blocks could consolidate into one.

Core logic correct and regression-safe for existing scenarios. Please run the **full split e2e + all PR checks** and get them green before merge.

ahmedhamouda78:
@sarayev thanks for the thorough review. Pushed `5491fd6` addressing all four points:

**1. Full inline snapshot for the per-user test** — Replaced the `toMatch()` regex + `normalized.slice(indexOf(...))` positives with a single `toMatchInlineSnapshot()` on the complete `writtenFile('resource.ts')`, per the repo guideline. This also removes the slice-to-EOF fragility you flagged (no longer sensitive to path ordering / `postRefactor`/`applyEscapeHatches` capture). Kept the scoped `not.toMatch` write/delete negative guards on the `private`/`protected` segments as-is.

**2. Write-only `protected/` edge case** — Added a test `scopes write-only authenticated protected access and retains authenticated read` with `authAccess: ['CREATE_AND_UPDATE']`. It locks the owner rule as `allow.entity('identity').to(['write'])` and confirms `protected/{entity_id}/*` still gets the unconditional `allow.authenticated.to(['read'])` (Gen1 cross-user read), while `private/{entity_id}/*` stays owner-only write.

**3. `createEntityPattern` comment** — Switched the multi-line JSDoc on the `private static` helper to `//` comments per the style guide.

**4. `buildAccessProperty` comments** — Consolidated the three inline comment blocks into one block describing the full per-path auth mapping (public shared / private+protected per-user via `entity('identity')` / protected authenticated-read overlap).

No production logic changed — all test-quality and comment-style. Local verification: `yarn jest s3.generator` -> 13 passed (incl. the new edge case), 11 snapshots passed; eslint clean. Full split e2e + PR checks re-running on the new commit.

sarayev (approved):
Thanks for addressing the feedback — approving, this looks good to me.

One non-blocking thing for a future pass (don't need to fix here): for read-only auth on `protected/`, the `allow.entity('identity').to(['read'])` rule is fully subsumed by the `allow.authenticated.to(['read'])` that also gets emitted, so it's effectively a redundant rule that slightly inflates the generated policy. It's a deliberate uniformity trade-off (and documented in the comment), just worth keeping in mind.

soberm (approved):

PR #14957, opened 2026-07-08
fix: use execFileSync for git commands in preInitSetup to prevent shell injection

svidgen:
## Summary

Replace `execSync` with `execFileSync` (array form) for the two git commands in `packages/amplify-cli/src/init-steps/preInitSetup.ts`:

- `git ls-remote <url>` in `validateGithubRepo()`
- `git clone <url> .` in `cloneRepo()`

## Motivation

Using the array-form `execFileSync` avoids spawning a shell and passes arguments directly to the executable. This is the recommended Node.js `child_process` API when constructing commands with variable inputs, as it avoids potential issues with special characters in URLs being interpreted by the shell.

## Changes

- Import `execFileSync` from `child_process`
- Replace `execSync(\`git ls-remote ${repoUrl}\`)` with `execFileSync('git', ['ls-remote', repoUrl])`
- Replace `execSync(\`git clone ${repoUrl} .\`)` with `execFileSync('git', ['clone', repoUrl, '.'])`

## Testing

- Pre-commit hooks (prettier, eslint, commitlint) pass
- No functional change in behavior for valid URLs

### DIFF
--- a/packages/amplify-cli/src/init-steps/preInitSetup.ts
+++ b/packages/amplify-cli/src/init-steps/preInitSetup.ts
@@ -1,5 +1,5 @@
 import { $TSContext, AmplifyError, getPackageManager, LocalEnvInfo, pathManager } from '@aws-amplify/amplify-cli-core';
-import { execSync } from 'child_process';
+import { execFileSync, execSync } from 'child_process';
 import * as fs from 'fs-extra';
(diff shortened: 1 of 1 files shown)


svidgen:
Closing — superseded by #14960 which was merged on Jul 14.

PR #14959, opened 2026-07-13
chore: add patch-package postinstall for node24 inquirer prompts

sarayev:
#### Description of changes

Fixes the amplify-cli e2e harness — plus one shipped runtime shim — so the suite runs correctly under Node 24, replacing fragile `patch-package` patches with code-owned fixes that live in the repository and compile with the rest of the codebase.

##### Inquirer Node 24 crash fix (ships to customers)

Adds `packages/amplify-cli/src/node24-inquirer-shim.ts`, a guarded, idempotent override of inquirer's base `UI.prototype.close`. On Node 24, tearing down a prompt could call `rl.pause()` on an already-closed readline interface and throw `ERR_USE_AFTER_CLOSE` (on Node <= 22 that call was a silent no-op). The shim delegates to the original `close()` while the readline is open and otherwise replicates the teardown without touching `pause()`. It is installed as the first statement of `run()` in `index.ts` so it is in place before any prompt fires. Because it lives in `src/` and ships in compiled `lib/`, it protects real CLI customers on Node 24, not just CI.

##### node-pty ConPTY fix (test harness)

Sets `useConpty: false` at the `pty.spawn()` call in `amplify-e2e-core`'s `asciinema-recorder.ts`. ConPTY's `AttachConsole` path can throw during PTY teardown on Node 24 + Windows; opting out keeps e2e recordings stable. This is a code-owned, version-agnostic replacement for a previous patch-package patch.

##### Tolerant prompt handling for the GraphQL codegen prompt

Adds a `sendConfirmNoIfPrompt` step and an `optional` step flag to `amplify-e2e-core`'s `nexpect.ts`. `nexpect` now drops leading optional steps that never matched instead of hanging on stdin or failing the chain. This is applied across the `amplifyPush.ts` helpers and `rebuildApi` in `categories/api.ts` to answer "N" to the "generate code for your newly created GraphQL API (Y/n)" prompt that only appears when an API exists — fixing silent hangs (api_6c, searchable_datastore, geo shards) without breaking non-API push flows that never emit the prompt.

##### SMS shard ordering

Extends `KNOWN_RED_SHARD_FRAGMENTS` in `scripts/split-e2e-tests-codebuild.ts` to pin the Pinpoint-SMS notifications shards into terminal (no-successor) slots, so their known environmental flakes don't cascade a STOP onto dependent shards. The regenerated `codebuild_specs/e2e_workflow_{linux,windows}_generated.yml` reflect the new ordering.

##### patch-package and CI tooling

The e2e fixes above are now code-owned rather than carried as `patch-package` patches. `patch-package` is added as a devDependency and invoked from `postinstall`, but the `patches/` directory ships no patch files, so the postinstall step is effectively a no-op. `scripts/cloud-pr.sh` also gains a `CLOUD_PR_SKIP_MWINIT` escape hatch for already-authenticated / non-interactive runs and drops `--debug-session-enabled` from the batch build call.

#### Issue #, if available

N/A

A split e2e run on this branch (Linux , plus Windows) is green except for pre-existing failures. Using a clean dev-HEAD baseline (commit `4853342dec`), all remaining failures were confirmed to pre-exist on `dev` and to not be introduced by this PR:

- Transformer v1 → v2 `GraphQLSchema` logical-ID collision (function_migration, searchable_migration).
- Flaky migration tests (auth_migration_push_pull_2, auth_5f_model_migration) — the same commit passed one `dev` run and failed another.
- Known Pinpoint-SMS environmental flakes.

This PR additionally fixes the Node 24 harness failures it targets (api_9a / api_6c, searchable_datastore, and the geo prompt-hangs). No regressions were observed.

#### Checklist

- [x] PR description included
- [ ] `yarn test` passes


soberm (approved):

svidgen (approved):
Seems OK. But, the patch dep change is actually a bit confusing. 🤔

PR #14962, opened 2026-07-15
fix(amplify-category-storage): make IAM policy names environment-specific

### LINKED ISSUE #14961: Storage deploy fails with 'Policy resource was already managed by another stack' after CFN enforcement change (Jan 2026)
## Description

As of January 30, 2026, CloudFormation enforces single-stack ownership of IAM inline policies. This breaks `amplify push` for Gen1 projects with **multiple environments** that share storage resources targeting the **same IAM role** (e.g., when using imported Identity Pool/auth roles).

The deploy fails with:
```
The following resource(s) failed to update: [S3AuthUploadPolicy, S3AuthPublicPolicy, S3AuthProtectedPolicy, S3AuthReadPolicy, S3AuthPrivatePolicy]

Policy resource was already managed by another stack or another resource in the current stack. Found stacks: [amplify-<app>-<env2>-storage...|S3AuthPrivatePolicy]
```

## Root Cause

When `amplify env add` creates a new environment, it copies `cli-inputs.json` verbatim — including the `policyUUID` field. This UUID is generated once (in `s3-defaults.ts` via `buildShortUUID()`) when storage is first added to the project. All environments therefore produce identical policy names (e.g., `Private_policy_ad0f22e8`).

This only becomes a problem when **both** conditions are met:
1. **Same PolicyName** — guaranteed by the shared `policyUUID` across envs
2. **Same target IAM role** — occurs when environments share auth roles (e.g., imported Identity Pool with shared authenticated/unauthenticated roles, or manually configured shared roles)

In the common case where each environment has its own auth roles (the default for `amplify add auth`), the identical policy names land on *different* roles, and CFN does not conflict. This is why relatively few customers hit it — it requires a shared-role configuration.

## Affected Code

- `packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts` (lines 115-121) — policy names use `policyUUID` without envName differentiation
- `amplify env add` clones `cli-inputs.json` (containing `policyUUID`) to new environments unchanged

## Workaround

Use `amplify override storage` in each affected environment to make policy names unique:

```typescript
// override.ts
import { AmplifyS3ResourceTemplate } from "@aws-amplify/cli-extensibility-helper";

export function override(resources: AmplifyS3ResourceTemplate) {
  const envName = "YOUR_ENV_NAME"; // e.g., "dev", "prod"
  resources.s3AuthPrivatePolicy.policyName += `-${envName}`;
  resources.s3AuthProtectedPolicy.policyName += `-${envName}`;
  resources.s3AuthPublicPolicy.policyName += `-${envName}`;
  resources.s3AuthReadPolicy.policyName += `-${envName}`;
  resources.s3AuthUploadPolicy.policyName += `-${envName}`;
}
```

Then run `amplify push`. Repeat for each environment with a different `envName` value.

## Suggested Fix

Append the environment name to the `policyUUID` when generating CFN parameters in `s3-stack-transform.ts`:

```typescript
// Before:
this.cfnInputParams.s3PrivatePolicy = `Private_policy_${userInput.policyUUID}`;

// After:
this.cfnInputParams.s3PrivatePolicy = `Private_policy_${userInput.policyUUID}_${envName}`;
```

This prevents new occurrences. Existing affected stacks still need the override workaround above.

## Related

- #10098 — same root cause (shared policyUUID + shared roles), different symptom (policy content overwriting rather than CFN rejection). A fix that appends envName to policy names would address both issues.

## Environment

- Amplify CLI: current (latest on `dev`)
- Trigger: CFN enforcement change Jan 30, 2026
- Frequency: Low — only affects projects with shared auth roles across environments (imported Identity Pool or manual role configuration)

w3lld1:
#### Description of changes

I made generated S3 IAM policy names environment-specific by appending the current Amplify environment name to the existing policy UUID. This prevents two environments that share imported auth roles from trying to manage identically named inline policies.

I also added a focused regression test for the generated policy names and updated the existing stack-transform expectations.

#### Issue #, if available

Fixes #14961

- `yarn lerna run build --scope @aws-amplify/amplify-category-storage --include-dependencies --concurrency 4`
- `node ../../node_modules/jest/bin/jest.js --runInBand --coverage=false` from `packages/amplify-category-storage` (15 suites, 48 tests)
- Prettier and ESLint checks for all changed TypeScript files

I did not run cloud E2E tests or the macOS CI job locally.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes for the changed package

### DIFF
--- a/packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
+++ b/packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
@@ -112,13 +112,14 @@ export class AmplifyS3ResourceStackTransform {
     if (userInput.adminTriggerFunction?.triggerFunction && userInput.adminTriggerFunction.triggerFunction !== 'NONE') {
       this.cfnInputParams.adminTriggerFunction = userInput.adminTriggerFunction.triggerFunction;
     }
-    this.cfnInputParams.s3PrivatePolicy = `Private_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.s3ProtectedPolicy = `Protected_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.s3PublicPolicy = `Public_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.s3ReadPolicy = `read_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.s3UploadsPolicy = `Uploads_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.authPolicyName = `s3_amplify_${userInput.policyUUID}`;
-    this.cfnInputParams.unauthPolicyName = `s3_amplify_${userInput.policyUUID}`;
+    const policyNameSuffix = `${userInput.policyUUID}_${this.context.amplify.getEnvInfo().envName}`;
+    this.cfnInputParams.s3PrivatePolicy = `Private_policy_${policyNameSuffix}`;
+    this.cfnInputParams.s3ProtectedPolicy = `Protected_policy_${policyNameSuffix}`;
+    this.cfnInputParams.s3PublicPolicy = `Public_policy_${policyNameSuffix}`;
+    this.cfnInputParams.s3ReadPolicy = `read_policy_${policyNameSuffix}`;
+    this.cfnInputParams.s3UploadsPolicy = `Uploads_policy_${policyNameSuffix}`;
+    this.cfnInputParams.authPolicyName = `s3_amplify_${policyNameSuffix}`;
+    this.cfnInputParams.unauthPolicyName = `s3_amplify_${policyNameSuffix}`;
     this.cfnInputParams.AuthenticatedAllowList = this._getAuthGuestListPermission(S3PermissionType.READ, userInput.authAccess);
     this.cfnInputParams.GuestAllowList = this._getAuthGuestListPermission(S3PermissionType.READ, userInput.guestAccess);
     this.cfnInputParams.s3PermissionsAuthenticatedPrivate = this._getPublicPrivatePermissions(
--- a/packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
+++ b/packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
@@ -112,20 +112,39 @@ export class AmplifyS3ResourceStackTransform {
     if (userInput.adminTriggerFunction?.triggerFunction && userInput.adminTriggerFunction.triggerFunction !== 'NONE') {
       this.cfnInputParams.adminTriggerFunction = userInput.adminTriggerFunction.triggerFunction;
     }
-    // Policy names must be unique per environment. When multiple environments share the
-    // same IAM role (e.g. an imported Identity Pool with shared auth/unauth roles), reusing
-    // the same policyUUID across environments produces identically named inline policies on
-    // the same role, which CloudFormation's single-stack ownership enforcement rejects with
-    // "Policy resource was already managed by another stack". The environment name is read
-    // once and validated so the suffix never silently degrades to a shared "_undefined".
-    const envName = this.context.amplify.getEnvInfo()?.envName;
-    if (typeof envName !== 'string' || envName.length === 0) {
-      throw new AmplifyError('EnvironmentNotInitializedError', {
-        message: 'Cannot determine the current Amplify environment name while generating S3 IAM policy names.',
-        resolution: `Run 'amplify init' or 'amplify env checkout <env>' in the root of your app directory to select an environment, then try again.`,
-      });
+    // Policy names carry the app-wide policyUUID. That alone is only insufficient when the
+    // authenticated/unauthenticated IAM roles are SHARED across environments, which happens
+    // when auth is an imported Cognito Identity Pool: two environments' storage stacks then
+    // try to manage identically named inline policies on the same role, and CloudFormation's
+    // single-stack ownership enforcement rejects the deploy ("Policy resource was already
+    // managed by another stack"). Appending the environment name keeps them distinct.
+    //
+    // For the default (managed) auth case each environment owns its own auth roles, so the
+    // identical policy names land on different roles and never collide. We deliberately keep
+    // the legacy `${policyUUID}` name there: a PolicyName change is replace-on-update for
+    // AWS::IAM::Policy, so suffixing unconditionally would force a policy replacement on
+    // EVERY existing storage app's next push -- churn (and risk) far beyond the shared-role
+    // bug this fixes. Gating on imported auth limits the rename to the environments that
+    // actually need it (and whose deploys are already failing).
+    const authResources = this.context.amplify.getProjectMeta?.()?.auth ?? {};
+    const hasImportedAuth = Object.values(authResources).some(
+      (authResource: $TSAny) => authResource?.service === 'Cognito' && authResource?.serviceType === 'imported',
+    );
+    let policyNameSuffix = userInput.policyUUID;
+    if (hasImportedAuth) {
+      // Imported auth shares roles across environments, so the env name is required to keep
+      // policy names distinct. generateCfnInputParameters only runs for a checked-out
+      // environment (push / add / update / override / export), so envName is expected here;
+      // fail fast with a clear error rather than silently producing a shared "_undefined".
+      const envName = this.context.amplify.getEnvInfo()?.envName;
+      if (typeof envName !== 'string' || envName.length === 0) {
+        throw new AmplifyError('EnvironmentNotInitializedError', {
+          message: 'Cannot determine the current Amplify environment name while generating S3 IAM policy names for imported auth.',
+          resolution: `Run 'amplify init' or 'amplify env checkout <env>' in the root of your app directory to select an environment, then try again.`,
+        });
+      }
+      policyNameSuffix = `${userInput.policyUUID}_${envName}`;
     }
-    const policyNameSuffix = `${userInput.policyUUID}_${envName}`;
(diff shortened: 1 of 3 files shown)


packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
```diff
@@ -112,13 +112,14 @@ export class AmplifyS3ResourceStackTransform {
...
       this.cfnInputParams.adminTriggerFunction = userInput.adminTriggerFunction.triggerFunction;
     }
-    this.cfnInputParams.s3PrivatePolicy = `Private_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.s3ProtectedPolicy = `Protected_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.s3PublicPolicy = `Public_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.s3ReadPolicy = `read_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.s3UploadsPolicy = `Uploads_policy_${userInput.policyUUID}`;
-    this.cfnInputParams.authPolicyName = `s3_amplify_${userInput.policyUUID}`;
-    this.cfnInputParams.unauthPolicyName = `s3_amplify_${userInput.policyUUID}`;
+    const policyNameSuffix = `${userInput.policyUUID}_${this.context.amplify.getEnvInfo().envName}`;
```
sharonyajain: **New unguarded getEnvInfo().envName dependency yields '_undefined' names or a throw when env info is absent**

Name generation now hard-depends on `this.context.amplify.getEnvInfo().envName`, a call this method never made before (the existing test had to add `getEnvInfo` to its mock context). There is no check that getEnvInfo() returns an object or that envName is non-empty.

```
const policyNameSuffix = `${userInput.policyUUID}_${this.context.amplify.getEnvInfo().envName}`;
```

**Why it matters:** getEnvInfo() returns no/blank envName (uninitialized env, headless/export/mock paths) → suffix becomes `${uuid}_undefined` → all such deployments again share one policy name and re-collide (the very bug being fixed), OR getEnvInfo() throws → transform now fails on a path that previously succeeded (regression).

**Suggestion:** Read envName once, assert it is a non-empty string, and fail fast with a clear AmplifyError if it is missing, we can add a test for the missing-envName case alongside the happy-path test.

packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
```diff
@@ -112,13 +112,27 @@ export class AmplifyS3ResourceStackTransform {
...
+    // "Policy resource was already managed by another stack". The environment name is read
+    // once and validated so the suffix never silently degrades to a shared "_undefined".
+    const envName = this.context.amplify.getEnvInfo()?.envName;
+    if (typeof envName !== 'string' || envName.length === 0) {
+      throw new AmplifyError('EnvironmentNotInitializedError', {
+        message: 'Cannot determine the current Amplify environment name while generating S3 IAM policy names.',
+        resolution: `Run 'amplify init' or 'amplify env checkout <env>' in the root of your app directory to select an environment, then try again.`,
+      });
+    }
+    const policyNameSuffix = `${userInput.policyUUID}_${envName}`;
```
sarayev: Thanks for tracking this down, the root cause on #14961 is spot on and appending the env name is the right way to make the policy names unique per environment.

One thing I want to flag before we ship: this also changes the policy names for environments that are *already* deployed, and I don't think that's free. When a push runs after the upgrade, CloudFormation sees the inline policy name change (e.g. `s3_amplify_abc123` becomes `s3_amplify_abc123_dev`) and treats it as a replace: it deletes the old policy and creates the new one, rather than editing it in place. During that swap there's a brief window where the auth/unauth role has no S3 policy attached, so live users can get 403s mid-deploy. And it hits *every* existing storage app on the first push, not just the shared-role case we're trying to fix.

Could we make this env-aware so existing environments aren't silently affected? A couple of options:
- Only apply the `_${envName}` suffix to new resources (or specifically the shared-role scenario), and keep the old name for environments already using it.
- Or, if we do want it applied everywhere, call out the transient disruption in the changeset so people know to push during a quiet window.

Not blocking the idea at all, just want to make sure we're not turning a fix for a few customers into a brief outage for everyone on upgrade. (major)

packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
```diff
@@ -112,13 +112,27 @@ export class AmplifyS3ResourceStackTransform {
...
-    this.cfnInputParams.unauthPolicyName = `s3_amplify_${userInput.policyUUID}`;
+    // Policy names must be unique per environment. When multiple environments share the
+    // same IAM role (e.g. an imported Identity Pool with shared auth/unauth roles), reusing
+    // the same policyUUID across environments produces identically named inline policies on
+    // the same role, which CloudFormation's single-stack ownership enforcement rejects with
+    // "Policy resource was already managed by another stack". The environment name is read
+    // once and validated so the suffix never silently degrades to a shared "_undefined".
+    const envName = this.context.amplify.getEnvInfo()?.envName;
+    if (typeof envName !== 'string' || envName.length === 0) {
+      throw new AmplifyError('EnvironmentNotInitializedError', {
```
sarayev: Nice touch guarding against the empty env name instead of letting it silently become `_undefined`.

The one case I'd double-check: is `generateCfnInputParameters` ever called when there legitimately isn't an active environment? I'm thinking `amplify export`, headless/CI runs, or template generation before an env is fully initialized. Before this change those paths would just produce a template with the old-style name and keep working; now they'd throw and fail. If we know for sure this only ever runs during a real push, then the throw is perfect, and maybe a one-line comment saying so would save the next person the same question. If it can run without an env, a softer fallback (use the legacy name plus a warning, and only throw on the deploy path) would keep those flows working. (minor)

packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
sharonyajain: Thank you for pointing this out. You're right that suffixing every policy name would force an AWS::IAM::Policy replacement on the next push for every existing storage app, since PolicyName is replace-on-update. That's disproportionate to a bug that only bites shared-role setups, so I've made it backward-compatible in b9e589cc48.

The suffix is now gated on imported auth. The collision only happens when the authenticated/unauthenticated IAM roles are shared across environments, which is the imported Cognito Identity Pool case (meta.auth.<name>.serviceType === 'imported'). So:

- Managed auth (each env owns its roles — the default and the vast majority): we keep the legacy s3_amplify_${uuid} name. No rename, no policy replacement on upgrade.
- Imported auth (shared roles — the case that actually collides, and whose deploys are already failing): we append _${envName} to make the names unique per environment.

packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
sharonyajain: I traced the callers: `generateCfnInputParameters` only runs for a checked-out environment — push (transformCategoryStack → transformS3ResourceStack), amplify add/update storage, amplify override storage, and amplify export all operate on the current env, so envName is present on every real path.

With the backward-compat change in b9e589cc48 this is now moot for the common case anyway: the env name is only read on the imported-auth path, so managed-auth flows never touch getEnvInfo() and behave exactly as before. On the imported path the env name is genuinely required (a blank one would degrade to a shared _undefined and re-collide), so failing fast with a clear error is the right call there. I've added a one-line comment saying so.

packages/amplify-category-storage/src/provider-utils/awscloudformation/cdk-stack-builder/s3-stack-transform.ts
```diff
@@ -112,13 +112,46 @@ export class AmplifyS3ResourceStackTransform {
...
+    // For the default (managed) auth case each environment owns its own auth roles, so the
+    // identical policy names land on different roles and never collide. We deliberately keep
+    // the legacy `${policyUUID}` name there: a PolicyName change is replace-on-update for
+    // AWS::IAM::Policy, so suffixing unconditionally would force a policy replacement on
+    // EVERY existing storage app's next push -- churn (and risk) far beyond the shared-role
+    // bug this fixes. Gating on imported auth limits the rename to the environments that
+    // actually need it (and whose deploys are already failing).
+    const authResources = this.context.amplify.getProjectMeta?.()?.auth ?? {};
+    const hasImportedAuth = Object.values(authResources).some(
+      (authResource: $TSAny) => authResource?.service === 'Cognito' && authResource?.serviceType === 'imported',
```
sarayev, line 131: Small one, non-blocking: the `$TSAny` on the `authResource` param is the only cast in the new block. If `getProjectMeta().auth` has a shared type for its entries (an auth-resource meta interface), typing it as that (or even `{ service?: string; serviceType?: string }`) would drop the cast and keep the optional-chaining checks. Fine to leave as-is if there's no ready type. (nit)

sarayev (commented):
Thanks for the quick turnaround, both of my concerns are addressed nicely.

Gating the suffix on imported auth is exactly the backward-compatible shape I was hoping for: managed-auth apps keep the legacy name and see no policy replacement on upgrade, and only the shared-role environments (whose deploys are already failing) get the env-scoped name. Scoping the fail-fast to the imported path and tracing the callers to confirm envName is always present there also resolves the export/headless worry. The three tests (imported gets the suffix, managed keeps the legacy name, imported + blank env throws) lock the behavior in well. Looks good to me.

Left one small nit inline on the `$TSAny` cast, non-blocking.

sharonyajain (approved):

PR #14963, opened 2026-07-22
fix(amplify-provider-awscloudformation): preserve GraphQLSchema logical id on v1 to v2 migration

sarayev:
#### Description of changes

Deploy-time, migration-gated fix in the Gen1 CLI for the v1 to v2
`AWS::AppSync::GraphQLSchema` logical-ID collision.

On a v1 to v2 GraphQL transformer migration the schema resource's
logical ID changes from v1's hard-coded `GraphQLSchema` to v2's
CDK-hashed `GraphQLAPITransformerSchema<hash>`. Because the schema's
physical ID is `<apiId>GraphQLSchema` (one per API), CloudFormation
attempts a create-before-delete on the rename and fails with
`CREATE_FAILED: <apiId>GraphQLSchema already exists`, rolling back the
migration deployment.

The fix adds a whole-template pass in `preProcessCFNTemplate` that runs
**only** when the previously-deployed API template still has the schema
at logical ID `GraphQLSchema`. In that case it rewrites the hashed
logical ID back to `GraphQLSchema` and fixes every reference (`Ref`,
`DependsOn`, `Fn::GetAtt`, `Fn::Sub`) — so CloudFormation performs an
in-place update instead of a colliding replacement.

The rewrite is migration-gated. Born-v2 Gen1 apps, brand-new APIs,
non-API/nested-stack templates, and Gen2/CDK consumers are unaffected
and produce byte-identical output. The change is confined to
`amplify-provider-awscloudformation`; transformer packages are
untouched.

#### Issue #, if available

N/A

- 5 new unit tests for `preserveGraphQLSchemaLogicalId`, including a
  born-v2 byte-identical (no-op) case; all pass.
- Existing `amplify-provider-awscloudformation` suite green.
- Live v1 to v2 migration e2e on a real account reached
  `UPDATE_COMPLETE`, the schema resource stayed at
  `LogicalResourceId=GraphQLSchema`, and no logical-ID collision
  occurred.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes

### DIFF
--- a/packages/amplify-provider-awscloudformation/src/__tests__/pre-push-cfn-processor/graphql-schema-logical-id-preserver.test.ts
+++ b/packages/amplify-provider-awscloudformation/src/__tests__/pre-push-cfn-processor/graphql-schema-logical-id-preserver.test.ts
@@ -1,0 +1,147 @@
+import { pathManager, readCFNTemplate } from '@aws-amplify/amplify-cli-core';
+import { Template } from 'cloudform-types';
+import * as fs from 'fs-extra';
+import * as path from 'path';
+import { preserveGraphQLSchemaLogicalId } from '../../pre-push-cfn-processor/graphql-schema-logical-id-preserver';
+
+jest.mock('@aws-amplify/amplify-cli-core');
+jest.mock('fs-extra');
+
+const readCFNTemplate_mock = readCFNTemplate as jest.MockedFunction<typeof readCFNTemplate>;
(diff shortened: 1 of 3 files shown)


soberm (approved):

svidgen (commented):
This is odd. What's the reason this broke? This v1 -> v2 flow has been around for years.

What's the implication on running `amplify push --force`? (IIRC, this can be used to skip a missing or corrupt `#current-cloud-backend/` state.)

Issue #14810, opened 2026-04-24
(gen2-migration) `generate` does not preserve "Email based user registration/forgot password: Disabled" setting


dgandhi62:
### Amplify CLI Version

NA

### What operating system are you using?

NA

### Did you make any manual changes to the cloud resources managed by Amplify? Please describe the changes made.

NA

### Summary

When migrating an Amplify Gen1 auth resource that had **email-based verification disabled** (using SMS/TOTP as an alternative), the Gen2 migration tool generates a configuration that **re-enables email-based verification**, breaking feature parity with the original Gen1 setup.

---

### Gen1 Configuration (Source of Truth)

**CLI selection:**
```
Email based user registration/forgot password: Disabled (Uses SMS/TOTP as an alternative)
```

**`cli-inputs.json`:**
```json
{
  "cognitoConfig": {
    "autoVerifiedAttributes": ["phone_number"],
    "mfaConfiguration": "OFF",
    "smsVerificationMessage": "Your verification code is {####}",
    "emailVerificationSubject": "Your verification code",
    "emailVerificationMessage": "Your verification code is {####}",
    "requiredAttributes": ["email"]
  }
}
```

**CloudFormation template — `UserPool` resource:**
```json
{
  "AutoVerifiedAttributes": ["phone_number"],
  "SmsVerificationMessage": { "Ref": "smsVerificationMessage" },
  "SmsConfiguration": {
    "ExternalId": "userpo7ed36813_role_external_id",
    "SnsCallerArn": { "Fn::GetAtt": ["SNSRole", "Arn"] }
  },
  "UserAttributeUpdateSettings": {
    "AttributesRequireVerificationBeforeUpdate": ["phone_number"]
  }
}
```

**CloudFormation template — `SNSRole` resource (created specifically for SMS verification):**
```json
{
  "SNSRole": {
    "Type": "AWS::IAM::Role",
    "Properties": {
      "AssumeRolePolicyDocument": {
        "Statement": [{
          "Effect": "Allow",
          "Principal": { "Service": "cognito-idp.amazonaws.com" },
          "Action": ["sts:AssumeRole"]
        }]
      },
      "Policies": [{
        "PolicyDocument": {
          "Statement": [{
            "Effect": "Allow",
            "Action": ["sns:Publish"],
            "Resource": "*"
          }]
        }
      }]
    }
  }
}
```

**Gen1 behavior:**
- Account verification: SMS code sent to phone number
[...]

PR #14969, opened 2026-08-10
fix(cli): preserve phone-based sign-in in gen2 migration

sharonyajain:
#### Description of changes

  The Gen2 migration auth renderer (`packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts`) unconditionally emitted `loginWith.email` and never emitted `loginWith.phone`. As a result, Gen1 Cognito user pools that verify via `phone_number` (email verification disabled, using SMS/TOTP) were migrated to email-based login, breaking parity for sign-in, auto-verification, and forgot-password.

  This change derives the login mechanisms from the source pool's `UsernameAttributes`, `AliasAttributes`, and `AutoVerifiedAttributes`, emitting `phone` and/or `email` to match the Gen1 configuration. It falls back to email only when neither is present (since `defineAuth` requires at least one). Email is preserved as a required user attribute when it is not a sign-in mechanism.

  #### Issue #, if available

Addresses part of #14810 (the `loginWith` mechanism). The SMS configuration / SNS caller-role carry-over described in that issue is a separate concern (Gen2 `defineAuth` does not auto-provision an SNS role from `loginWith: { phone: true }`)
  and is tracked as a follow-up.

  - Corrected the existing `generates phone login` unit test snapshot, which previously asserted the buggy `loginWith: { email: true }` output for a phone-based pool.
  - Added a regression test for the exact issue scenario (email verification disabled, `AutoVerifiedAttributes: ['phone_number']`, email as a required attribute) asserting `phone: true` and no email login.
  - Updated the `discussions` and `media-vault` migration-app snapshots to reflect correct phone/combined login.
  - `yarn test` for the gen2-migration suite: 432/432 tests pass across 48 suites.

  #### Checklist

  - [x] PR description included
  - [x] `yarn test` passes

### DIFF
--- a/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/auth/auth.generator.test.ts
+++ b/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/auth/auth.generator.test.ts
@@ -298,7 +298,84 @@ describe('AuthGenerator', () => {
     expect(content).toContain('phone: true');
     expect(content).not.toContain('email: true');
     // email is preserved as a required user attribute, not a login mechanism
-    expect(content).toContain('email');
+    expect(content).toMatch(/userAttributes:\s*\{[^}]*email:\s*\{[^}]*required:\s*true/s);
+  });
+
+  // Login mechanisms should be derived from AliasAttributes when that is the
+  // populated sign-in signal (not just UsernameAttributes/AutoVerifiedAttributes).
+  it('generates phone login from AliasAttributes', async () => {
+    const gen1App = await createGen1App({
+      providers: { awscloudformation: { StackName: 'amplify-test-main-123456', Region: 'us-east-1' } },
+      auth: {
+        testAuth: {
+          service: 'Cognito',
+          output: {
+            UserPoolId: 'us-east-1_abc123',
+            AppClientIDWeb: 'webclient123',
+            AppClientID: 'client123',
+            IdentityPoolId: 'us-east-1:idpool',
+          },
+        },
+      },
+    });
+    jest.spyOn(gen1App.aws, 'fetchUserPool').mockResolvedValue({
+      AliasAttributes: ['phone_number'],
+      SchemaAttributes: [],
+    });
+    jest.spyOn(gen1App.aws, 'fetchMfaConfig').mockResolvedValue({});
+    jest.spyOn(gen1App.aws, 'fetchUserPoolClient').mockResolvedValue({});
+    jest.spyOn(gen1App.aws, 'fetchIdentityProviders').mockResolvedValue([]);
+    jest.spyOn(gen1App.aws, 'fetchIdentityGroups').mockResolvedValue([]);
+    jest.spyOn(gen1App.aws, 'fetchIdentityPool').mockResolvedValue({
+      IdentityPoolId: 'us-east-1:idpool',
+      IdentityPoolName: 'test-pool',
+      AllowUnauthenticatedIdentities: true,
+    });
+
+    const generator = new AuthGenerator(gen1App, backendGenerator, outputDir, authResource, logger);
+    const ops = await generator.plan();
+    await ops[0].execute();
+    const content = writtenFile('auth/resource.ts');
+    expect(content).toContain('phone: true');
+    expect(content).not.toContain('email: true');
+  });
+
+  // A pool that supports both email and phone sign-in must enable both in loginWith.
+  it('generates both email and phone login when the pool supports both', async () => {
+    const gen1App = await createGen1App({
+      providers: { awscloudformation: { StackName: 'amplify-test-main-123456', Region: 'us-east-1' } },
+      auth: {
+        testAuth: {
+          service: 'Cognito',
+          output: {
+            UserPoolId: 'us-east-1_abc123',
+            AppClientIDWeb: 'webclient123',
+            AppClientID: 'client123',
+            IdentityPoolId: 'us-east-1:idpool',
+          },
+        },
+      },
+    });
+    jest.spyOn(gen1App.aws, 'fetchUserPool').mockResolvedValue({
+      UsernameAttributes: ['email', 'phone_number'],
+      SchemaAttributes: [],
+    });
+    jest.spyOn(gen1App.aws, 'fetchMfaConfig').mockResolvedValue({});
+    jest.spyOn(gen1App.aws, 'fetchUserPoolClient').mockResolvedValue({});
+    jest.spyOn(gen1App.aws, 'fetchIdentityProviders').mockResolvedValue([]);
+    jest.spyOn(gen1App.aws, 'fetchIdentityGroups').mockResolvedValue([]);
+    jest.spyOn(gen1App.aws, 'fetchIdentityPool').mockResolvedValue({
+      IdentityPoolId: 'us-east-1:idpool',
+      IdentityPoolName: 'test-pool',
+      AllowUnauthenticatedIdentities: true,
+    });
+
+    const generator = new AuthGenerator(gen1App, backendGenerator, outputDir, authResource, logger);
+    const ops = await generator.plan();
+    await ops[0].execute();
+    const content = writtenFile('auth/resource.ts');
+    expect(content).toContain('email: true');
+    expect(content).toContain('phone: true');
   });
 
   it('generates email with verification options', async () => {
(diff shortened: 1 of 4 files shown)


soberm (commented):
Went through the diff. The fix looks solid overall. good regression test tied directly to #14810, and a sensible call to scope the SNS/SMS follow-up out rather than cramming it into this PR.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
```diff
@@ -659,10 +659,41 @@ export class AuthRenderer {
...
+   *
+   * `defineAuth` requires at least one of email/phone, so we fall back to email
+   * when the pool exposes neither (e.g. username-only sign-in), matching the
+   * previous default behavior.
+   */
+  private static deriveLoginMechanisms(userPool: UserPoolType): { readonly email: boolean; readonly phone: boolean } {
+    const signals = new Set<string>([
+      ...(userPool.UsernameAttributes ?? []),
+      ...(userPool.AliasAttributes ?? []),
+      ...(userPool.AutoVerifiedAttributes ?? []),
```
soberm: [question] `AutoVerifiedAttributes` isn't quite the same signal as `UsernameAttributes`/`AliasAttributes` above. Per the Cognito docs, auto-verify just controls whether Cognito sends a verification code on signup/update, independent of whether the attribute is actually usable to sign in. A pool could auto-verify both email and phone (e.g. for SMS-based password recovery) while only ever supporting plain username sign-in, and this would now emit both `email: true` and `phone: true` even though neither was a real login mechanism in Gen1. Looking at #14810, the repro does seem to only have `AutoVerifiedAttributes: ['phone_number']` with no alias/username attrs mentioned, so this matches what was asked for there. I just can't tell from the outside whether "auto-verified only" reliably implies "was actually usable to sign in" for the wider set of pools this needs to generalize to. Might be worth weighting `UsernameAttributes`/`AliasAttributes` higher and only falling back to `AutoVerifiedAttributes` when neither of those is populated at all, instead of OR-ing all three unconditionally.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
```diff
@@ -659,10 +659,41 @@ export class AuthRenderer {
...
+   * rather than being forced onto email verification.
+   *
+   * `defineAuth` requires at least one of email/phone, so we fall back to email
+   * when the pool exposes neither (e.g. username-only sign-in), matching the
+   * previous default behavior.
+   */
+  private static deriveLoginMechanisms(userPool: UserPoolType): { readonly email: boolean; readonly phone: boolean } {
+    const signals = new Set<string>([
+      ...(userPool.UsernameAttributes ?? []),
+      ...(userPool.AliasAttributes ?? []),
```
soberm: [minor] Since this heuristic is the crux of the fix, a few more direct test cases would help: nothing in the suite exercises this `AliasAttributes` source specifically (only `UsernameAttributes` and `AutoVerifiedAttributes` are covered), and there's no tight assertion for "both email and phone enabled together" — that's only implied by the media-vault snapshot fixture.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
```diff
@@ -671,10 +702,18 @@ export class AuthRenderer {
...
+        assignments.push(
+          factory.createPropertyAssignment(factory.createIdentifier('email'), this.createEmailDefinitionObject(emailOptions)),
+        );
+      } else {
+        assignments.push(factory.createPropertyAssignment(factory.createIdentifier('email'), factory.createTrue()));
+      }
+    }
+
+    if (phoneEnabled) {
+      assignments.push(factory.createPropertyAssignment(factory.createIdentifier('phone'), factory.createTrue()));
```
soberm, line 721: [question] Two related questions on the phone side. First: is there a Gen2 `loginWith.phone` equivalent to email's `verificationEmailSubject`/`verificationEmailBody` for customizing the SMS body? Gen1's `SmsVerificationMessage`/`SmsAuthenticationMessage` aren't carried over here — fine if Gen2 has no such option, just flagging in case there's an `smsMessage`-ish field that deserves the same treatment as email. Second: the PR description says Gen2 `defineAuth` doesn't auto-provision an SNS role from `phone: true` and defers that as follow-up — worth double-checking against the CDK `aws-cognito` `UserPool` construct, since it's historically auto-created a default SMS role when phone verification is configured and no `smsRole` is explicitly passed. If that holds here, part of the follow-up issue might already be resolved by this PR.

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/auth/auth.generator.test.ts
```diff
@@ -258,6 +258,49 @@ describe('AuthGenerator', () => {
...
+    });
+
+    const generator = new AuthGenerator(gen1App, backendGenerator, outputDir, authResource, logger);
+    const ops = await generator.plan();
+    await ops[0].execute();
+    const content = writtenFile('auth/resource.ts');
+    expect(content).toContain('phone: true');
+    expect(content).not.toContain('email: true');
+    // email is preserved as a required user attribute, not a login mechanism
+    expect(content).toContain('email');
```
soberm: [nit] This last assertion is pretty loose — `'email'` as a substring matches almost anything containing that word, so it doesn't really pin down that `userAttributes: { email: {...} }` came out right. Something like checking for `email:` inside the `userAttributes` block (or a small regex) would tie it more directly to what the comment above says it's testing.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
sharonyajain: Thank you for calling this out. Reworked deriveLoginMechanisms to consult UsernameAttributes/AliasAttributes first and only fall back to AutoVerifiedAttributes when neither is populated. The real fixture pools all have UsernameAttributes, so their snapshots are unchanged.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
sharonyajain: Added two direct cases: generates phone login from AliasAttributes, and generates both email and phone login when the pool supports both (asserting both email: true and phone: true rather than leaving it implied by the media-vault snapshot)

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/auth/auth.generator.test.ts
sharonyajain: Replaced the toContain('email') with `toMatch(/userAttributes:\s*\{[^}]*email:\s*\{[^}]*required:\s*true/s)` so it actually pins email inside the userAttributes block as a required attribute

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
sharonyajain: 1. SMS message body: Yes, PhoneNumberLogin's object form supports `verificationMessage?: (createCode: () => string) => string (plus otpLogin)` from amplify-backend/packages/auth-construct/src/types.ts. Unlike email it's a single combined message (no subject), and it's a function that must call createCode() rather than embed Gen1's {####}. Carrying Gen1's `SmsVerificationMessage` over therefore needs a placeholder→createCode() transform plus synth-time validation, so I've kept it in the SMS follow-up rather than this PR.

2.  Yes. CDK's UserPool construct auto-creates a default SMS caller role (assumed by cognito-idp.amazonaws.com, inline sns:Publish) whenever signInAliases.phone or autoVerify.phone is true and no smsRole is passed — 
   aws-cdk-lib/aws-cognito/lib/user-pool.ts, smsConfiguration(): roleRequired = mfa.includes('SMS_MFA') || phoneVerification. Since this PR now emits loginWith: { phone: true } (→ signInAliases.phone/autoVerify.phone), the SNS role is provisioned automatically. So the "missing SNS role" concern is effectively resolved by this fix; the follow-up narrows to just the custom SMS message templates. I've updated the follow-up scope

soberm (approved):

Simone319 (approved):

PR #14971, opened 2026-08-18
fix(gen2-migration): skip Identity Pool when Gen1 auth is User Pool-only

### LINKED ISSUE #14742: (gen2-migration) `generate` adds Identity Pool when Gen1 auth was configured as `User Sign-Up & Sign-In only` (User Pool only)

_No response_

_No response_

### Amplify CLI Version

NA

NA

NA

### Description:

When migrating a Gen1 Amplify project to Gen2 using amplify gen2-migration generate, the migration tool creates an Identity Pool even though the Gen1 project was configured with "User Sign-Up & Sign-In only" — which provisions only a Cognito User Pool with no Identity Pool.

My Gen1 setup used a Cognito User Pool authorizer on API Gateway - the user signs in, gets a JWT, and passes it in the Authorization header. No IAM credentials needed, no Identity Pool needed.

### Expected behavior:

The generated Gen2 code should reflect the original Gen1 auth configuration — a User Pool with no Identity Pool. The defineAuth output should not include an Identity Pool, matching the "User Sign-Up & Sign-In only" selection.

### Actual behavior:

The generated backend.ts includes Identity Pool configuration:

```typescript
const cfnIdentityPool = backend.auth.resources.cfnResources.cfnIdentityPool;
cfnIdentityPool.allowUnauthenticatedIdentities = false;
```

This changes the auth architecture from what was originally configured. The Gen1 project used a Cognito User Pool authorizer for the REST API, but the migration switches to IAM authorization, which requires an Identity Pool.

### Additional findings

**Orphaned IAM roles:** Removing the Identity Pool from the construct tree does not automatically remove the associated authenticated/unauthenticated IAM roles. These become orphaned resources that still exist in the account but serve no purpose, since they were only needed for Identity Pool credential vending.

**Codegen breaks without Identity Pool:** `ampx generate outputs` (and related codegen) expects an Identity Pool to exist. If you remove it via CDK escape hatches, codegen will throw an error because it can't resolve the `identityPoolId` output. This means even if you successfully remove the Identity Pool from the stack, the developer workflow breaks.

**Authorization model mismatch:** The migration switches the authorization model from JWT-based (User Pool authorizer) to IAM-based (which requires an Identity Pool). This is architecturally different from the original Gen1 setup.

### CDK Escape Hatch Workaround

For anyone hitting this, here's the workaround to remove the Identity Pool at the CDK level in `backend.ts`:

```typescript
import { CfnOutput, CfnParameter } from 'aws-cdk-lib';

// --- Remove Identity Pool ---
const amplifyAuth = backend.auth.stack.node.findChild('amplifyAuth');
const cfnIdentityPool = backend.auth.resources.cfnResources.cfnIdentityPool;

// Replace the Identity Pool with a dummy CfnParameter using the same logical ID
// to prevent CloudFormation from erroring on the missing resource during update
const logicalId = backend.auth.stack.resolve((cfnIdentityPool as any).logicalId) as string;
new CfnParameter(backend.auth.stack, 'IdentityPoolIdParam', {
  type: 'String',
  default: 'NONE',
});
const param = backend.auth.stack.node.findChild('IdentityPoolIdParam') as CfnParameter;
param.overrideLogicalId(logicalId);

// Remove Identity Pool and its role attachment from the construct tree
amplifyAuth.node.tryRemoveChild('IdentityPool');
amplifyAuth.node.tryRemoveChild('IdentityPoolRoleAttachment');

// Override CfnOutput values instead of removing them —
// removing them breaks downstream references and codegen
const idPoolOutput = backend.stack.node.findChild('identityPoolId') as CfnOutput;
(idPoolOutput as any).value = 'NONE';
const allowUnauthOutput = backend.stack.node.findChild('allowUnauthenticatedIdentities') as CfnOutput;
(allowUnauthOutput as any).value = 'false';
```
Note: Codegen may error - `ampx generate outputs` expects a real Identity Pool ID. Setting it to `'NONE'` avoids deploy failures but codegen tooling may not handle it gracefully.

### Reproduction steps

`Cognito User Pool (no Identity Pool)`

- Sign-in: Email
- MFA: Off
- Email verification: Enabled, code-based ("Your verification code is {####}")
- Password policy: Default (min 8 chars, no special requirements)
- Required sign-up attributes: Email
- Refresh token: 30 days
- OAuth: Disabled
- Lambda triggers: None
- User Pool Groups: None
- Admin queries API: None

`Lambda function (app2f4d22a3b)`

- Runtime: Node.js
- Returns authenticated user's Cognito claims (email, sub, auth_time) from the API Gateway request context
- Returns 401 if no claims present

`REST API via API Gateway (api3c3877d7)`

- Single path: /auth-test
- Backed by the Lambda function above
- Authorization: Cognito User Pool authorizer (JWT-based, not IAM)
- Authenticated users: GET access only
- CORS: Enabled, all origins

### Project Identifier

_No response_

### Log output

<details>

```


```

</details>

### Additional information

_No response_

sharonyajain:
## Problem

When migrating a Gen1 Amplify project to Gen2 using `amplify gen2-migration generate`, the migration tool creates Identity Pool configuration even when the Gen1 project was configured with "User Sign-Up & Sign-In only" — which provisions only a Cognito User Pool with no Identity Pool dependency.

This changes the auth architecture from JWT-based (User Pool authorizer) to IAM-based (which requires an Identity Pool), creates orphaned IAM roles, and breaks `ampx generate outputs` if the Identity Pool is later removed via CDK escape hatches.

Closes #14742

## Root Cause

`auth.generator.ts` unconditionally called `this.gen1App.resourceMetaOutput(this.resource, 'IdentityPoolId')` which throws if the key is absent. Even when present (Gen1 always creates an Identity Pool), the renderer emitted Identity Pool escape-hatch code without checking whether the pool was actually in use.

## Changes

- **`gen1-app.ts`**: Add `tryResourceMetaOutput()` — a non-throwing variant that returns `undefined` for absent output keys
- **`auth.generator.ts`**: Use `tryResourceMetaOutput` for `IdentityPoolId`; skip `fetchIdentityPool` when absent
- **`auth.renderer.ts`**: Guard all Identity Pool escape hatches (`cfnIdentityPool`, `cognitoIdentityProviders` push, `addPropertyDeletionOverride`) on `!!options.identityPool`
- **`auth.generator.test.ts`**: Add regression test for User Pool-only configuration (no `IdentityPoolId` in output)

## Testing

- New test: `generates auth without Identity Pool when IdentityPoolId is absent`
- Verifies `fetchIdentityPool` is never called
- Verifies generated code contains no `cfnIdentityPool`, `cognitoIdentityProviders`, or `cognitoProviders` references
- Existing tests still pass (they all provide `IdentityPoolId` in their meta)

### DIFF
--- a/packages/amplify-cli/package.json
+++ b/packages/amplify-cli/package.json
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "test": "jest --logHeapUsage",
+    "test": "jest --logHeapUsage --forceExit",
     "postinstall": "node scripts/post-install.js",
     "watch": "tsc -w",
     "clean": "rimraf ./lib tsconfig.tsbuildinfo",
--- a/packages/amplify-cli/package.json
+++ b/packages/amplify-cli/package.json
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "test": "jest --logHeapUsage --forceExit",
+    "test": "jest --logHeapUsage",
     "postinstall": "node scripts/post-install.js",
     "watch": "tsc -w",
     "clean": "rimraf ./lib tsconfig.tsbuildinfo",
--- a/packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
@@ -218,8 +218,7 @@ export class Gen1App {
    * Returns a resource output value from amplify-meta.json.
    */
   public resourceMetaOutput(resource: DiscoveredResource, outputKey: string): string {
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- untyped amplify-meta.json
-    const value = (this._meta as any)[resource.category]?.[resource.resourceName]?.output?.[outputKey];
+    const value = this.tryResourceMetaOutput(resource, outputKey);
     if (value === undefined) {
       throw new AmplifyError('ResourceMetaOutputNotFoundError', {
         message: `Missing output '${outputKey}' for resource '${resource.resourceName}' in category '${resource.category}'`,
--- a/packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
@@ -229,6 +229,15 @@ export class Gen1App {
   }
 
   /**
+   * Returns a resource output value from amplify-meta.json, or undefined if absent.
+   * Use for optional outputs that may not exist in all Gen1 configurations.
+   */
+  public tryResourceMetaOutput(resource: DiscoveredResource, outputKey: string): string | undefined {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- untyped amplify-meta.json
+    return (this._meta as any)[resource.category]?.[resource.resourceName]?.output?.[outputKey];
+  }
+
+  /**
    * Returns the name of the single resource in a category matching a service type.
    */
   public singleResourceName(category: string, service: string): string {
--- a/packages/amplify-cli/src/commands/gen2-migration/_common/spinning-logger.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/_common/spinning-logger.ts
@@ -8,6 +8,8 @@ import chalk from 'chalk';
  */
 export class SpinningLogger {
   private static readonly SEPARATOR = ' → ';
+  private static readonly instances = new Set<SpinningLogger>();
+  private static sigintRegistered = false;
   private readonly segments: string[] = [];
   private readonly spinner: AmplifySpinner;
   private readonly debugMode: boolean;
@@ -17,14 +19,23 @@ export class SpinningLogger {
     this.debugMode = options?.debug ?? globalIsDebug;
     this.spinner = new AmplifySpinner();
 
-    // Restore the cursor if the process is interrupted while the spinner is active
-    process.on('SIGINT', () => {
-      if (this.spinnerActive) {
-        this.spinner.stop();
-      }
-      // Registering a SIGINT handler disables the default exit behavior, so we must exit explicitly.
-      process.exit(130);
-    });
+    SpinningLogger.instances.add(this);
+
+    // Register the SIGINT handler only once to avoid accumulating listeners
+    // across multiple SpinningLogger instances (which causes
+    // MaxListenersExceededWarning and prevents clean process exit in tests).
+    if (!SpinningLogger.sigintRegistered) {
+      SpinningLogger.sigintRegistered = true;
+      process.on('SIGINT', () => {
+        for (const instance of SpinningLogger.instances) {
+          if (instance.spinnerActive) {
+            instance.spinner.stop();
+          }
+        }
+        // Registering a SIGINT handler disables the default exit behavior, so we must exit explicitly.
+        process.exit(130);
+      });
+    }
   }
 
   /**
--- a/packages/amplify-cli/src/commands/gen2-migration/_common/spinning-logger.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/_common/spinning-logger.ts
@@ -19,8 +19,6 @@ export class SpinningLogger {
     this.debugMode = options?.debug ?? globalIsDebug;
     this.spinner = new AmplifySpinner();
 
-    SpinningLogger.instances.add(this);
-
     // Register the SIGINT handler only once to avoid accumulating listeners
     // across multiple SpinningLogger instances (which causes
     // MaxListenersExceededWarning and prevents clean process exit in tests).
@@ -49,6 +47,7 @@ export class SpinningLogger {
       return;
     }
     this.spinnerActive = true;
+    SpinningLogger.instances.add(this);
     this.spinner.start(this.buildSpinnerText());
   }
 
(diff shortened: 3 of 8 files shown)


sharonyajain:
## CI Failure Note

The `build-and-test` check failure is a **pre-existing repo-wide issue** unrelated to this PR's changes. The same failure reproduces on other PRs (e.g. `feat/gen2-migration-non-js-functions` — [run 32198932303](https://github.com/aws-amplify/amplify-cli/actions/runs/32198932303)).

### Root cause

GitHub Actions recently deprecated Node 20 on runners and forces all actions to run under Node 24. Under Node 24, Jest's process exit behavior is stricter with open async handles — specifically `Socket` listeners from AWS SDK client mocks in the gen2-migration snapshot tests prevent `cli-internal` from exiting cleanly (exit code 1 from open handles, not from test failures).

**All `cli-internal` test suites PASS** (including the new regression test added in this PR). The exit code 1 comes from Jest being unable to shut down the process after tests complete.

### What this PR includes to mitigate

- **SpinningLogger singleton SIGINT handler** — fixes the `MaxListenersExceededWarning: 11 SIGINT listeners` accumulation (each `SpinningLogger` instance was registering a new handler; now uses a static singleton pattern)
- **`--forceExit --detectOpenHandles`** on the `cli-internal` test script — matches the pattern already used by `amplify-appsync-simulator` and `amplify-provider-awscloudformation` in this repo

### What remains (out of scope for this PR)

The Node 24 runner upgrade requires either:
1. Fixing the remaining Socket handle leaks in the AWS SDK mock clients used by gen2-migration tests, OR
2. Adding `timeout-minutes` to the workflow, OR  
3. Splitting the `cli-internal` test suite into shards

This blocks all PRs to this repo equally and should be tracked as a separate infrastructure issue.

packages/amplify-cli/package.json
```diff
@@ -24,7 +24,7 @@
   },
   "scripts": {
     "build": "tsc",
-    "test": "jest --logHeapUsage",
+    "test": "jest --logHeapUsage --forceExit",
```
soberm: [minor] This only adds `--forceExit`, not `--detectOpenHandles`. The PR comment about the CI failure says both were added and that it matches `amplify-appsync-simulator` and `amplify-provider-awscloudformation`, but `detectOpenHandles` isn't actually here (and doesn't show up anywhere else in the repo). Also, of those two packages, only `amplify-appsync-simulator` uses `--forceExit`; `amplify-provider-awscloudformation` uses neither flag. Might be worth adding `--detectOpenHandles` for real so the next handle leak shows up in test output instead of just getting forced shut.

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
```diff
@@ -228,6 +228,15 @@ export class Gen1App {
     return value;
   }
 
+  /**
+   * Returns a resource output value from amplify-meta.json, or undefined if absent.
+   * Use for optional outputs that may not exist in all Gen1 configurations.
+   */
+  public tryResourceMetaOutput(resource: DiscoveredResource, outputKey: string): string | undefined {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any -- untyped amplify-meta.json
+    return (this._meta as any)[resource.category]?.[resource.resourceName]?.output?.[outputKey];
```
soberm, line 236: [nit] Same lookup expression as `resourceMetaOutput` right above, just without the throw. Could simplify by having `resourceMetaOutput` call this one and throw when it comes back `undefined`, so there's only one place doing the `_meta` traversal instead of two copies that have to stay in sync.

packages/amplify-cli/src/commands/gen2-migration/_common/spinning-logger.ts
```diff
@@ -17,14 +19,23 @@ export class SpinningLogger {
...
 
-    // Restore the cursor if the process is interrupted while the spinner is active
-    process.on('SIGINT', () => {
-      if (this.spinnerActive) {
-        this.spinner.stop();
-      }
-      // Registering a SIGINT handler disables the default exit behavior, so we must exit explicitly.
-      process.exit(130);
-    });
+    SpinningLogger.instances.add(this);
```
soberm: [minor] `instances` only grows, nothing removes an entry once a logger's done being used. Not a problem for a one-shot CLI run, but if this pattern gets reused somewhere longer-lived (or a migration ends up creating one of these per resource), they'll pile up for the life of the process. Maybe a `WeakSet`, or have `stop()`/`succeed()`/`failed()` unregister the instance.

packages/amplify-cli/package.json
sharonyajain: Rather than adding --detectOpenHandles, I removed --forceExit entirely — the actual exit-code-1 was a missing test-helper binding (now fixed) plus the SIGINT-listener/mock-client handle leaks addressed in this PR.

packages/amplify-cli/src/commands/gen2-migration/_common/gen1-app.ts
sharonyajain: Done

packages/amplify-cli/src/commands/gen2-migration/_common/spinning-logger.ts
sharonyajain: Fixed — instances now register on spinner activation (start/resume) and unregister on deactivation (stop/succeed/failed/pause), so the set only ever holds loggers with a live spinner instead of growing for the life of the process. 0f699b0.

soberm (dismissed):

sarayev:
The destination branch should be `dev`, `main` is for releases. See here: https://github.com/aws-amplify/amplify-cli/blob/dev/CONTRIBUTING.md?plain=1#L120

soberm (approved):

sarayev (approved):

PR #14972, opened 2026-08-18
feat(gen2-migration): generate code for non-JS Lambda functions

### LINKED ISSUE #14535: (gen2-migration) `generate` command should handle non js lambda functions

_No response_

_No response_

In Gen1, its possible to create functions in various languages:

```console
? Choose the runtime that you want to use: (Use arrow keys)
  .NET 8
  Go
  Java
❯ NodeJS
  Python
```

We currently only support the `NodeJS` runtime - we should support all.

In Gen2, non JS runtimes are supported via a flavor of the `defineFunction` API:

```ts
export const sayHelloFunctionHandler = defineFunction(
  (scope) =>
    new Function(scope, "say-hello", {
      handler: "index.handler",
      runtime: Runtime.PYTHON_3_9, // or any other python version
      timeout: Duration.seconds(20), //  default is 3 seconds
      code: ...
);
```

More information here: https://docs.amplify.aws/vue/build-a-backend/functions/custom-functions/. 

### Implementation Notes

- Since the CDK `Function` construct exposes the same API regardless of runtime - we expect common logic to support all different languages. If however we see that special handling is needed per language, we should prioritize python and go first.

None

### Additional context

_No response_

sharonyajain:
#### Description of changes

Resolves #14535 — `gen2-migration generate` now produces Gen2 code for non-JS Lambda functions (Python, Go, Java, .NET, Ruby) instead of throwing or skipping them.

Previously, the generate command only supported Node.js runtimes. PR #14744 added a skip path (via `--skip-validations`), but left users to hand-code the Gen2 definition. This PR completes the feature by generating the full `resource.ts` using Gen2's [custom functions](https://docs.amplify.aws/react/build-a-backend/functions/custom-functions/) pattern.

**What it generates** (example for Python):
```ts
import { execSync } from 'node:child_process';
import * as path from 'node:path';
import { fileURLToPath } from 'node:url';
import { defineFunction } from '@aws-amplify/backend';
import { DockerImage, Duration } from 'aws-cdk-lib';
import { Code, Function, Runtime } from 'aws-cdk-lib/aws-lambda';

const functionDir = path.dirname(fileURLToPath(import.meta.url));

export const myPythonFunc = defineFunction(
  (scope) =>
    new Function(scope, 'myPythonFunc', {
      handler: 'index.handler',
      runtime: Runtime.PYTHON_3_11,
      timeout: Duration.seconds(20),
      memorySize: 256,
      code: Code.fromAsset(functionDir, {
        bundling: {
          image: DockerImage.fromRegistry('dummy'),
          local: {
            tryBundle(outputDir: string) {
              execSync(`python3 -m pip install ...`);
              execSync(`cp -r ...`);
              return true;
            },
          },
        },
      }),
    }),
);
```

**Key changes:**
- `FunctionAssessor`: All runtimes are now `supported` for generate (removed the `isNonJsRuntime` gate)
- `FunctionRenderer`: New `renderCustomFunction()` method emits the CDK Function construct pattern with runtime-specific bundling
- `FunctionGenerator`: Removed the throw for non-JS runtimes; branches into custom path when `!runtime.startsWith('nodejs')`
- `copyFunctionSource`: Preserves dependency files (requirements.txt, go.mod, pom.xml) for non-JS functions
- `mapToCdkRuntime()`: Maps Lambda runtime strings → CDK `Runtime.*` enum members

**Runtime support:**
| Lambda Runtime | CDK Enum | Bundling |
|---|---|---|
| python3.x | PYTHON_3_X | pip install + copy |
| go1.x / provided.al2023 | PROVIDED_AL2023 | go build |
| java* | JAVA_* | mvn package |
| dotnet* | DOTNET_* | dotnet publish |
| ruby* | RUBY_* | copy |

#### Issue #, if available

Closes #14535

- Runtime mapping unit tests (7 runtimes → correct CDK enum)
- Custom function rendering tests for Python, Go, Java, .NET
- Assessor tests updated: all runtimes pass `validFor('generate')`
- Manual verification of generated output matches [Gen2 docs](https://docs.amplify.aws/react/build-a-backend/functions/custom-functions/)
- Pre-existing test for `does not render runtime for non-nodejs` still passes (Node.js path unchanged)

#### Checklist

- [x] PR description included
- [x] `yarn test` passes (gen2-migration tests)

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

### DIFF
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
@@ -80,16 +80,13 @@ export class FunctionGenerator implements Planner {
       });
 
     const runtime = config.Runtime;
-    if (runtime && !runtime.startsWith('nodejs')) {
-      throw new AmplifyError('UnsupportedRuntimeError', {
-        message: `Function '${deployedName}' uses unsupported runtime '${runtime}'. Gen 2 migration only supports Node.js functions.`,
-        resolution: 'Migrate the function to a Node.js runtime before running the Gen 2 migration.',
-      });
-    }
 
     this.logger.debug(`Fetching Lambda function schedule '${deployedName}'`);
     const schedule = await this.gen1App.aws.fetchFunctionSchedule(deployedName);
-    const entry = TS.extractFilePathFromHandler(config.Handler ?? 'index.js');
+    const entry =
+      runtime && !runtime.startsWith('nodejs')
+        ? config.Handler ?? 'index.handler'
+        : TS.extractFilePathFromHandler(config.Handler ?? 'index.js');
     const { literalEnvVars, dynamicEnvVars } = classifyEnvVars(config.Environment?.Variables ?? {}, this.readModelNames());
 
     const dynamoActions = this.extractDynamoActions();
@@ -141,21 +138,72 @@ export class FunctionGenerator implements Planner {
         describe: async () => [`Generate amplify/function/${resourceName}/resource.ts`],
         execute: async () => {
           const dirPath = path.join(this.outputDir, 'amplify', 'function', resourceName);
-
-          this.logger.info(`Rendering function/${resourceName}/resource.ts`);
-          const nodes = this.renderer.render(renderOpts);
-          const content = TS.printNodes(nodes);
-
           await fs.mkdir(dirPath, { recursive: true });
-          await fs.writeFile(path.join(dirPath, 'resource.ts'), content, 'utf-8');
-          await this.copyFunctionSource(resourceName, dirPath);
 
           const alias = resourceName;
-          this.backendGenerator.addNamespaceImport(alias, `./function/${resourceName}/resource`);
-          this.backendGenerator.addDefineBackendEntry(resourceName, alias, resourceName);
 
-          const escapeHatchArgs = this.deriveApplyEscapeHatchArguments(hasAnalytics, dynamicEnvVars);
-          this.backendGenerator.addApplyEscapeHatchesCall({ alias, extraArgs: escapeHatchArgs });
+          if (runtime && !runtime.startsWith('nodejs')) {
+            // Non-JS runtime: emit the Gen2 custom function pattern
+            // (defineFunction((scope) => new Function(...))). These functions
+            // have no applyEscapeHatches() export, so no escape-hatch call is
+            // registered on the backend.
+            this.logger.info(`Rendering custom function/${resourceName}/resource.ts`);
+            const manualMigrationNotes: string[] = [];
+            if (dynamicEnvVars.length > 0) {
+              manualMigrationNotes.push(
+                `Dynamic environment variables (branch/secret/cross-resource references) were not migrated: ${dynamicEnvVars
+                  .map((v) => v.name)
+                  .join(', ')}. Wire them via backend.${resourceName}.resources.lambda.addEnvironment(...) in backend.ts.`,
+              );
+            }
+            if (schedule) {
+              manualMigrationNotes.push(
+                `A schedule ('${schedule}') was configured on the Gen1 function but was not migrated. Add an EventBridge schedule in backend.ts.`,
+              );
+            }
+            const ungrantedPermissions: string[] = [];
+            if (dynamoActions.length > 0) ungrantedPermissions.push('DynamoDB table access');
+            if (appSyncPermissions.hasMutation || appSyncPermissions.hasQuery) ungrantedPermissions.push('AppSync (GraphQL API) access');
+            if (hasAnalytics) ungrantedPermissions.push('Kinesis (analytics) access');
+            if (unMappedAuthActions.length > 0) ungrantedPermissions.push('additional IAM policy actions');
+            if (dataTriggerModels.length > 0) ungrantedPermissions.push('data (@model) trigger wiring');
+            if (storageTriggerTables.length > 0) ungrantedPermissions.push('storage (DynamoDB) trigger wiring');
+            if (ungrantedPermissions.length > 0) {
+              manualMigrationNotes.push(
+                `Permission grants / triggers were not migrated: ${ungrantedPermissions.join(
+                  ', ',
+                )}. Re-add them on the function's CDK resource in backend.ts.`,
+              );
+            }
+            const content = this.renderer.renderCustomFunction({
+              resourceName,
+              handler: entry,
+              runtime,
+              architecture: config.Architectures?.[0],
+              timeoutSeconds: config.Timeout,
+              memoryMB: config.MemorySize,
+              environment: literalEnvVars,
+              manualMigrationNotes,
+            });
+            await fs.writeFile(path.join(dirPath, 'resource.ts'), content, 'utf-8');
+            await this.copyFunctionSource(resourceName, dirPath, runtime);
+
+            this.backendGenerator.addNamespaceImport(alias, `./function/${resourceName}/resource`);
+            this.backendGenerator.addDefineBackendEntry(resourceName, alias, resourceName);
+          } else {
+            this.logger.info(`Rendering function/${resourceName}/resource.ts`);
+            const nodes = this.renderer.render(renderOpts);
+            const content = TS.printNodes(nodes);
+
+            await fs.writeFile(path.join(dirPath, 'resource.ts'), content, 'utf-8');
+            await this.copyFunctionSource(resourceName, dirPath);
+
+            this.backendGenerator.addNamespaceImport(alias, `./function/${resourceName}/resource`);
+            this.backendGenerator.addDefineBackendEntry(resourceName, alias, resourceName);
+
+            const escapeHatchArgs = this.deriveApplyEscapeHatchArguments(hasAnalytics, dynamicEnvVars);
+            this.backendGenerator.addApplyEscapeHatchesCall({ alias, extraArgs: escapeHatchArgs });
+          }
         },
       },
     ];
@@ -220,20 +268,21 @@ export class FunctionGenerator implements Planner {
     }
   }
 
-  private async copyFunctionSource(resourceName: string, destDir: string): Promise<void> {
+  private async copyFunctionSource(resourceName: string, destDir: string, runtime?: string): Promise<void> {
     const srcDir = path.join('amplify', 'backend', 'function', resourceName, 'src');
+    const isNonJs = runtime !== undefined && !runtime.startsWith('nodejs');
     await fs.cp(srcDir, destDir, {
       recursive: true,
       filter: (src) => {
         const b = path.basename(src);
-        return (
-          b !== 'node_modules' &&
-          b !== '.yarn' &&
-          b !== 'package.json' &&
-          b !== 'package-lock.json' &&
-          b !== 'yarn.lock' &&
-          b !== 'pnpm-lock.yaml'
-        );
+        if (b === 'node_modules' || b === '.yarn') return false;
+        // For Node.js functions, filter out JS dependency/lock files so they
+        // don't leak into the generated Gen2 project. Non-JS functions keep
+        // their dependency manifests (requirements.txt, go.mod, pom.xml, etc.).
+        if (!isNonJs) {
+          return b !== 'package.json' && b !== 'package-lock.json' && b !== 'yarn.lock' && b !== 'pnpm-lock.yaml';
+        }
+        return true;
       },
     });
   }
--- a/packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
+++ b/packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
@@ -1,5 +1,6 @@
 import ts, { ObjectLiteralElementLike } from 'typescript';
 import { AmplifyError } from '@aws-amplify/amplify-cli-core';
+import { Runtime as CdkRuntime } from 'aws-cdk-lib/aws-lambda';
 import { newLineIdentifier, TS } from '../../ts';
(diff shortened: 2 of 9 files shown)


Simone319:
## CodeBuild Batch FAILED — Attempt 1

**Batch:** [`AmplifyCLI-PR-Testing:b5e4dc18-299d-4c87-b244-c74b7608ce7d`](https://us-east-1.console.aws.amazon.com/codesuite/codebuild/671107461633/projects/AmplifyCLI-PR-Testing/batch/AmplifyCLI-PR-Testing:b5e4dc18-299d-4c87-b244-c74b7608ce7d?region=us-east-1)
**Source SHA:** `268ade11e067737f26630d185833e3080808183d`
**Result:** 2 of 17 build groups FAILED (15 succeeded)

### ❌ `lint` — FAILED
- **875 ESLint problems (0 errors, 875 warnings)** were reported — not the direct blocker, since these are warnings.
- The actual failure comes from **Prettier formatting check** (`Checking formatting...`):
  ```
  [warn] packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/function/function.renderer.test.ts
  [warn] packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
  [warn] packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
  [warn] Code style issues found in 3 files. Forgot to run Prettier?
  ```
  → Command exited non-zero: `source ./shared-scripts.sh && _lint` (exit status 1)
- **Fix:** run `yarn prettier --write` (or equivalent repo lint-fix script) on those 3 files under `packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/`.

### ❌ `verify_versions_match` — FAILED
```
@aws-amplify/cli-internal version: 14.6.0-dev.268ade11e067737f26630d185833e3080808183d.0
@aws-amplify/cli version: 14.5.2-dev.268ade11e067737f26630d185833e3080808183d.0
```
- The dev-version bump for `@aws-amplify/cli-internal` (14.6.0) and `@aws-amplify/cli` (14.5.2) are out of sync — they must match.
- Tool's own guidance from the log:
  > Manual fix: add a proper conventional commit that touches the `amplify-cli-npm` package to correct its version bump. Example: https://github.com/aws-amplify/amplify-cli/pull/13759/commits/15dcd96feae925ff26ca51abfb4a0477890af745

---
_Reported automatically from the CodeBuild batch run above. Commit status `AWS CodeBuild BuildBatch us-east-1 (AmplifyCLI-PR-Testing)` has been set to `failure`._

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
```diff
@@ -126,21 +123,42 @@ export class FunctionGenerator implements Planner {
...
 
-          const escapeHatchArgs = this.deriveApplyEscapeHatchArguments(hasAnalytics, dynamicEnvVars);
-          this.backendGenerator.addApplyEscapeHatchesCall({ alias, extraArgs: escapeHatchArgs });
+          if (runtime && !runtime.startsWith('nodejs')) {
+            // Non-JS runtime: emit the Gen2 custom function pattern
+            // (defineFunction((scope) => new Function(...))). These functions
+            // have no applyEscapeHatches() export, so no escape-hatch call is
+            // registered on the backend.
+            this.logger.info(`Rendering custom function/${resourceName}/resource.ts`);
+            const content = this.renderer.renderCustomFunction({
```
Simone319, line 178: **Blocking:** This branch passes only handler/runtime/timeout/memory to `renderCustomFunction`, so the environment variables, schedule, and permission grants the Node.js path emits are dropped for non-JS functions — even though all three are fetched just above (`classifyEnvVars(config.Environment...)`, `fetchFunctionSchedule`, `extractDynamoActions`) and are configurable on Gen1 functions of any runtime. Env vars are the clear must-fix (add `environment:` to the generated `new Function({...})`); for the schedule and permissions, at minimum emit a warning / TODO so the migrated function doesn't silently lose them.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
```diff
@@ -942,3 +1005,66 @@ function convertScheduleExpression(raw: string): string | undefined {
...
+  if (runtime.startsWith('python')) {
+    return [
+      `execSync(\`python3 -m pip install -r \${path.join(functionDir, 'requirements.txt')} -t \${path.join(outputDir)} --platform manylinux2014_x86_64 --only-binary=:all:\`);`,
+      `execSync(\`cp -r \${functionDir}/* \${path.join(outputDir)}\`);`,
+    ];
+  }
+  if (runtime.startsWith('go') || runtime.startsWith('provided')) {
+    return [
+      `execSync(\`rsync -rLv \${functionDir}/* \${path.join(outputDir)}\`);`,
+      `execSync(\`cd \${path.join(outputDir)} && GOARCH=amd64 GOOS=linux go build -tags lambda.norpc -o \${path.join(outputDir)}/bootstrap \${functionDir}/main.go\`);`,
```
Simone319: **Blocking:** The Go build always outputs `bootstrap` (`-o .../bootstrap`), but the emitted `handler:` comes from the Gen1 config, so a `go1.x` function whose handler isn't literally `bootstrap` generates code that won't deploy. The renderer test masks this by passing `handler: 'bootstrap'`; consider forcing `handler: 'bootstrap'` for `provided`/`go` runtimes.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
```diff
@@ -942,3 +1005,66 @@ function convertScheduleExpression(raw: string): string | undefined {
...
+  if (runtime.startsWith('ruby')) {
+    // 'ruby3.3' → 'RUBY_3_3'
+    const version = runtime.replace('ruby', '').replace('.', '_');
+    return `RUBY_${version}`;
+  }
+  // Go and custom runtimes use provided.al2023
+  if (runtime.startsWith('go') || runtime.startsWith('provided')) {
+    return 'PROVIDED_AL2023';
+  }
+  // Fallback for unknown runtimes
```
Simone319, line 1128: **Important:** An unrecognized runtime falls through to `PROVIDED_AL2023` + copy-only bundling, emitting a `resource.ts` that looks valid but won't deploy; the previous behavior threw `UnsupportedRuntimeError`. Prefer failing loudly (or warning) for genuinely unknown runtimes rather than generating broken code.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
```diff
@@ -942,3 +1005,66 @@ function convertScheduleExpression(raw: string): string | undefined {
 
   return undefined;
 }
+
+/**
+ * Maps a Lambda runtime string to the CDK Runtime enum member name.
+ * 'python3.9' → 'PYTHON_3_9', 'go1.x' → 'PROVIDED_AL2023', etc.
+ */
+export function mapToCdkRuntime(runtime: string): string {
```
Simone319, line 1103: **Important:** The string-munging assumes a clean `<lang><version>` shape, so Gen1 runtime IDs with suffixes break: `java8.al2` → `JAVA_8.al2`, `dotnetcore3.1` → `DOTNET_core3.1`, neither a valid CDK enum member. An explicit lookup table keyed on known runtime strings (routing the rest to the loud-failure path above) would be more robust.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
```diff
@@ -54,6 +65,58 @@ export class FunctionRenderer {
...
+    lines.push(`export const ${opts.resourceName} = defineFunction(`);
+    lines.push(`  (scope) =>`);
+    lines.push(`    new Function(scope, '${opts.resourceName}', {`);
+    lines.push(`      handler: '${opts.handler}',`);
+    lines.push(`      runtime: Runtime.${cdkRuntime},`);
+    lines.push(`      timeout: Duration.seconds(${timeout}),`);
+    lines.push(`      memorySize: ${memorySize},`);
+    lines.push(`      code: Code.fromAsset(functionDir, {`);
+    lines.push(`        bundling: {`);
+    lines.push(`          image: DockerImage.fromRegistry('dummy'),`);
```
Simone319: **Important:** `DockerImage.fromRegistry('dummy')` is a non-functional Docker fallback — it only never fires because `tryBundle` always returns `true`; if the local toolchain is missing, `execSync` throws and synth dies with a cryptic error instead of falling back. Consider a real per-runtime image (as the Gen2 custom-function docs show), or documenting that generated bundling requires the local toolchain.

Simone319 (commented):
Thanks for this — the runtime-mapping coverage and the dependency-manifest preservation are nice touches. One overall question before this is mergeable: has the generated output been **deployed and invoked** end-to-end for at least one non-JS runtime (e.g. run `gen2-migration generate` on a Python or Go function, then `npx ampx sandbox` and a live invoke)? The validation listed is at the transformer/schema level, and several of the inline points below (the Go `bootstrap` handler, the dropped env vars/permissions, the `dummy` bundling image) would only surface at deploy/runtime — a single real deploy would confirm the generated function actually stands up.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.generator.ts
sharonyajain: **Fixed** in `d35e8cfe0a`. The custom path now emits literal env vars as `environment: { ... }` on the `new Function(...)`, and surfaces everything that can't be auto-migrated as a `// TODO:` block in the generated `resource.ts` so nothing is silently lost:
- **Env vars** — `renderCustomFunction` receives `environment: literalEnvVars` and renders them.
- **Dynamic env vars / schedule / permission grants & triggers** — the generator assembles `manualMigrationNotes` from `dynamicEnvVars`, `schedule`, `dynamoActions`, `appSyncPermissions`, `unMappedAuthActions`, and the data/storage trigger detectors, emitted as TODO lines pointing the user to re-wire them in `backend.ts`.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
sharonyajain: **Fixed** in `d35e8cfe0a`. For `go`/`provided` runtimes the handler is now forced to `'bootstrap'` regardless of the Gen1 handler string, matching the `-o .../bootstrap` build output. Added a regression test that passes a non-bootstrap handler (`'main'`, `go1.x`) and asserts the generated `handler: 'bootstrap'`.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
sharonyajain: **Fixed** in `d35e8cfe0a`. A genuinely unrecognized runtime family now emits a loud `// WARNING:` in the generated file stating the bundling only copies files and won't build/deploy as-is, telling the user to set a real runtime/build step. (Deprecated/EOL versions that aren't members of CDK's `Runtime` enum get a separate `// NOTE:` + `PROVIDED_AL2023` fallback.) I kept it as a prominent warning rather than a hard throw so partial/`--skip-validations` migrations still produce an editable file — but it's no longer silent.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
sharonyajain: Good catch — and the broken-output symptom is now fixed in `d35e8cfe0a`: the mapped member is validated against CDK's actual `Runtime` enum, so `java8.al2` → `JAVA_8.al2` and `dotnetcore3.1` → `DOTNET_core3.1` (neither a real member) no longer emit an invalid `Runtime.*` — they fall back to `PROVIDED_AL2023` with a `// NOTE:`. I agree an explicit lookup table keyed on known runtime strings would be more robust than the string-munging; since the guard already prevents the incorrect output, I'd propose that as a follow-up cleanup rather than widening this PR. Happy to file the issue if you'd like it tracked.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/function/function.renderer.ts
sharonyajain: **Fixed** in `d35e8cfe0a`. Replaced `DockerImage.fromRegistry('dummy')` with the runtime's real CDK bundling image (`Runtime.<member>.bundlingImage`), so if the local toolchain is missing CDK has a working Docker fallback instead of a cryptic `execSync` failure. Also added a comment documenting that local bundling runs the runtime toolchain on the host, with the image as the fallback.

sharonyajain:
On the end-to-end verification question: no — the generated output has **not** yet been deployed and invoked live for a non-JS runtime. Verification so far is unit/renderer-level (46 function-suite tests, 450 gen2-migration tests, prettier + version-match green). A real `gen2-migration generate` → `npx ampx sandbox` → live invoke on a Python/Go function needs an AWS account and a non-JS Gen1 app, which I haven't run.

Flagging this as the open **manual-verification** item. The inline fixes just pushed (env vars, forced Go `bootstrap` handler, real bundling image, loud warning for unknown runtimes) are exactly the things a live deploy would most likely have surfaced, so they narrow the risk — but a maintainer live pass (or a sandbox account to use) is still the right gate before merge. Head is now `d35e8cfe0a`.

Simone319 (approved):

PR #14981, opened 2026-08-27
fix(cli): gen2-migration forward-direction function grants avoid circular deps

### LINKED ISSUE #14727: (gen2-migration) `generate` command produces stacks with circular dependencies

Local

_No response_

### Amplify CLI Version

Local

Mac

No

### Describe the bug

Running `npx amplify gen2-migration generate` can result in a Gen2 app that cannot be deployed due to circular dependencies issues:

```console
🛑 [CloudformationStackCircularDependencyError] circular dependency found between nested stacks [storage, auth, data, function]
```

### Expected behavior

Resources should be placed in the correct stacks such that circular dependencies don't exist.

### Reproduction steps

[Fitness tracker](https://github.com/aws-amplify/amplify-cli/blob/gen2-migration/amplify-migration-apps/fitness-tracker/README.md) suffers from this. We currently compensate by instructing the user to change the `resourceGroupName` property of the function.

### Project Identifier

_No response_

### Log output

<details>

```


```

</details>

### Additional information

In addition, when a function accesses multiple resources, the `access` property on `defineAuth` creates a dependency `auth` -> `function`. If that function also accesses another category, say `data`, we now have an `auth` -> `data` dependency which creates a circle. Same thing happens with storage. So `resourceGroupName` alone doesn't fix the problem.

If however direct `function.addToRolePolicy` CDK is used, this dependency is avoided because it preserves the `function` -> `auth` direction, not the reverse.

sharonyajain:
#### Description of changes

`amplify gen2-migration generate` could produce a Gen2 app that fails to deploy with:

```
🛑 [CloudformationStackCircularDependencyError] circular dependency found between nested stacks [storage, auth, data, function]
```

**Root cause.** The generator emitted a function's access to auth/storage using
*reverse-direction* constructs:

- auth: `access: (allow) => [allow.resource(fn).to([...])]` on `defineAuth`
- storage: `allow.resource(fn).to([...])` entries inside the `defineStorage` access block

Both make the **auth/storage** stack reference the function's role — an
`auth → function` / `storage → function` cross-stack dependency. When the same
function also accesses another category (e.g. `data`), and that category
references auth (`data → auth`), the reverse-direction grant closes a cross-stack
cycle. Setting `resourceGroupName` only co-locates resources in one stack and
cannot break a genuine cross-category cycle.

**Fix.** Emit the grants in the **forward** direction in `backend.ts`, on the
underlying CDK constructs, so the dependency stays `function → auth` /
`function → storage`:

- auth → `backend.auth.resources.userPool.grant(backend.<fn>.resources.lambda, ...cognito-idp actions)`, using a documented Amplify-permission → `cognito-idp:*` action mapping (the CDK grant works on raw IAM actions, not Amplify permission names).
- storage → `backend.storage.resources.bucket.grant{Read,Write,ReadWrite,Delete}(backend.<fn>.resources.lambda)`, picking the narrowest grant that covers the consolidated permission set.

Identity-based storage access (guest / authenticated / group) is unchanged and
still rendered as the per-path `defineStorage` access block; a bucket whose only
Gen1 access was function access now emits no access block at all.

#### Issue #, if available

Fixes #14727

- Unit suites `auth.generator.test.ts` and `s3.generator.test.ts` rewritten to assert the forward-direction grants (spy on `addPostDefineBackendStatement`) — 39/39 pass, 28/28 snapshots.
- End-to-end migration-app snapshots regenerated and passing, including **fitness-tracker** (the repro named in the issue) — the `admin` function's `access` block on `defineAuth` is replaced by `backend.auth.resources.userPool.grant(backend.admin.resources.lambda, ...)` in `backend.ts`. `media-vault` and `store-locator` snapshots also updated.
- `tsc --noEmit` clean; eslint 0 errors on touched files; prettier clean.

Not yet validated: a live `ampx sandbox` deploy of the regenerated output confirming the circular-dependency error is gone. That requires an AWS account and a multi-category Gen1 app; flagging it as the remaining manual-verification gate.

One thing worth a maintainer's eye: the auth permission → `cognito-idp:*` mapping duplicates expansion logic that lives inside `@aws-amplify/backend-auth`. It needs a sanity-check against that source and a note to keep the table in sync.

#### Checklist

- [x] PR description included
- [x] `yarn test` passes (targeted gen2-migration generate suites; see above)

### DIFF
--- a/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
+++ b/packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
@@ -121,10 +121,7 @@ describe('S3Generator', () => {
 
       const branchName = process.env.AWS_BRANCH ?? 'sandbox';
 
-      export const storage = defineStorage({
-        name: \`myBucket-main-\${branchName}\`,
-        access: (allow) => ({}),
-      });
+      export const storage = defineStorage({ name: \`myBucket-main-\${branchName}\` });
 
       export function postRefactor(backend: Backend) {
         const s3Bucket = backend.storage.resources.cfnResources.cfnBucket;
@@ -596,7 +593,7 @@ describe('S3Generator', () => {
     `);
   });
 
-  it('renders function access patterns with imports', async () => {
+  it('emits function access as forward-direction grants in backend.ts', async () => {
     const gen1App = await createGen1App({
       providers: { awscloudformation: { StackName: 'amplify-test-main-123456', Region: 'us-east-1' } },
       storage: { myBucket: { service: 'S3', output: { BucketName: 'myBucket-main-abc123' } } },
@@ -620,52 +617,37 @@ describe('S3Generator', () => {
     );
     generator.addFunctionAccess('processImages', ['read', 'write']);
 
+    const addPostDefineBackendStatementSpy = jest.spyOn(backendGenerator, 'addPostDefineBackendStatement');
+
     const ops = await generator.plan();
     await ops[0].execute();
 
-    expect(writtenFile('resource.ts')).toMatchInlineSnapshot(`
-      "import { defineStorage } from '@aws-amplify/backend';
-      import { processImages } from '../function/processImages/resource';
-      import { CfnResource } from 'aws-cdk-lib';
-      import type { Backend } from '../backend';
+    // Function->storage access must NOT be emitted as a reverse-direction
+    // allow.resource(fn) entry in the defineStorage access block (that creates a
+    // storage->function dependency and causes circular dependencies). With no
+    // guest/authenticated/group access, resource.ts has no access block and does
+    // not import the function.
+    const resourceTs = writtenFile('resource.ts');
+    expect(resourceTs).not.toContain('access:');
+    expect(resourceTs).not.toContain('allow.resource');
+    expect(resourceTs).not.toContain("from '../function/processImages/resource'");
 
-      const branchName = process.env.AWS_BRANCH ?? 'sandbox';
-
-      export const storage = defineStorage({
-        name: \`myBucket-main-\${branchName}\`,
-        access: (allow) => ({
-          'public/*': [allow.resource(processImages).to(['read', 'write'])],
-          'protected/{entity_id}/*': [
-            allow.resource(processImages).to(['read', 'write']),
-          ],
-          'private/{entity_id}/*': [
-            allow.resource(processImages).to(['read', 'write']),
-          ],
-        }),
-      });
-
-      export function postRefactor(backend: Backend) {
-        const s3Bucket = backend.storage.resources.cfnResources.cfnBucket;
-        s3Bucket.bucketName = 'myBucket-main-abc123';
-      }
-
-      export function applyEscapeHatches(backend: Backend) {
-        const s3Bucket = backend.storage.resources.cfnResources.cfnBucket;
-        for (const cfnResource of backend.storage.stack.node
-          .findAll()
-          .filter(
-            (c) =>
-              CfnResource.isCfnResource(c) &&
-              ['AWS::S3::Bucket', 'Custom::S3AutoDeleteObjects'].includes(
-                c.cfnResourceType
-              )
-          )) {
-          (cfnResource as CfnResource).addOverride('UpdateReplacePolicy', 'Retain');
-          (cfnResource as CfnResource).addOverride('DeletionPolicy', 'Retain');
-        }
-      }
-      "
-    `);
+    // read + write emits an explicit IAM policy with the EXACT Amplify action set
+    // (GetObject + PutObject + ListBucket) and crucially NO s3:DeleteObject — a
+    // CDK bucket.grantReadWrite would have bundled DeleteObject in, over-granting
+    // a function whose Gen1 access was read+write only.
+    expect(addPostDefineBackendStatementSpy).toHaveBeenCalledWith(
+      'backend.processImages.resources.lambda.addToRolePolicy(\n' +
+        '  new PolicyStatement({\n' +
+        '    effect: Effect.ALLOW,\n' +
+        "    actions: ['s3:GetObject', 's3:PutObject', 's3:ListBucket'],\n" +
+        "    resources: [backend.storage.resources.bucket.arnForObjects('*'), backend.storage.resources.bucket.bucketArn],\n" +
+        '  }),\n' +
+        ');',
+    );
+    // The emitted statement must not carry a delete action.
+    const [[emitted]] = addPostDefineBackendStatementSpy.mock.calls;
+    expect(emitted).not.toContain('s3:DeleteObject');
   });
 
   it('renders triggers with function imports', async () => {
@@ -707,7 +689,6 @@ describe('S3Generator', () => {
 
       export const storage = defineStorage({
         name: \`myBucket-main-\${branchName}\`,
-        access: (allow) => ({}),
         triggers: {
           onUpload: onUploadFn,
           onDelete: onDeleteFn,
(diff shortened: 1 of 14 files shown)


sharonyajain:
Thanks for the review. Dispositions below, pushed as `7387f96fdf`.

### Critical

**1. `FunctionAccess` type consistency (auth `resourceName` vs storage `functionName`) — rebutted (no runtime error possible).**
The two `FunctionAccess` interfaces are declared in separate modules (`auth/auth.renderer.ts` and `storage/s3.renderer.ts`), each exported and consumed only within its own category. They are never imported across the boundary or assigned to one another, and their `permissions` fields are *different types* — auth is `AuthPermissions` (a boolean-keyed object), storage is `readonly Permission[]` (a string array). They model different things, so there is no structural-typing path that produces a runtime error, and "unifying" them isn't coherent. Both interfaces predate this PR. Left unchanged — the naming difference is a pre-existing cosmetic point outside this fix's scope.

**2. Validate generated IAM actions for the unknown-permission fallback — partially addressed; runtime-validation framework declined as out of scope.**
`func.permissions` is typed `AuthPermissions` (a fixed ~30-key interface), and every key is present in the mapping table, so the `cognito-idp:<permission>` fallback is unreachable for real Gen1 input — no deployment-failure path exists on the normal flow. Rather than add an IAM-action validator (speculative, and there is no authoritative client-side action list to validate against), I took the legitimate kernel: the generator now calls `AuthRenderer.unknownPermissions(...)` and emits a `logger.warn` naming any permission that fell through to the raw-action fallback, so drift surfaces loudly at generate time. The fallback still emits the raw action rather than dropping it silently.

**3. Test coverage for edge cases — fixed.**
Added to `auth.generator.test.ts`:
- unknown-permission fallback → raw `cognito-idp:<permission>` action **and** a `logger.warn` (new `unknownPermissions` path);
- no function access registered → no grant, no warning;
- a function with empty permissions → no grant emitted (strengthened the existing "skips empty access" test);
- overlapping permissions across duplicate entries for one function → a single grant with de-duplicated, sorted actions.
(Empty-array and multi-function cases were already covered by the existing "multiple functions" test.) Malformed function names aren't a meaningful case here — the resource name is the Amplify resource name from Gen1 metadata, not free-form input, and it's interpolated as an identifier the same way the existing generator already does.

### Architectural

**4. Permission-mapping maintenance / import from `@aws-amplify/backend-auth` — accepted, cannot import; added a drift guard instead.**
`@aws-amplify/backend-auth` does not export this permission→action table (the expansion happens inside `allow.resource().to()`), so there is nothing to import. I hoisted the table to a module-level `PERMISSION_ACTION_MAP` const with a doc comment stating the sync obligation, and the new `unknownPermissions` warning gives the automated-sync-validation signal you asked for: if upstream adds a permission not in the map, generate-time output warns rather than silently guessing. Flagged in the PR body as a maintainer sanity-check item.

**5. Extract shared `buildFunctionAccessBackendStatements` between auth and storage — rebutted (premature abstraction).**
The two builders share only a method name. Inputs differ (auth: boolean-map `AuthPermissions` → `cognito-idp` action strings via a mapping table; storage: `Permission[]` → CDK `grant{Read,Write,ReadWrite,Delete}` method selection) and so do outputs (`userPool.grant(..., ...actions)` vs `bucket.grant*(...)`). A shared utility over two things that differ in input, output, and mapping logic would add indirection without removing real duplication. Left as two focused methods.

### Minor

**6. Unused imports in `media-vault/.../auth/resource.ts` — already resolved.**
Those are regenerated migration-app snapshot fixtures; the removed function imports were dropped by regeneration and the files are self-contained generated output (nothing else imports from them). Verified no dangling references.

**7. Comment → JSDoc in `auth.generator.ts` — fixed.**
Converted the forward-direction inline comment to a `/** ... */` block.

Verification: `tsc --noEmit` clean, eslint 0 errors on touched files, prettier clean; `auth.generator` suite 29/29 (4 new tests), full `gen2-migration/generate` dir 23 suites / 258 tests / 97 snapshots pass (fitness-tracker + all migration-app snapshots unchanged by these edits).

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/storage/s3.renderer.ts
```diff
@@ -376,22 +366,69 @@ export class S3Renderer {
...
+    const statements: string[] = [];
+    for (const [functionName, permSet] of Object.entries(permsByFunction)) {
+      const lambda = `backend.${functionName}.resources.lambda`;
+      const grantee = `backend.storage.resources.bucket`;
+      const canRead = permSet.has('read');
+      const canWrite = permSet.has('write') || permSet.has('create');
+      const canDelete = permSet.has('delete');
+
+      if (canRead && canWrite) {
+        statements.push(`${grantee}.grantReadWrite(${lambda});`);
```
soberm: grantWrite / grantReadWrite include s3:DeleteObject* in CDK's action set, but Amplify's `write` permission maps to PutObject only (delete is a separate permission). A function whose Gen1 access was ['read','write'] (no delete) now receives DeleteObject it didn't have — a destructive scope-widening. It also makes the explicit grantDelete partly redundant (the read+write+delete case's grantDelete is dead, since grantReadWrite already grants delete). Consider an explicit addToRolePolicy with the exact Amplify action set, or document the intentional over-grant.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
```diff
@@ -113,6 +113,82 @@ const factory = ts.factory;
...
+ *
+ * This mirrors the permission -> action expansion that `allow.resource(fn).to()`
+ * performs internally in `@aws-amplify/backend-auth`. It is duplicated here
+ * because the forward-direction CDK grant (backend.auth.resources.userPool.grant)
+ * works on raw IAM actions rather than Amplify permission names. `@aws-amplify/backend-auth`
+ * does not export this table, so it cannot be imported; keep it in sync if the
+ * upstream permission set changes (the generator warns on any permission missing
+ * from this map, so a drift surfaces at generate time rather than silently).
+ */
+const PERMISSION_ACTION_MAP: Readonly<Record<string, readonly string[]>> = {
```
soberm, line 141: PERMISSION_ACTION_MAP is a hand-maintained duplicate of @aws-amplify/backend-auth's internal permission->action expansion. unknownPermissions() only warns when a permission NAME is missing from the map — it won't detect upstream ADDING an action to an existing permission (e.g. manageUsers), which would silently under-grant. Suggest pinning the upstream version/commit this table was copied from and calling out that drift on known-permission action sets is not guarded.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
```diff
@@ -594,69 +679,101 @@ export class AuthRenderer {
...
+    const statements: string[] = [];
+    for (const [resourceName, actions] of Object.entries(actionsByFunction)) {
+      if (actions.size === 0) {
+        continue;
       }
+      const actionArgs = Array.from(actions)
+        .sort()
+        .map((a) => `'${a}'`)
+        .join(', ');
+      statements.push(`backend.auth.resources.userPool.grant(backend.${resourceName}.resources.lambda, ${actionArgs});`);
```
soberm, line 821: resourceName/functionName are interpolated raw into generated TypeScript (backend.${name}.resources.lambda) with no sanitization in the generate path. A Gen1 name containing hyphens/dots would emit invalid TS. Mitigating factor: the defineBackend key derives from the same value so key and grantee break together (visible at generate time, not silent), but an identifier assertion would harden this.

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
```diff
@@ -620,52 +617,27 @@ describe('S3Generator', () => {
...
+    const resourceTs = writtenFile('resource.ts');
+    expect(resourceTs).not.toContain('access:');
+    expect(resourceTs).not.toContain('allow.resource');
+    expect(resourceTs).not.toContain("from '../function/processImages/resource'");
+
+    // read + write consolidates into a single forward-direction grantReadWrite
+    // on the bucket construct in backend.ts, keeping the dependency
+    // function->storage.
+    expect(addPostDefineBackendStatementSpy).toHaveBeenCalledWith(
+      'backend.storage.resources.bucket.grantReadWrite(backend.processImages.resources.lambda);',
```
soberm: Coverage gap: storage grant variants other than grantReadWrite (delete-only -> grantDelete, read+delete, write-only -> grantWrite) have no focused unit test, and the 'identity access retained AND function grant emitted' mixed case is only covered by the media-vault fixture. Auth side is well covered.

soberm (commented):
Reviewed the forward-direction grant fix — approach is directionally correct and breaks the cross-stack cycle (grants land on the function role, referencing only the auth/storage ARN); consolidation/de-dup/sort logic is sound and the three regenerated snapshots match the generator output. A few non-blocking notes inline. Also worth flagging: validation is unit + snapshot only — a one-off live `ampx sandbox` deploy of a regenerated app would confirm the CloudformationStackCircularDependencyError is actually gone end-to-end.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/storage/s3.renderer.ts
sharonyajain: Fixed in `20af9e18`. I confirmed against aws-cdk-lib@2.189.1 that `Bucket.writeActions = [...BUCKET_DELETE_ACTIONS, ...putActions]`, so both `grantWrite` and `grantReadWrite` include `s3:DeleteObject`, and the separate `grantDelete` in the read+write+delete case was dead (grantReadWrite already granted it).

Replaced the coarse `bucket.grant*` helpers with an explicit `addToRolePolicy(new PolicyStatement({...}))` carrying the EXACT Amplify Gen1 action set:
- `read` → `s3:GetObject` (objects) + `s3:ListBucket` (bucket)
- `write`/`create` → `s3:PutObject` (objects)
- `delete` → `s3:DeleteObject` (objects)

A read+write function no longer receives DeleteObject. Verified in the regenerated media-vault fixture: `thumbnailgen` (read+write+delete) now emits one statement `['s3:DeleteObject','s3:GetObject','s3:PutObject','s3:ListBucket']` instead of grantReadWrite + a redundant grantDelete. New tests assert the delete-only, read+delete, and write-only variants each carry exactly their actions.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
sharonyajain: Fixed in `20af9e18` (documentation). Correct — `unknownPermissions()` only catches a missing permission *name*, not an action added to an existing permission's set, which would silently under-grant. I pinned the upstream reference the table was transcribed from (`@aws-amplify/backend-auth` ~v1.6.x) in the `PERMISSION_ACTION_MAP` doc and called out both drift risks explicitly:

1. New permission NAME upstream → caught (unknownPermissions warns + raw-action fallback keeps it visible).
2. New ACTION on an existing permission → NOT caught; on any upstream bump, re-check the action sets below, not just the names.

The package doesn't export the raw-action map, so an import isn't possible; the pinned reference + the explicit "re-check action sets on bump" note is the guard against silent under-granting.

packages/amplify-cli/src/commands/gen2-migration/generate/amplify/auth/auth.renderer.ts
sharonyajain: Fixed in `20af9e18`. Added a shared `TS.assertValidIdentifier(name, context)` helper (single source in `ts.ts`, used by both the auth and storage grant builders) that throws a clear generate-time error if a resource/function name isn't a valid JS identifier (`/^[A-Za-z_$][A-Za-z0-9_$]*$/`), naming the interpolation site. So a hyphenated/dotted name fails loudly with an actionable message instead of emitting invalid TypeScript. Added a test asserting a `my-func` name throws at generate time.

packages/amplify-cli/src/__tests__/commands/gen2-migration/generate/amplify/storage/s3.generator.test.ts
sharonyajain: Fixed in `20af9e18`. Added focused storage tests:
- `emits exact per-variant actions for delete-only, read+delete, and write-only access` — each asserts the exact action set (delete-only → DeleteObject only, no ListBucket; read+delete → GetObject+DeleteObject+ListBucket, no PutObject; write-only → PutObject only, no ListBucket/DeleteObject).
- `emits both identity access AND a function grant when a bucket has both` — a guest-read + function-grant bucket, asserting the `access` block (guest) stays in defineStorage AND the forward-direction function grant lands in backend.ts (the mixed case that was only fixture-covered before).

Storage generator suite is now 15/15.

